PR #85, opened 2023-05-22
Define the `SimpleReplace` API.

cqc-alec:


specification/hugr.md
```diff
@@ -1063,7 +1063,46 @@ in {a}\* (i.e. there is no hierarchy relation between them).
...
+  - $\textrm{inp}(P)$ be the set of successor ports of edges from output ports
+    of the Input node of $P$;
+  - $\textrm{out}(P)$ be the set of input ports of the Output node of $P$.
+
+The method takes as input:
+
+  - the ID of a DFG node $P$ in $\Gamma$;
+  - a DFG-convex set $S$ of IDs of leaf nodes that are children of $P$ (not
+    including the Input and Output nodes);
+  - a hugr whose root is a DFG node $R$ with only leaf nodes as children;
```
acl-cqc, line 1091: just phrasing-wise, I think I'd prefer "a hugr $R$ whose root is a DFG node and which has only leaf nodes as children"

specification/hugr.md
```diff
@@ -1063,7 +1063,46 @@ in {a}\* (i.e. there is no hierarchy relation between them).
...
+Given a DFG-convex set $S$ of IDs of leaf nodes in a DSG, let:
+
+  - $\textrm{inp}(S)$ be the set of input ports of nodes in $S$ whose source is
+    in $\Gamma \setminus S$;
+  - $\textrm{out}(S)$ be the set of input ports of nodes in $\Gamma \setminus S$
+    whose source is in $S$.
+
+Given a DFG node $P$, let:
+
+  - $\textrm{inp}(P)$ be the set of successor ports of edges from output ports
```
acl-cqc, line 1082: Ok, so this is actually `$\textrm{inp}(S)$` (the other definition) taking S as `children(P) \ {Input}`, is that right? (I've been looking for a way to unify these two, or else give them different names. Maybe that's it. Ah, it doesn't account for intergraph edges - the `inp(P)` here would not include intergraph edges coming in, whereas `inp(S)` would include them.)

specification/hugr.md
```diff
@@ -1063,7 +1063,46 @@ in {a}\* (i.e. there is no hierarchy relation between them).
...
+    of the Input node of $P$;
+  - $\textrm{out}(P)$ be the set of input ports of the Output node of $P$.
```
acl-cqc, line 1084: ...and that's equivalent to `$\textrm{out}(S)$` (the other `out`) taking `S` as the singleton set of the Output node of P

specification/hugr.md
```diff
@@ -1063,7 +1063,46 @@ in {a}\* (i.e. there is no hierarchy relation between them).
...
+  - the ID of a DFG node $P$ in $\Gamma$;
+  - a DFG-convex set $S$ of IDs of leaf nodes that are children of $P$ (not
+    including the Input and Output nodes);
+  - a hugr whose root is a DFG node $R$ with only leaf nodes as children;
+  - a map $\nu_\textrm{inp}: \textrm{inp}(R) \to \textrm{inp}(S)$;
+  - a map $\nu_\textrm{out}: \textrm{out}(S) \to \textrm{out}(R)$.
+  
+The new hugr is then derived by:
+  
+  - adding all children of $R$, and all edges between them, to $\Gamma$;
```
acl-cqc, line 1097: I think you do not want to add the new `Input` and `Output` nodes from $R$

acl-cqc: Oh, sorry, you remove these at the end, but maybe it makes sense just not to add them

acl-cqc (commented):
Thanks Alec. The input and output maps make sense now, nice :).
1. I would still like to see if we can include entire subtrees rather than just leaves. If it complicates much then OK but I *think* that the only issue is....
2. Intergraph edges. I think we need to do something here, even if just to rule them out. Note that these will necessarily be edges from the outside into replaced subtrees. Since intergraph edges must be classical and can just be deleted, it feels like this shouldn't be a problem, and we might just fix this by appropriately defining the two `$\mathrm{inp}$` (which as I commented might be combined and that might deal with this)
3. Does it make sense to restrict these to DFGs? Basically the same routine could be used for CFGs: the replacement CFG would have an entry node (just a BB) and a distinguished exit node; the in/out mapping would route control edges incoming to the replaced subgraph all to successors of the "fake" entry node, and edges to the new/fake exit node, to successors of the replaced subgraph. So there are some terminology changes but the only real thing here is that we could potentially have two sets of incoming edges (control and data), and the same for outgoing - does that mean two inp/out maps or one combined map? Or, if this is best left as DFG only, maybe we should call it dfg_replace?

specification/hugr.md
cqc-alec: Yes, in fact in the implementation I didn't add them. I will update the spec. (Just have to be careful about distinguishing the nodes of R from the newly-added copies.)

cqc-alec: Done in https://github.com/CQCL-DEV/hugr/pull/85/commits/7d1535e1197dace3027fc359b24f0107c9c996c9 .

acl-cqc:
> Thanks Alec. The input and output maps make sense now, nice :).
> 
> 1. I would still like to see if we can include entire subtrees rather than just leaves. If it complicates much then OK but I _think_ that the only issue is....

To summarize discussion just now:-
* We believe intergraph edges "just work". The replacement will have to include corresponding endpoints, with edges to/from the Input/Output node, but these edges can then be rerouted to/from the original intergraph destinations/targets, just like other edges in/out of the replacement.
* However, dealing with non-leaf nodes does complicate the implementation, so let's skip that (for now or evermore)
* And, since we are dealing only with leaf nodes, CFGs are out-of-scope.

acl-cqc:
Ah - intergraph edges don't "just work" - for outgoing edges we would also need to deal with the associated Order edge. I guess they would work for incoming edges (the Order edge goes to the *parent* of the replaced nodes, which stays in the Hugr), but uniformity may be better than expressivity here....

specification/hugr.md
```diff
@@ -1063,7 +1063,41 @@ in {a}\* (i.e. there is no hierarchy relation between them).
...
+##### Replacement methods
+
+###### `SimpleReplace`
+
+This method is used for simple replacement of dataflow subgraphs consisting of
+leaf nodes.
+
+Given a set $S$ of nodes in a hugr $H$, let:
+
+  - $\textrm{inp}_H(S)$ be the set of input ports of nodes in $S$ whose source
```
acl-cqc, line 1075: Oh, neat - love the subscripting

specification/hugr.md
```diff
@@ -1063,7 +1063,41 @@ in {a}\* (i.e. there is no hierarchy relation between them).
...
+The method takes as input:
+
+  - the ID of a DFG node $P$ in $\Gamma$;
+  - a DFG-convex set $S$ of IDs of leaf nodes that are children of $P$ (not
+    including the Input and Output nodes), and that have no incoming or outgoing
+    Ext edges;
+  - a hugr $H$ whose root is a DFG node $R$ with only leaf nodes as children --
+    let $T$ be the set of non-Input/Output children of $R$;
+  - a map $\nu_\textrm{inp}: \textrm{inp}_H(T) \to \textrm{inp}_{\Gamma}(S)$;
+  - a map $\nu_\textrm{out}: \textrm{out}_{\Gamma}(S) \to \textrm{out}_H(R)$.
```
acl-cqc, line 1089: Does the final one there need to be \textrm{out}_H(T) rather than (R) ?

specification/hugr.md
```diff
@@ -1063,7 +1063,41 @@ in {a}\* (i.e. there is no hierarchy relation between them).
...
+
+The method takes as input:
+
+  - the ID of a DFG node $P$ in $\Gamma$;
+  - a DFG-convex set $S$ of IDs of leaf nodes that are children of $P$ (not
+    including the Input and Output nodes), and that have no incoming or outgoing
+    Ext edges;
+  - a hugr $H$ whose root is a DFG node $R$ with only leaf nodes as children --
+    let $T$ be the set of non-Input/Output children of $R$;
+  - a map $\nu_\textrm{inp}: \textrm{inp}_H(T) \to \textrm{inp}_{\Gamma}(S)$;
```
acl-cqc, line 1088: I wonder if we should allow the `T` here to include the Output node. In fact maybe we must - supposing the replacement had an Input linked directly to the Output?

specification/hugr.md
acl-cqc: or a T that includes the Input (but not the Output) - maybe we need two different T's rather than a common one

acl-cqc (approved):
Nice! :-)

specification/hugr.md
cqc-alec: > I wonder if we should allow the `T` here to include the Output node. In fact maybe we must - supposing the replacement had an Input linked directly to the Output?

Ah, good point. An edge case I had not considered...

acl-cqc: lol - well, me neither! But I think it's what you get if you remove a double-CX, say

cqc-alec: I think this was why I defined the `RemoveIdentity` primitive. Somehow I assumed it wasn't needed in this case but I think I was wrong. Perhaps the easiest thing is to disallow this situation for for `SimpleReplace`, but add an implementation of `RemoveIdentity`. (I think if `T` includes the Output node it messes up everything else.)

cqc-alec: That might just be a stroke of genius.

acl-cqc: I think the trick here is working out which bits are symmetric for inputs vs outputs, and which bits don't change - like that both use "predecessors" and there is no use of "successors", which (yours) might have been the first such stroke :)

PR #103, opened 2023-05-30
[new] Typechecking for Const nodes

croyzor:
Add a validation check that const values match the type of the wires coming out of them

### DIFF
--- a/src/hugr/typecheck.rs
+++ b/src/hugr/typecheck.rs
@@ -14,7 +14,7 @@ use std::fmt::{self, Display};
 
 /// Errors that arise from typechecking constants
 #[derive(Clone, Debug, Eq, PartialEq, Error)]
-pub enum TypeError {
+pub enum ConstTypeError {
     /// This case hasn't been implemented. Possibly because we don't have value
     /// constructors to check against it
     Unimplemented(ClassicType),
@@ -45,37 +45,37 @@ pub enum TypeError {
     TypeRowMismatch(TypeRow, TypeRow),
 }
 
-impl Display for TypeError {
+impl Display for ConstTypeError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let str = match self {
-            TypeError::Unimplemented(ty) => {
+            ConstTypeError::Unimplemented(ty) => {
                 format!("Const type checking unimplemented for {}", ty)
             }
-            TypeError::Failed(typ) => format!("Invalid const value for type {}", typ),
-            TypeError::IntTooLarge(width, val) => {
+            ConstTypeError::Failed(typ) => format!("Invalid const value for type {}", typ),
+            ConstTypeError::IntTooLarge(width, val) => {
                 format!("Const int {} too large for type I{}", val, width)
             }
-            TypeError::IntWidthTooLarge(w) => format!("Int type too large: I{}", w),
-            TypeError::IntWidthInvalid(w) => {
+            ConstTypeError::IntWidthTooLarge(w) => format!("Int type too large: I{}", w),
+            ConstTypeError::IntWidthInvalid(w) => {
                 format!(
                     "The int type I{} is invalid, because {} is not a power of 2",
                     w, w
                 )
             }
-            TypeError::IntWidthMismatch(exp, act) => format!(
+            ConstTypeError::IntWidthMismatch(exp, act) => format!(
                 "Type mismatch for int: expected I{}, but found I{}",
                 exp, act
             ),
-            TypeError::ConstCantBeVar => "Type of a const value can't be Var".to_string(),
-            TypeError::TupleWrongLength => "Tuple of wrong length".to_string(),
-            TypeError::LinearTypeDisallowed => {
+            ConstTypeError::ConstCantBeVar => "Type of a const value can't be Var".to_string(),
+            ConstTypeError::TupleWrongLength => "Tuple of wrong length".to_string(),
+            ConstTypeError::LinearTypeDisallowed => {
                 "Linear types not allowed in const nodes".to_string()
             }
-            TypeError::InvalidSumTag => "Tag of Sum value is invalid".to_string(),
-            TypeError::TypeMismatch(exp, act) => {
+            ConstTypeError::InvalidSumTag => "Tag of Sum value is invalid".to_string(),
+            ConstTypeError::TypeMismatch(exp, act) => {
                 format!("Type mismatch for const - expected {}, found {}", exp, act)
             }
-            TypeError::TypeRowMismatch(exp, act) => {
+            ConstTypeError::TypeRowMismatch(exp, act) => {
                 format!("Type mismatch for const - expected {}, found {}", exp, act)
             }
         };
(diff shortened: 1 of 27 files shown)


ss2165:
This PR does a few things to work around the fact that a `ConstValue::Int` doesn't actually know what type it is. I think this is worth fixing (values should be able to say what type they are), so this is done in #118 .
I think basing the validation in this PR on the changes in that one is the way to go.
This should mean hopefully that the changes to builder in this PR are not required.

src/ops/module.rs
```diff
@@ -40,7 +40,7 @@ pub enum ModuleOp {
         definition: SimpleType,
     },
     // A constant value definition.
-    Const(ConstValue),
```
ss2165, line 43: since you can always get the type from the value, seems no need to store it here?

src/hugr/validate.rs
```diff
@@ -374,7 +375,25 @@ impl<'a> ValidationContext<'a> {
 
         match from_optype.port_kind(from_offset).unwrap() {
             // Inter-graph constant wires do not have restrictions
-            EdgeKind::Const(_) => return Ok(()),
+            EdgeKind::Const(typ) => match from_optype {
+                // We don't need to check the second argument of ModuleOp::Const
+                // because it's what's providing the information for `port_kind`
+                OpType::Module(ModuleOp::Const(val, _)) => {
+                    println!("Const typechecking: {:?} in {}", val, typ);
```
ss2165, line 382: remove print

src/hugr/typecheck.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// A mismatch between the embedded type and the type we're checking
+    /// against, as above, but for rows instead of simple types
+    TypeRowMismatch(TypeRow, TypeRow),
+}
+
+impl Display for TypeError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let str = match self {
+            TypeError::Unimplemented(ty) => {
+                format!("Const type checking unimplemented for {}", ty)
```
ss2165, line 50: `thiserror` makes this much easier,

see for example https://github.com/CQCL-DEV/hugr/blob/f80dce7105c4d99dec904d6ef33930a584d3dc9a/src/hugr.rs#L238

src/hugr/typecheck.rs
```diff
@@ -0,0 +1,163 @@
...
+                format!("Type mismatch for const - expected {}, found {}", exp, act)
+            }
+        };
+        f.write_str(&str)
+    }
+}
+
+/// Per the spec, valid widths for integers are 2^n for all n in [0,7]
+fn check_valid_width(width: u8) -> Result<(), TypeError> {
+    if width > 128 {
```
ss2165, line 86: I would suggest adding a type alias for the width somewhere
`pub type HUGR_INT_WIDTH = u8;`

and then using things like 
```suggestion
    if width > (1<<HUGR_INT_WIDTH::BITS) {
```

ss2165: this should make it easier for us to change the width in future

src/hugr/typecheck.rs
```diff
@@ -0,0 +1,163 @@
...
+        return Err(TypeError::IntWidthTooLarge(width));
+    }
+
+    let valid_widths: Vec<u8> = (0..8).map(|a| u8::pow(2, a)).collect();
```
ss2165, line 90: this can be a static array, or hashset for speed

src/hugr/typecheck.rs
```diff
@@ -0,0 +1,163 @@
...
+    } else {
+        Err(TypeError::IntWidthInvalid(width))
+    }
+}
+
+/// Typecheck a constant value - for ensuring that it's valid before creating
+/// a const node
+pub fn typecheck_const(typ: &ClassicType, val: &ConstValue) -> Result<(), TypeError> {
+    match (typ, val) {
+        // N.B. If the width is larger than u6, our const value (backed by an i64)
```
ss2165, line 102: this comment seems out of date?

src/hugr/typecheck.rs
```diff
@@ -0,0 +1,163 @@
...
+        // wont be able to accomodate the value anyway.
+        (ClassicType::Int(exp_width), ConstValue::Int { value, width }) => {
+            // Check that the types make sense
+            check_valid_width(*exp_width)?;
+            check_valid_width(*width)?;
+            // Check that the terms make sense against the types
+            if exp_width == width {
+                let max_value = u128::pow(2, *width as u32);
```
ss2165, line 110: similarly the backing type should be declared somewhere as `pub type HUGR_INT_STORE = u128;`

ss2165, line 110: the compiler/stdlib might optimise using this anyway, but use `1<<pow` in general to do powers of two

src/hugr/typecheck.rs
```diff
@@ -0,0 +1,163 @@
...
+                if value < &max_value {
+                    Ok(())
+                } else {
+                    Err(TypeError::IntTooLarge(*width, *value))
+                }
+            } else {
+                Err(TypeError::IntWidthMismatch(*exp_width, *width))
+            }
+        }
+        (ty @ ClassicType::F64, _) => Err(TypeError::Unimplemented(ty.clone())),
```
ss2165, line 120: what's the trouble with floats?

ss2165 (changes requested):

src/hugr/typecheck.rs
croyzor: I've opted for `HugrIntWidthStore` and `HugrIntValueStore` because otherwise rust complains about non camel case type names

src/hugr/typecheck.rs
croyzor: We don't have any float constructor for `ConstValue` yet!

src/hugr/typecheck.rs
croyzor: I've just updated the logic here a bit, but it seems good to me to have `HugrIntValueStore::pow`when we're being generic over what output type is

src/hugr/typecheck.rs
```diff
@@ -1,163 +1,149 @@
...
             // Check that the types make sense
             check_valid_width(*exp_width)?;
             check_valid_width(*width)?;
             // Check that the terms make sense against the types
             if exp_width == width {
-                let max_value = u128::pow(2, *width as u32);
-                if value < &max_value {
+                let max_value = if *width == 128 {
+                    u128::MAX
+                } else {
```
ss2165, line 91:
```suggestion
                let max_value = if *width == HUGR_MAX_INT_WIDTH {
                    HUGR_MAX_INT_WIDTH ::MAX
                } else {
```

src/ops/module.rs
```diff
@@ -101,11 +101,15 @@ impl ModuleOp {
...
             )),
-            ModuleOp::Const(_, ty) => Some(EdgeKind::Const(ty.clone())),
+            ModuleOp::Const(tm) => Some(EdgeKind::Const(tm.const_type())),
         }
     }
 }
 
+pub(crate) type HugrIntValueStore = u128;
+pub(crate) type HugrIntWidthStore = u8;
+pub(crate) const HUGR_MAX_INT_WIDTH: HugrIntWidthStore = 128;
```
ss2165, line 111:
```suggestion
pub(crate) const HUGR_MAX_INT_WIDTH: HugrIntWidthStore = HugrIntValueStore::BITS;
```

?

src/hugr/validate.rs
```diff
@@ -700,7 +697,7 @@ pub enum ValidationError {
     InterGraphEdgeError(#[from] InterGraphEdgeError),
     /// Type error for constant values
     #[error("Type error for constant value: {0}.")]
-    ConstTypeError(#[from] TypeError),
+    ConstConstTypeError(#[from] ConstTypeError),
```
ss2165, line 700:
```suggestion
    ConstTypeError(#[from] ConstTypeError),
```

ss2165 (changes requested):
Looks good, just spotted a couple of things (mainly just need to change ConstConst)

ss2165 (approved):
🙏

PR #111, opened 2023-06-01
[fix] Display each port's type, rather than whole signature in dot

croyzor:
I reproduced @cqc-alec 's circuit (minus a Hadamard gate) and here's what it looks like now:

graphviz


cqc-alec (approved):
Looks good, thanks!

PR #112, opened 2023-06-01
Allow any op as root

ss2165:
to allow for building non-module root HUGR

Closes #82

### DIFF
--- a/src/algorithm/nest_cfgs.rs
+++ b/src/algorithm/nest_cfgs.rs
@@ -429,7 +429,7 @@ pub(crate) mod test {
         // entry -> split            > merge -> head -> tail -> exit
         //               \-> right -/             \-<--<-/
         let mut builder = HugrBuilder::new();
-        let mut module_builder = builder.module_builder();
+        let mut module_builder = builder.module_hugr_builder();
         let main = module_builder.declare("main", Signature::new_df(vec![NAT], type_row![NAT]))?;
         let pred_const = module_builder.constant(ConstValue::simple_predicate(0, 2))?; // Nothing here cares which
         let const_unit = module_builder.constant(ConstValue::simple_unary_predicate())?;
@@ -483,7 +483,7 @@ pub(crate) mod test {
         // Here we would like two consecutive regions, but there is no *edge* between
         // the conditional and the loop to indicate the boundary, so we cannot separate them.
         let mut builder = HugrBuilder::new();
-        let mut module_builder = builder.module_builder();
+        let mut module_builder = builder.module_hugr_builder();
         let main = module_builder.declare("main", Signature::new_df(vec![NAT], type_row![NAT]))?;
         let pred_const = module_builder.constant(ConstValue::simple_predicate(0, 2))?; // Nothing here cares which
         let const_unit = module_builder.constant(ConstValue::simple_unary_predicate())?;
@@ -685,7 +685,7 @@ pub(crate) mod test {
         //let sum2_type = SimpleType::new_predicate(2);
 
         let mut builder = HugrBuilder::new();
-        let mut module_builder = builder.module_builder();
+        let mut module_builder = builder.module_hugr_builder();
         let main = module_builder.declare("main", Signature::new_df(vec![NAT], type_row![NAT]))?;
         let pred_const = module_builder.constant(ConstValue::simple_predicate(0, 2))?; // Nothing here cares which
         let const_unit = module_builder.constant(ConstValue::simple_unary_predicate())?;
--- a/src/algorithm/nest_cfgs.rs
+++ b/src/algorithm/nest_cfgs.rs
@@ -449,7 +449,7 @@ pub(crate) mod test {
         cfg_builder.branch(&merge, 0, &head)?;
         let exit = cfg_builder.exit_block();
         cfg_builder.branch(&tail, 0, &exit)?;
-        let cfg_id = cfg_builder.finish();
+        let cfg_id = cfg_builder.finish()?;
 
         func_builder.finish_with_outputs(cfg_id.outputs())?;
         module_builder.finish()?;
@@ -501,7 +501,7 @@ pub(crate) mod test {
         cfg_builder.branch(&merge, 0, &tail)?; // trivial "loop body"
         let exit = cfg_builder.exit_block();
         cfg_builder.branch(&tail, 0, &exit)?;
-        let cfg_id = cfg_builder.finish();
+        let cfg_id = cfg_builder.finish()?;
 
         func_builder.finish_with_outputs(cfg_id.outputs())?;
         module_builder.finish()?;
@@ -717,7 +717,7 @@ pub(crate) mod test {
         cfg_builder.branch(&entry, 0, &head)?;
         cfg_builder.branch(&tail, 0, &exit)?;
 
-        let cfg_id = cfg_builder.finish();
+        let cfg_id = cfg_builder.finish()?;
 
         func_builder.finish_with_outputs(cfg_id.outputs())?;
         module_builder.finish()?;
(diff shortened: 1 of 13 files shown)


src/algorithm/nest_cfgs.rs
```diff
@@ -452,7 +453,7 @@ pub(crate) mod test {
 
         func_builder.finish_with_outputs(cfg_id.outputs())?;
 
-        let h = module_builder.finish()?;
+        let h = builder.finish()?;
```
cqc-alec, line 456: Do we not need to call `module_builder.finish()`?

src/algorithm/nest_cfgs.rs
```diff
@@ -503,7 +505,7 @@ pub(crate) mod test {
...
```
cqc-alec, line 508: Do we not need to call `module_builder.finish()`?

src/algorithm/nest_cfgs.rs
```diff
@@ -717,7 +720,7 @@ pub(crate) mod test {
...
```
cqc-alec, line 723: Do we not need to call `module_builder.finish()`?

src/builder.rs
```diff
@@ -67,14 +69,60 @@ pub enum BuildError {
...
+        }
+    }
+
+    /// Use this builder to build a module HUGR
+    pub fn module_builder(&mut self) -> ModuleBuilder {
+        ModuleBuilder(&mut self.base)
+    }
+
+    /// Use this builder to build a DFG HUGR
+    pub fn root_dfg_builder(
```
cqc-alec, line 93: Should we call it `dfg_builder` (by analogy: `module_builder` returns a `ModuleBuilder`; this returns a `DFGBuilder`)? Edit: hmm, no that's already taken. How about `module_hugr_builder` and `dfg_hugr_builder`?

src/builder.rs
```diff
@@ -93,11 +141,12 @@ mod test {
...
     ) -> Result<Hugr, BuildError> {
-        let mut module_builder = ModuleBuilder::new();
+        let mut builder = HugrBuilder::new();
+        let mut module_builder = builder.module_builder();
         let f_builder = module_builder.declare_and_def("main", signature)?;
 
         f(f_builder)?;
 
-        module_builder.finish()
+        builder.finish()
```
cqc-alec, line 150: Again, do we need to call `module_builder.finish()`? (Probably I'm missing something.)

src/algorithm/nest_cfgs.rs
aborgna-q: It'd be nice to have relevant types to prevent this.

Perhaps we could call `finish` on the builder wrapper's `drop`? (ofc that only works if finish doesn't take any parameter.)

ss2165: what are you trying to prevent?

aborgna-q: Forgetting the call to `module_builder.finish()`

cqc-alec (approved):
magic

src/builder.rs
```diff
@@ -67,14 +69,77 @@ pub enum BuildError {
...
+    // TODO: CFG, BasicBlock, Def, Conditional, TailLoop, Case
+
+    /// Complete building and return HUGR, performing validation.
+    pub fn finish(self) -> Result<Hugr, BuildError> {
+        Ok(self.base.finish()?)
+    }
+}
+
+// macro to implement empty drop for builders
+// forces use of value (through calling finish) before automatic drop
```
aborgna-q, line 118: Where does it call `finish`?

src/builder/dataflow.rs
```diff
@@ -82,7 +82,8 @@ impl<'f> Dataflow for DFGBuilder<'f> {
 }
 
 /// Wrapper around [`DFGBuilder`] used to build other dataflow regions.
-pub struct DFGWrapper<'b, T>(DFGBuilder<'b>, PhantomData<T>);
+// Stores option of DFGBuilder so it can be taken out without moving.
```
aborgna-q, line 85: Is it possible to have a wrapper with the builder taken out? I only see that being done in `finish()`, which immediately drops the wrapper.

src/hugr/validate.rs
```diff
@@ -51,12 +51,6 @@ impl<'a> ValidationContext<'a> {
                 node: self.hugr.root(),
```
aborgna-q, line 51: > // Root node must be a root in the hierarchy, and a module root operation.

There's a leftover comment in line 48

src/hugr/validate.rs
```diff
@@ -95,6 +89,7 @@ impl<'a> ValidationContext<'a> {
         let flags = optype.validity_flags();
 
         // The Hugr can have only one root node.
+        // The root node has no edges.
```
aborgna-q, line 92: Perhaps we should check. An edge to a child would be caught on the child's validation,
but a loop edge on the root node would not be checked.

aborgna-q (commented):
I think there are some `expect`s in the validation that may fail when unwrapping a parent if the node is the root (esp. in `validate_intergraph_edge`).
We could check that the root has no connections, but what is checked first depend on the traversal order for the node validation.


Also, I don't see where the new `Option`s on the builder wrappers come into play.

Otherwise, this looks nice.

src/builder.rs
ss2165: forces the user to call finish, the previous lines that Alec flagged as "should we call module finish here?" would now throw compiler errors because rust would try to call "drop" at the end of the scope - which requires a mutable reference to be alive. By calling "finish" you ensure the builder leaves scope.

src/builder/dataflow.rs
ss2165: this was a hack to make it possible to call "finish" on the inside builder (I wasn't allowed to move it out of the wrapper to call finish because it implements Drop - this may be a more serious)

src/builder.rs
aborgna-q: It doesn't prevent things like
```rust
let mut builder = HugrBuilder::new();
{
  let mut module_builder = builder.module_hugr_builder();
}
builder.finish()
```
But I guess it avoids some errors.

A more drastic approach would be calling finish or `panic!`ing on the drop, but I'm not sure we want to go there.

src/builder/dataflow.rs
aborgna-q: Fair.
I think the only way to get around that would be some `MaybeUninit` and `ManuallyDrop` trickeries, but I wouldn't do that unless there's a perf problem.

src/builder.rs
ss2165: https://github.com/CQCL-DEV/hugr/issues/115 made an issue here

ss2165:
Have left compositional building via graph inserting as a follow up task https://github.com/CQCL-DEV/hugr/issues/120

src/builder/build_traits.rs
```diff
@@ -58,11 +56,36 @@ pub trait Container {
...
+}
+
+/// Types implementing this trait can be used to build complete HUGRs
+/// (with varying root node types)
+pub trait HugrBuilder: Container {
+    /// Finish building the HUGR, perform any validation checks and return it.
+    fn finish_hugr(self) -> Result<Hugr, ValidationError>;
+}
+
+/// Type implementing this trait build a container graph region by borrowing a HUGR
```
aborgna-q, line 80:
```suggestion
/// Types implementing this trait build a container graph region by borrowing a HUGR
```

src/builder/module_builder.rs
```diff
@@ -171,14 +164,15 @@ mod test {
...
-            Dataflow,
+            Dataflow, DataflowSubContainer,
         },
         type_row,
     };
 
     use super::*;
     #[test]
     fn basic_recurse() -> Result<(), BuildError> {
+        // let mut builder = HugrBuilder::new();
```
aborgna-q, line 175:
```suggestion
```

aborgna-q (approved):
Awesome!

PR #114, opened 2023-06-05
Implement `SimpleReplace`

cqc-alec:
This PR includes and implements #85 .

Since we don't yet have DFG-rooted HUGRs we use a workaround where the replacement graph is specified as a HUGR with a unique DFG node. Once #112 is merged we can remove this workaround.

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1072,12 +1072,12 @@ There are the following primitive operations.
 This method is used for simple replacement of dataflow subgraphs consisting of
 leaf nodes.
 
-Given a set $S$ of nodes in a hugr $H$, let:
+Given a set $X$ of nodes in a hugr $G$, let:
 
-  - $\textrm{inp}_H(S)$ be the set of input ports of nodes in $S$ whose source
-    is in $H \setminus S$;
-  - $\textrm{out}_H(S)$ be the set of input ports of nodes in $H \setminus S$
-    whose source is in $S$.
+  - $\textrm{inp}_G(X)$ be the set of input ports of nodes in $X$ whose source
+    is in $G \setminus X$;
+  - $\textrm{out}_G(X)$ be the set of input ports of nodes in $G \setminus X$
+    whose source is in $X$.
 
 Notation: given an input port $p$, let $p^-$ be its unique predecessor port.
 
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1079,6 +1079,8 @@ Given a set $S$ of nodes in a hugr $H$, let:
   - $\textrm{out}_H(S)$ be the set of input ports of nodes in $H \setminus S$
     whose source is in $S$.
 
+Notation: given an input port $p$, let $p^-$ be its unique predecessor port.
+
 The method takes as input:
 
   - the ID of a DFG node $P$ in $\Gamma$;
@@ -1090,16 +1092,20 @@ The method takes as input:
   - a map $\nu\_\textrm{inp}: \textrm{inp}\_H(T \setminus \\{\texttt{Input}\\}) \to \textrm{inp}\_{\Gamma}(S)$;
   - a map $\nu_\textrm{out}: \textrm{out}_{\Gamma}(S) \to \textrm{out}_H(T \setminus \\{\texttt{Output}\\})$.
   
-The new hugr is then derived by:
+The new hugr is then derived as follows:
   
-  - adding copies of all children of $R$, except for Input and Output nodes, to
-    $\Gamma$, and make them all children of $P$;
-  - adding edges between all newly added nodes matching those in $R$;
-  - for each $p \in \textrm{inp}\_H(T)$, adding an edge from the predecessor of
-    $\nu\_\textrm{inp}(p)$ to the new copy of $p$;
-  - for each $p \in \textrm{out}\_{\Gamma}(S)$, adding an edge from the new copy
-    of the predecessor of $\nu\_\textrm{out}(p)$ to $p$.
-  - removing all nodes in $S$ and edges between them.
+  1. Make a copy in $\Gamma$ of all children of $R$, excluding Input and Output,
+     and all edges between them. Make all the newly added nodes children of $P$.
+     Notation: if $p$ is a port of a node in $R$, write $p^*$ for the copy of
+     the port in $\Gamma$.
+  2. For each $(q, p = \nu_\textrm{inp}(q))$ such that $q \notin \texttt{Output}$,
+     add an edge from $p^-$ to $q^*$.
+  3. For each $(p, q = \nu_\textrm{out}(p))$ such that $q^- \notin \texttt{Input}$,
+     add an edge from $(q^-)^*$ to $p$.
+  4. For each $p_1, q, p_0$ such that
+     $q = \nu_\textrm{out}(p_1), p_0 = \nu_\textrm{inp}(q)$, add an edge from
+     $p_0^-$ to $p_1$. (Sanity check: $q^-$ must be an Input node in this case.)
+  5. Remove all nodes in $S$ and edges between them.
 
 ###### `Replace`
 
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1063,7 +1063,45 @@ in {a}\* (i.e. there is no hierarchy relation between them).
 
 There are the following primitive operations.
 
-##### Replacement method
+##### Replacement methods
+
+###### `SimpleReplace`
+
+This method is used for simple replacement of dataflow subgraphs consisting of
+leaf nodes.
+
+Given a DFG-convex set $S$ of IDs of leaf nodes in a DSG, let:
+
+  - $\textrm{inp}(S)$ be the set of input ports of nodes in $S$ whose source is
+    in $\Gamma \setminus S$;
+  - $\textrm{out}(S)$ be the set of input ports of nodes in $\Gamma \setminus S$
+    whose source is in $S$.
+
+Given a DFG node $P$, let:
+
+  - $\textrm{inp}(P)$ be the set of successor ports of edges from output ports
+    of the Input node of $P$;
+  - $\textrm{out}(P)$ be the set of input ports of the Output node of $P$.
+
+The method takes as input:
+
+  - the ID of a DFG node $P$ in $\Gamma$;
+  - a DFG-convex set $S$ of IDs of leaf nodes that are children of $P$;
+  - a "modular hugr" with DFG root $R$ and only leaf nodes as children;
+  - a map $\nu_\textrm{inp}: \textrm{inp}(S) \to \textrm{inp}(R)$;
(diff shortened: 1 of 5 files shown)


src/hugr.rs
```diff
@@ -73,6 +77,127 @@ impl Hugr {
...
+            index_map.insert(*n_index, new_node_index);
+        }
+        // 3.2. Add edges between all newly added nodes matching those in n_dfg_node.
+        for n_index in n_non_io_indices {
+            let n_node = Node { index: *n_index };
+            let new_node_index = index_map.get(n_index).unwrap();
+            for n_node_succ in r.n.output_neighbours(n_node) {
+                if r.n.get_optype(n_node_succ).tag() != OpTag::Output {
+                    let new_node_succ_index = index_map.get(&n_node_succ.index).unwrap();
+                    for connection in r.n.graph.get_connections(*n_index, n_node_succ.index) {
```
acl-cqc, line 137: Hmmm, there is a lot of `.index` going on here - it's strange that PortGraph `port_offset` returns an `Option<PortOffset>` but `link_nodes` takes `usize`s (that it converts to PortOffsets internally).

I wonder if you can do this more easily with `HugrView::node_outputs`...ah, but we can't mutate the Hugr while we hold a view, so we can't use that.

Bah. Ok, fine. Might be worth a PR to update PortGraph.link_nodes to use PortOffset, but I'm not gonna hold this up for that.

src/hugr.rs
```diff
@@ -73,6 +77,127 @@ impl Hugr {
...
+            }
+            let new_node_index = self.graph.add_node(sig.input.len(), sig.output.len());
+            self.op_types[new_node_index] = op.clone();
+            // Make r.p the parent
+            self.hierarchy
+                .insert_after(new_node_index, self_input_node_index)
+                .ok();
+            index_map.insert(*n_index, new_node_index);
+        }
+        // 3.2. Add edges between all newly added nodes matching those in n_dfg_node.
```
acl-cqc, line 130: IIUC the spec was written assuming implicit copies so I'm guessing this may change when that's ready.

src/replacement/simple_replace.rs
```diff
@@ -0,0 +1,228 @@
...
+
+use crate::{hugr::Node, Hugr, Port};
+use thiserror::Error;
+
+/// Specification of a simple replacement operation.
+#[derive(Debug, Clone)]
+pub struct SimpleReplacement {
+    /// The common DFG parent of all nodes to be replaced.
+    pub p: Node,
+    /// The set of nodes to remove (a convex set of leaf children of `p`).
```
acl-cqc, line 11: How about calling this `remove` (with comment "called S in the spec" if you like!)

src/replacement/simple_replace.rs
```diff
@@ -0,0 +1,228 @@
...
+    use crate::ops::tag::OpTag;
+    use crate::ops::{DataflowOp, LeafOp, OpType};
+    use crate::types::{LinearType, Signature, SimpleType};
+    use crate::{type_row, Port};
+
+    use super::SimpleReplacement;
+
+    const QB: SimpleType = SimpleType::Linear(LinearType::Qubit);
+
+    /// Creates a hugr with a DFG.
```
acl-cqc, line 75: When we have rooted-at-any-op Hugrs, we might as well skip much of this as well, right? (Just to simplify the test)

src/replacement/simple_replace.rs
```diff
@@ -0,0 +1,228 @@
...
+        let _f_id = {
+            let mut func_builder = module_builder.declare_and_def(
+                "main",
+                Signature::new_df(type_row![QB, QB], type_row![QB, QB]),
+            )?;
+
+            let [qb0, qb1] = func_builder.input_wires_arr();
+
+            let mut inner_builder =
+                func_builder.dfg_builder(vec![(QB, qb0), (QB, qb1)], type_row![QB, QB])?;
```
acl-cqc, line 138: and ASCII art again

src/hugr.rs
```diff
@@ -73,6 +77,127 @@ impl Hugr {
...
+            let new_out_portindex = self
+                .graph
+                .port_index(*new_out_nodeindex, n_preexit_portoffset)
+                .unwrap();
+            self.graph.unlink_port(s_exit_portindex);
+            self.graph
+                .link_ports(new_out_portindex, s_exit_portindex)
+                .ok();
+        }
+        // 3.5. Remove all nodes in r.s and edges between them.
```
aborgna-q, line 193: Can we do this before inserting the new nodes, to allow reusing memory?

aborgna-q (changes requested):
I strongly suggest using more meaningful names in the implementation. Having to go to an internal definition in another module to parse what `r.s` adds a lot of friction.

src/hugr.rs
aborgna-q: > it's strange that PortGraph port_offset returns an Option<PortOffset> but link_nodes takes usizes

Yes, I'm changing that soon™️

src/replacement/simple_replace.rs
cqc-alec: I propose renaming:

* `p` --> `region`
* `s` --> `removal`
* `n` --> `replacement`

src/hugr.rs
cqc-alec: No, because we then panic at the line
```
let rem_inp_predecessor_port_index = self.graph.port_link(rem_inp_port_index).unwrap();
```
where we need to locate the predecessor node.

cqc-alec:
Addressed comments in the PR (except the ones about code location, which I'll return to). Moved to draft since it implements the old version of `SimpleReplace` (which didn't handle empty wires in the replacement), so needs updating to match the new version...

src/hugr.rs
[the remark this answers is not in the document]
acl-cqc: This is in progress in #119

src/replacement/simple_replace.rs
acl-cqc: I love `removal` and agree with `replacement`. Less sure about `region` - maybe just `parent`?

cqc-alec:
I think this now handles empty wires correctly. Note the updated spec.

cqc-alec:
Some fixes needed following merge from latest `main`; update to follow...

src/replacement/simple_replace.rs
cqc-alec: Changed to `parent`.

src/hugr.rs
```diff
@@ -70,6 +75,193 @@ pub struct Wire(Node, usize);
...
+        }
+        // 3. Do the replacement.
+        // 3.1. Add copies of all replacement nodes and edges to self. Exclude Input/Output nodes.
+        // Create map from old NodeIndex (in r.replacement) to new NodeIndex (in self).
+        let mut index_map: HashMap<NodeIndex, NodeIndex> = HashMap::new();
+        let replacement_nodes = r
+            .replacement
+            .hierarchy
+            .children(r.replacement.root().index)
+            .map_into::<Node>()
```
aborgna-q, line 104:
```suggestion
            .children(r.replacement.root())
```

src/hugr.rs
```diff
@@ -70,6 +75,193 @@ pub struct Wire(Node, usize);
...
+                .get_optype(node)
+                .signature()
+                .const_input
+                .is_empty()
+            {
+                return Err(SimpleReplacementError::InvalidReplacementNode());
+            }
+        }
+        let self_input_node_index = self.hierarchy.first(r.parent.index).unwrap();
+        let replacement_output_node = replacement_nodes[replacement_sz - 1];
```
aborgna-q, line 123: Nitpicky, but it's slightly clearer
```suggestion
        let replacement_output_node = *replacement_nodes.last().unwrap();
```

src/hugr.rs
aborgna-q, line 1: Is the idea to move `apply_simple_replacement` to `simple_replace.rs` once #119 is done?
The code could be moved over there now, so it's less noisy to change that afterwards.

aborgna-q (commented):
`Result::ok` just casts the value to an `Option`, discarding any error information. In most cases calling that before `unwrap` will give less debug info.

specification/hugr.md
```diff
@@ -1090,16 +1092,20 @@ The method takes as input:
   - a map $\nu\_\textrm{inp}: \textrm{inp}\_H(T \setminus \\{\texttt{Input}\\}) \to \textrm{inp}\_{\Gamma}(S)$;
   - a map $\nu_\textrm{out}: \textrm{out}_{\Gamma}(S) \to \textrm{out}_H(T \setminus \\{\texttt{Output}\\})$.
```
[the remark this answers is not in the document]
cqc-alec: I think they are correct? T is a set in H, and S is a set in Gamma.

cqc-alec:
> `Result::ok` just casts the value to an `Option`, discarding any error information. In most cases calling that before `unwrap` will give less debug info.

Thanks, yes. It seems I can't change all instances of `.ok().unwrap()` though because of errors like:

```
error[E0277]: `NodePortOffsets` doesn't implement `Debug`
    --> src/replacement/simple_replace.rs:230:14
     |
230  |             .unwrap();
     |              ^^^^^^ `NodePortOffsets` cannot be formatted using `{:?}` because it doesn't implement `Debug`
     |
     = help: the trait `Debug` is not implemented for `NodePortOffsets`
     = help: the trait `Debug` is implemented for `ExactlyOneError<I>`
     = note: required for `itertools::adaptors::map::MapSpecialCase<NodePortOffsets, itertools::adaptors::map::MapSpecialCaseFnInto<Port>>` to implement `Debug`
     = note: 1 redundant requirement hidden
     = note: required for `ExactlyOneError<MapSpecialCase<NodePortOffsets, MapSpecialCaseFnInto<Port>>>` to implement `Debug`
     = note: the full type name has been written to '/home/alec/r/hugr/target/debug/deps/hugr-653fda1ba5647c44.long-type-8370086427390340701.txt'
note: required by a bound in `Result::<T, E>::unwrap`
    --> /home/alec/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:1085:12
     |
1085 |         E: fmt::Debug,
     |            ^^^^^^^^^^ required by this bound in `Result::<T, E>::unwrap`
```

Will change what I can.

src/hugr.rs
cqc-alec: Yeah I was going to wait until that was done, since there isn't really a framework for it yet.

specification/hugr.md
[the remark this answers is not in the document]
aborgna-q: Just above it says:

> Given a set $S$ of nodes in a hugr $H$, let:

aborgna-q (approved):

specification/hugr.md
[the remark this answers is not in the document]
cqc-alec: I see the confusion. But that is a different H (scoped to the definition). I will change the definition to use a different letter!

PR #100, opened 2023-05-30
MultiPortGraph layer for implicit copies

aborgna-q:
This is a work in progress for implementing implicit copies as a thin layer an top of a portgraph.

Switching the graph in a hugr for this should be almost frictionless, except `LinkError::AlreadyConnected` should not be thrown when connecting already-connected ports.
There's just some slight API changes when compared to portgraph, and some new methods needed to manipulate the subports.

Ideally this should be upstreamed to portgraph in the future.

Implementation progress:
- [x] Basic wrapper with automatic creation of copy nodes.
- Iterators:
  - [x] `Nodes`, `NodeSubports`, `Neighbours`, `PortLinks`
  - [x] `NodeConnections`
  - [x] `NodeLinks`
  - [x] `Ports`
- [x] `port_count` method
- [x] Node and port compacting
- [x] Drop copy nodes when `set_num_ports` deletes some ports
- [x] Unit tests

PR #116, opened 2023-06-05
Base definition for the MultiPortGraph

aborgna-q:
This PR introduces a layer on top of a portgraph that allows multiple edges to be connected to the same port.
Internally, it automatically creates and deletes "copy nodes" associated with the multi-connected ports (_multiports_).

To be able to distinguish between different connections to the same port, there's a new `SubPortIndex` that contains the main PortIndex plus the offset of the port in the copy node (or 0 if there is no copy node).

Most of the interface of a `portgraph` has been carried over, replacing `PortIndex` with `SubPortIndex` where appropriate, or adding specialized variations (e.g. `unlink_port` vs `unlink_subports`).

This PR is the first split change from #100. It does not contain any iterator definitions.

### DIFF
--- a/src/hugr/multiportgraph.rs
+++ b/src/hugr/multiportgraph.rs
@@ -1,0 +1,771 @@
+//! Wrapper around a portgraph providing multiports via implicit copy nodes
+
+use bitvec::vec::BitVec;
+
+use itertools::Itertools;
+use portgraph::{
+    portgraph::{NodePortOffsets, NodePorts},
+    Direction, LinkError, NodeIndex, PortGraph, PortIndex, PortOffset, SecondaryMap,
+};
+
+use serde::{Deserialize, Serialize};
+
+/// An unlabelled port graph that allows multiple links to the same ports.
+///
+/// A port graph consists of a collection of nodes identified by a [`NodeIndex`].
+/// Each node has an ordered sequence of input and output ports, identified by a [`PortIndex`] that is unique within the graph.
+/// To optimize for the most common use case, the number of input and output ports of a node must be specified when the node is created.
+/// Multiple connections to the same [`PortIndex`] can be distinguished by their [`SubportIndex`].
+///
+/// When a node and its associated ports are removed their indices will be reused on a best effort basis
+/// when a new node is added.
+/// The indices of unaffected nodes and ports remain stable.
+#[derive(Clone, PartialEq, Default, Debug, Serialize, Deserialize)]
+pub struct MultiPortGraph {
+    graph: PortGraph,
+    /// Flags marking the internal ports of a multiport. That is, the ports connecting the main node and the copy nodes.
+    multiport: BitVec,
+    /// Flags marking the implicit copy nodes.
+    copy_node: BitVec,
+    /// Number of implicit copy nodes.
+    copy_node_count: usize,
+    /// Number of subports in the copy nodes of the graph.
+    subport_count: usize,
+}
+
+/// Index of a multi port within a `PortGraph`.
+///
+/// Note that the offsets of the subport indices are not guaranteed to be
+/// contiguous nor well-ordered. They are not invalidated by adding or removing
+/// other links to the same port.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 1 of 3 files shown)


src/hugr/multiportgraph.rs
```diff
@@ -0,0 +1,771 @@
...
+        self.multiport.shrink_to_fit();
+        self.copy_node.shrink_to_fit();
+    }
+}
+
+/// Internal helper methods
+impl MultiPortGraph {
+    /// Remove an internal copy node.
+    fn remove_copy_node(&mut self, copy_node: NodeIndex, from: PortIndex) {
+        let dir = self.port_direction(from).unwrap();
```
ss2165, line 480:
```suggestion
        let dir = self.port_direction(from).unwrap();
        debug_assert!(self.copy_node.get(copy_node));
```

src/hugr/multiportgraph.rs
```diff
@@ -0,0 +1,771 @@
...
+        dir: Direction,
+    ) -> Result<(), LinkError> {
+        match dir {
+            Direction::Incoming => self.graph.link_ports(port2, port1),
+            Direction::Outgoing => self.graph.link_ports(port1, port2),
+        }
+    }
+
+    /// Returns the PortIndex from the main node that connects to this copy node.
+    #[inline]
```
ss2165, line 593: seems big to inline

src/hugr/multiportgraph.rs
```diff
@@ -0,0 +1,771 @@
...
+        let link = self.graph.port_link(port_index)?;
+        self.graph.port_node(link)
+    }
+
+    /// Get the `PortIndex` in the copy node for a SubportIndex.
+    ///
+    /// If the port is not a multiport, returns the port index in the operation node.
+    fn get_subport_index(&self, subport: SubportIndex) -> Option<PortIndex> {
+        let port_index = subport.port();
+        if self.is_multiport(subport.port()) {
```
ss2165, line 653:
```suggestion
        if self.is_multiport(port_index) {
```

src/hugr/multiportgraph.rs
```diff
@@ -0,0 +1,771 @@
...
+    #[test]
+    fn link_ports() {
+        let mut g = MultiPortGraph::new();
+        let node0 = g.add_node(2, 1);
+        let node1 = g.add_node(1, 2);
+        let node0_output = g.output(node0, 0).unwrap();
+        let node1_input = g.input(node1, 0).unwrap();
+        assert_eq!(g.link_count(), 0);
+
+        // Link the same ports twice
```
ss2165, line 753:
```suggestion
        // Link the same ports thrice
```

ss2165 (commented):
Looks good, just a few comments/queries

src/hugr/multiportgraph.rs
```diff
@@ -0,0 +1,771 @@
...
+            .get_subport_index(subport_from)
+            .expect("subport_from does not exist");
+        let to_index = self
+            .get_subport_index(subport_to)
+            .expect("subport_to does not exist");
+        self.graph.link_ports(from_index, to_index)
+    }
+
+    /// Unlinks all connections to the `port`. Return `false` if the port was not linked.
+    pub fn unlink_port(&mut self, port: PortIndex) -> bool {
```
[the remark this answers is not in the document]
aborgna-q: It's a bit annoying, because since the copy port is removed you have to allocate a new container with those ports even if the user doesn't need them.

src/hugr/multiportgraph.rs
```diff
@@ -0,0 +1,771 @@
...
+            true
+        } else {
+            self.graph.unlink_port(port).is_some()
+        }
+    }
+
+    /// Unlinks the `port` and returns the port it was linked to. Returns `None`
+    /// when the port was not linked.
+    ///
+    /// TODO: Remove copy nodes when they are no longer needed?
```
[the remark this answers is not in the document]
aborgna-q: I'll put it as a new issue since it's only a performance optimization.

src/hugr/multiportgraph.rs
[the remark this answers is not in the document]
ss2165: fair enough

PR #119, opened 2023-06-05
Refactor: move `rewrite` inside `hugr`, `Rewrite` -> `Replace` implementing new 'Rewrite' trait

acl-cqc:
* Hugr::apply_rewrite now just "bounces" off a method implemented by the trait. So we could actually remove Hugr::apply_rewrite altogether??
* Renamed Rewrite to Replace, although I expect it to be entirely replaced by SimpleReplace, Replace, etc. etc.
* The old Hugr::apply_rewrite code is now moved into an implementation of the trait method....
* ....which can see internals of Hugr by virtue of being in the new `hugr::replace` package

An alternative scheme is to use a fixed `enum` of the allowable operations: https://github.com/CQCL-DEV/hugr/compare/refactor/replace - but I think the trait version is better because:
* while trait implementations outside the hugr module won't be able to do much, the trait at least allows someone to construct "compound" operations containing a sequence of RewriteOp's that we provide, so it's not pointless;
* The trait avoids a big match, as calling the trait method implicitly does that for us, so I think modularity/legibility is improved
* However, as per first point, we could also just drop the `trait` altogether and have a bunch of "unrelated" methods...


src/hugr.rs
```diff
@@ -74,29 +75,8 @@ impl Hugr {
...
-            |_| {},
-            node_inserted,
-            |_, _| {},
-            |_, _| {},
-        )?;
-
-        // TODO: Check types
-
-        Ok(())
+    pub fn apply<E>(&mut self, op: impl RewriteOp<E>) -> Result<(), E> {
```
aborgna-q, line 78: I'd keep the `apply_rewrite` name here, since it's the public API.
Otherwise the name won't mean much to the user.

src/hugr/replace.rs
```diff
@@ -0,0 +1,14 @@
+//! Rewrite operations on the HUGR.
+
+#[allow(clippy::module_inception)] // TODO: Rename?
+pub mod rewrite;
+use crate::Hugr;
+pub use rewrite::{OpenHugr, Rewrite, RewriteError};
+
+/// An operation that can be applied to mutate a Hugr
+pub trait RewriteOp<E> {
```
aborgna-q, line 9: We should make this a [sealed trait](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed), since we don't want user writing their own rewrite ops (for now at least).

src/hugr/replace.rs
aborgna-q, line 1: We are mixing `Rewrite` with `Replace`. I think the `RewriteOp` trait should share the name with its module, but either word is ok.

aborgna-q (changes requested):

src/hugr/replace.rs
acl-cqc: Yeah, ok, the naming was not consistent. I've had to do some more renaming to make it so, specifically the old Rewrite is now Replace (as I expect this to be rewritten to match the replacement API in the spec). However "Rewrite" is now a generic term for a (Replace or {Inline,Outline}{CFG,DFG}, ...etc...).

src/hugr/replace.rs
acl-cqc: I am not entirely convinced here. I think I could be persuaded, but especially given making a sealed trait involves playing around with some mildly complex arrangement of namespaces etc. (not just added a `sealed` between `pub` and `trait`, say), I'm still pondering what this achieves and why.

* Implementing the trait doesn't give you any power to see inside the Hugr. That power only comes from being defined inside the hugr module.
* The only point of implementing the trait is to be able to do `h.apply_rewrite(r)` rather than `r.apply(&mut h)`. Is that such a signal that we want to restrict it to "our" classes? It may be....

In practice many transformations one might want to apply to a Hugr are going to involve a series of the Rewrite's that we will define here. For example, an outlining, following by a replace, followed by an inlining. If we seal the trait, such transformations are distinguished from the "primitives" that we provide. Is that useful? *Maybe*. But it is certainly less compositional (one loses the ability to parameterize code over Rewrites including custom/compound ones, for example)....

src/hugr/rewrite.rs
```diff
@@ -0,0 +1,14 @@
...
+#[allow(clippy::module_inception)] // TODO: Rename?
+pub mod replace;
+use crate::Hugr;
+pub use replace::{OpenHugr, Replace, ReplaceError};
+
+/// An operation that can be applied to mutate a Hugr
+pub trait Rewrite<E> {
+    /// Mutate the specified Hugr, or fail with an error.
+    /// Implementations are strongly encouraged not to mutate the Hugr
+    /// if they return an Err.
```
aborgna-q, line 12: That's quite restrictive. Normally we'd use `?` to resurface any internal errors during the rewriting. Those may happen at any point, so this would require doing (potentially costly) dry-checks before running.

aborgna-q: Transactional semantics is it's own beast. Anything beyond 1. will incur a hit to performance, that we may perhaps prefer the user to have control over.

2. 3. 4. don't seem too practical without lots of extra overhead.
5. May be an option for region-focused rewrites, but not necessarily for every rewrite.
6. The pre-execution predicates are a thing in tket, and I think something we wanted to move away from.

I'd say the general rewrite interface shouldn't require this. I propose:

7. Leave these assurances to each Replace operation. Add `Transactional<>` wrappers or perhaps specialized versions (e.g. `struct TransactionalSimpleReplace(SimpleReplace)`) so the users that actually want to pay the cost can choose to.

@cqc-alec and @ss2165 may have opinions, but we could move this to a discussion too.

acl-cqc: Hmmm, perhaps I should state that `apply` may `panic` if-and-only-if the Hugr doesn't validate()?

acl-cqc: Ok, I think I've enacted as discussed

src/hugr/rewrite.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// If this call succeeds, [self.apply] should also succeed on the same `h`
+    /// If this calls fails, [self.apply] would fail with the same error.
+    fn verify(&self, h: &Hugr) -> Result<(), E>;
+
+    /// Mutate the specified Hugr, or fail with an error.
+    /// If [self.unchanged_on_failure] is true, then `h` must be unchanged if Err is returned.
+    /// See also [self.verify]
+    /// # Panics
+    /// May panic if-and-only-if `h` would have failed [Hugr::validate]; that is,
+    /// implementations may begin with `assert!(h.validate())`, and *should* begin with `debug_assert!(h.validate())`
```
acl-cqc, line 26: @ss2165 this feels a bit strong. `are encouraged to use debug_assert instead` (i.e. if they want to do anything)?

ss2165: yeah I think that works, or somewhat equivalently "may begin with `debug_assert!(h.validate())`"

src/hugr/rewrite.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// An operation that can be applied to mutate a Hugr
+pub trait Rewrite<E> {
+    /// If `true`, [self.apply]'s of this rewrite guarantee that they do not mutate the Hugr when they return an Err.
+    /// If `false`, there is no guarantee; the Hugr should be assumed invalid when Err is returned.
+    const UNCHANGED_ON_FAILURE: bool;
+
+    /// Checks whether the rewrite would succeed on the specified Hugr.
+    /// If this call succeeds, [self.apply] should also succeed on the same `h`
+    /// If this calls fails, [self.apply] would fail with the same error.
```
acl-cqc, line 18: An alternative is to return `Result<(), Option<E>>` where returning Err(None) means, no guarantee can be given about `apply`. If a rewrite really has to get partway through in order to complete its own validity checks then said alternative would support that. But I think we should leave as `Result<(), E>` until/unless we find a Rewrite where that's difficulty.

(Compound Rewrites is one case. `verify` in such cases might have to involve cloning the hugr.)

acl-cqc:
Ok, I've now merged Alec's SimpleReplace with this, thus applying the same refactoring to his code. I haven't managed to implement verify.

I'm tempted to make `verify` return a "proof" of ability-to-perform rewrite rather than `()`, that is:
```
trait TransactionCommitter { // One would like to use FnOnce here but there are Rust problems
  fn commit(h: &mut Hugr) -> ();
}

trait Rewrite<E> {
  type Committer: TransactionCommitter;
  fn verify(&self, h: &Hugr) -> Result<Self::Committer, E>;
  fn apply(&self, h:&mut Hugr) -> Result<(), E> {
    self.verify(h)?.commit()
  }
}
```
but I think that can be another PR (as it's somewhat contentious) - I'm hoping this isn't anymore

src/hugr/rewrite/simple_replace.rs
```diff
@@ -41,6 +48,198 @@ impl SimpleReplacement {
...
+        for &node in replacement_inner_nodes {
+            // Add the nodes.
+            let op: &OpType = self.replacement.get_optype(node);
+            let sig = op.signature();
+            let new_node_index = h.graph.add_node(sig.input.len(), sig.output.len());
+            h.op_types[new_node_index] = op.clone();
+            // Make self.parent the parent
+            h.hierarchy
+                .insert_after(new_node_index, self_input_node_index)
+                .ok();
```
acl-cqc, line 106: @cqc-alec all this code was basically moved here, modulo renaming `self` -> `h` and `r` -> `self`. I'm not sure about this `.ok()` though - that converts the Result to an Option, and then throws away. Should we '.unwrap'? Happy to make a driveby change in this PR if you agree

src/hugr/rewrite/simple_replace.rs
```diff
@@ -41,6 +48,198 @@ impl SimpleReplacement {
...
+                            .unwrap()
+                            .index();
+                        h.graph
+                            .link_nodes(
+                                *new_node_index,
+                                src_offset,
+                                *new_node_successor_index,
+                                tgt_offset,
+                            )
+                            .ok();
```
acl-cqc, line 140: @cqc-alec and here?

src/hugr/rewrite/simple_replace.rs
cqc-alec: Yes happy with the drive-by change.

src/hugr/rewrite.rs
aborgna-q: I'd say custom enums with `{Recoverable, NonRecoverable}` variants rather than `Option<E>`.
But yeah, let's keep it simple for now.

src/hugr/rewrite.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    fn verify(&self, h: &Hugr) -> Result<(), E> {
+        self.underlying.verify(h)
+    }
+
+    fn apply(self, h: &mut Hugr) -> Result<(), E> {
+        if R::UNCHANGED_ON_FAILURE {
+            return self.underlying.apply(h);
+        }
+        // note that if underlying.may_fail_destructively(h) is false, we don't need a backup...
```
aborgna-q, line 51: Outdated comment

src/hugr/rewrite/replace.rs
```diff
@@ -114,30 +116,41 @@ impl Rewrite {
...
     }
 
+    pub fn verify_convexity(&self) -> Result<(), ReplaceError> {
+        todo!()
+    }
+
+    pub fn verify_boundaries(&self) -> Result<(), ReplaceError> {
+        todo!()
+    }
+}
```
aborgna-q, line 126:
```suggestion
    pub fn verify_convexity(&self) -> Result<(), ReplaceError> {
        unimplemented!()
    }

    pub fn verify_boundaries(&self) -> Result<(), ReplaceError> {
        unimplemented!()
    }
}
```

src/hugr/rewrite.rs
```diff
@@ -0,0 +1,60 @@
...
+pub mod replace;
+pub mod simple_replace;
+use std::mem;
+
+use crate::Hugr;
+pub use replace::{OpenHugr, Replace, ReplaceError};
+pub use simple_replace::{SimpleReplacement, SimpleReplacementError};
+
+/// An operation that can be applied to mutate a Hugr
+pub trait Rewrite<E> {
```
aborgna-q, line 13: Each rewrite has an specific error type. `E` should be an associated type (`Error` ?) instead of a generic parameter.

aborgna-q (approved):
Main comment is replacing the generic parameter with an associated error type in `Rewrite`.

Merging `HEAD` may break some things.

src/hugr/rewrite.rs
acl-cqc: Doubly so (the mentioned function has been renamed, and it no longer applies), oops

src/hugr/rewrite/replace.rs
acl-cqc: I feel I want to leave this code untouched as much as possible given it is likely destined to the bin, but otoh - the plan is probably not "to do" these, so ok :-)

src/hugr/rewrite.rs
acl-cqc: Yes, good call, thanks

PR #126, opened 2023-06-07
Use multiportgraph in Hugr

aborgna-q:
Replace the PortGraphs with MultiPortGraph. This required some changes to the validation and things interfacing with the low-level definition.

In the validation, output edges with a non-linear port are now allowed to have multiple connections.

This PR _does not_ remove the explicit copy nodes. The builder remains untouched for now; the next step is actually modifying it to use the multiports instead of creating explicit copy nodes.
A quick test modifying `if_copy_add_port` and `discard` showed that this should be straighforward.


src/hugr/multiportgraph.rs
```diff
@@ -70,6 +73,14 @@ impl MultiPortGraph {
         }
     }
 
+    /// Returns a reference to the internal plain portgraph.
+    ///
+    /// This graph exposes the copy nodes as well as the main nodes.
+    pub fn as_graph(&self) -> &PortGraph {
```
ss2165, line 79: `as_port_graph` is clearer I think

ss2165 (approved):
👍

PR #131, opened 2023-06-07
Improve copy node rendering in dotstr

aborgna-q:
This is a temporary fix until we upstream MultiPortGraph to Portgraph, and add proper support for rendering multigraphs.

For now the copy nodes are shown, but all labels are stripped.

image


ss2165 (approved):
the hackery 😢

PR #132, opened 2023-06-07
Make HugrMut a trait

acl-cqc:
This turned out significantly harder than I'd expected, I'd not realized how much parameterization over hugr/ref/hugrmut there was in the builder. But I think I got there....
* Anything returning a (new) HugrMut has to return an actual Hugr, it's the only concrete implementation of HugrMut we have.
* `Container` having both `base()` and `hugr()` methods now seems a bit strange. But one is mutable and one is not....so left as is for now (TODO's can be removed if we are happy)?
* Other things are variously parameterized over HugrMut and HugrMutRef...this was quite hard to satisfy `rustc` and I'm not sure I've got all these as consistent as I could have. The acid test will be when we have mutable views onto subregions and want to use HugrMut methods on those, but that's not in this PR!
* Using `Hugr::default()` for a new module-rooted hugr is perhaps slightly tricky, could expose something there. (This was what the old HugrMut was using.)
* Also made HugrMut be only `pub(crate)`, but that's a separate commit (1bb79d2df4c398aa4cb7778338d7c1e7878cab3c)

### DIFF
--- a/src/algorithm/nest_cfgs.rs
+++ b/src/algorithm/nest_cfgs.rs
@@ -399,7 +399,7 @@ impl<T: Copy + Clone + PartialEq + Eq + Hash> EdgeClassifier<T> {
 pub(crate) mod test {
     use super::*;
     use crate::builder::{
-        BuildError, CFGBuilder, Container, Dataflow, DataflowSubContainer, HugrBuilder, HugrMutRef,
+        BuildError, CFGBuilder, Container, Dataflow, DataflowSubContainer, HugrBuilder,
         ModuleBuilder, SubContainer,
     };
     use crate::ops::{
@@ -611,7 +611,7 @@ pub(crate) mod test {
         dataflow_builder.finish_with_outputs([u].into_iter().chain(w))
     }
 
-    fn build_if_then_else_merge<T: HugrMutRef>(
+    fn build_if_then_else_merge<T: AsMut<Hugr> + AsRef<Hugr>>(
         cfg: &mut CFGBuilder<T>,
         const_pred: &ConstID,
         unit_const: &ConstID,
@@ -624,7 +624,7 @@ pub(crate) mod test {
         Ok((split, merge))
     }
 
-    fn build_then_else_merge_from_if<T: HugrMutRef>(
+    fn build_then_else_merge_from_if<T: AsMut<Hugr> + AsRef<Hugr>>(
         cfg: &mut CFGBuilder<T>,
         unit_const: &ConstID,
         split: BasicBlockID,
@@ -649,7 +649,7 @@ pub(crate) mod test {
     }
 
     // Returns loop tail - caller must link header to tail, and provide 0th successor of tail
-    fn build_loop_from_header<T: HugrMutRef>(
+    fn build_loop_from_header<T: AsMut<Hugr> + AsRef<Hugr>>(
         cfg: &mut CFGBuilder<T>,
         const_pred: &ConstID,
         header: BasicBlockID,
@@ -663,7 +663,7 @@ pub(crate) mod test {
     }
 
     // Result is header and tail. Caller must provide 0th successor of header (linking to tail), and 0th successor of tail.
-    fn build_loop<T: HugrMutRef>(
+    fn build_loop<T: AsMut<Hugr> + AsRef<Hugr>>(
         cfg: &mut CFGBuilder<T>,
         const_pred: &ConstID,
         unit_const: &ConstID,
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -2,7 +2,7 @@
 //!
 use thiserror::Error;
 
-use crate::hugr::{HugrError, HugrMut, Node, ValidationError, Wire};
+use crate::hugr::{Hugr, HugrError, HugrMut, Node, ValidationError, Wire};
 use crate::ops::handle::{BasicBlockID, CfgID, ConditionalID, DfgID, FuncID, TailLoopID};
 
 use crate::types::LinearType;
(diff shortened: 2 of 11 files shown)


src/builder/build_traits.rs
```diff
@@ -641,11 +640,11 @@ fn wire_up<T: Dataflow + ?Sized>(
 /// Return None if Const kind
 /// Panics if port not valid for Op or port is not Const/Value
 fn check_classical_value(
-    base: &HugrMut,
+    base: &impl AsRef<Hugr>,
```
acl-cqc, line 643: This could probably also be `&impl HugrMut`, but `AsRef<Hugr>` seems more general (? or is it)

acl-cqc: At present `HugrMut` has `AsRef<Hugr>` as a prereq, so that'd be strictly less general - some `AsRef<Hugr>` things don't have the HugrMut methods (but probably could have if we wanted). However, in the future I hope that HugrMut will not require being able to retrieve an actual Hugr (even a reference), rather, HugrMut will be able to return an immutable *view* onto what might be only part of a Hugr. So if we wanted to plan ahead for that, I should change this here to HugrMut, but that will be another PR(-series)...

acl-cqc:
@aborgna-q @ss2165 please co-ordinate who reviews this! :)

src/hugr/hugrmut.rs
```diff
@@ -221,38 +116,154 @@ impl HugrMut {
     ///
     ///  - If the sibling node does not have a parent.
     ///  - If the attachment would introduce a cycle.
-    pub fn add_op_after(
+    fn add_op_after(&mut self, sibling: Node, op: impl Into<OpType>) -> Result<Node, HugrError>;
+
+    /// Build the HUGR, returning an error if the graph is not valid.
+    fn finish(self) -> Result<Hugr, ValidationError>;
```
aborgna-q, line 122: This shouldn't be in the trait.
We can already dereference; finishing the hugr should be done by whatever builder struct you are using (or if you already have a `Hugr` just call `validate`).

src/builder.rs
```diff
@@ -70,26 +70,13 @@ pub enum BuildError {
...
-}
-
-/// Trait allowing treating type as (im)mutable reference to [`HugrMut`]
-pub trait HugrMutRef: AsMut<HugrMut> + AsRef<HugrMut> {}
-impl HugrMutRef for HugrMut {}
-impl HugrMutRef for &mut HugrMut {}
+/// Trait allowing treating type as (im)mutable reference to [`Hugr`]
+pub trait HugrMutRef: AsMut<Hugr> + AsRef<Hugr> {}
+impl<H: HugrMut> HugrMutRef for H {}
+impl HugrMutRef for &mut Hugr {}
```
aborgna-q, line 76: This can also be moved with the rest of the logic in `hugr.rs`.

aborgna-q, line 76: Can we implement `AsMut` and `AsRef` instead?

src/builder/build_traits.rs
```diff
@@ -37,9 +37,9 @@ use crate::hugr::HugrMut;
 pub trait Container {
     /// The container node.
     fn container_node(&self) -> Node;
-    /// The underlying [`HugrMut`] being used to build the HUGR.
-    fn base(&mut self) -> &mut HugrMut;
-    /// Immutable reference to HUGR being built.
+    /// The underlying [`Hugr`] being built...TODO: should we just require `AsMut<Hugr>`?
+    fn base(&mut self) -> &mut Hugr;
+    /// Immutable reference to HUGR being built...TODO: should we just require `AsRef<Hugr>`? Or combine with previous?
     fn hugr(&self) -> &Hugr;
```
aborgna-q, line 43: It's probably nicer for the borrow checker to keep the separate methods (I'd vote for renaming `base` to `hugr_mut` though).

Requiring `Container : AsMut<Hugr> + AsRef<Hugr>` may be confusing for the user, as it'd expose all the low-level mutation methods along with the builder ones.

aborgna-q (approved):
LGTM

I'd centralize the AsRef traits in the main hugr module.
Probably also dropping `DerefHugr` in `HugrView` in favour of `AsRef<Hugr>`.

src/builder/build_traits.rs
ss2165: agree with renaming, and yes, the separate method makes things simpler I believe.

src/hugr/hugrmut.rs
acl-cqc: Yes, I meant to call this out, sorry - it looked a bit dubious. Removing it entirely has meant that many of the `impl<T: HugrMut> Container for SomeBuilder<T>` have become `impl Container for SomeBuilder<Hugr>` which is to say, you can only call `.finish_hugr` when building a whole Hugr. Given that we are not doing mutable subregions yet, this is good :)

src/hugr/hugrmut.rs
```diff
@@ -221,38 +116,154 @@ impl HugrMut {
...
+    fn as_ref(&self) -> &Hugr {
+        self
+    }
+}
+
+impl AsMut<Hugr> for Hugr {
+    fn as_mut(&mut self) -> &mut Hugr {
+        self
+    }
+}
```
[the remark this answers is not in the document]
acl-cqc: Done

src/builder/build_traits.rs
acl-cqc: Yes, if you call `hugr_mut()` to get a mutable ref and then "upcast" to a immutable ref, the mutable borrow on the receiver of the `hugr_mut()` is presumably still live.

Also agreed that `Container: AsMut<Hugr> + AsRef<Hugr>` could be confusing.

So, renamed, and TODOs removed.

In time I guess these will return (im)mutable views onto subregions.

src/builder.rs
acl-cqc: At present HugrMutRef is a trait that one implements *on top of* AsMut and AsRef. AFAICS, the only purpose is to make it easy to parameterize (i.e. over `T: HugrMutRef` rather than `T: AsMut<Hugr> + AsRef<Hugr>`).

So, removed via bulk replace. (I have not gone through and checked whether any of them can be just `: AsMut<Hugr>` or `: AsRef<Hugr>` rather than `+` of both.)

This is a bit messy textually even if I think makes sense. (Sadly you can't define `type HugrMutRef = AsRef<Hugr> + AsMut<Hugr>` - `type` must really be a *type* not merely a *bound*.)

However, have kept in separate commit b48e2f6a7175edb68bf7e4b14b34aa72a0275adb in case you think this is too messy...

acl-cqc: HugrMutRef deleted entirely!

acl-cqc:
Somewhere there was a suggestion of dropping DerefHugr (in view.rs) in favour of AsRef<Hugr> - this might be good, but I think is separate to this PR.

src/builder.rs
acl-cqc: @aborgna-q can you have a quick look at that b48e2f6a7175edb68bf7e4b14b34aa72a0275adb in case you think better of it....

src/algorithm/nest_cfgs.rs
```diff
@@ -624,7 +624,7 @@ pub(crate) mod test {
         Ok((split, merge))
     }
 
-    fn build_then_else_merge_from_if<T: HugrMutRef>(
```
ss2165, line 627: do these actually need both Mut and Ref or is the Mut requirement sufficient?

ss2165 (approved):
👍 builder code base much nicer with this

src/algorithm/nest_cfgs.rs
acl-cqc: They need it because the `impl` block is also parameterized like that...as I say I think we could probably go through and replace quite a few `AsRef + AsMut`s with one or the other, but I've not done that thorough an analysis here; consistency is helpful in itself and we can reconsider later if we really want.

PR #135, opened 2023-06-08
Rename LICENCE -> LICENSE

croyzor:


aborgna-q (approved):
I guess it makes sense to use the _default_ spelling.

croyzor:
I think I had a bit of a brain fart here - I didn't realise that license vs licence is a dialect issue! I'm happy to leave as is

PR #139, opened 2023-06-08
Replace `linked_port` with `linked_ports` in HugrView

aborgna-q:
Since we now support multiple links to the same port

### DIFF
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -604,7 +604,7 @@ fn wire_up<T: Dataflow + ?Sized>(
         }
     }
 
-    if let Some((connected, connected_offset)) = base.hugr().linked_port(src, src_offset) {
+    if let Some((connected, connected_offset)) = base.hugr().linked_ports(src, src_offset).next() {
         if let Some(copy_port) = if_copy_add_port(base, src) {
             src_port = copy_port;
             src = connected;
--- a/src/hugr/view.rs
+++ b/src/hugr/view.rs
@@ -1,6 +1,7 @@
 #![allow(unused)]
 //! A Trait for "read-only" HUGRs.
 
+use std::iter::FusedIterator;
 use std::ops::Deref;
 
 use itertools::{Itertools, MapInto};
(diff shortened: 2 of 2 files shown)


src/hugr/view.rs
```diff
@@ -67,7 +64,12 @@ pub trait HugrView {
     fn all_node_ports(&self, node: Node) -> NodePorts;
 
     /// Return node and port connected to provided port, if not connected return None.
-    fn linked_port(&self, node: Node, port: Port) -> Option<(Node, Port)>;
+    fn linked_ports(&self, node: Node, port: Port) -> PortLinks<'_>;
```
cqc-alec, line 67: Fix docstring.

src/hugr/view.rs
```diff
@@ -227,3 +226,43 @@ where
...
+
+/// Iterator over the links of a port
+#[derive(Clone)]
+pub struct PortLinks<'a> {
+    hugr: &'a Hugr,
+    links: crate::hugr::multiportgraph::PortLinks<'a>,
+}
+
+impl<'a> Iterator for PortLinks<'a> {
+    type Item = (Node, Port);
```
cqc-alec, line 238: Wonder if this is worth a more general type definition (not in this PR though).

src/builder/build_traits.rs
```diff
@@ -604,7 +604,7 @@ fn wire_up<T: Dataflow + ?Sized>(
         }
     }
 
-    if let Some((connected, connected_offset)) = base.hugr().linked_port(src, src_offset) {
+    if let Some((connected, connected_offset)) = base.hugr().linked_ports(src, src_offset).next() {
```
cqc-alec, line 607: Should we be iterating over _all_ linked ports here?

cqc-alec (changes requested):

src/builder/build_traits.rs
aborgna-q: This is for the explicit-copies builder. This part of the code is directly dropped in the next PR #140

src/hugr/view.rs
aborgna-q: The iterator story is quite cumbersome. Hopefully we can clean it up in the future.

cqc-alec (approved):

PR #140, opened 2023-06-08
Drop the Copy operation type

### LINKED ISSUE #127: Drop explicit copies
Use multiports (#126) to implicitly copy non-classical values.

aborgna-q:
Closes #127
Depends on #139


ss2165 (approved):
🙏

PR #143, opened 2023-06-08
[refactor] Add resources to signatures

croyzor:
Resource requirements for a dataflow node can be represented by a pair of the resources requirements of the inputs and the resource requirements of the outputs (see the very bottom of the diff).

Inside of a dataflow node, the input requirements are uniformly associated with the wires coming out of the source (Input) node, and the output requirements are associated with the sinks in the Output node

In this PR:
* Add input & output resources to signatures
* Change function interfaces to deal with signatures when possible, so that resources aren't forgotten about
* Include resource requirements in Input/Output nodes, so that their signatures can be computed

### DIFF
--- a/src/builder/tail_loop.rs
+++ b/src/builder/tail_loop.rs
@@ -22,10 +22,7 @@ impl<B: HugrMutRef> TailLoopBuilder<B> {
         loop_node: Node,
         tail_loop: &ops::TailLoop,
     ) -> Result<Self, BuildError> {
-        let signature = Signature::new_df(
-            tail_loop.body_input_row(),
-            tail_loop.body_output_row(),
-        );
+        let signature = Signature::new_df(tail_loop.body_input_row(), tail_loop.body_output_row());
         let dfg_build = DFGBuilder::create_with_io(base, loop_node, signature)?;
 
         Ok(TailLoopBuilder::from_dfg_builder(dfg_build))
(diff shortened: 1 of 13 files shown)


src/ops/validate.rs
```diff
@@ -513,9 +514,11 @@ mod test {
 
         let input_node = OpType::Dataflow(DataflowOp::Input {
             types: in_types.clone(),
+            resources: ResourceSet::new(),
```
ss2165, line 517: on main `Input` and `Output` are now standalone structs, so you can define constructors for them to make this easier.
I suggest 
```rust 
fn new(types: TypeRow) -> Self;
fn with_resources(self, resources: ResourceSet) -> Self;
```
 So you can do `ops::Input::new(type_row![QB]).with_resources(...)`

croyzor: Done!

src/builder/tail_loop.rs
```diff
@@ -20,14 +20,13 @@ impl<B: HugrMutRef> TailLoopBuilder<B> {
     pub(super) fn create_with_io(
         base: B,
         loop_node: Node,
-        tail_loop: &ops::TailLoop,
+        tail_loop_sig: &ops::TailLoop,
```
ss2165, line 23: this shouldn't have been renamed

src/ops/dataflow.rs
```diff
@@ -26,25 +27,64 @@ pub(super) trait DataflowOpTrait {
     }
 }
 
+pub(crate) trait IOTrait {
```
ss2165, line 30: I think this can be `pub`, generally useful for construction

ss2165 (approved):

PR #144, opened 2023-06-09
Store function name in Def and Declare nodes

### LINKED ISSUE #136: Store function name in Def/Declare nodes
### Discussed in https://github.com/CQCL-DEV/hugr/discussions/72

<sup>Originally posted by **ss2165** May 19, 2023</sup>
@ss2165
16 May 2023

Linking happens by looking up functions by name - therefore it seems the name should not just be metadata (I don’t think linking should rely on metadata)

@ss2165
16 May 2023

I think this means that declare/alias_declare should hold names, and that these operations are only valid at module level. 

@acl-cqc
16 May 2023

I’d be happy with that. The alternative would be - the metadata may contain data that affects other tools (that operate on the Hugr), right? So we just have to say that the linker is an”other tool”….

@acl-cqc
16 May 2023

Also, we will need some kind of namespacing on these names, right?</div>

ss2165:
Closes #136

### DIFF
--- a/src/builder/module_builder.rs
+++ b/src/builder/module_builder.rs
@@ -73,18 +73,24 @@ impl<T: HugrMutRef> ModuleBuilder<T> {
         f_id: &FuncID<false>,
     ) -> Result<FunctionBuilder<&mut HugrMut>, BuildError> {
         let f_node = f_id.node();
-        let (inputs, outputs) =
-            if let OpType::Declare(ops::Declare { signature }) = self.hugr().get_optype(f_node) {
-                (signature.input.clone(), signature.output.clone())
-            } else {
-                return Err(BuildError::UnexpectedType {
-                    node: f_node,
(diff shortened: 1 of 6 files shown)


src/builder/module_builder.rs
```diff
@@ -117,11 +123,14 @@ impl<T: HugrMutRef> ModuleBuilder<T> {
     /// [`OpType::Declare`] node.
     pub fn declare(
         &mut self,
-        _name: impl Into<String>,
+        name: impl Into<String>,
         signature: Signature,
     ) -> Result<FuncID<false>, BuildError> {
-        // TODO add name and param names to metadata
-        let declare_n = self.add_child_op(ops::Declare { signature })?;
+        // TODO add param names to metadata
```
croyzor, line 129: nice

croyzor (approved):

PR #145, opened 2023-06-09
Update spec to require that I/O + Entry/Exit are first two children

ss2165:
As discussed in #121


cqc-alec:
Should we update the implementation at the same time, to keep them in sync?

ss2165:
> Should we update the implementation at the same time, to keep them in sync?

I'm working on it now

ss2165:
Implementation PR #148

PR #151, opened 2023-06-09
[multiportgraph] Fix stale data when resizing copy nodes

aborgna-q:
On some cases when a copy node was resized and their ports reallocated the MultiPortGraph was left with stale port metadata which didn't get cleaned when another (non-copy) node was allocated.

This took a while to track down :)

In this PR there's also a new helper method `pg_main_node` to be used in a future PR, and some cleanup of `remove_copy_node`.


ss2165 (approved):
👍

PR #152, opened 2023-06-09
Use multiports for State, output Const, and input Const edges

### LINKED ISSUE #130: Use a single multiport for _other edges_

aborgna-q:
With this we can now compute the exact number of ports required by an `OpType`.
Now the only part of the code that needs to set or modify the number of ports in an operation is `HugrMut::add_op` (and some tests doing `replace_op` shenigans).

Closes #130 #129
Depends on #156

### DIFF
--- a/src/ops.rs
+++ b/src/ops.rs
@@ -136,10 +136,9 @@ pub trait OpTrait {
         let signature = self.signature();
         let port = port.into();
         let dir = port.direction();
-        match port.index().cmp(&signature.port_count(dir)) {
-            std::cmp::Ordering::Less => signature.get(port),
-            std::cmp::Ordering::Equal => self.other_port(dir),
-            std::cmp::Ordering::Greater => None,
+        match port.index() < signature.port_count(dir) {
+            true => signature.get(port),
+            false => self.other_port(dir),
         }
     }
 }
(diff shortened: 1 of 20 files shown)


src/ops.rs
```diff
@@ -59,6 +59,65 @@ impl Default for OpType {
...
+impl OpType {
+    /// The edge kind for the non-dataflow or constant-input ports of the
+    /// operation, not described by the signature.
+    ///
+    /// If not None, a single extra multiport of that kind will be present on
+    /// the given direction.
+    pub fn other_port(&self, dir: Direction) -> Option<EdgeKind> {
+        match dir {
+            Direction::Incoming => self.other_input(),
+            Direction::Outgoing => self.other_output(),
```
ss2165, line 71: seems worth replacing `other_input` and `other_output` with a function parametrized over Direction too?

aborgna-q: Many `OpTrait` implementations only override the output, so putting the parametrized function in the trait would just copy the `match` there.

I'm ok with either option.

ss2165: fair enough, happy to leave as is

ss2165 (approved):

PR #153, opened 2023-06-11
Spec edits

cqc-alec:
Some minor clarifications and typo fixes in the "Functional description" section.

I'll raise Discussions for a couple of points I'm not sure about.

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -425,13 +425,9 @@ cycles. The common parent is a CFG-node.
 
 **Dataflow Sibling Graph (DSG)**: nodes are operations, `CFG`,
 `Conditional`, `TailLoop` and `DFG` nodes; edges are value and order and
-must be acyclic. The common parent may be a `Def`, `TailLoop`, `DFG`,
-`Case` or `BasicBlock` node.
-
-In a dataflow sibling graph, the edges (value and order considered
-together) must be acyclic. There is a unique Input node and Output node.
-All nodes must be reachable from the Input node, and must reach the
-Output node.
+must be acyclic. There is a unique Input node and Output node. All nodes must be
+reachable from the Input node, and must reach the Output node. The common parent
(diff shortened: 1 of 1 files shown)


specification/hugr.md
```diff
@@ -243,7 +251,7 @@ regions:
     `AliasDef`.
 
 
-A **loadable HUGR** is a module HUGR where all edges are connected and there are
+A **loadable HUGR** is a module HUGR where all ports are connected and there are
```
aborgna-q, line 254: This is not quite correct anymore. The multiports (state edge ports, output classical value ports) can remain disconnected.

We still have to specify which ports can have multiple/no connection, so that can come in another PR.

ss2165 (approved):
just an indexing preference :p

specification/hugr.md
```diff
@@ -369,7 +377,7 @@ children of a CFG-node.
     the entry block.
 
 The first output of the DSG contained in a `BasicBlock` has type
-`Predicate(#t0, #t1,...#tn)`, where the node has `N` successors, and the
+`Predicate(#t1,...#tn)`, where the node has `n` successors, and the
```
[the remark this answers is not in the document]
ss2165: ah yes that is the correct version of what I meant

specification/hugr.md
cqc-alec: OK yes, we should at least allow discarded classical values to be unconnected ports, for self-consistency. Perhaps "all output ports are connected" will suffice?

aborgna-q: "all input ports" except input state edges

cqc-alec: Sorry I missed this comment before merging. Do state (order) edges have ports (at the spec level I mean)?

aborgna-q: Good question. I guess we should define how to talk about the "other edge" ports. (see #159)

PR #166, opened 2023-06-13
Don't encode other_port offsets on serialization

### LINKED ISSUE #133: Serialize edge as pair of (node, option[port])
### Discussed in https://github.com/CQCL-DEV/hugr/discussions/97

<sup>Originally posted by **ss2165** May 30, 2023</sup>
In the serialization spec currently all edges are defined as pairs of (node, port). However in the main spec ports are only used for input/const_input/output

I propose instead to move to pairs of (node, Option[port]) such that control flow edges can be expressed as ((node, port), (node, null)) and order edges as ((node, null), (node, null)).

@mark-koch does that work for guppy?</div>

aborgna-q:
Closes #133

### DIFF
--- a/src/hugr/serialize.rs
+++ b/src/hugr/serialize.rs
@@ -4,12 +4,13 @@
 use std::collections::HashMap;
 use thiserror::Error;
 
+use crate::ops::OpTrait;
 use crate::{hugr::Hugr, ops::OpType};
 use portgraph::hierarchy::AttachError;
 use portgraph::{Direction, Hierarchy, LinkError, NodeIndex, UnmanagedDenseMap};
 use serde::{Deserialize, Deserializer, Serialize};
 
-use super::multiportgraph::{MultiPortGraph, SubportIndex};
+use super::multiportgraph::MultiPortGraph;
 
 /// A wrapper over the available HUGR serialization formats.
 ///
@@ -35,7 +36,7 @@ struct SerHugrV0 {
     /// For each node: (parent, num_inputs, num_outputs, node_operation)
     nodes: Vec<(NodeIndex, usize, usize, OpType)>,
     /// for each edge: (src, src_offset, tgt, tgt_offset)
-    edges: Vec<[(NodeIndex, usize); 2]>,
+    edges: Vec<[(NodeIndex, Option<u16>); 2]>,
     root: NodeIndex,
 }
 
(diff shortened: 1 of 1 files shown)


src/hugr/serialize.rs
```diff
@@ -35,7 +36,7 @@ struct SerHugrV0 {
     /// For each node: (parent, num_inputs, num_outputs, node_operation)
     nodes: Vec<(NodeIndex, usize, usize, OpType)>,
     /// for each edge: (src, src_offset, tgt, tgt_offset)
```
cqc-alec, line 38:
```suggestion
    /// for each edge: (src, src_subport_offset, tgt, tgt_subport_offset)
```
?

src/hugr/serialize.rs
```diff
@@ -114,23 +123,31 @@ impl TryFrom<&Hugr> for SerHugrV0 {
...
 
-        let find_offset = |p: SubportIndex| {
-            (
-                node_rekey[&graph.port_node(p).unwrap()],
-                graph.port_offset(p).unwrap().index(),
-            )
+        let find_offset = |node: NodeIndex, offset: usize, dir: Direction| {
+            let sig = &op_types[node].signature();
+            let offset = match offset < sig.port_count(dir) {
+                true => Some(offset as u16),
```
cqc-alec, line 129: This cast worries me a bit -- is there some inconsistency in integer types that we should iron out?

src/hugr/serialize.rs
```diff
@@ -158,7 +175,24 @@ impl TryFrom<SerHugrV0> for Hugr {
             op_types_sec[ni] = typ;
         }
 
+        let unwrap_offset = |node, offset, dir| -> Result<usize, Self::Error> {
+            let offset = match offset {
+                Some(offset) => offset as usize,
```
cqc-alec, line 180: (and here)

src/hugr/serialize.rs
aborgna-q: We don't encode the specific subports, just the port in the operation node.
A serialization-deserialization roundtrip could potentially reorder the subports without changing the interpretation of the graph (In practice that won't happen because we encode things sequentially).

src/hugr/serialize.rs
aborgna-q: Portgraph supports up to `u16::MAX` ports per node in each direction for performance reasons. Any PortOffset higher than 65k would panic during runtime.

(Note that that limit is _per node_, the graph itself can support up to `usize`)

cqc-alec (approved):

PR #167, opened 2023-06-13
Describe ports.

### LINKED ISSUE #164: Add precise description of ports to spec
Since the word is somewhat overloaded and the implementation now uses "multiports" as well, there is scope for confusion.

See #159 .

cqc-alec:
Closes #164 .

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -90,7 +90,13 @@ represent (typed) data or control dependencies.
 
 ## Functional description
 
-A HUGR is a directed graph with nodes and edges. The nodes represent
+A HUGR is a directed graph. There are several different types of node, and
+several different types of edge, with different semantics, described below.
+Nodes usually have additional data associated with them. Some edges have a
+"port" at each end (source and target), containing some additional data; others
+are simply arrows.
+
+The nodes represent
 processes that produce values - either statically, i.e. at compile time,
 or at runtime. Each node is uniquely identified by its **node index**,
 although this may not be stable under graph structure modifications.
@@ -138,6 +144,15 @@ EdgeKind ::= Hierarchy | Value(Locality, SimpleType) | Static(Locality, ClassicT
 
 Locality ::= Local | Ext | Dom
 ```
+
+Ports always exist on `Value` and `Static` edges; never on `Hierarchy`, `Order `
+or `ControlFlow` edges. Every port has an associated `SimpleType`. The source
+and target ports of an edge have the same `SimpleType`.
+
+A source port with a `ClassicType` may have any number of edges associated with
+it (including zero). A port with a `LinearType`, and a target port of any type,
+must have exactly one edge associated with it.
+
 #### Hierarchy
 
 A **Hierarchy** edge from node *a* to *b* encodes that *a* is the direct parent


specification/hugr.md
```diff
@@ -138,6 +144,15 @@ EdgeKind ::= Hierarchy | Value(Locality, SimpleType) | Static(Locality, ClassicT
 
 Locality ::= Local | Ext | Dom
 ```
+
+Ports always exist on `Value` and `Static` edges; never on `Hierarchy`, `Order `
+or `ControlFlow` edges. Every port has an associated `SimpleType`. The source
```
aborgna-q, line 149: What about the ordered side of `ControlFlow` edges? Don't we require ports there?

ss2165: We need an ordering there yes, but we may wish to use different language in the spec than "Ports". I'm in two minds tbh.

specification/hugr.md
```diff
@@ -138,6 +144,15 @@ EdgeKind ::= Hierarchy | Value(Locality, SimpleType) | Static(Locality, ClassicT
...
+and target ports of an edge have the same `SimpleType`.
+
+A source port with a `ClassicType` may have any number of edges associated with
+it (including zero). A port with a `LinearType`, and a target port of any type,
```
ss2165, line 153: maybe worth noting explicitly that this is the encoding of the "value used exactly once"

specification/hugr.md
cqc-alec: In two minds as well. It is only the order that matters so this seems different from ports as carriers of specific types. But indeed this ordering is not otherwise captured...

aborgna-q (approved):
LGTM

specification/hugr.md
```diff
@@ -90,7 +90,20 @@ represent (typed) data or control dependencies.
 
 ## Functional description
 
-A HUGR is a directed graph with nodes and edges. The nodes represent
+A HUGR is a directed graph. There are several different types of node, and
+several different types of edge, with different semantics, described below.
+
+Nodes usually have additional data associated with them.
+
+Edge types are of three kinds:
```
ss2165, line 98: I think it is confusing to use the term "kind" confusing here since we use `EdgeKind` later. I suggest "Connectivity". 

I also don't think this is a good place to discuss this, it would work better under Edges? I think I would introduce the concept of connectivity and immediately state the connectivity of each kind of edge, at the bottom of the Edges section (after describing the kinds)

ss2165: Yeah I'm not sure about that simple example, nearly removed it last time, I guess it works as a "soon you will understand what all of this means!" teaser.

"Category" works, but given we don't need to talk about it again, I think it's just simpler to add a section on "Ports" under "Edges" and describe the port/ordering constraints of each edge kind there.

cqc-alec: Have created a subsection "Ports and ordering of edges" and rejigged the headings below.

specification/hugr.md
```diff
@@ -153,7 +185,7 @@ A **sibling graph** is a subgraph of the HUGR containing all nodes with
 a particular parent, plus the Order, Value and ControlFlow edges between
 them.
 
-#### Value
+##### Value
```
ss2165, line 188: heading level 5 results in text that is even smaller than the body on github rendering, I think either move it back up a level or just use Bold for these Kind names

cqc-alec: OK. (Using excessive levels of heading is a bad habit of mine.)

ss2165 (approved):
👍

PR #171, opened 2023-06-13
Update spec for `BasicBlock` subytypes.

### LINKED ISSUE #169: Make all CFG children `BasicBlock`, of two subtypes
### Discussed in https://github.com/CQCL-DEV/hugr/discussions/155

<sup>Originally posted by **cqc-alec** June 11, 2023</sup>
In the current implementation, there are two kinds of `BasicBlock`: `Block` and `Exit`. In the spec, `BasicBlock` and `ExitBlock` are distinguished:

> All children except the second are BasicBlock-nodes, the first is the entry block. The second child is an ExitBlock node

Which is the more natural approach?</div>

cqc-alec:
Fixes #169 .

Implementation update to follow.


ss2165 (approved):
👍

PR #178, opened 2023-06-15
Even more ergonomic serialization

### LINKED ISSUE #177: Remove in/out port counts from serialized format
### Discussed in https://github.com/CQCL-DEV/hugr/discussions/159

<sup>Originally posted by **aborgna-q** June 12, 2023</sup>
With multiports we can now define the exact number of ports each operation will have. There's a number of issues that are not currently specified in the spec:

- We should clarify the order of dataflow/constant/other edges (#23)
- We should specify which ports must have exactly one connection vs which can have any number of edges (including none).

In a related note, having a fixed number of ports means that the port counts in the node serialization are redundant. Should we drop them from the serialization specification?</div>

ss2165:
* remove port counts
* use hugr interfaces for serialization
* proptest does not produce valid hugrs so remove
* Encode types more simply
* store parent along with op type
* use internal tagging for clarity

Closes #177 

Before:
```json
{
    "nodes": [
        [
            0,
            0,
            0,
            {
                "Module": []
            }
        ],
        [
            0,
            0,
            1,
            {
                "Def": {
                    "name": "main",
                    "signature": {
                        "input": [
                            {
                                "l": {
                                    "Container": {
                                        "Sum": [
                                            {
                                                "c": {
                                                    "Int": 64
                                                }
                                            },
                                            {
                                                "l": "Qubit"
                                            }
                                        ]
                                    }
                                }
                            }
                        ],
                        "output": [
                            {
                                "l": {
                                    "Container": {
                                        "Sum": [
                                            {
                                                "c": {
                                                    "Int": 64
                                                }
                                            },
                                            {
                                                "l": "Qubit"
                                            }
                                        ]
                                    }
                                }
                            }
                        ],
                        "const_input": [],
                        "input_resources": [],
                        "output_resources": []
                    }
                }
            }
        ],
        [
            1,
            0,
            2,
            {
                "Input": {
                    "types": [
                        {
                            "l": {
                                "Container": {
                                    "Sum": [
                                        {
                                            "c": {
                                                "Int": 64
                                            }
                                        },
                                        {
                                            "l": "Qubit"
                                        }
                                    ]
                                }
                            }
                        }
                    ],
                    "resources": []
                }
            }
        ],
        [
            1,
            2,
            0,
            {
                "Output": {
                    "types": [
                        {
                            "l": {
                                "Container": {
                                    "Sum": [
                                        {
                                            "c": {
                                                "Int": 64
                                            }
                                        },
                                        {
                                            "l": "Qubit"
                                        }
                                    ]
                                }
                            }
                        }
                    ],
                    "resources": []
                }
            }
        ],
        [
            1,
            2,
            2,
            {
                "LeafOp": {
                    "Noop": {
                        "l": {
                            "Container": {
                                "Sum": [
                                    {
                                        "c": {
                                            "Int": 64
                                        }
                                    },
                                    {
                                        "l": "Qubit"
                                    }
                                ]
                            }
                        }
                    }
                }
            }
        ]
    ],
    "edges": [
        [
            [
                2,
                0
            ],
            [
                4,
                0
            ]
        ],
        [
            [
                4,
                0
            ],
            [
                3,
                0
            ]
        ]
    ],
    "root": 0
}
```

After:
```json
{
  "nodes":[
    {
      "parent":0,
      "op":"Module"
    },
    {
      "parent":0,
      "op":"Def",
      "name":"main",
      "signature":{
        "input":[
          {
            "t":"Sum",
            "row":[
              {
                "t":"I",
                "width":64
              },
              {
                "t":"Q"
              }
            ],
            "l":true
          }
        ],
        "output":[
          {
            "t":"Sum",
            "row":[
              {
                "t":"I",
                "width":64
              },
              {
                "t":"Q"
              }
            ],
            "l":true
          }
        ],
        "const_input":[
          
        ],
        "input_resources":[
          
        ],
        "output_resources":[
          
        ]
      }
    },
    {
      "parent":1,
      "op":"Input",
      "types":[
        {
          "t":"Sum",
          "row":[
            {
              "t":"I",
              "width":64
            },
            {
              "t":"Q"
            }
          ],
          "l":true
        }
      ],
      "resources":[
        
      ]
    },
    {
      "parent":1,
      "op":"Output",
      "types":[
        {
          "t":"Sum",
          "row":[
            {
              "t":"I",
              "width":64
            },
            {
              "t":"Q"
            }
          ],
          "l":true
        }
      ],
      "resources":[
        
      ]
    },
    {
      "parent":1,
      "op":"LeafOp",
      "lop":"Noop",
      "ty":{
        "t":"Sum",
        "row":[
          {
            "t":"I",
            "width":64
          },
          {
            "t":"Q"
          }
        ],
        "l":true
      }
    }
  ],
  "edges":[
    [
      [
        2,
        0
      ],
      [
        4,
        0
      ]
    ],
    [
      [
        4,
        0
      ],
      [
        3,
        0
      ]
    ]
  ]
}
```

### DIFF
--- a/src/hugr/serialize.rs
+++ b/src/hugr/serialize.rs
@@ -288,7 +288,9 @@ pub mod test {
                 .map(|in_wire| {
                     f_build
                         .add_dataflow_op(
-                            LeafOp::Noop(f_build.get_wire_type(in_wire).unwrap()),
+                            LeafOp::Noop {
+                                ty: f_build.get_wire_type(in_wire).unwrap(),
+                            },
                             [in_wire],
                         )
                         .unwrap()
@@ -301,7 +303,6 @@ pub mod test {
         };
 
         let ser_hugr: SerHugrV0 = (&hugr).try_into().unwrap();
-
         // HUGR internal structures are not preserved across serialization, so
         // test equality on SerHugrV0 instead.
         assert_eq!(ser_roundtrip(&ser_hugr), ser_hugr);
--- a/src/hugr/serialize.rs
+++ b/src/hugr/serialize.rs
@@ -5,13 +5,17 @@ use std::collections::HashMap;
 use thiserror::Error;
 
 use crate::ops::OpTrait;
-use crate::{hugr::Hugr, ops::OpType};
-use portgraph::hierarchy::AttachError;
-use portgraph::multiportgraph::MultiPortGraph;
-use portgraph::{
-    Direction, Hierarchy, LinkError, LinkView, NodeIndex, PortView, UnmanagedDenseMap,
+use crate::Node;
+use crate::{
+    hugr::{Hugr, HugrMut},
+    ops::OpType,
 };
+use portgraph::hierarchy::AttachError;
+
+use portgraph::{Direction, LinkError, NodeIndex};
 use serde::{Deserialize, Deserializer, Serialize};
+
+use super::{HugrError, HugrView};
 
 /// A wrapper over the available HUGR serialization formats.
 ///
@@ -35,10 +39,9 @@ enum Versioned {
 #[derive(Serialize, Deserialize, PartialEq, Debug)]
 struct SerHugrV0 {
     /// For each node: (parent, num_inputs, num_outputs, node_operation)
-    nodes: Vec<(NodeIndex, usize, usize, OpType)>,
+    nodes: Vec<(Node, OpType)>,
     /// for each edge: (src, src_offset, tgt, tgt_offset)
-    edges: Vec<[(NodeIndex, Option<u16>); 2]>,
-    root: NodeIndex,
+    edges: Vec<[(Node, Option<u16>); 2]>,
 }
 
 /// Errors that can occur while serializing a HUGR.
@@ -54,10 +57,13 @@ pub enum HUGRSerializationError {
     #[error("Cannot connect an edge without port offset to node {node:?} with operation type {op_type:?}.")]
     MissingPortOffset {
         /// The node that has the port without offset.
-        node: NodeIndex,
+        node: Node,
         /// The operation type of the node.
         op_type: OpType,
     },
+    /// Error building HUGR.
+    #[error("HugrError: {0:?}")]
+    HugrError(#[from] HugrError),
 }
 
 impl Serialize for Hugr {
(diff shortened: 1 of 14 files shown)


src/hugr/serialize.rs
```diff
@@ -89,120 +95,99 @@ impl<'de> Deserialize<'de> for Hugr {
...
     }
 }
 
 impl TryFrom<SerHugrV0> for Hugr {
     type Error = HUGRSerializationError;
-    fn try_from(SerHugrV0 { nodes, edges, root }: SerHugrV0) -> Result<Self, Self::Error> {
-        let mut hierarchy = Hierarchy::new();
-
+    fn try_from(SerHugrV0 { nodes, edges }: SerHugrV0) -> Result<Self, Self::Error> {
+        // Root must be first node
```
aborgna-q, line 157: We don't explicitly require this anywhere else. We should either:

- State this as a property of the Hugr, checked during validation.
- Allow `root != 0`, and could logically swap `NodeIndex(0)` and `root` during the encoding

src/hugr/serialize.rs
```diff
@@ -35,10 +39,9 @@ enum Versioned {
 #[derive(Serialize, Deserialize, PartialEq, Debug)]
 struct SerHugrV0 {
     /// For each node: (parent, num_inputs, num_outputs, node_operation)
```
aborgna-q, line 41: Update docs

src/hugr/serialize.rs
```diff
@@ -89,120 +95,99 @@ impl<'de> Deserialize<'de> for Hugr {
...
         // an underestimate
-        let mut graph = MultiPortGraph::with_capacity(nodes.len(), edges.len() * 2);
-        let mut op_types_sec = UnmanagedDenseMap::with_capacity(nodes.len());
-        for (parent, incoming, outgoing, typ) in nodes {
-            let ni = graph.add_node(incoming, outgoing);
-            if parent != ni {
-                hierarchy.push_child(ni, parent)?;
-            }
-            op_types_sec[ni] = typ;
+        let mut hugr = Hugr::with_capacity(nodes.next().unwrap().1, nodes.len(), edges.len() * 2);
```
aborgna-q, line 161: There could be a check for
```suggestion
        let root = nodes.next().unwrap();
        assert_eq!(root.0, NodeIndex::new(0).into());
        let mut hugr = Hugr::with_capacity(root.1, nodes.len(), edges.len() * 2);
```
(Or a new `HUGRSerializationError`)

aborgna-q (commented):
Nice!

---

Are we optimizing for uncompressed size? If not we could do a sightly friendlier schema using internal tagging

> ```json
>"input": [
>     {
>         "Sum": [
>             [
>                 {
>                     "I": 64
>                 },
>                 "Q"
>             ], true
>         ]
>     }
> ],
> ```

=>

```json
"input": [
    {   "op": "Sum",
        "v": [
            {
                "op": "I",
                "v":  64
            },
            "Q"
        ],
        "l": true
    }
],
```

(or something similar)

src/hugr/serialize.rs
ss2165: option 2: this would mean root index isn't preserved, but that's ok because indices get messed about anyway? 
I think option 2 sounds good

ss2165: done

aborgna-q (approved):
Nice!

PR #179, opened 2023-06-16
Rewrite Replace specification

cqc-alec:
This is a substantial rewrite.

The old version wasn't consistent with having multiple outputs from ports.

We may need to design an interface to allow the user to construct a "partial hugr".

There are definitely footguns here for the user and we could do variable amounts of validation before (and/or while) applying the replacement.

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1168,44 +1168,78 @@ The new hugr is then derived as follows:
 
 ###### `Replace`
 
-This takes as input:
+This is the general subgraph-replacement method.
 
-  - a set S of IDs of nodes in Γ, all of which are separated;
+A _partial hugr_ is is a graph $G$ having the same node and edge types as a hugr
+and satisfying all the same local constraints except that it may be missing edges.
 
-  - a partial hugr G;
+In particular:
 
-  - a map T from top(G) to IDs of container nodes in Γ\\S\*;
+  - it has no root node (the set of nodes without a parent is
+    denoted $\mathrm{top}(G)$ );
+  - it may have empty container nodes (the set of these is
+    denoted $\mathrm{bot}(G)$ ).
 
-  - a map B from bot(G) to IDs of container nodes in S\*, such that B(x)
-    is separated from B(y) unless x == y. Let X be the set of children
-    of values in B, and R be S\*\\X\*.
+A partial hugr describes a set of nodes and well-formed edges between
+them that potentially forms a subgraph of a hugr.
 
-  - a bijection μ<sub>inp</sub> between inp(G) and the set of input
-    ports of nodes in R whose source is not in R;
+Given a set $S$ of nodes in a hugr, let $S^\*$ be the set of all nodes
+descended from nodes in $S$ (i.e. reachable from $S$ by following hierarchy edges),
+including $S$ itself.
 
-  - a bijection μ<sub>out</sub> between out(G) and the set of output
-    ports of nodes in R whose target is not in R;
+Call two nodes $a, b \in \Gamma$ _separated_ if $a \notin \\{b\\}^\*$ and
+$b \notin \\{a\\}^\*$ (i.e. there is no hierarchy relation between them).
 
-  - disjoint subsets Init and Term of top(G);
+We define two structures to specify edge insertion.
 
-The new hugr is then derived by:
+A `NewInpEdgeSpec` specifies an edge inserted from an existing node to a (given) new node.
+It contains the following fields:
 
-1.  adding the new nodes from G;
+  - `EdgeKind`: may be `Value`, `Order`, `Static` or `ControlFlow`.
+  - `SrcNode`: the source node of the new edge.
+  - `SrcPort`: (for `Value` and `Static` edges only) the source port.
+  - `ToPos`: (for `Value` and `Static` edges only) the desired position among
+    the incoming ports to the new node.
 
-2.  connecting the ports according to the bijections μ<sub>inp</sub> and
-    μ<sub>out</sub>;
+A `NewOutEdgeSpec` specifies an edge inserted from a new node to an existing node.
+It contains the following fields:
 
-3.  for each node n in top(G), adding a hierarchy edge from t(n) to n,
-    placing n in the first position among children of t(n) if n is in
-    Init and in the second position if n is in Term;
+  - `EdgeKind`: may be `Value`, `Order`, `Static` or `ControlFlow`.
+  - `TgtNode`: the target node of the new edge.
+  - `TgtPort`: (for `Value` and `Static` edges only) the target port (whose
+    existing incoming edge will be removed).
+  - `FromPos` (for `Value`, `Static` and `ControlFlow` edges only) the desired
+    position among the outgoing ports from the new node.
 
-4.  for each node n in bot(G), and for each child m of b(n), adding a
-    hierarchy edge from n to m (replacing m’s existing parent edge)
+The `Replace` method takes as input:
 
-5.  removing all nodes in R
+  - a set $S$ of mutually-separated nodes in $\Gamma$;
+  - a partial hugr $G$;
+  - a map $T : \mathrm{top}(G) \to \Gamma \setminus S^*$ whose image consists of container nodes;
+  - a map $B : \mathrm{bot}(G) \to S^\*$ whose image consists of container nodes, such that $B(x)$
+    is separated from $B(y)$ unless $x = y$. Let $X$ be the set of children
+    of nodes $B(y)$ for $y \in \mathrm{bot}(G)$, and $R = S^\* \setminus X^\*$.
+  - a list $\mu\_\textrm{inp}$ of pairs $(n, \sigma\_\mathrm{inp})$ where $n$ is a node in $G$ and $\sigma\_\mathrm{inp}$ is a `NewInpEdgeSpec`;
+  - a list $\mu\_\textrm{out}$ of pairs $(n, \sigma\_\mathrm{out})$ where $n$ is a node in $G$ and $\sigma\_\mathrm{out}$ is a `NewOutEdgeSpec`.
 
-6.  If any edges inserted in step 2 are inter-graph (i.e DFG.
-    non-sibling), inserting any `Order` edges required to validate them.
+The new hugr is then derived as follows:
+
+1.  Make a copy in $\Gamma$ of all the nodes in $G$, and all edges between them.
+2.  For each $(n, \sigma\_\mathrm{inp}) \in \mu\_\textrm{inp}$, insert a new edge going into the new
+    copy of $n$ according to the specification $\sigma\_\mathrm{inp}$. (Note: some
+    bookkeeping will be required to ensure the eventual input port orderings of newly added nodes
+    are correct.)
+3.  For each $(n, \sigma\_\mathrm{out}) \in \mu\_\textrm{out}$, insert a new edge going out of the new
+    copy of $n$ according to the specification $\sigma\_\mathrm{out}$. (Note: it should
+    be verified that the predecessor node of the existing target port is in $R$,
+    and so will be removed at step 6.)
+4.  For each $(n, t = T(n))$, append the copy of $n$ to the list
+    of children of $t$ (adding a hierachy edge from $t$ to $n$).
+5.  For each node $(n, b = B(n))$ and for each child $m$ of $b$, replace $m$ with
+    the new copy of $n$ in the list of children of $b$ (replacing the existing hierarchy edge).
+6.  Remove all nodes in $R$ and edges adjoining them.
+7.  If any edges inserted in step 2 are inter-graph (i.e. DFG,
+    non-sibling), insert any `Order` edges required to validate them.
 
 ##### Outlining methods
 


specification/hugr.md
```diff
@@ -1168,44 +1168,78 @@ The new hugr is then derived as follows:
 
 ###### `Replace`
 
-This takes as input:
+This is the general subgraph-replacement method.
 
-  - a set S of IDs of nodes in Γ, all of which are separated;
+A _partial hugr_ is is a graph $G$ having the same node and edge types as a hugr
+and satisfying all the same local constraints except that it may be missing edges.
```
acl-cqc, line 1174: Are "local constraints" defined somewhere? I mean, I get the general idea, but wondering if there is room for ambiguity. Or is what follows the precise definition of which constraints these do include?
(What follows certainly changed my initial reading - "empty container nodes" - so does "local constraints" mean those referring to value/order/static/controlflow edges?)

acl-cqc: Rather than "it may be missing edges" consider "some value (/controlflow??) ports may not have edges" (to/from/linked to/from/etc.)

specification/hugr.md
```diff
@@ -1168,44 +1168,78 @@ The new hugr is then derived as follows:
...
+We define two structures to specify edge insertion.
 
-The new hugr is then derived by:
+A `NewInpEdgeSpec` specifies an edge inserted from an existing node to a (given) new node.
+It contains the following fields:
 
-1.  adding the new nodes from G;
+  - `EdgeKind`: may be `Value`, `Order`, `Static` or `ControlFlow`.
+  - `SrcNode`: the source node of the new edge.
+  - `SrcPort`: (for `Value` and `Static` edges only) the source port.
```
acl-cqc, line 1200: Should we include this for control-flow also (as *which successor* matters wrt. unpacking of the "sum" type determining which way to branch)?

Or, is it possible that we do not need this operation to apply to CFGs? (A quick consideration suggests that we probably do, but I haven't thought hard about it)

Or perhaps this InpEdgeSpec cannot be a ControlFlow after all? Hmmm, it seems like we *do* need to specify what happens to each incoming CF edge, as simply removing a CF edge from the hugr will (generally speaking) invalidate it

specification/hugr.md
```diff
@@ -1168,44 +1168,78 @@ The new hugr is then derived as follows:
...
-    μ<sub>out</sub>;
+A `NewOutEdgeSpec` specifies an edge inserted from a new node to an existing node.
+It contains the following fields:
 
-3.  for each node n in top(G), adding a hierarchy edge from t(n) to n,
-    placing n in the first position among children of t(n) if n is in
-    Init and in the second position if n is in Term;
+  - `EdgeKind`: may be `Value`, `Order`, `Static` or `ControlFlow`.
+  - `TgtNode`: the target node of the new edge.
+  - `TgtPort`: (for `Value` and `Static` edges only) the target port (whose
```
acl-cqc, line 1209: good qualification (about removal), that helps :)

specification/hugr.md
```diff
@@ -1168,44 +1168,78 @@ The new hugr is then derived as follows:
 
 ###### `Replace`
 
-This takes as input:
+This is the general subgraph-replacement method.
 
-  - a set S of IDs of nodes in Γ, all of which are separated;
+A _partial hugr_ is is a graph $G$ having the same node and edge types as a hugr
```
acl-cqc, line 1173: Ah hang on (writing this after some of the comments below) - this is the second definition of 'partial hugr'. There's another one just *before* the API Methods section. That one's slightly more precise, but is it correct?

specification/hugr.md
```diff
@@ -1168,44 +1168,78 @@ The new hugr is then derived as follows:
...
 
-4.  for each node n in bot(G), and for each child m of b(n), adding a
-    hierarchy edge from n to m (replacing m’s existing parent edge)
+The `Replace` method takes as input:
 
-5.  removing all nodes in R
+  - a set $S$ of mutually-separated nodes in $\Gamma$;
+  - a partial hugr $G$;
+  - a map $T : \mathrm{top}(G) \to \Gamma \setminus S^*$ whose image consists of container nodes;
+  - a map $B : \mathrm{bot}(G) \to S^\*$ whose image consists of container nodes, such that $B(x)$
```
acl-cqc, line 1219: Note that AFAICS `bot` and `top` are defined in the *first* version of partial hugr, [here](https://github.com/CQCL-DEV/hugr/blob/96cac0e9064d7ec3ee98a351330bc4577c05dde2/specification/hugr.md?plain=1#L1111)

specification/hugr.md
```diff
@@ -1168,44 +1168,78 @@ The new hugr is then derived as follows:
...
+3.  For each $(n, \sigma\_\mathrm{out}) \in \mu\_\textrm{out}$, insert a new edge going out of the new
+    copy of $n$ according to the specification $\sigma\_\mathrm{out}$. (Note: it should
+    be verified that the predecessor node of the existing target port is in $R$,
+    and so will be removed at step 6.)
+4.  For each $(n, t = T(n))$, append the copy of $n$ to the list
+    of children of $t$ (adding a hierachy edge from $t$ to $n$).
+5.  For each node $(n, b = B(n))$ and for each child $m$ of $b$, replace $m$ with
+    the new copy of $n$ in the list of children of $b$ (replacing the existing hierarchy edge).
+6.  Remove all nodes in $R$ and edges adjoining them.
+7.  If any edges inserted in step 2 are inter-graph (i.e. DFG,
```
acl-cqc, line 1241: "inter-graph" -> "value/static edges with locality `ext`"
But, I wonder if we could instead require these to be included in the `\mu_\textrm{in}` / `\mu_\textrm{out}`?

specification/hugr.md
acl-cqc: Ok, I go and read the spec ;). "Ports" are used only for Value/Static, whereupon we need both `SrcPort` and `DestPort`.

For ControlFlow, we need an integer position at the source, but not the target. For order, we don't need any position at either end.

So
* the source exists for Value, Static and Control Flow; it is a port for Value/Static, but an integer position for Controlflow. That might well be an int in all 3 cases.
* the dest exists for Value and Static only - as you write. But given we have "the source port" why not have "the dest port" similarly rather than "the desired position among the incoming ports").

specification/hugr.md
```diff
@@ -1168,44 +1168,78 @@ The new hugr is then derived as follows:
...
+It contains the following fields:
 
-3.  for each node n in top(G), adding a hierarchy edge from t(n) to n,
-    placing n in the first position among children of t(n) if n is in
-    Init and in the second position if n is in Term;
+  - `EdgeKind`: may be `Value`, `Order`, `Static` or `ControlFlow`.
+  - `TgtNode`: the target node of the new edge.
+  - `TgtPort`: (for `Value` and `Static` edges only) the target port (whose
+    existing incoming edge will be removed).
+  - `FromPos` (for `Value`, `Static` and `ControlFlow` edges only) the desired
```
acl-cqc, line 1211: Similar to above. I wonder whether the best option might be to combine the two EdgeSpec structs by removing SrcNode/TgtNode (and then including those in the `\mu_\mathrm{in}` / `\mu_\mathrm{out}` below)

specification/hugr.md
```diff
@@ -1168,44 +1168,78 @@ The new hugr is then derived as follows:
...
+    of nodes $B(y)$ for $y \in \mathrm{bot}(G)$, and $R = S^\* \setminus X^\*$.
+  - a list $\mu\_\textrm{inp}$ of pairs $(n, \sigma\_\mathrm{inp})$ where $n$ is a node in $G$ and $\sigma\_\mathrm{inp}$ is a `NewInpEdgeSpec`;
+  - a list $\mu\_\textrm{out}$ of pairs $(n, \sigma\_\mathrm{out})$ where $n$ is a node in $G$ and $\sigma\_\mathrm{out}$ is a `NewOutEdgeSpec`.
 
-6.  If any edges inserted in step 2 are inter-graph (i.e DFG.
-    non-sibling), inserting any `Order` edges required to validate them.
+The new hugr is then derived as follows:
+
+1.  Make a copy in $\Gamma$ of all the nodes in $G$, and all edges between them.
+2.  For each $(n, \sigma\_\mathrm{inp}) \in \mu\_\textrm{inp}$, insert a new edge going into the new
```
acl-cqc, line 1228: I wonder if it'd be easier for the spec (not the implementation, but the impl could do something differently) to say, "where these edges are from ports that currently have edges to nodes in R, the existing edges are replaced". Otherwise we are talking about some rather weird intermediates where the number of edges etc. don't line up. (That is to say - the spec doesn't have to talk about how we get from the start to the end, just what the end looks like)

acl-cqc (commented):
Good stuff - I like what you've done with input/output lists of edges rather than maps, also the hierarchy system (looks both irreducible and complete). I've a few ideas ;-) for the presentation, esp. the edge spec's and I think we should have at most one definition of "partial hugr"...

specification/hugr.md
cqc-alec: Sorry, I forgot to delete that first one. Agree this second needs making more precise.

specification/hugr.md
cqc-alec: Yes I think we can, that's a good idea.

specification/hugr.md
cqc-alec: Good idea.

acl-cqc (commented):
Thanks Alec - I think I've found a few more local tweaks but generally this looks good :)

specification/hugr.md
```diff
@@ -1170,44 +1160,71 @@ The new hugr is then derived as follows:
...
-    Init and in the second position if n is in Term;
+The `Replace` method takes as input:
 
-4.  for each node n in bot(G), and for each child m of b(n), adding a
-    hierarchy edge from n to m (replacing m’s existing parent edge)
+  - a set $S$ of mutually-separated nodes in $\Gamma$;
+  - a partial hugr $G$;
+  - a map $T : \top(G) \to \Gamma \setminus S^*$ whose image consists of container nodes;
+  - a map $B : \bot(G) \to S^\*$ whose image consists of container nodes, such that $B(x)$
+    is separated from $B(y)$ unless $x = y$. Let $X$ be the set of children
```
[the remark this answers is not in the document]
cqc-alec: I assume you mean "children of nodes that are values of B". However I wonder if this use of "values" may be a bit specialized? I think I prefer the current form.

specification/hugr.md
```diff
@@ -1170,44 +1160,71 @@ The new hugr is then derived as follows:
...
-    placing n in the first position among children of t(n) if n is in
-    Init and in the second position if n is in Term;
+The `Replace` method takes as input:
 
-4.  for each node n in bot(G), and for each child m of b(n), adding a
-    hierarchy edge from n to m (replacing m’s existing parent edge)
+  - a set $S$ of mutually-separated nodes in $\Gamma$;
+  - a partial hugr $G$;
+  - a map $T : \top(G) \to \Gamma \setminus S^*$ whose image consists of container nodes;
+  - a map $B : \bot(G) \to S^\*$ whose image consists of container nodes, such that $B(x)$
```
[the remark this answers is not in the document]
cqc-alec: Yes they are, but we don't need to say it here because it's part of the definition of bot(G).

Wikipedia says they are synonyms but that "range" should be avoided because it's also sometimes used to mean the codomain (so they'd only be the same for surjective functions).

specification/hugr.md
```diff
@@ -1170,44 +1160,71 @@ The new hugr is then derived as follows:
...
+    copy of the `TgtNode` of $\sigma\_\mathrm{inp}$ according to the specification $\sigma\_\mathrm{inp}$.
+    Where these edges are from ports that currently have edges to nodes in $R$,
+    the existing edges are replaced.
+3.  For each $\sigma\_\mathrm{out} \in \mu\_\textrm{out}$, insert a new edge going out of the new
+    copy of the `SrcNode` of $\sigma\_\mathrm{out}$ according to the specification $\sigma\_\mathrm{out}$.
+    The predecessor node of the existing target port must be in $R$; the
+    edge from it is removed.
+4.  For each $(n, t = T(n))$, append the copy of $n$ to the list
+    of children of $t$ (adding a hierachy edge from $t$ to $n$).
+5.  For each node $(n, b = B(n))$ and for each child $m$ of $b$, replace $m$ with
```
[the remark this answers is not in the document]
cqc-alec: Yeah, good point, I think this is a mistake: we are removing b, but not m...

specification/hugr.md
```diff
@@ -127,14 +127,39 @@ as *control* and the second as *target*.
...
+- `Hierarchy` edges express the relationship between container nodes and their
+  children.
+
+`Value` and `Static` edges are sometimes referred to as _dataflow_ edges.
+A `Value` edge can carry data of any `SimpleType`: either a `ClassicType`
+(ordinary classical data) or a `LinearType` (data which cannot be copied,
+including quantum data). A `Static` edge can only carry a `ClassicType`. For
+more details see the [Type System](#type-system) section.
+
+A `Value` edge can carry any `SimpleType`; a `Static` edge can only carry a
```
acl-cqc, line 151: Ok, you can remove the first sentence here now (it repeats the previous paragraph almost exactly)

specification/hugr.md
```diff
@@ -1170,44 +1159,71 @@ The new hugr is then derived as follows:
...
-  - a set S of IDs of nodes in Γ, all of which are separated;
+This is the general subgraph-replacement method.
 
-  - a partial hugr G;
+A _partial hugr_ is a graph formed by a subset of nodes of a valid hugr together
+with a subset of their adjoining edges. It must not include a `Module` node.
 
-  - a map T from top(G) to IDs of container nodes in Γ\\S\*;
+Informally, a partial hugr is is a graph $G$ having the same node and edge types
+as a hugr but that may be missing some edges or nodes.
```
acl-cqc, line 1168: nit: `or' -> `and/or`

acl-cqc: In fact I'm not sure this sentence adds anything to the previous. Although maybe reiterating that helps for it to sink in ;)

specification/hugr.md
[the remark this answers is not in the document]
acl-cqc: ...."children of nodes in the image of B", perhaps?

acl-cqc (approved):
Looks good to me, thanks Alec - a few more ideas but I'm not sure any of them are better!

PR #180, opened 2023-06-19
Rename "Inter-graph edges" to "non-local edges"

acl-cqc:
And try to clarify a bit.

Hierarchy edges don't fit that neatly (they don't fit any of the localities), ideas welcome - do we need to write "non-hierarchy edges"? I'm hoping this here is fine...

Also add a glossary entry for "dataflow edge".


ss2165 (approved):
👍

PR #181, opened 2023-06-19
[new] Add simple resource checking of edges to validation

croyzor:
Add a check to validation which gathers resource requirements of each dataflow edge from its node's signature, and checks that all linked edges have the same resource requirements.


src/hugr/validate.rs
```diff
@@ -63,6 +71,33 @@ impl<'a> ValidationContext<'a> {
         Ok(())
     }
 
+    /// Use the signature supplied by a dataflow node to work out the resource
+    /// resource requirements for all of its input and output edges, then put
```
aborgna-q, line 75:
```suggestion
    /// Use the signature supplied by a dataflow node to work out the
    /// resource requirements for all of its input and output edges, then put
```

src/hugr/validate.rs
```diff
@@ -63,6 +71,33 @@ impl<'a> ValidationContext<'a> {
...
+        }
+
+        let output_ports = self.hugr.node_ports(node.clone(), Direction::Outgoing);
+        for p in output_ports.into_iter() {
+            let port_ix = self.hugr.graph.port_index(node.index, p.offset).unwrap();
+            assert!(self
+                .resources
+                .insert(port_ix, sig.output_resources.clone())
+                .is_none());
+        }
```
aborgna-q: If you add a `Signature::get_resources(dir: Direction)` (similar to the other methods) this should be doable without repetition
```rust
        for dir in Direction::BOTH {
            for p in self.hugr.node_ports(node.clone(), dir) {
                let port_ix = self.hugr.graph.port_index(node.index, p.offset).unwrap();
                assert!(self
                    .resources
                    .insert(port_ix, sig.get_resources(dir).clone())
                    .is_none());
            }
        }
```

src/hugr/validate.rs
```diff
@@ -146,6 +181,67 @@ impl<'a> ValidationContext<'a> {
...
+    /// has already been done.
+    fn check_resources_compatible(
+        &self,
+        from_node: &Node,
+        to_node: &Node,
+        port: (&Port, &PortIndex),
+        link: (&Port, &PortIndex),
+    ) -> Result<(), ValidationError> {
+        let d1 = self.hugr.graph.port_direction(*port.1).unwrap();
+        let d2 = self.hugr.graph.port_direction(*link.1).unwrap();
```
aborgna-q, line 200:
```suggestion
        let d1 = port.0.direction();
        let d2 = link.0.direction();
```
(can be inlined)

src/hugr/validate.rs
```diff
@@ -173,6 +269,19 @@ impl<'a> ValidationContext<'a> {
             });
         }
 
+        // Ignore subport indices assuming that we only care about dataflow nodes
+        for (port_index, link_index) in links.clone().into_iter().map(|(a, b)| (a.port(), b.port()))
```
aborgna-q: Why not move this to the `for (subport, link) in links {` loop below?
Then you know that the edge goes from `subport` to `link`, so you don't need the extra logic handling directions in `check_resources_compatible`.

src/hugr/validate.rs
```diff
@@ -720,19 +849,31 @@ pub enum InterGraphEdgeError {
         from_offset: Port,
         typ: ClassicType,
     },
+    #[error("Two linked ports both have the same direction")]
```
aborgna-q, line 851: This is part of portgraph guarantees, it should be safe to `unwrap` it instead of propagating.

src/hugr/validate.rs
```diff
@@ -27,6 +28,8 @@ struct ValidationContext<'a> {
     hugr: &'a Hugr,
     /// Dominator tree for each CFG region, using the container node as index.
     dominators: HashMap<Node, DominatorTree>,
+    /// Resource requirements associated with each edge
+    resources: HashMap<PortIndex, ResourceSet>,
```
aborgna-q, line 32: In the Hugr level we normally address ports by `(Node, Port)` instead of the internal portgraph's `NodeIndex`. Maybe using that would avoid some extra queries?

src/hugr/validate.rs
croyzor: In this case I don't think there's anything to unwrap - I need to fill the case or else rust will complain about it being missing, so I guess I could panic instead?

src/hugr/validate.rs
croyzor: Yes, that's much nicer!

src/hugr/validate.rs
croyzor: Good spot!
Are you saying that `graph.port_links(x)` always returns pairs in order (i.e. `(source, target)`)? Even if `x` is the target?

src/hugr/validate.rs
croyzor: Ahh, nice

src/hugr/validate.rs
```diff
@@ -27,6 +28,8 @@ struct ValidationContext<'a> {
     hugr: &'a Hugr,
     /// Dominator tree for each CFG region, using the container node as index.
     dominators: HashMap<Node, DominatorTree>,
+    /// Resource requirements associated with each edge
+    resources: HashMap<(Node, Port), ResourceSet>,
```
aborgna-q, line 32: I didn't see it before, but will same-direction ports in the same node ever have different ResourceSets? Would `(Node, Direction)` work?

croyzor: Hmm I guess it would, I was thinking using the specific port offers the flexibility of knowing where to put a lift node in some cases

aborgna-q: If it's useful leave it as is. It's a trade-of between memory/perf and flexibility.

croyzor: I think I'll change it to `(Node, Direction)` after all

acl-cqc: `(Node, Direction)` will require more Lift nodes I think, as with `(Node, Port)` there is an implicit union over inputs? I guess the ideal would be `(Node, Port)` for inputs and `Node` for outputs....

acl-cqc:
What's the high-level overview here? You mentioned unification, so....
* I would think (not having looked into the details) that *given Lift nodes* it was possible to write a (most general) unification-based checker. At which point one does not need explicit resource annotations anymore, only to be given a graph with lift nodes....right???
* Alternatively, if you're given a Hugr with resource requirements everywhere, it is straightforward to compute where the lift nodes are needed (as you do here). Does that mean we could free clients from the chore of having to insert lift nodes themselves?

croyzor:
> What's the high-level overview here?

It's a simple check on graphs which could have come out of a rewrite or guppy or BRAT, where it's assumed that lift nodes and resource annotations are present because they were added in some other step.

> I would think (not having looked into the details) that given Lift nodes it was possible to write a (most general) unification-based checker. At which point one does not need explicit resource annotations anymore, only to be given a graph with lift nodes....right???

I think it's more likely that we'll want to pass in a graph which has resource information in its signature, and add lift nodes from that

>   Alternatively, if you're given a Hugr with resource requirements everywhere, it is straightforward to compute where the lift nodes are needed (as you do here). Does that mean we could free clients from the chore of having to insert lift nodes themselves?

Yes, that's the next task (TKET-3246)

src/hugr/validate.rs
croyzor: > `(Node, Direction)` will require more Lift nodes I think, as with `(Node, Port)` there is an implicit union over inputs? I guess the ideal would be `(Node, Port)` for inputs and `Node` for outputs....

I don't think this applies here, as we're not trying to add lift nodes

aborgna-q (approved):
LGTM 👍

PR #182, opened 2023-06-20
Make HugrView iterators generic

aborgna-q:
In preparation for new hugr abstractions.

Replaces the hand-written `PortLinks` with `context-iterators` primitives.

### DIFF
--- a/src/hugr/view.rs
+++ b/src/hugr/view.rs
@@ -8,6 +8,8 @@ use context_iterators::{ContextIterator, IntoContextIterator, MapCtx, WithCtx};
 use itertools::{Itertools, MapInto};
 use portgraph::{multiportgraph, LinkView, PortView};
 
+type MapWithCtx<I, Ctx, O> = MapCtx<WithCtx<I, Ctx>, fn(<I as Iterator>::Item, &Ctx) -> O>;
+
 use super::Hugr;
 use super::{Node, Port};
 use crate::ops::OpType;
@@ -151,10 +153,8 @@ where
     type Neighbours<'a> = MapInto<multiportgraph::Neighbours<'a>, Node> where Self: 'a;
 
     /// Iterator over the children of a node
-    type PortLinks<'a> = MapCtx<
-        WithCtx<multiportgraph::PortLinks<'a>, &'a Hugr>,
-        fn((multiportgraph::SubportIndex, multiportgraph::SubportIndex), &&'a Hugr) -> (Node, Port),
-    > where
+    type PortLinks<'a> = MapWithCtx<multiportgraph::PortLinks<'a>, &'a Hugr, (Node, Port)>
+    where
         Self: 'a;
 
     #[inline]
--- a/src/hugr/view.rs
+++ b/src/hugr/view.rs
@@ -4,29 +4,43 @@
 use std::iter::FusedIterator;
 use std::ops::Deref;
 
+use context_iterators::{ContextIterator, IntoContextIterator, MapCtx, WithCtx};
 use itertools::{Itertools, MapInto};
-use portgraph::{LinkView, PortView};
+use portgraph::{multiportgraph, LinkView, PortView};
 
 use super::Hugr;
 use super::{Node, Port};
 use crate::ops::OpType;
 use crate::Direction;
 
-/// An Iterator over the nodes in a Hugr(View)
-pub type Nodes<'a> = MapInto<portgraph::multiportgraph::Nodes<'a>, Node>;
-
-/// An Iterator over (some or all) ports of a node
-pub type NodePorts = MapInto<portgraph::portgraph::NodePortOffsets, Port>;
-
-/// An Iterator over the children of a node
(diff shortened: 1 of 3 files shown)


src/hugr/view.rs
```diff
@@ -77,35 +97,66 @@ pub trait HugrView {
...
+    /// An Iterator over the children of a node
+    type Children<'a> = MapInto<portgraph::hierarchy::Children<'a>, Node> where Self: 'a;
+
+    /// An Iterator over (some or all) the nodes neighbouring a node
+    type Neighbours<'a> = MapInto<multiportgraph::Neighbours<'a>, Node> where Self: 'a;
+
+    /// Iterator over the children of a node
+    type PortLinks<'a> = MapCtx<
+        WithCtx<multiportgraph::PortLinks<'a>, &'a Hugr>,
+        fn((multiportgraph::SubportIndex, multiportgraph::SubportIndex), &&'a Hugr) -> (Node, Port),
```
acl-cqc, line 156: Oh, blimey. Are you sure this is better than PortLinks? The old PortLinks seems a lot easier to read and understand IMHO, and you could keep that without changing the rest of the PR...

src/hugr/view.rs
```diff
@@ -4,29 +4,43 @@
...
-pub type Children<'a> = MapInto<portgraph::hierarchy::Children<'a>, Node>;
-
-/// An Iterator over (some or all) the nodes neighbouring a node
-pub type Neighbours<'a> = MapInto<portgraph::multiportgraph::Neighbours<'a>, Node>;
-
 /// A trait for inspecting HUGRs.
 /// For end users we intend this to be superseded by region-specific APIs.
 pub trait HugrView {
+    /// An Iterator over the nodes in a Hugr(View)
+    type Nodes<'a>: Iterator<Item = Node>
```
acl-cqc, line 20: Yup, all good here

src/hugr/view.rs
acl-cqc: I guess this might get us the "win" of being a DoubleEndedIterator for free? Which is something, I admit. Note that HugrView doesn't specify that the iterators are double-ended, though - only this particular implementation does - do we want it to?

aborgna-q: I pushed a way cleaner version instead. What do you think about it?
(`MapWithCtx` will be in the funky library at some point :) )

src/hugr/view.rs
```diff
@@ -132,36 +183,33 @@ where
...
-    fn linked_ports(&self, node: Node, port: Port) -> PortLinks<'_> {
+    fn linked_ports(&self, node: Node, port: Port) -> Self::PortLinks<'_> {
         let hugr = self.as_ref();
         let port = hugr.graph.port_index(node.index, port.offset).unwrap();
-        let links = hugr.graph.port_links(port);
-        PortLinks { hugr, links }
+        hugr.graph
+            .port_links(port)
+            .with_context(hugr)
+            .map_with_context(|(_, link), hugr| {
```
acl-cqc, line 207: Hmmm, alternatively, are we sure this shouldn't be a method on `hugr` rather than accessing the `hugr.graph` directly? Generally speaking I think the grand mixture of Node and Port vs NodeIndex and PortIndex that one sees when using a Hugr is a bit of a pain and it'd be nicer to standardize on Node/Port. Doing that everywhere is clearly beyond the scope of this PR but we could make a start....

aborgna-q: I don't think we want to expose anything handling `NodeIndex` or `SubportIndex` at the Hugr level.

Here we are wrapping the internal iterator from portgraph with works with `SubportIndex`, so we necessarily need some portgraph calls. (The context could have been a `&MultiPortgraph` instead of `&Hugr`, but it's the same here)

src/hugr/view.rs
aborgna-q: > I guess this might get us the "win" of being a DoubleEndedIterator for free?

The main goal is dropping the hand-written iterators, which just bloat the code and always end up missing some trait implementation.

This is not as big of a problem here as it is in portgraph (which has 800+ lines of trivial iterator impls), but the future region views will also require the same iterator-wrapping code.

acl-cqc (approved):
Yes, that's better. Comment that MapWithCtx thing somehow and you're good to go :)

src/hugr/view.rs
```diff
@@ -4,29 +4,45 @@
 use std::iter::FusedIterator;
 use std::ops::Deref;
 
+use context_iterators::{ContextIterator, IntoContextIterator, MapCtx, WithCtx};
 use itertools::{Itertools, MapInto};
-use portgraph::{LinkView, PortView};
+use portgraph::{multiportgraph, LinkView, PortView};
+
+type MapWithCtx<I, Ctx, O> = MapCtx<WithCtx<I, Ctx>, fn(<I as Iterator>::Item, &Ctx) -> O>;
```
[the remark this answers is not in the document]
aborgna-q: Aaand it's gone 🪄

PR #183, opened 2023-06-20
update to simpler serialization spec

ss2165:


specification/hugr.md
```diff
@@ -1433,27 +1433,34 @@ conversion to/from the binary serialised form.
...
 
-// (parent, #incoming, #outgoing)
-struct Node = (Optional<Int>, Int, Int)
+struct Node{
+  // parent node
+  parent: Int,
+  // name of operation
+  op: String
+  //other op-specific fields
+  ...
```
acl-cqc, line 1457: args (for custom ops)?

specification/hugr.md
```diff
@@ -1433,27 +1433,34 @@ conversion to/from the binary serialised form.
...
-  node_weights: map<Int, Operation>
+  nodes: [Node],
+  edges: [Edge],
 }
 
-// (parent, #incoming, #outgoing)
-struct Node = (Optional<Int>, Int, Int)
+struct Node{
+  // parent node
+  parent: Int,
```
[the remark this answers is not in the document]
ss2165: Added line "The unique root node of the HUGR reports itself as the parent."

specification/hugr.md
ss2165: are you suggesting that field should be mandated on all nodes or that it should
be mentioned as a possible extra field?

specification/hugr.md
```diff
@@ -1441,27 +1441,35 @@ conversion to/from the binary serialised form.
...
+  // name of operation
+  op: String
+  //other op-specific fields
+  ...
+}
 // ((source, offset), (target, offset)
 struct Edge = ((Node, Optional<Int>), (Node, Optional<Int>))
```
acl-cqc, line 1469: Also, I think this definition of `Edge` needs updating? Presumably you do not mean to inline the Node structs from each end into the Edge

specification/hugr.md
ss2165: Yeah currently it's part of the `...`
Maybe we should come back to this once we know what the spec for defining args in this format is? #186 and any follow on. I could leave an issue?

acl-cqc: Yeah OK; agreed we can't necessarily/completely define that now. OTOH perhaps mentioning `args` as a field would say what we do know now even without completing the picture with the bits we don't yet. Ok, up to you

acl-cqc (approved):

specification/hugr.md
ss2165: Yeah I've added a line above.

PR #186, opened 2023-06-20
Sketch out resources + extension mechanism, update CustomType

acl-cqc:
Lots of gaps here, and many things are not as neat as they might be eventually. But various OpDef and CustomOp classes are removed and/or combined, as per the spec. Also update the spec re. "Int".

I haven't implemented the loading code here, beyond that CustomOp's should serialize only the names of the OpDefs to which they refer (the loading code is TODO atm 'coz I haven't figured out the necessary serde). A function `resolve_extension_ops` can be called after deserializing (which produces just names) to find the actual OpDefs from a "registry" (a Map<String, Resource>, see #188)

We have now ruled out Ops parameterized by float(s); these will have to be passed in as static inputs.

Other questions/follow-up issues...
* #189 (ResourceSet in Signature) remains unresolved, but can become an issue after this is merged, as *something* needs doing.
* #198 Parameterization of CustomTypes now handled here too, but (following also #221) there is jira TKET-3305 it refactor the type hierarchy.
* Make new issue to move many operations out into their own resource (and add arithmetic)
* Make new issue for registering Extensions / building the "registry". This needs to combine both loading YAML and also binary code (perhaps through dynamic loading) whereby extensions "self-register" (the spec says). The current "API" expects any binary `compute_signature` functions to have registered themselves *before* the YAML is loaded, whereas try_lower implementations could be added after loading YAML - this is a bit inconsistent, but follows from that every OpDef much have some way of computing Signatures, but may not provide any way of lowering.


src/resource.rs
```diff
@@ -2,35 +2,291 @@
...
-#[derive(Clone, Debug, Default, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
 pub struct Resource {
     /// Unique identifier for the resource.
-    pub name: ResourceId,
-    /// Set of resource dependencies required by this resource.
-    pub resource_reqs: ResourceSet,
+    name: ResourceId,
+    // Set of resource dependencies required by this resource.
+    // TODO I haven't seen where these are required yet. If they are,
```
croyzor, line 283: What about the case where someone wants to define a resource in terms of types provided by another resource?

This doesn't necessarily require any knowledge of the implementation of the other resource - one could have an operation on types from another resource which is defined in terms of that resources operations

acl-cqc: Yes, indeed. I'm not quite sure what this should look like. It may fall under the heading of "namespacing", which is one of the unresolved issues in this PR. Good point, thanks :)

src/types.rs
```diff
@@ -31,8 +32,6 @@ pub enum EdgeKind {
     Const(ClassicType),
     /// Explicitly enforce an ordering between nodes in a DDG.
     StateOrder,
-    /// An edge specifying a resource set.
-    Resource(ResourceSet),
```
croyzor, line 35: :+1:

src/types/type_arg.rs
```diff
@@ -0,0 +1,79 @@
...
+    Type,
+    /// Node must provide a [TypeArgValue::ClassicType]
+    ClassicType,
+    /// Node must provide a [TypeArgValue::F64] floating-point value
+    F64,
+    /// Node must provide a [TypeArgValue::Int] integer
+    Int,
+    /// Node must provide some Opaque data in an [TypeArgValue::Opaque].
+    /// TODO is the typerow here correct?
+    Opaque(String, Box<TypeRow>),
```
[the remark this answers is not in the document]
acl-cqc: Ah, hmmm, indeed it does match, so this should probably be a CustomType....

So IIUC the idea is a resource might declare a type `MyContainer<T>`. MyContainer *of some TypeRow* is then a type. (That is: types can only be parameterized by other types; opdef's can be parameterized by values too. Hmmm, perhaps custom types should be parameterized by TypeParam's too, allowing MyContainerOfSize<n> for some int n)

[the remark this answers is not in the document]
acl-cqc: and if MyContainer<Sum<Int,Float>> is a type, then we allow the OpDef to require a parameter of type `MyContainer<Sum<Int,Float>>` meaning a particular op==node would name the OpDef and provide *an instance of MyContainer<Sum<Int,Float>>* as a serialized blob....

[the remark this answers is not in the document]
ss2165: turning in to discussion

src/ops/custom.rs
```diff
@@ -1,210 +1,157 @@
...
+use super::{LeafOp, OpName, OpTrait, OpType};
 
-    /// The custom operation.
-    custom: Box<dyn CustomOp>,
+/// An instantiation of an [`OpDef`] with values for the type arguments
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[serde(into = "OpaqueOp", from = "OpaqueOp")]
+pub enum ExternalOp {
+    Resource {
+        def: Rc<OpDef>,
```
aborgna-q, line 20: I don't think we want to lose `Send` on the `Hugr`. We probably should use an `Arc` here.

In the future we _could_ add a lifetime parameter to the Hugr and make this a reference... but that would take some work.

acl-cqc: @aborgna-q thanks, `Rc` vs `Arc` was something else I meant to bring up that got lost among the other discussions. Will do.
Is making that change sufficient to restore `Send`ness? (And, why isn't there a compile-time error if we've just lost that...can we add one?)

acl-cqc: We should probably make some change to enforce `Send`ability in a preliminary PR, given this one is still being discussed!

aborgna-q: Yes, `Arc` is `Send + Sync`.
There's nothing requiring `Send` on `Hugr` at the moment, perhaps we could have a simple test to check for it.

src/types/type_arg.rs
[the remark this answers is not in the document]
acl-cqc: Yes, this isn't right here, but I don't think CustomType is either. Await outcome of discussion :)

src/ops/custom.rs
acl-cqc: So there's example code for a test here: https://github.com/rust-lang/rust/issues/44363
But rustdoc at least is able to work this out automagically: https://github.com/rust-lang/rust/issues/17606
The latter thus seems preferable, but how to turn this into a CI test? Scanning/parsing the generated docs sounds worse than writing the code snippet in the first of those issues.

aborgna-q: In our case we can add a "test" that fails to compile if the traits are missing:

(in `hugr.rs`)
```rust

#[cfg(test)]
mod test {
    use super::*;

    #[test]
    fn impls_send_sync() {
        trait Test: Send + Sync {}
        impl Test for Hugr {}
    }
}
```

src/resource.rs
acl-cqc: I've kept this in as an "upper bound" of all signatures that *might* be returned by compute_signature functions, which can't be determined by inspection.

However, I'm not entirely sure how it's used. If I declare an operation that takes a MyType<Int,6> without the resource declaring the type MyType being around (even in YAML!), seems like I don't lose anything - it's just an early-warning that I the end-user-programmer *may* not be able to get very far. But the HUGR is still capable of passing such values around as serialized blobs, if I have operations that both create and consume those values. Really, the only information added by the "declaration" of the custom type is validity checking of the type arguments?!

src/types/type_arg.rs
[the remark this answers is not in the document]
acl-cqc: No CustomTypes in TypeArgs, and CustomType now parameterized by TypeArgs the same way.

specification/hugr.md
```diff
@@ -719,7 +719,7 @@ resources:
   types:
   - name: QubitVector
     # Opaque types can take type arguments, with specified names
-    args: [["size", u64]]
+    args: [["size", Int]]
```
acl-cqc, line 722: I changed all the `u64`s to `Int` to reflect what's in `ClassicType` and `TypeParam` - and the general policy that the difference between signed and unsigned is in the operations, not the bits flowing around the wires of the Hugr.

However, there is an argument that the specification here is not just about the (signedness-independent) serialized form of the parameters, but also an instruction as to how int constants should be parsed / treated by the frontend. If this is left as Int, then *some other* data somewhere needs to tell any frontend that the value should be unsigned...

src/extensions/rotation.rs
```diff
@@ -79,31 +90,32 @@ impl CustomConst for Constant {
...
 
     fn const_type(&self) -> ClassicType {
-        match self {
+        let t: Type = match self {
             Constant::Angle(_) => Type::Angle,
             Constant::Quaternion(_) => Type::Quaternion,
-        }
-        .classic_type()
+        };
+        let t2: CustomType = t.into();
```
acl-cqc, line 97: This is a bit of a pain. Unfortunately a single `.into()` doesn't do a conversion requiring two calls of `.from` :(.

acl-cqc:
I've made this no-longer-a-draft but the question is, what do we want to do. Merge as it stands, fix all the TODOs, ??

specification/hugr.md
cqc-alec: Are there any cases where we actually want signed semantics in the uses of `Int` in the core, or should we just say that it's always unsigned (and maybe rename to `Nat` or something)?

specification/hugr.md
```diff
@@ -788,9 +799,11 @@ Serialization section). It is also trivial to serialize these
 definitions in to the overall HUGR serialization format.
 
 Note the only required fields are `name` and `description`. `signature` is optional, but if present
-must have children `inputs` and `outputs`, each lists. The optional `misc` field is used for arbitrary
-YAML, which is read in as-is and passed to compiler passes and (if no `signature` is present) the
-`compute_signature` function; e.g. a pass can use the `basis` information to perform commutation.
+must have children `inputs` and `outputs`, each lists, and may have `resources`.
+
+The optional `misc` field is used for arbitrary YAML, which is read in as-is and passed to compiler
+ passes and (if no `signature` is present) the`compute_signature` function; e.g. a pass can use the `basis` information to perform commutation.
```
ss2165, line 805: I thought the conclusion of the discussion was that data in `misc` would not be used to compute signature?

src/types/custom.rs
```diff
@@ -1,27 +1,35 @@
...
 impl CustomType {
     /// Creates a new opaque type.
-    pub fn new(id: SmolStr, params: impl Into<TypeRow>) -> Self {
+    pub fn new(id: SmolStr, params: impl Into<Vec<TypeArg>>) -> Self {
         Self {
             id,
-            params: Box::new(params.into()),
+            params: params.into(),
         }
     }
```
aborgna-q, line 34:
```rust
    pub fn new(id: impl Into<SmolStr>, params: impl Into<Vec<TypeArg>>) -> Self {
        Self {
            id: id.into(),
            params: params.into(),
        }
    }
```

and since it's a common case, it would be nice to have a const-constructor

```rust
    pub const fn new_simple(id: SmolStr) -> Self {
        Self {
            id,
            params: Vec::new(),
        }
    }
```

specification/hugr.md
acl-cqc: I admit I haven't seen a use for a signed int in a type yet.

specification/hugr.md
acl-cqc: IIRC we said that if you wanted statically-known floats then you could put that in `misc` if you didn't want Static wires to every node. But I don't recall saying that `misc` couldn't affect signature. `misc` is a fairly generic mechanism to say "I want to reuse the same (Rust) code for different ops, I can express the difference with data" - I don't see why that shouldn't include/allow reusing Rust code that computes type signatures.

src/types/custom.rs
acl-cqc: Yes, nice :-) - done.

src/types/custom.rs
```diff
@@ -1,27 +1,35 @@
...
-    id: SmolStr,
-    params: Box<TypeRow>,
+    /// Same as the corresponding [`TypeDef`]
+    ///
+    /// [`TypeDef`]: crate::resource::TypeDef
+    pub id: SmolStr,
+    /// Arguments that fit the [`TypeParam`]s declared by the typedef
+    ///
+    /// [`TypeParam`]: super::type_param::TypeParam
+    pub params: Vec<TypeArg>,
```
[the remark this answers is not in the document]
acl-cqc: We don't have any current uses, but I'd be surprised if custom code handling these types (giving these "external"/"opaque" declarations actual meaning) didn't need to see the params! We don't really know what that looks like (it's probably within some binary code that performs a custom OpDef *on* those types), so maybe there's another way. I could have a public accessor method instead, but then I feel like I should have a public mutator too, and...accessor/mutator is a bit of a YAGNI (preserving the interface if we change the implementation not to store the actual params, or to store some representation thereof), right?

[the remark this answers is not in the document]
aborgna-q: Yeah, I don't like exposing complex internals for API stability reasons (that `Vec` may be a target for perf experiments in a distant future).
Though having the `Serialize` derive kind of exposes everything anyways :P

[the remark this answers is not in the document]
acl-cqc: Yeah, that is definitely a downside of this serialization scheme...

I think it might be better to review what's public and isn't when we do a ("stable") release. A quick regex search suggests we have quite a lot of public fields all over the place, which is to say, trying to keep such APIs consistent (in the absence of making such a release), tends to be somewhat patchy. But I've fixed these two. There are a bunch more in resource.rs too tho...

src/types/custom.rs
```diff
@@ -1,37 +1,46 @@
...
+    }
+
     /// Returns the unique identifier of the opaque type.
     pub fn id(&self) -> &str {
         &self.id
     }
 
     /// Returns the parameters of the opaque type.
-    pub fn params(&self) -> &TypeRow {
+    pub fn params(&self) -> &Vec<TypeArg> {
```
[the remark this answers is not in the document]
aborgna-q: Or a `Cow<'_, [TypeArg]>` if you want both :)

But as you say in the other comment, this could be a consideration for the public release.

src/types/type_param.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Where an argument has type [TypeParam::List]`<T>` - all elements will implicitly
+    /// be of the same variety of TypeArg, representing a `T`.
+    List(Vec<TypeArg>),
+}
+
+/// Checks a [TypeArg] is as expected for a [TypeParam]
+pub fn check_arg(arg: &TypeArg, param: &TypeParam) -> Result<(), String> {
+    match (arg, param) {
+        (TypeArg::Type(_), TypeParam::Type) => (),
+        (TypeArg::ClassicType(_), TypeParam::ClassicType) => (),
```
croyzor, line 49: I think this might need some more thought. We've specifically said that we don't want floats to be allowed here, so we only want to accept this `TypeArg` (and the one on the line above) if they're in our set of allowed static types

croyzor: Nvm, I was looking at the wrong level

src/extensions/rotation.rs
acl-cqc: But it can be at least a bit easier by using `.custom_type()`

src/ops/custom.rs
```diff
@@ -1,219 +1,186 @@
...
-
-impl Deref for OpaqueOp {
-    type Target = dyn CustomOp;
+/// An opaquely-serialized op that refers to an as-yet-unresolved [`OpDef`]
+#[derive(Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct OpaqueOp {
+    resource: ResourceId,
+    op_name: SmolStr,
+    #[serde(skip)]
+    description: String, // cache in advance as description() returns &str
```
croyzor, line 65: Why should we skip this in serialisation? Wont we wont it to be exposed to/from guppy?

src/ops/custom.rs
```diff
@@ -1,219 +1,186 @@
...
-}
 
-impl_downcast!(CustomOp);
-impl_box_clone!(CustomOp, CustomOpBoxClone);
-
-/// Serializable definition for dynamically loaded operations.
-///
-/// TODO: Define a way to construct new CustomOps from a serialized definition.
-#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
-pub struct OpDef {
```
croyzor, line 108: Is OpDef now ExternalOp? This isn't reflected in the spec afaict

src/resource.rs
```diff
@@ -2,35 +2,325 @@
...
+pub trait CustomSignatureFunc: Send + Sync {
+    /// Compute signature of node given the operation name,
+    /// values for the type parameters,
+    /// and 'misc' data from the resource definition YAML
+    fn compute_signature(
+        &self,
+        name: &SmolStr,
+        arg_values: &[TypeArg],
+        misc: &HashMap<String, serde_yaml::Value>,
+    ) -> Result<(TypeRow, TypeRow, ResourceSet), SignatureError>;
```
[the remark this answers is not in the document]
acl-cqc: Well, firstly there's the question of whether external/resource ops should be able to have `static_inputs` - currently the belief is, no they should not be.

Secondly there's the `input_resources` question, and when that's resolved (i.e. Signature stores the delta) this will be a stage closer to a Signature. But I don't think we intend to allow const inputs.

So, post #189, we could end up with something like
```
struct GraphType {inputs: TypeRow, outputs: TypeRow, resource_delta: ResourceSet};
struct Signature { ty: GraphType, static_inputs: TypeRow }
```
but not there yet.

src/types/type_param.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/// A statically-known argument value to an operation.
+#[derive(Clone, Debug, PartialEq, Eq, serde::Deserialize, serde::Serialize)]
+pub enum TypeArg {
+    /// Where the TypeDef declares that an argument is a [TypeParam::Type]
+    Type(SimpleType),
+    /// Where the TypeDef declares that an argument is a [TypeParam::ClassicType],
+    /// it'll get one of these (rather than embedding inside a Type)
+    ClassicType(ClassicType),
```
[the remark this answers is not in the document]
acl-cqc: So this boils down to whether the resource-implementor writes
```
let TypeArg::ClassicType(elem_type) = arg else {panic!("Illegal argument for parameter");}
```
vs
```
let TypeArg::Type(SimpleType::Classic(elem_type)) = arg else {panic!("Illegal argument for parameter!);}
```
(of course the *calling* code has already ruled out the situations that would give rise to the panic, but the type system doesn't let us say that).

However, I think it's good to keep a 1:1 correspondence between "The TypeParam that I declare" and "The TypeArg that I expect back". Rather than TypeParam::ClassicType documenting that "you'll get back a TypeArg::Type whose contents are a SimpleType::Classic".

src/ops/custom.rs
```diff
@@ -1,219 +1,186 @@
...
+use super::{LeafOp, OpName, OpTrait, OpType};
 
-    /// The custom operation.
-    custom: Box<dyn CustomOp>,
+/// An instantiation of an [`OpDef`] with values for the type arguments
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[serde(into = "OpaqueOp", from = "OpaqueOp")]
+pub enum ExternalOp {
+    Resource { def: Arc<OpDef>, args: Vec<TypeArg> },
+    Opaque(OpaqueOp),
```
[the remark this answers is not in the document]
acl-cqc: See comment on OpaqueOp, but I've added here

src/ops/custom.rs
```diff
@@ -1,219 +1,186 @@
...
-        Self {
-            id: id.into(),
-
-            custom: Box::new(custom),
+impl From<ExternalOp> for OpaqueOp {
+    fn from(value: ExternalOp) -> Self {
+        match value {
+            ExternalOp::Opaque(op) => op,
+            ExternalOp::Resource { def, args } => {
+                // There's no way to report a panic here, but serde requires Into not TryInto. Eeeek!
```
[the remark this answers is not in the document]
acl-cqc: Yah.

So I don't generally like `panic` at *deserialization* time, 'coz at that point you've loaded up the Hugr, done who knows what to it - and if we panic then you've lost all of that. (Panicking at load time seems less traumatic!)

However, ok, you force me to stay honest :). The right solution here is to cache the signature field in the ExternalOp::Resource, to which end I've broken out a `ResourceOp; struct. Done, I think :)

src/ops/custom.rs
acl-cqc: Op*Def* is now in `resource.rs`. ExternalOp is the instantiation of such a def into a concrete `OpType` for a node.

src/ops/custom.rs
acl-cqc: Yes, good point. I was thinking that when we turn this back into a ResourceOp we'll lose this (the `OpDef` defines the description method instead). So we could probably benefit from some mechanism for serializing different parts at different times, but I've not done that here in this "sketch"! However, description now saved and restored.

When we turn back into a ResourceOp we'll just discard the serialized description - the only other thing we could really do would be to raise an error if they were different, but I haven't bothered to do that here.

src/resource.rs
```diff
@@ -2,35 +2,325 @@
...
+    #[serde(serialize_with = "elide_arcs", deserialize_with = "reinstate_arcs")]
+    operations: HashMap<SmolStr, Arc<OpDef>>,
+}
+
+fn elide_arcs<S: Serializer>(
+    _ops: &HashMap<SmolStr, Arc<OpDef>>,
+    _serializer: S,
+) -> Result<S::Ok, S::Error> {
+    // serde doesn't like Arc - write it out as if it were a HashMap<SmolStr, OpDef>
+    todo!()
```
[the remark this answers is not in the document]
acl-cqc: Well, they do affect serialization of *resources*, but not of *ops* or Hugrs; we expect to pass the serialized YAML around with the Hugr, but that is very much a TODO at the moment. I'm not sure we necessarily need a non-YAML Resource format, or whether we do any serialization of Resources atm. @mark-koch perhaps?

If we do need this now, this may have to wait awhile, I've not figured out how to make serde comply yet...

src/resource.rs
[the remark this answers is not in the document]
acl-cqc: Agustin showed me how to do this, admittedly it requires enabling a feature of Serde globally (which we might not want everywhere), but if we can live with that then the rest is easy :)

src/types/custom.rs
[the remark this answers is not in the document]
acl-cqc: Hmmmm, so I've switched this (and an argument type that was `&Vec<TypeArg>` too) to `&[TypeArg]` as that seems the most general thing.

Is there any advantage in declaring the return type `Cow` (as you suggest) rather than the caller constructing their own `Cow` around the `&[TypeArg]` here? The answer may be, "maybe but only if" we construct+store the `Cow` internally and share it between multiple calls (all returning the same `Cow`), rather than many clients each building their own `Cow` around the same `&TypeArg`, but I'm not even sure about that - isn't building a `Cow` zero-cost anyway?

acl-cqc:
This has been sitting here awhile and @ss2165 is now away, but perhaps if @croyzor could review for direction and @aborgna-q for rustiness then that would be reasonable grounds to merge?

src/ops/custom.rs
croyzor: Oops I missed that!

src/ops/custom.rs
```diff
@@ -1,219 +1,193 @@
...
+        match self {
+            Self::Opaque(op) => op.signature.clone().unwrap(),
+            Self::Resource(ResourceOp { signature, .. }) => signature.clone(),
+        }
     }
+}
 
-    /// The signature of the operation.
-    fn signature(&self) -> Signature;
+// Note *not* Serializable: container (ExternalOp) should have serialized as an OpaqueOp instead.
```
aborgna-q, line 82:
```suggestion
// Note *not* Serializable: container (ExternalOp) should be serialized as an OpaqueOp instead.
```

src/ops/custom.rs
```diff
@@ -1,219 +1,193 @@
...
+#[derive(Clone, Debug)]
+pub struct ResourceOp {
```
aborgna-q, line 84: Docs!

src/ops.rs
```diff
@@ -2,7 +2,7 @@
 
 pub mod constant;
 pub mod controlflow;
-pub mod custom;
+mod custom;
```
aborgna-q, line 5: We don't want to expose `ExternalOp`?

src/types/type_param.rs
```diff
@@ -0,0 +1,61 @@
...
+//! [`TypeDef`]: crate::resource::TypeDef
+
+use crate::ops::constant::HugrIntValueStore;
+
+use super::{ClassicType, SimpleType};
+
+/// A parameter declared by an OpDef. Specifies a value
+/// that must be provided by each operation node.
+/// TODO any other 'leaf' types? We specifically do not want float.
+/// bool should eventually be a Sum type (Container).
```
aborgna-q, line 14:
```suggestion
// TODO any other 'leaf' types? We specifically do not want float.
// bool should eventually be a Sum type (Container).
```

src/types/type_param.rs
```diff
@@ -0,0 +1,61 @@
...
+    ClassicType(ClassicType),
+    /// Where the TypeDef declares a [TypeParam::Int]
+    Int(HugrIntValueStore),
+    /// Where an argument has type [TypeParam::List]`<T>` - all elements will implicitly
+    /// be of the same variety of TypeArg, representing a `T`.
+    List(Vec<TypeArg>),
+}
+
+/// Checks a [TypeArg] is as expected for a [TypeParam]
+pub fn check_arg(arg: &TypeArg, param: &TypeParam) -> Result<(), String> {
```
aborgna-q, line 46: Either
```rust
pub fn check_arg(arg: &TypeArg, param: &TypeParam) -> bool {
```
or define an actual error type to return the internal mismatching types.

aborgna-q (approved):
👍 

Most of these comments reduce to "please add more docs". There's a load of new {Module, External, Custom} {Type, Op, Def, Param, Arg} that may confuse a first time reader. Perhaps a short resume of the structure as a top-level comment in `resource` would help...

croyzor (approved):

src/resource.rs
```diff
@@ -2,35 +2,313 @@
...
+
+#[derive(serde::Deserialize, serde::Serialize)]
+enum SignatureFunc {
+    // Note: I'd prefer to make the YAML version just implement the same CustomSignatureFunc trait,
+    // and then just have a Box<dyn CustomSignatureFunc> instead of this enum, but that seems less likely
+    // to serialize well.
+    /// TODO: these types need to be whatever representation we want of a type scheme encoded in the YAML
+    #[serde(rename = "signature")]
+    FromYAML { inputs: String, outputs: String },
+    #[serde(skip)]
```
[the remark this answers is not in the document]
acl-cqc: So I'm only really thinking of the YAML "serialization format" outlined in `specification/hugr.md`. The idea is that if the Resource uses binary code, then we *don't* serialize that - I guess we could try and serialize a reference to a custom trait impl, but the spec only says that such Resources "self-register".

So ideally I'd just prevent `bincode` from even trying to serialize this at all. I guess the alternative is to somehow require `impl`s of the trait to be serializable - is that some simple prerequisite of `trait CustomLowerFunc` / `trait CustomSignatureFunc`?

src/resource.rs
```diff
@@ -2,35 +2,313 @@
...
+                .finish(),
+            Self::CustomFunc(_) => f.write_str("<custom sig>"),
+        }
+    }
+}
+
+#[derive(serde::Deserialize, serde::Serialize)]
+enum LowerFunc {
+    #[serde(skip)]
+    None,
```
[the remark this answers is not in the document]
acl-cqc: Again just thinking of the YAML - if there's not a fixed hugr, which can be given (e.g. filename) in the YAML file, then either the Resource has binary code that self-registers the custom function, or there is no possibility of lowering.

src/resource.rs
```diff
@@ -2,35 +2,313 @@
...
+}
 
 /// A unique identifier for a resource.
 ///
 /// The actual [`Resource`] is stored externally.
 pub type ResourceId = SmolStr;
 
 /// A resource is a set of capabilities required to execute a graph.
-#[derive(Clone, Debug, Default, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
```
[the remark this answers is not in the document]
acl-cqc: Yah, I am not sure about this. I do not *expect* to clone Resources; they'll be pretty big, and may contain third-party implementations of traits amongst others, so the intention is that the design does not require cloning. (This was also the motivation for `Arc<OpDef>`)

src/ops/custom.rs
acl-cqc: A late addition, that one. Oops.

src/resource.rs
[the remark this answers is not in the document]
acl-cqc: ON second thoughts, yes I guess we should. Someone will want to build a Resource by cloning another and mutating the copy... (indeed, this is even cheaper than it was before the use of `Arc<OpDef>`!). I would like to avoid multiple Resources with the same name, but that will have to be enforced via the "registry" (effectively a `Map<String, Resource>`, here only used in `resolve_extension_ops` with the rest a TODO). And I think we probably should not be cloning Resources within the Hugr crate, but I guess that's not a reason not to derive Clone.

What would you think to a method `clone_and_rename(new_name: String) -> Self` instead/as well, tho?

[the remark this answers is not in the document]
aborgna-q: Yes, we will never clone a resource and re-use it with the same Hugr.

But in general having clone is more for some meta-tasks (e.g. testing). If
- there's not logical error in having multiple copies (as long as you don't use it on the same registry)
- they do not manage external resources (so no handles / destructor shenanigans)
- they are safe and cheap to move
there's no reason not to have `Clone`.

src/resource.rs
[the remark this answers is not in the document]
aborgna-q: Fair enough. I think you can disregard this comment.

src/types/type_param.rs
acl-cqc: Defined an actual error type - rather skeletal for now but TypeArg/Param will become more complete in a later PR

src/ops.rs
acl-cqc: Oops, not sure how that crept in

src/resource.rs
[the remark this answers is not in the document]
acl-cqc: Yes, I was thinking of testing and heading for the same conclusion. Done. (It's quite nice that the Arc means cloning a Resource doesn't try to clone the CustomSignatureFunc's :-) )

PR #187, opened 2023-06-21
Update portgraph dep

### LINKED ISSUE #160: Dot string visualisation draws unconnected ports for StateOrder
So all nodes have extra unconnected ports
I believe this is due to changes from #152 
image

aborgna-q:
Closes #160

### DIFF
--- a/src/hugr.rs
+++ b/src/hugr.rs
@@ -16,7 +16,7 @@ pub use rewrite::{Replace, ReplaceError, Rewrite, SimpleReplacement, SimpleRepla
 
 use portgraph::dot::{DotFormat, EdgeStyle, NodeStyle, PortStyle};
 use portgraph::multiportgraph::MultiPortGraph;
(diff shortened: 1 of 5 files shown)


src/hugr.rs
```diff
@@ -121,10 +121,10 @@ impl Hugr {
                         PortStyle::new(html_escape::encode_text(&format!("{}", ty)))
                     }
                     EdgeKind::StateOrder => match self.graph.port_links(port).count() > 0 {
-                        true => PortStyle::Text("".to_string()),
+                        true => PortStyle::text("", false),
```
croyzor, line 124: Wont we still have graphviz boxes for these ports which just don't have any text in them?

aborgna-q: Yes, but they don't weight in the cell widths, and they do not have a border.

image

croyzor (approved):

PR #191, opened 2023-06-21
Region views of a portgraph

### LINKED ISSUE #125: Expose Toposort/Dominators algorithms on Hugr regions
Using portgraph's algorithms, but filtered to Hugr regions

aborgna-q:
Adds a `RegionView` and `FlatRegionView` hugr views that show a single region in the Hugr, either containing all the descendant nodes or just the direct children of the root.

These regions implement `petgraph`'s visit traits, exposing the internal graph. Closes #125 

### Not in this PR:

- We do not implement `petgraph::visit::IntoEdgeReferences` nor the `IntoEdge*` derivates.

  The weight of an edge should be an `EdgeKind`s, but petgraph requires the iterators to return an `EdgeRef : Copy` which has a method that returns an `&EdgeKind`.
  Since the edge kinds are computed on the fly from the signature we have to store them somewhere temporarily, but `EdgeRef` is `Copy` so it cannot contain it.
  I tried some alternatives with cached structures on the iterators, but it quickly became lifetime hell.
  I'll leave that for later; the algorithms that we currently want to use do not require those traits.
  
- The region's petgraph-exposed graph contains the root and all edges.

  We may want to remove the root when running algorithms, and maybe filter the kind of edges (parametrically, as it depends on the problem). The current formulation works for toposorting and dominators (#192), but we could add some extra options in the future.

### DIFF
--- a/src/hugr/region.rs
+++ b/src/hugr/region.rs
@@ -14,8 +14,7 @@ use super::HugrView;
 
 type FlatRegionGraph<'g> = portgraph::view::FlatRegion<'g, MultiPortGraph>;
 
-/// Single region view of a HUGR. Ignores any nodes that are not direct children
-/// of the root.
+/// Single region view of a HUGR. Includes only the root node and its direct children.
 ///
 /// For a view that includes all the descendants of the root, see [`RegionView`].
 #[derive(Clone, Debug)]
@@ -34,8 +33,8 @@ pub struct FlatRegionView<'g> {
 }
 
 impl<'g> FlatRegionView<'g> {
-    /// Create a new region view of a HUGR containing only the direct children
-    /// of the root.
+    /// Create a new region view of a HUGR containing only a root and its direct
+    /// children.
     pub fn new(hugr: &'g Hugr, root: Node) -> Self {
         let Hugr {
             graph,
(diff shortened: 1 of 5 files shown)


src/hugr/region.rs
```diff
@@ -0,0 +1,289 @@
...
+    fn edge_count(&self) -> usize {
+        // Faster implementation than filtering all the nodes in the internal graph.
+        self.nodes()
+            .map(|n| self.output_neighbours(n).count())
+            .sum()
+    }
+
+    fn nodes(&self) -> Self::Nodes<'_> {
+        // Faster implementation than filtering all the nodes in the internal graph.
+        self.hierarchy.children(self.root.index).map_into()
```
ss2165, line 105: this doesn't include the root? But `node_count` does?

src/hugr/region.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    fn num_ports(&self, node: Node, dir: Direction) -> usize {
+        self.graph.num_ports(node.index, dir)
+    }
+
+    fn children(&self, node: Node) -> Self::Children<'_> {
+        let mut iter = self.hierarchy.children(node.index).map_into();
+        if node != self.root {
+            // Eagerly empty the iterator.
+            // Ideally we would construct an empty iterator directly, but
```
ss2165, line 137: could you construct a new hierarchy with just one node and return the children of that?

ss2165 (commented):
Bit confused about whether the regions include the root or not?

src/hugr/region.rs
aborgna-q: Good catch, I updated it and added some simple tests.

src/hugr/region.rs
aborgna-q: Unfortunately no because of the lifetime requirement :(

I may add a `Children::new_empty(&'a Hierarchy) -> Self<'a>` method, though it seems a bit hacky.

ss2165: if it's a hack define it as a non-member function locally?

aborgna-q: It requires accessing the internals, so it has to be done from petgraph's side.
https://github.com/CQCL/portgraph/pull/81

ss2165 (approved):
👍

PR #196, opened 2023-06-21
Fixes #194

### LINKED ISSUE #194: `SimpleReplacement` after a copy fails
The following test fails. In the presence of copies it seems an edge is being
removed twice.

```rust
fn test_replace_after_copy() {
    let one_bit: Vec<SimpleType> = vec![ClassicType::bit().into()];
    let two_bit: Vec<SimpleType> = vec![ClassicType::bit().into(), ClassicType::bit().into()];

    let mut builder = DFGBuilder::new(one_bit.clone(), one_bit.clone()).unwrap();
    let inw = builder.input_wires().into_iter().exactly_one().unwrap();
    let outw = builder
        .add_dataflow_op(LeafOp::Xor, [inw, inw])
        .unwrap()
        .outputs();
    let [input, _] = builder.io();
    let mut h = builder.finish_hugr_with_outputs(outw).unwrap();

    let mut builder = DFGBuilder::new(two_bit, one_bit).unwrap();
    let inw = builder.input_wires();
    let outw = builder.add_dataflow_op(LeafOp::Xor, inw).unwrap().outputs();
    let [repl_input, repl_output] = builder.io();
    let repl = builder.finish_hugr_with_outputs(outw).unwrap();

    let orig = h.clone();

    let parent = h.root();
    let removal = h
        .nodes()
        .filter(|&n| h.get_optype(n).tag() == OpTag::Leaf)
        .collect();

    let first_out_p = h.node_outputs(input).next().unwrap();
    let embedded_inputs = h.linked_ports(input, first_out_p);
    let repl_inputs = repl
        .node_outputs(repl_input)
        .map(|p| repl.linked_ports(repl_input, p).next().unwrap());
    let inputs = embedded_inputs.zip(repl_inputs).collect();

    let outputs = repl
        .node_inputs(repl_output)
        .filter(|&p| repl.get_optype(repl_output).signature().get(p).is_some())
        .map(|p| ((repl_output, p), p))
        .collect();

    h.apply_rewrite(SimpleReplacement::new(
        parent, removal, repl, inputs, outputs,
    ))
    .unwrap();

    // Nothing changed
    assert_eq!(h.node_count(), orig.node_count());
}
```

lmondada:


lmondada:
I've removed `unlink_port` on the source port as in the presence of copy nodes this was removing additional links. It's technically superfluous as the nodes will be removed anyways.

The only drawback of not unlinking the port explicitly is that by adding additional edges before removing them, the copy nodes cannot reuse the ports that will be freed up by the old edge for the new edge.

Related to this, `validate` was only checking accepting copy nodes with a single outport, instead of accepting copy nodes where only one of the outputs was connected.

cqc-alec:
Happy with the change to `simple_replace.rs`. @aborgna-q Do you agree with the change to `validate.rs`?

aborgna-q:
Did the current replacement function fail that validation step? It seems to me that it shouldn't be creating copy nodes since it unlinks before re-linking.

But yes, the validation changes makes sense at least until we drop the unnecessary copy nodes in portgraph (when there's a single connection left).

lmondada:
> Did the current replacement function fail that validation step? It seems to me that it shouldn't be creating copy nodes since it unlinks before re-linking.

It did before I used `unlink_subport` (you can get rid of the unlinking altogether and it still works, as nodes are deleted later on anyways)

aborgna-q (approved):

PR #201, opened 2023-06-22
s/EdgeKind::Const/EdgeKind::Static/

### LINKED ISSUE #150: Rename EdgeKind::Const -> EdgeKind::Static
Spec was updated in #108

aborgna-q:
Closes #150

### DIFF
--- a/src/hugr/validate.rs
+++ b/src/hugr/validate.rs
@@ -228,7 +228,9 @@ impl<'a> ValidationContext<'a> {
         // Input ports and output linear ports must always be connected
         let mut links = self.hugr.graph.port_links(port_index).peekable();
         let must_be_connected = match dir {
-            Direction::Incoming => port_kind.is_linear() || matches!(port_kind, EdgeKind::Const(_)),
+            Direction::Incoming => {
+                port_kind.is_linear() || matches!(port_kind, EdgeKind::Static(_))
+            }
             Direction::Outgoing => port_kind.is_linear(),
         };
         if must_be_connected && links.peek().is_none() {
@@ -470,7 +472,7 @@ impl<'a> ValidationContext<'a> {
 
         match from_optype.port_kind(from_offset).unwrap() {
             // Inter-graph constant wires do not have restrictions
-            EdgeKind::Const(typ) => {
+            EdgeKind::Static(typ) => {
                 if let OpType::Const(ops::Const(val)) = from_optype {
                     return typecheck_const(&typ, val).map_err(ValidationError::from);
                 } else {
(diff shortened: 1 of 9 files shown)


src/hugr/validate.rs
```diff
@@ -228,7 +228,9 @@ impl<'a> ValidationContext<'a> {
         // Input ports and output linear ports must always be connected
         let mut links = self.hugr.graph.port_links(port_index).peekable();
         let must_be_connected = match dir {
-            Direction::Incoming => port_kind.is_linear() || matches!(port_kind, EdgeKind::Const(_)),
+            Direction::Incoming => {
```
acl-cqc, line 231: Whoa, hang on. Drive-by I realize, but shouldn't this be `Direction::Incoming => true` ?

src/types.rs
```diff
@@ -27,8 +27,8 @@ pub enum EdgeKind {
     ControlFlow,
     /// Data edges of a DDG region, also known as "wires".
     Value(SimpleType),
-    /// A reference to a constant value definition, used in the module region.
-    Const(ClassicType),
+    /// A reference to a static value definition, used in the module region.
```
acl-cqc, line 30: Drive-by again, but I think now that we have local constants you could remove ", used in the module region" ?

src/types.rs
```diff
@@ -46,16 +46,16 @@ impl EdgeKind {
...
 #[derive(Clone, Default, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct Signature {
     /// Value inputs of the function.
     pub input: TypeRow,
     /// Value outputs of the function.
     pub output: TypeRow,
-    /// Possible constE input (for call / load-constant).
-    pub const_input: TypeRow,
+    /// Possible static input (for call / load-constant).
+    pub static_input: TypeRow,
```
acl-cqc, line 58: Given we're renaming this, perhaps we should call it `static_inputs` as `TypeRow` allows more than one

acl-cqc: Oh, no, sorry, we call inputs `input`, fair enough

src/hugr/validate.rs
aborgna-q: Nop, incoming order ports can be disconnected.

acl-cqc (approved):
Looks good to me, thanks @aborgna-q . Just a couple of drive-bys, really...

src/hugr/validate.rs
aborgna-q: Ah, but the check is still wrong as we are allowing more things besides order edges. I'll fix it.

acl-cqc: Oh, yes, we use ports for those in the implementation, but not the spec.

In which case let's update the comment two lines above; the word "ports" there must be interpreted as ports in the spec, not ports in the implementation.

The word we're looking for might be "dataflow" (at least the spec defines dataflow edges as value or static)

PR #203, opened 2023-06-23
Fix edge serialization not compacting the nodes

### LINKED ISSUE #202: Serialisation fails with copy before node
The following serialisation round-trip fails (you need to add `serde_json` for this particular example). Is it due to having the same input to both arguments?

```rust
let tp: Vec<SimpleType> = vec![ClassicType::bit().into(); 2];
let mut dfg = DFGBuilder::new(tp.clone(), tp.clone()).unwrap();
let mut params: [_; 2] = dfg.input_wires_arr();
for p in params.iter_mut() {
    *p = dfg
        .add_dataflow_op(LeafOp::Xor, [*p, *p])
        .unwrap()
        .out_wire(0);
}
let h = dfg.finish_hugr_with_outputs(params).unwrap();
// Nothing wrong here
h.validate().unwrap();

let ser = serde_json::to_vec(&h).unwrap();
// This fails
let _: Hugr = serde_json::from_slice(&ser).unwrap();
```

aborgna-q:
If a copy node made the op nodes non-contiguous, the edge serialization didn't translate the endpoints `Node`s.

This is a one-liner fix + a new check on the node index when deserializing + the previously-failing test.

Closes #202

### DIFF
--- a/src/hugr/serialize.rs
+++ b/src/hugr/serialize.rs
@@ -9,7 +9,7 @@ use crate::ops::OpTrait;
 use crate::ops::OpType;
 use crate::Node;
 use portgraph::hierarchy::AttachError;
-use portgraph::{Direction, LinkError, NodeIndex};
+use portgraph::{Direction, LinkError, NodeIndex, PortView};
 
 use serde::{Deserialize, Deserializer, Serialize};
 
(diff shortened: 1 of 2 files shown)


src/hugr/serialize.rs
```diff
@@ -330,4 +342,23 @@ pub mod test {
         // test equality on SerHugrV0 instead.
         assert_eq!(ser_roundtrip(&ser_hugr), ser_hugr);
     }
+
+    #[test]
+    fn dgf_roundtrip() -> Result<(), Box<dyn std::error::Error>> {
```
lmondada, line 347: I'm being petty
```suggestion
    fn dfg_roundtrip() -> Result<(), Box<dyn std::error::Error>> {
```

lmondada (approved):
Thank you so much, it's super appreciated!! Would just fix the typo.

PR #210, opened 2023-06-24
Preserve sibling ordering in serialisation

lmondada:
Feel free to disagree with how I've solved this. Happy with any solution that works.


aborgna-q (approved):
Looks alright.

- I'd add some doc to `set_index` explaining the traversal and why it is needed.
- The implicit sibling order is under-specified in the spec. That should be a new Issue.

lmondada:
Man @aborgna-q it's the weekend you shouldn't be reading this! We'll see on Monday. I've got a fix, the rest is not urgent

cqc-alec:
>     * The implicit sibling order is under-specified in the spec. That should be a new Issue.

The spec [does say](https://github.com/cqc-alec/hugr/blob/f60fcffcd16f4491087fa61afd4b9bf45e6eb036/specification/hugr.md?plain=1#L154) that the outgoing hierarchy edges from a node have a definite linear ordering (i.e. siblings are ordered). Is something else missing?

lmondada:
@aborgna-q I've added some comments. Does it reflect what you had in mind?

aborgna-q:
Yes, that works!

I'm not sure it's a BFS, if you have a graph like
```mermaid
graph TD;
    root-->A;
    root-->B;
    A-->A1;
    A-->A2;
    B-->B1;
    B-->B2;
```
and call `set_index` on `B1` you'll set the indices `[root, A, B, B1, B2]`, ignoring `A1` and `A2`.

This works for the current serialization, so it's ok for now. I'll probably alter it a bit for #223

aborgna-q (approved):

PR #211, opened 2023-06-26
Add test that Hugr implements Send and Sync

acl-cqc:
Thanks Agustin for [the suggestion](https://github.com/CQCL-DEV/hugr/pull/186#discussion_r1241875055), definitely an easier form than the one I found  :-)


aborgna-q (approved):
👍

PR #214, opened 2023-06-26
Remove comments.

cqc-alec:
I think these are probably obsolete, but if any are not we should capture them as discussions.


ss2165 (approved):

acl-cqc:
I was about to say, do we not need `lookup`? But maybe the builder shows us that we don't! :) [and, very happy to see it go!]

PR #215, opened 2023-06-26
There is no alternative.

cqc-alec:
Or is there?

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -349,26 +349,6 @@ with those values unpacked. The graph may additionally take in a row
 `#x` (appended to `#i`) and return the same row (appended to `#o`). The
 contained graph may thus be evaluated more than once.
 
-**Alternate TailLoop**
-
-It is unclear whether this exact formulation of TailLoop is the most
-natural or useful. It may be that compilation typically uses multiple.
-Another is:
-
-A node with type `I -> O` with three children of types `A: I -> p + F`
-(the output is the row formed by extending the row `F` with the boolean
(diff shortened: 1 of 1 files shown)


acl-cqc:
This is not intended to suggest that any alternative to `TailLoop` exists right now. But I discussed a while back with @ss2165 how there are a great many alternative forms that we *could* have - typically, these allow to express more loops than `TailLoop` (loops that would otherwise have to be built as `Cfg` nodes), but are more complex to use.

I think we will only really see the motivation for this when we have a decent body/corpus of programs written in the Hugr, so it's not a decision we can take right now. @ss2165 agreed? But perhaps the doc just needs to be clearer that it's a note for the future rather than a description of the present - before we switched to `.md` it used to be "boxed out" so is there anything like that we can do in .md?

ss2165:
I wonder if it might be left as an issue for the future rather than a discussion in the spec?

cqc-alec:
Yeah I'm happy to create an issue for this. I think there is benefit to making this document a concise description of our current proposal. @acl-cqc would you be OK with moving it out to a separate issue?

acl-cqc:
Yes, absolutely. (An issue or a discussion? - neither of which were being used when we wrote this in the spec)

cqc-alec:
> Yes, absolutely. (An issue or a discussion? - neither of which were being used when we wrote this in the spec)

Probably a discussion actually.

cqc-alec:
#216

acl-cqc (approved):

acl-cqc:
Ok, thanks!

PR #220, opened 2023-06-27
Composable HUGRs

### LINKED ISSUE #120: Allow HUGR composition
E.g. build a DFG hugr then insert and wire in to a Def Hugr.

Requires `insert_graph` in portgraph.

aborgna-q:
### Adds 
- `HugrMut::insert_hugr` to insert a hugr as child of some node.
- `builder::Container::add_child_hugr` to do the same while building
- `builder::Dataflow::add_hugr_with_wires` to insert and automatically connect a dataflow hugr.

### Optional:

To call `portgraph`'s `insert_graph` we need to access the `LinkView` graph of the hugr. This works alright for `Hugr`s, but we didn't have such API for `RegionView` or `FlatRegionView`.
I added a pair of traits `view::AsPortgraph` and `view::sealed::AsPortgraph` that internally expose the graph views. The second is a private trait, so the external users cannot access this functionality.

- Do we want this? It makes the hugr insertion more flexible since it lets us use filtered graphs.
- Should we give the users access to the internal portgraph? It's annoying to expose the internals, but maybe some users may want to write their own views? (that could be an `internals` feature flag).

Closes #120


src/hugr/hugrmut.rs
```diff
@@ -115,6 +117,11 @@ pub(crate) trait HugrMut {
     /// In general this invalidates the ports, which may need to be resized to
     /// match the OpType signature.
     fn replace_op(&mut self, node: Node, op: impl Into<OpType>) -> OpType;
+
+    /// Insert another hugr into this one, under a given root node.
+    ///
+    /// Returns the root node of the inserted hugr.
+    fn insert_hugr(&mut self, root: Node, other: &impl AsPortgraph) -> Result<Node, HugrError>;
```
ss2165, line 124: I'm not sure making this generic over `impl AsPortgraph` is sound? The method is called `insert_hugr`, it would be helpful if the types stopped me inserting arbitrary portgraphs.

src/hugr/hugrmut.rs
```diff
@@ -267,6 +274,36 @@ where
...
+    fn insert_hugr(&mut self, root: Node, other: &impl AsPortgraph) -> Result<Node, HugrError> {
+        let node_map = self.as_mut().graph.insert_graph(other.as_portgraph())?;
+        let other_root = node_map[&other.root().index];
+
+        // Update hierarchy and optypes
+        self.as_mut().hierarchy.push_child(other_root, root.index)?;
+        for (&node, &new_node) in node_map.iter() {
+            other
+                .children(node.into())
+                .try_for_each(|child| -> Result<(), HugrError> {
```
ss2165, line 287: didn't know `try_for_each` existed, neat!

src/hugr/hugrmut.rs
```diff
@@ -267,6 +274,36 @@ where
...
+                    self.as_mut()
+                        .hierarchy
+                        .push_child(node_map[&child.index], new_node)?;
+                    Ok(())
+                })?;
+            let optype = other.get_optype(node.into());
+            self.as_mut().op_types.set(new_node, optype.clone());
```
ss2165, line 294: would it be more useful to have a way of moving the subgraph in to the bigger one rather than copying?

aborgna-q: Uhm, yeah.
Moving the graph and hierarchy contents is not that useful because they are mostly node pointers, but for the actual weights we should be able to `move` them.

src/hugr/hugrmut.rs
aborgna-q: We should be able to do `HugrView : sealed::AsPortgraph` and then just require `impl HugrView` here.

Users would still be able to implement `HugrView` on their wrappers if they have `AsRef<Hugr>`

src/hugr/hugrmut.rs
ss2165: do you want that to be a separate method or update  this one?

aborgna-q: I think we'll need to split
```rust
fn insert_hugr(&mut self, root: Node, other: AsMut<Hugr>) -> ...; // or just `other: Hugr`
fn insert_hugr_view(&mut self, root: Node, other: &impl HugrView) -> ...;
```
If we just update this one we won't be able to insert filtered regions (or other read-only views).

ss2165: i think `other: Hugr` is fine no? for `insert_hugr`

src/builder/build_traits.rs
```diff
@@ -90,10 +90,16 @@ pub trait Container {
     }
 
     /// Insert a finished HUGR as a child of the container.
-    fn add_child_hugr(&mut self, child: &impl HugrView) -> Result<Node, BuildError> {
+    fn add_hugr(&mut self, child: Hugr) -> Result<Node, BuildError> {
         let parent = self.container_node();
         Ok(self.hugr_mut().insert_hugr(parent, child)?)
     }
+
+    /// Insert a finished HUGR as a child of the container.
```
ss2165, line 98: update docstring

src/hugr/hugrmut.rs
```diff
@@ -115,6 +116,16 @@ pub(crate) trait HugrMut {
...
+
+    /// Insert another hugr into this one, under a given root node.
+    ///
+    /// Returns the root node of the inserted hugr.
+    fn insert_hugr(&mut self, root: Node, other: Hugr) -> Result<Node, HugrError>;
+
+    /// Insert another hugr into this one, under a given root node.
+    ///
+    /// Returns the root node of the inserted hugr.
+    fn insert_view(&mut self, root: Node, other: &impl HugrView) -> Result<Node, HugrError>;
```
ss2165, line 128:
```suggestion
    fn insert_from_view(&mut self, root: Node, other: &impl HugrView) -> Result<Node, HugrError>;
```

ss2165 (approved):

PR #223, opened 2023-06-27
Node canonicalization method in HugrMut

### LINKED ISSUE #208: Compact nodes in `HugrMut`
Or something similar, to ensure that a serialization roundtrip won't modify the node indices.

(Also requires making the root the first node)

aborgna-q:
Defines the canonical order as the BFS traversal of the hierarchy.
This updates the code from #210 to enforce the canonical order when serializing.

Closes #208

### DIFF
--- a/src/hugr/serialize.rs
+++ b/src/hugr/serialize.rs
@@ -1,7 +1,7 @@
 //! Serialization definition for [`Hugr`]
 //! [`Hugr`]: crate::hugr::Hugr
 
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 use thiserror::Error;
 
 use crate::hugr::{Hugr, HugrMut};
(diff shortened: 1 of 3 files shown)


src/hugr/serialize.rs
```diff
@@ -106,52 +106,25 @@ impl<'de> Deserialize<'de> for Hugr {
...
+        let mut node_rekey: HashMap<Node, Node> = HashMap::with_capacity(hugr.node_count());
+        // Generate a BFS-ordered list of nodes based on the hierarchy
+        let mut ordered = VecDeque::from([hugr.root()]);
         let mut index_counter = (0..).map(|i| NodeIndex::new(i).into());
+        while let Some(node) = ordered.pop_front() {
+            node_rekey.insert(node, index_counter.next().unwrap());
+            for child in hugr.children(node) {
+                ordered.push_back(child);
+            }
+        }
```
lmondada, line 124: Would it be possible to put this into a `fn canonical_rekey` that you can use both here and in `canonicalize_nodes`? Or is there a reason not to do that?

lmondada (changes requested):
Hey Agustin, thanks for the PR.

There is one bit of logic that you might be able to refactor out. I'm not quite sure where it would go, you might be able to put it in `HugrView`? Maybe there is a complication I am not seeing.

src/hugr/serialize.rs
aborgna-q: Good idea; I moved the loop into a `Hugr::canonical_order` method.

lmondada (approved):
Thank you!

PR #224, opened 2023-06-27
Test validation of cyclic graph + local consts, doc failures

acl-cqc:
* `test_cyclic` creates a cyclic DFG, and this validates :-(. We don't have anything in validation to look for this.
* `test_local_const` creates a graph with a LoadConstant node that is not in the causal cone of the Input. This passes validation, because the DfsPostOrder used in `validate_children_dag` currently traverses *both ways* along edges :-(, I think validation should fail here, and that the fix is in portgraph.
   * `test_local_const` then goes on to add the missing edge, and that still passes, but if we fixed the traversal to only go forwards, `validate_children_dag` would then fail as it wouldn't have visited the Const node (but would count it in the children); we'd then need the fix detailed inside `validate_children_dag`

I'd love to turn these into "expected-fail" tests but that looks quite hard, best plan appears to be to use [cargo-nextest](https://nexte.st/)

### DIFF
--- a/src/hugr/validate.rs
+++ b/src/hugr/validate.rs
@@ -731,7 +731,7 @@ mod test {
     use super::*;
     use crate::builder::{BuildError, ModuleBuilder};
     use crate::builder::{Container, Dataflow, DataflowSubContainer, HugrBuilder};
-    use crate::hugr::HugrMut;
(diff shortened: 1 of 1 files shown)


acl-cqc:
Closing in favour of #227 and #228

PR #225, opened 2023-06-27
Re-spec and Implement OutlineCfg rewrite

acl-cqc:
* I re-spec'd the rewrite as it was waay too complex, and the transformations I was doing to handle the more awkward cases can all be done via "replace" and "InsertIdentity" first. However, there's still a case where the region you want to nest includes the CFG Entry node (in which case there would be zero incoming edges, and the new basic block would be the entry node of the existing CFG). The spec rules this out but I don't think there is any workaround. I *could* implement it in the code but it would make it significantly more complicated (at least a couple of "if"s and a bunch more code). I'm wondering if we should add a special-case version of insertIdentity to add a new node before the entry node of a CFG; thoughts?
    * Update: see [comment here](https://github.com/CQCL-DEV/hugr/pull/225#issuecomment-1620480524)
* I also added a new method to the builder to allow me to create a CFGBuilder object on an existing CFG-node (so I can then use the CFGBuilder to construct a new Block and then a new CFG inside that).
* Tieing this all together with the `nest_cfgs.rs` analysis will be the next PR. At the least, I'll need a way to apply rewrites to a HugrView/HugrMut/subregion-of-a-Hugr...


src/hugr/rewrite/simple_replace.rs
```diff
@@ -227,8 +227,7 @@ impl Rewrite for SimpleReplacement {
         }
         // 3.5. Remove all nodes in self.removal and edges between them.
         for node in &self.removal {
-            h.graph.remove_node(node.index);
-            h.hierarchy.remove(node.index);
+            h.remove_node(*node).unwrap();
```
acl-cqc, line 230: IIRC I was looking at simple_replace for an example of "how to do this" and then found a better way...

cqc-alec:
> However, there's still a case where the region you want to nest includes the CFG Entry node (in which case there would be zero incoming edges, and the new basic block would be the entry node of the existing CFG). The spec rules this out but I don't think there is any workaround. I _could_ implement it in the code but it would make it significantly more complicated (at least a couple of "if"s and a bunch more code). I'm wondering if we should add a special-case version of insertIdentity to add a new node before the entry node of a CFG; thoughts?

Maybe this is the excuse we need to make the CFG's entry node special, exactly mirroring the exit node -- so it would not be a `DFB`, but just have outputs and no DFG inside it.

At the cost of introducing an extra node, if it makes the replacement API simpler I think it may be worth it. Thoughts?

src/hugr/rewrite/outline_cfg.rs
```diff
@@ -0,0 +1,273 @@
...
+}
+
+/// Errors that can occur in expressing an OutlineCfg rewrite.
+#[derive(Debug, Error)]
+pub enum OutlineCfgError {
+    /// The set of blocks were not siblings
+    #[error("The nodes did not all have the same parent")]
+    NotSiblings,
+    /// The parent node was not a CFG node
+    #[error("The parent node {0:?} was not a CFG but an {1:?}")]
```
cqc-alec, line 174:
```suggestion
    #[error("The parent node {0:?} was not a CFG but an {1:?}")]
```
"a" or "an"? I don't propose going to great lengths to solve this, unless you can think of a phrasing that avoids the question.

cqc-alec (changes requested):

acl-cqc:
> > However, there's still a case where the region you want to nest includes the CFG Entry node (in which case there would be zero incoming edges, and the new basic block would be the entry node of the existing CFG). The spec rules this out but I don't think there is any workaround. I _could_ implement it in the code but it would make it significantly more complicated (at least a couple of "if"s and a bunch more code). I'm wondering if we should add a special-case version of insertIdentity to add a new node before the entry node of a CFG; thoughts?
> 
> Maybe this is the excuse we need to make the CFG's entry node special, exactly mirroring the exit node -- so it would not be a `DFB`, but just have outputs and no DFG inside it.
> 
> At the cost of introducing an extra node, if it makes the replacement API simpler I think it may be worth it. Thoughts?

Yes, I had been wondering something similar. (Actually I'd been wishing we'd done that thing whereby Entry/Exit, and Input/Output for DFGs, were edges to/from the parent/container node, but never mind ;-) ). There are some possibilities:
* Require the entry node to have no predecessors. In which case there's no need to have any children, either (they could go outside the CFG and their outputs wired as inputs to the CFG). So...
* Add a new "Entry block" subtype (BasicBlock::Entry - we have BasicBlock::DFB and BasicBlock::Exit currently). I guess this has no inputs (implicitly these are the same as the CFG's inputs, just as the outputs of a BasicBlock::Exit are the same as the CFG's outputs), but the same `predicate_variants` and `other_outputs` as a normal DFB. [Note the Entry node could include a branch. If it doesn't branch, then really its single successor should have other predecessors, or else we could normalize by merging basic blocks.]
* Combine the proposed `BasicBlock::Entry` with `BasicBlock::Exit` into a `BasicBlock::Identity` and use that for both entry and exit. I guess this would have `predicate_variants` and `other_outputs`, with the idea that
   * for the entry block, the CFG inputs are equal to the Sum of the predicate_variants, plus the other outputs
   * for the exit block, the predicate_variants are the empty list (this is consistent with it having no successors) - that is, a 0-ary Sum, a type of which there are no instances. The inputs and other_outputs are then identical with the outputs of the CFG.

The latter seems both the most sensible and also a bit weird. Hmmm, I'll keep thinking.

I was also wondering if maybe we should just allow the set of blocks to include the entry block. Since the entry block may have other predecessors too, it may still share most of the code with the standard case (where entry block is *not* in the set of blocks) - I think I will code this up first and see how it looks.

cqc-alec:
> > > However, there's still a case where the region you want to nest includes the CFG Entry node (in which case there would be zero incoming edges, and the new basic block would be the entry node of the existing CFG). The spec rules this out but I don't think there is any workaround. I _could_ implement it in the code but it would make it significantly more complicated (at least a couple of "if"s and a bunch more code). I'm wondering if we should add a special-case version of insertIdentity to add a new node before the entry node of a CFG; thoughts?
> > 
> > 
> > Maybe this is the excuse we need to make the CFG's entry node special, exactly mirroring the exit node -- so it would not be a `DFB`, but just have outputs and no DFG inside it.
> > At the cost of introducing an extra node, if it makes the replacement API simpler I think it may be worth it. Thoughts?
> 
> Yes, I had been wondering something similar. (Actually I'd been wishing we'd done that thing whereby Entry/Exit, and Input/Output for DFGs, were edges to/from the parent/container node, but never mind ;-) ). There are some possibilities:
> 
>     * Require the entry node to have no predecessors. In which case there's no need to have any children, either (they could go outside the CFG and their outputs wired as inputs to the CFG). So...
> 
>     * Add a new "Entry block" subtype (BasicBlock::Entry - we have BasicBlock::DFB and BasicBlock::Exit currently). I guess this has no inputs (implicitly these are the same as the CFG's inputs, just as the outputs of a BasicBlock::Exit are the same as the CFG's outputs), but the same `predicate_variants` and `other_outputs` as a normal DFB. [Note the Entry node could include a branch. If it doesn't branch, then really its single successor should have other predecessors, or else we could normalize by merging basic blocks.]
> 
>     * Combine the proposed `BasicBlock::Entry` with `BasicBlock::Exit` into a `BasicBlock::Identity` and use that for both entry and exit. I guess this would have `predicate_variants` and `other_outputs`, with the idea that
>       
>       * for the entry block, the CFG inputs are equal to the Sum of the predicate_variants, plus the other outputs
>       * for the exit block, the predicate_variants are the empty list (this is consistent with it having no successors) - that is, a 0-ary Sum, a type of which there are no instances. The inputs and other_outputs are then identical with the outputs of the CFG.
> 
> 
> The latter seems both the most sensible and also a bit weird. Hmmm, I'll keep thinking.
> 
> I was also wondering if maybe we should just allow the set of blocks to include the entry block. Since the entry block may have other predecessors too, it may still share most of the code with the standard case (where entry block is _not_ in the set of blocks) - I think I will code this up first and see how it looks.

I'm thinking some sort of hybrid of these ideas:

* Have distinct node types `BasicBlock::{Entry,Exit,DFB}`. I think using the same node type for both entry and exit is a little confusing since the conditions on them are different, although operationally they are both the identity.
* Neither `Entry` nor `Exit` has children.
* Require the `Entry` node to have no predecessors, no inputs and a single successor; its outputs to that successor are just the inputs to the CFG. (We have a choice between doing the logic for the first branch in a DFB node or in a node preceding the CFG; here I choose the first option since it seems a bit less weird.)

acl-cqc:
Ok, I've gone with allowing the set to include the entry node (because it is, after all, just another DFB). You can see the changes necessary in this commit: ddf8fb62facc8a8294a2705c227ea3e5a0085419 - not that hard at all. Beyond that I've added a test of moving the entry node, refactoring some other test code in `nest_cfgs.rs`.

(Contrastingly allowing the exit block is much harder, somewhat because it's a separate block type, also the second node in children not the first, and if we move the exit block into the sub-cfg, we need a new exit block because the sub-cfg cannot *be* the exit block - whereas it can be the entry block.)

I wonder if we need a testutils file/class, maybe even a test_cfgs library - `nest_cfgs.rs` (in src/algorithm) is a long way away from src/hugr/rewrite/outline_cfg.rs...

acl-cqc:
If we want to add another BasicBlock type, etc., I think that's a wider change, and best in another PR - we could do that after this, but I don't think this is so bad that we have to do that first (and quite possibly not at all)

src/hugr/rewrite/simple_replace.rs
acl-cqc: Moved into #242

specification/hugr.md
```diff
@@ -1226,11 +1226,20 @@ nodes as children.
...
+originally specified. The set of basic blocks must satisfy constraints:
+* There must be exactly one node in the set with incoming (controlflow) edges
+ from nodes outside the set.
+* The set may not include the entry block
+* There must be exactly one edge from a node in the set to a node outside it.
+
+Situations in which multiple nodes have edges leaving the set, can be converted to
+this form by a combination of InsertIdentity operations and one Replace. Similarly,
+rather than movng the Exit block into the nested CFG, an identity node with a single
+successor can be added just before, with that edge used as the exitting edge.
```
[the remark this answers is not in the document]
acl-cqc: Ah yes! I was so surprised that I checked in a dictionary but quite right :)

specification/hugr.md
```diff
@@ -1226,11 +1226,20 @@ nodes as children.
...
+
+** Note ** This means there is no way to move the entry block (creating a new block to replace the entry block). Neither does `InsertIdentity` allow inserting a new identity block before the entry block ("inbetween" the containing CFG node and the entry block), as this is not an edge.
```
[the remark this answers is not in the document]
acl-cqc: This now gone as moving entry block allowed

src/hugr/rewrite/outline_cfg.rs
```diff
@@ -0,0 +1,273 @@
...
+        match (entry, exit_succ) {
+            (Some(e), Some((x, o))) => Ok((e, x, o)),
+            (None, _) => Err(OutlineCfgError::NoEntryNode),
+            (_, None) => Err(OutlineCfgError::NoExitNode),
+        }
+    }
+}
+
+impl Rewrite for OutlineCfg {
+    type Error = OutlineCfgError;
```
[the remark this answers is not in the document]
acl-cqc: It's required for the `Rewrite` trait, as the type of errors that can be returned by the `apply` method

src/hugr/rewrite/outline_cfg.rs
```diff
@@ -0,0 +1,273 @@
...
+    const UNCHANGED_ON_FAILURE: bool = true;
```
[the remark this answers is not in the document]
acl-cqc: again, required for `Rewrite` trait. (If you put anything in the `impl <Trait> for ....` block that *isn't* part of the Trait, you'll get an error. Methods "private" to OutlineCfg would be in `impl OutlineCfg` i.e. not for any trait.)

src/hugr/rewrite/outline_cfg.rs
acl-cqc: But let's at least be consistent ;-). Gone for "a".

src/builder/cfg.rs
```diff
@@ -95,6 +97,20 @@ impl<B: AsMut<Hugr> + AsRef<Hugr>> CFGBuilder<B> {
             inputs: Some(input),
         })
     }
+    pub(crate) fn from_existing(base: B, cfg_node: Node) -> Result<Self, BuildError> {
```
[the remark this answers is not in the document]
acl-cqc: It's not pub - but yes, docstring added.

I've clued about the entry node (the thing is that creating a CFGBuilder *doesn't* create an entry node - you have to do that explicitly later using e.g. `simple_entry_builder` and it will prepend that onto the list of children rather than append).

And added a test. (That creates a disconnected basic block, but it'll do for the test.)

src/hugr/rewrite/outline_cfg.rs
```diff
@@ -0,0 +1,273 @@
...
+            Ok(Some(n)) => n,
+            _ => return Err(OutlineCfgError::NotSiblings),
+        };
+        let o = h.get_optype(cfg_n);
+        if !matches!(o, OpType::CFG(_)) {
+            return Err(OutlineCfgError::ParentNotCfg(cfg_n, o.clone()));
+        };
+        let mut entry = None;
+        let mut exit_succ = None;
+        for &n in self.0.iter() {
```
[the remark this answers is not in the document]
acl-cqc: Yes, fair. Normally I wouldn't worry for a struct of only one element, but here it might be slightly non-obvious what the contents were, so done.

specification/hugr.md
```diff
@@ -1226,11 +1226,26 @@ nodes as children.
...
+   * *either* the set includes the CFG's entry node, and any edges from outside
+     the set (there may be none or more) target said entry node;
+   * *or* the set does not include the CFG's entry node, but contains exactly one
+     node which is the target of edges from outside the set.
+* The set may not include the exit block.
+* There must be exactly one edge from a node in the set to a node outside it.
+
+Situations in which multiple nodes have edges leaving the set, can be converted to
+this form by a combination of InsertIdentity operations and one Replace. Similarly/for example,
+rather than movng the Exit block into the nested CFG:
```
cqc-alec, line 1243:
```suggestion
rather than moving the Exit block into the nested CFG:
```

specification/hugr.md
```diff
@@ -1226,11 +1226,26 @@ nodes as children.
...
+ from nodes outside the set. Specifically,
+   * *either* the set includes the CFG's entry node, and any edges from outside
+     the set (there may be none or more) target said entry node;
+   * *or* the set does not include the CFG's entry node, but contains exactly one
+     node which is the target of edges from outside the set.
+* The set may not include the exit block.
+* There must be exactly one edge from a node in the set to a node outside it.
+
+Situations in which multiple nodes have edges leaving the set, can be converted to
+this form by a combination of InsertIdentity operations and one Replace. Similarly/for example,
```
cqc-alec, line 1242:
```suggestion
this form by a combination of InsertIdentity operations and one Replace. For example,
```

specification/hugr.md
```diff
@@ -1226,11 +1226,26 @@ nodes as children.
...
+rather than movng the Exit block into the nested CFG:
+1. an identity node with a single successor can be added onto each edge into the Exit
```
cqc-alec, line 1244:
```suggestion
1. An identity node with a single successor can be added onto each edge into the Exit.
```

specification/hugr.md
```diff
@@ -1226,11 +1226,26 @@ nodes as children.
...
+2. if there is more than one such incoming edge, these identity nodes can then all be combined
```
cqc-alec, line 1245:
```suggestion
2. If there is more than one such incoming edge, these identity nodes can then all be combined
```

specification/hugr.md
```diff
@@ -1226,11 +1226,26 @@ nodes as children.
...
+   by a Replace operation that removes them all and adds a single node (keeping the same number
+   of in-edges, but reducing to one out-edge to the Exit)
```
cqc-alec, line 1247:
```suggestion
   of in-edges, but reducing to one out-edge to the Exit).
```

cqc-alec (approved):
Looks good, just a few textual suggestions.

specification/hugr.md
acl-cqc: Yes, I thought "Similarly/for example" was a bit of a stretch ;). I've removed similarly but extended the class of things (just before) to include the example :)

specification/hugr.md
acl-cqc: Also, "identity" -> "Identity"

PR #227, opened 2023-06-28
Fix petgraph to traverse forwards-only, add test of local Const

acl-cqc:
Turns out that the DfsPostOrder in validate_children_dag was walking in both directions, not just forwards. This meant that it "found" nodes that should not be reachable, and validation passed on graphs containing nodes not reachable from the Input.

Changing the direction to forwards-only makes validation correctly reject such graphs, but then we need to special-case validation not to expect local Const/Def nodes to be forwards-reachable (the spec says these "causally precede" dataflow operations rather than requiring explicit Order edges as for e.g. loadConstant). I did this by adding a new OpTag::ScopedDefn (and removed OpTag::ConstInput while I was about it).

Note the first commit passes tests on its own so you can see the old behaviour and the *expected* behaviour in comments.


src/hugr/validate.rs
```diff
@@ -392,7 +392,13 @@ impl<'a> ValidationContext<'a> {
 
         let postorder = DfsPostOrder::new(&region, entry_node);
         let nodes_visited = postorder.iter(&region).filter(|n| *n != parent).count();
-        if nodes_visited != self.hugr.hierarchy.child_count(parent.index) {
+        // ModuleOp's (local Consts and Defs) should not be reachable from the Input node, so discount them
+        let non_module_op_count = self
+            .hugr
+            .children(parent)
+            .filter(|n| !OpTag::ModuleOp.contains(self.hugr.get_optype(*n).tag()))
```
ss2165, line 399: it seems somewhat incidental from the spec that `ModuleOp`s are the ones we wish to ignore, are we happy that it makes sense?

Also note if we go with this the `ConstInput` tag introduced in https://github.com/CQCL-DEV/hugr/blob/93c677c51c43ab37a09c8a118935b1d7566bb723/src/hugr/validate.rs#L342 can be removed.

acl-cqc: Well, I was thinking that all the other ModuleOp's *could* be "promoted" to be allowed locally - and it would make sense to, but we might choose not to because it makes life much harder for the linker - so this was a bit of future-proofing. Ick, indeed.

So I've removed ConstInput, and added a "ScopedDefn" category that I think says what we want quite neatly. How's thath?

ss2165 (approved):

PR #228, opened 2023-06-28
Test cyclic graph, doc failure

acl-cqc:
This creates a cyclic DFG, and demonstrates that it validates ok :-(. We don't have anything in validation to look for this yet.

It'd be great to turn this into an "expected-fail" tests but that looks quite hard, best plan appears to be to use [cargo-nextest](https://nexte.st/) and I've not done that here.

Or, could extend this PR to include a fix, i.e. actually detect cycles.

So maybe this isn't ready yet


acl-cqc:
Superceded (and fixed) by #468

PR #232, opened 2023-06-30
[refactor] Rename Alias/Func Def/Decl nodes

acl-cqc:
Def => FuncDefn
Declare => FuncDeclare (there are a couple of uses of "Declare" as a verb that I've preserved)
Alias => AliasDefn
AliasDeclare => AliasDecl

Also update spec.


ss2165 (approved):
I like Decl/Defn for length

ss2165 (approved):
👍

PR #234, opened 2023-06-30
[Tests] Rrefactor nest_cfg tests to use CFG-rooted Hugrs (and RegionView's)

acl-cqc:
The first commit shows we no longer need to pass the parent of the Cfg into the "CfgView" wrapper classes, because FlatRegionView does that work for us :-). (We still need this extra level of indirection, over a view onto the Hugr, because using types other than hugr::Node will allow us to increase the power of the analysis, as `HalfNodeView` demonstrates; but it's a little easier now).

The second commit rewrites the tests to use CFG-rooted Hugrs, avoiding the enclosing module/function/etc. stuff. This greatly simplifies the tests, and means that (for these tests) we don't even need FlatRegionView (but we would if performing this analysis on a real Hugr).

### DIFF
--- a/src/algorithm/half_node.rs
+++ b/src/algorithm/half_node.rs
@@ -2,7 +2,6 @@ use std::hash::Hash;
 
 use super::nest_cfgs::CfgView;
 use crate::hugr::view::HugrView;
-use crate::ops::handle::{CfgID, NodeHandle};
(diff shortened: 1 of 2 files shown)


aborgna-q (approved):
Nice!

PR #236, opened 2023-07-03
Cleanup some code

aborgna-q:
Some minor cleanup changes.

- Add more #[inline]s in `region.rs`.
- Merge `remove_node` and `remove_op` in HugrMut.
- Remove the unused `replacement.rs` file.

### DIFF
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -21,9 +21,6 @@ pub(crate) trait HugrMut {
     /// Panics if the node is the root node.
     fn remove_op(&mut self, node: Node) -> Result<(), HugrError>;
 
-    /// Remove a node from the graph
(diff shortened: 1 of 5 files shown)


src/hugr/hugrmut.rs
```diff
@@ -21,9 +21,6 @@ pub(crate) trait HugrMut {
     /// Panics if the node is the root node.
     fn remove_op(&mut self, node: Node) -> Result<(), HugrError>;
 
-    /// Remove a node from the graph
-    fn remove_node(&mut self, node: Node) -> Result<(), HugrError>;
```
croyzor, line 25: I think `remove_node` makes more sense as the name to keep

aborgna-q: I was mimicking `add_op`, I guess it makes more sense looking at the parameter name.

croyzor (approved):

PR #237, opened 2023-07-03
Node metadata

### LINKED ISSUE #235: Support for arbitrary metadata
From the [specification](https://github.com/CQCL-DEV/hugr/blob/main/specification/hugr.md#extensible-metadata):

> Each node in the HUGR may have arbitrary metadata attached to it. This
is preserved during graph modifications, and copied when rewriting.
Additionally the metadata may record references to other nodes; these
references are updated along with node indices.

This also needs [serialization ](https://github.com/CQCL-DEV/hugr/blob/main/specification/hugr.md#schema) support:

> Metadata could also be included as a map keyed by node index

aborgna-q:
This is a proposal for basic metadata support in the HUGR.
It adds a new SecondaryMap from nodes to a `OpMetadata` type, currently aliased to `serde_json::Value`.

The specification doesn't specify whether this map should be in or alongside the Hugr, but doing it like this lets us seamlessly integrate it into building, serializing, and rewriting (the serialization remains backwards-compatible).

The metadata itself is fully unstructured. We may want to enforce some minimal schema in the future, perhaps requiring a dictionary at the head so we can more easily merge different values, but that's left as future work.

Closes #235.

### DIFF
--- a/src/hugr.rs
+++ b/src/hugr.rs
@@ -41,6 +41,9 @@ pub struct Hugr {
 
     /// Operation types for each node.
     op_types: UnmanagedDenseMap<portgraph::NodeIndex, OpType>,
+
+    /// Operation metadata
+    metadata: UnmanagedDenseMap<portgraph::NodeIndex, OpMetadata>,
 }
 
 impl Default for Hugr {
@@ -149,6 +152,9 @@ impl Hugr {
     }
 }
 
+/// Arbitrary metadata for an operation.
+pub type OpMetadata = serde_json::Value;
+
 /// Internal API for HUGRs, not intended for use by users.
 impl Hugr {
     /// Create a new Hugr, with a single root node.
@@ -169,6 +175,7 @@ impl Hugr {
             hierarchy,
             root,
             op_types,
+            metadata: UnmanagedDenseMap::with_capacity(nodes),
         }
     }
 
--- a/src/hugr.rs
+++ b/src/hugr.rs
@@ -42,8 +42,8 @@ pub struct Hugr {
     /// Operation types for each node.
     op_types: UnmanagedDenseMap<portgraph::NodeIndex, OpType>,
 
-    /// Operation metadata
-    metadata: UnmanagedDenseMap<portgraph::NodeIndex, OpMetadata>,
+    /// Node metadata
+    metadata: UnmanagedDenseMap<portgraph::NodeIndex, NodeMetadata>,
 }
 
 impl Default for Hugr {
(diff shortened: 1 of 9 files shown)


src/hugr.rs
```diff
@@ -149,6 +152,9 @@ impl Hugr {
     }
 }
 
+/// Arbitrary metadata for an operation.
+pub type OpMetadata = serde_json::Value;
```
cqc-alec, line 156: This is really metadata for a _node_, which may or may not be an _operation_. I suggest calling it `NodeMetadata` instead?

src/hugr/hugrmut.rs
```diff
@@ -21,6 +23,14 @@ pub(crate) trait HugrMut {
     /// Panics if the node is the root node.
     fn remove_node(&mut self, node: Node) -> Result<(), HugrError>;
 
+    /// Returns the metadata associated with a node.
+    fn get_metadata_mut(&mut self, node: Node) -> &mut OpMetadata;
+
+    /// Returns the metadata associated with a node.
```
cqc-alec, line 29:
```suggestion
    /// Sets the metadata associated with a node.
```

src/hugr/region.rs
```diff
@@ -30,6 +30,9 @@ pub struct FlatRegionView<'g> {
 
     /// Operation types for each node.
     op_types: &'g UnmanagedDenseMap<portgraph::NodeIndex, OpType>,
+
+    /// Operation types for each node.
```
cqc-alec, line 34:
```suggestion
    /// Metadata for each node.
```

src/hugr/region.rs
```diff
@@ -181,6 +191,9 @@ pub struct RegionView<'g> {
...
```
cqc-alec, line 195:
```suggestion
    /// Metadata for each node.
```

cqc-alec (commented):
Nice

cqc-alec:
May also need to add metadata handling to the new `OutlineCFG` rewrite.

aborgna-q:
From what I gather, `OutlineCFG` should preserve the metadata of the moved children (since they keep their indices).
There's the question of whether we should add some metadata to the created entry/exit nodes, but I think that's rather for the user to decide.

cqc-alec:
> From what I gather, `OutlineCFG` should preserve the metadata of the moved children (since they keep their indices). There's the question of whether we should add some metadata to the created entry/exit nodes, but I think that's rather for the user to decide.

OK, yes that makes sense.

cqc-alec (approved):

PR #240, opened 2023-07-03
[new] Add lift nodes

croyzor:
Add lift nodes, which add a resource requirement to their inputs, as a type of leaf node.
Note that the representation of resources may be subject to change later in accordance with #189


src/ops/leaf.rs
```diff
@@ -71,6 +72,15 @@ pub enum LeafOp {
         /// The variants of the sum type.
         variants: TypeRow,
     },
+    /// A lift node, which adds a resource to an edge
+    Lift {
+        /// The type of the edge
```
aborgna-q, line 77:
```suggestion
        /// The types of the edges
```

src/ops/tag.rs
```diff
@@ -60,6 +60,8 @@ pub enum OpTag {
     BasicBlock,
     /// A control flow exit node.
     BasicBlockExit,
+    /// A lift node, which adds resource requirements to a wires type
+    Lift,
```
aborgna-q, line 64: Is this needed? Currently all leaf ops have `OpTag::Leaf`.

If the new tag is required:
- Change its parent to `OpTag::Lift => &[OpTag::Leaf],` (it's still a leaf).
- Return in in `LeafOp::tag`.

aborgna-q (approved):
👍

acl-cqc (commented):
I wonder if it's worth noting that these generalize Noop - in that a Lift with no additional resources is a Noop.

src/builder/dataflow.rs
```diff
@@ -363,4 +365,74 @@ mod test {
...
+    fn lift_node() -> Result<(), BuildError> {
+        let mut module_builder = ModuleBuilder::new();
+
+        let ab_resources = ResourceSet::from_iter(["A".into(), "B".into()]);
+        let c_resources = ResourceSet::singleton(&"C".into());
+        let abc_resources = ab_resources.clone().union(&c_resources);
+
+        let mut parent_sig = Signature::new_df(type_row![BIT], type_row![BIT]);
+        parent_sig.output_resources = abc_resources.clone();
+        let mut parent = module_builder.define_function("parent", parent_sig)?;
```
acl-cqc, line 379: Note that you can start the Hugr with a DFG (of arbitrary signature), skipping the ModuleBuilder altogether. See https://github.com/CQCL-DEV/hugr/pull/234 for example; I think here you'd want `DFGBuilder::new`

src/ops/tag.rs
croyzor: Removed

PR #241, opened 2023-07-04
[refactor] Add a more general constructor for DFGBuilder

croyzor:
Simplify the `lift_node` test case as suggested in #240 by @acl-cqc. I.e. remove the ModuleBuilder and using a DFGBuilder as the root of the graph.

In order to do this, a more general constructor for DFGBuilder is needed, which allows the resource requirements for the DFG to be specified

### DIFF
--- a/src/builder/dataflow.rs
+++ b/src/builder/dataflow.rs
@@ -59,19 +59,27 @@ impl DFGBuilder<Hugr> {
     /// # Errors
     ///
     /// Error in adding DFG child nodes.
-    pub fn new(
-        input: impl Into<TypeRow>,
-        output: impl Into<TypeRow>,
-    ) -> Result<DFGBuilder<Hugr>, BuildError> {
-        let input = input.into();
-        let output = output.into();
-        let signature = Signature::new_df(input, output);
+    pub fn new(signature: Signature) -> Result<DFGBuilder<Hugr>, BuildError> {
         let dfg_op = ops::DFG {
             signature: signature.clone(),
         };
         let base = Hugr::new(dfg_op);
         let root = base.root();
         DFGBuilder::create_with_io(base, root, signature)
+    }
+
+    /// Begin building a new DFG rooted HUGR, which doesn't contain any static
+    /// inputs or resource requirements
+    ///
+    /// # Errors
+    ///
+    /// Error in adding DFG child nodes.
+    pub fn new_df(
+        input: impl Into<TypeRow>,
+        output: impl Into<TypeRow>,
+    ) -> Result<DFGBuilder<Hugr>, BuildError> {
+        let signature = Signature::new_df(input.into(), output.into());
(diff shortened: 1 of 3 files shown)


acl-cqc (commented):
So Signature also allows `const_input`, which I don't think you want here. I think the right way is to
* Define a GraphType that's just input/output TypeRows + Resource delta
* Use that as the contents of a `ClassicType::Graph` (which currently has a ResourceSet *and* a Signature, which seems like too many ResourceSets!)
* Define Signature as the new GraphType plus const_input (another TypeRow)
* Then have this new method here taking a GraphType

I appreciate this involves doing the "Signature contains Resource-delta" first. So/but if we agree that's the right direction, I'd be happy to see some things "wrong" in the meantime. I guess that includes the approach here....

In which case (if this is the desired route) then I think I'd
* keep `new` called `new`. It's a DFGBuilder, so I think the `_df` is somewhat implicit.
* call the new thing `new_from_sig`, and assert that the const_input is empty, as it must be

Also, include updating the tests from #240? I admit I haven't grokked the use of this yet...is another way (avoiding the need for this at all) to use some ExternalOp/OpaqueOps that add resources? (Ah, I guess that'd require merging #186 first?)

acl-cqc:
> Also, include updating the tests from #240?

Ah, sorry, yes I see this is there now! Not as much simplification as I'd hoped but OK, thanks

acl-cqc:
> I think the right way is to
> 
> * Define a GraphType that's just input/output TypeRows + Resource delta
> * Use that as the contents of a `ClassicType::Graph` (which currently has a ResourceSet _and_ a Signature, which seems like too many ResourceSets!)
> * Define Signature as the new GraphType plus const_input (another TypeRow)
> * Then have this new method here taking a GraphType

Hmmm, hang on - if this (the first three) is the direction we can aim in, do we actually need the new method at all?

acl-cqc:
> Hmmm, hang on - if this (the first three) is the direction we can aim in, do we actually need the new method at all?

I think this is the crux of it. I think I'm skeptical about the rest because I don't see why a DFGBuilder should have a signature. Obviously it does at the moment: https://github.com/CQCL-DEV/hugr/blob/17c24e3501199d5e3038884cb9520039983da145/src/builder/build_traits.rs#L78-L89

croyzor:
This can wait until after a Signature refactor, but:
> So Signature also allows const_input, which I don't think you want here.

Why? Can DFG blocks not take const inputs?

> keep new called new. It's a DFGBuilder, so I think the _df is somewhat implicit.

`new`/`new_from_sig` is probably better, but for the record I was mimicking the `new_df` constructor for `Signature`

> I don't see why a DFGBuilder should have a ResourceSet

The DFG that it's building should have a ResourceSet, no? It's transforming inputs to outputs, potentially adding resource requirements in the process

PR #247, opened 2023-07-05
Transform CFGs into nested form

acl-cqc:
* Rename CfgView -> CfgNodeMap, and SimpleCfgView -> IdentityNodeMap
* Add CfgNester subtrait with a method to perform a single nesting transform
* Change CfgNodeMap + HalfNodeView to take the <H:HugrView/Mut> rather than an `&'a (mut) H`
* Implement CfgNester for IdentityNodeMap using OutlineCfg rewrite (no impl for HalfNodeView yet)
* Add transform_cfg_to_nested that recursively traverses any CfgNodeMap and nests all regions discovered in the CFG, innermost-first
* And transform_all_cfgs that runs the previous with an IdentityNodeMap for each CFG found in the entire Hugr


acl-cqc:
@mark-koch this is not really "for review" - but, please take a look, and see if you can use it?

mark-koch:
I made a Guppy branch to play around with this: https://github.com/CQCL-DEV/guppy/tree/toy/nesting

To set everything up, checkout the branch and run
```sh
python -m venv .venv
source .venv/bin/activate
pip install -e '.[dev]'
cd validator
maturin develop
cd ..
```

Then you can play around with the file `nesting.py`. The current example in the file produces a failed assertion, but maybe I'm using it wrong?

acl-cqc:
> The current example in the file produces a failed assertion, but maybe I'm using it wrong?

As per description:

> all the nest_cfgs.rs analysis works only on whole CFG-rooted Hugrs (!)

This is a temporary measure until we solve https://github.com/CQCL-DEV/hugr/issues/246, but I could easily add a parameter to the CFG node to analyze.

acl-cqc:
@mark-koch I've added a new function `transform_all_cfgs` that we can play around with for the time being, but this is probably not a high priority

src/algorithm/nest_cfgs.rs
```diff
@@ -43,21 +43,23 @@ use std::hash::Hash;
...
+
+/// A "view" of a CFG in a Hugr which allows basic blocks in the underlying CFG to be split into
+/// multiple blocks in the view (or merged together).
+/// `T` is the type of basic block; this can just be a BasicBlock (e.g. [`Node`]) in the Hugr,
+/// for an [IdentityCfgMap] if the extra level of indirection is not required. However, since
+/// SESE regions are bounded by edges between pairs of such `T`, such splitting may allow the
+/// algorithm to identify more regions than existed in the underlying CFG
 /// (without mutating the underlying CFG perhaps in vain).
-pub trait CfgView<T> {
+pub trait CfgNodeMap<T> {
```
acl-cqc, line 63: I'm renaming this as a could-be-separate part of the PR ;). While I'm about it, I could make the `T` a `type` member of the trait (`NodeType`, say)...

src/algorithm/nest_cfgs.rs
```diff
@@ -71,6 +74,92 @@ pub trait CfgView<T> {
...
+    view: &mut impl CfgNodeMap<T>,
+) -> Result<(), String> {
+    let edge_classes = EdgeClassifier::get_edge_classes(view);
+    let mut rem_edges: HashMap<usize, HashSet<(T, T)>> = HashMap::new();
+    for (e, cls) in edge_classes.iter() {
+        rem_edges.entry(*cls).or_default().insert(*e);
+    }
+
+    // Traverse. Any traversal will encounter edges in SESE-respecting order.
+    fn traverse<T: Copy + Eq + Hash + std::fmt::Debug>(
```
acl-cqc, line 96: This has to be recursive because the calls with the `stop_at` parameter are too complex to use the internal `node_stack`. I tried using [fix_fn](https://crates.io/crates/fix_fn) to make a recursive closure (which would avoid passing `edge_classes` and `rem_edges`) but that doesn't allow declaring type parameters or using them from the containing method :-(

src/algorithm/nest_cfgs.rs
```diff
@@ -71,6 +75,92 @@ pub trait CfgView<T> {
...
     /// Returns an iterator over the predecessors of the specified basic block.
     fn predecessors(&self, node: T) -> Self::Iterator<'_>;
+
+    /// Given an entry edge and exit edge defining a SESE region, mutates the
+    /// Hugr such that all nodes between these edges are placed in a nested CFG.
+    /// Returns the newly-constructed block (containing a nested CFG).
+    ///
+    /// # Panics
+    /// May panic if the two edges do not constitute a SESE region.
+    fn nest_sese_region(&mut self, entry_edge: (T, T), exit_edge: (T, T)) -> T;
```
acl-cqc, line 85: It is possible we should return some kind of Result, but it's not really clear what. The algorithm here only calls this for single-entry-single-exit regions identified by the analysis, so this should not fail. Perhaps a more complex splitting/merging CfgNodeMap might want to fail in some cases - but *that* I think we should leave for when a fuller splitting/merging CfgNodeMap is developed, i.e. not this PR.

src/algorithm/nest_cfgs.rs
```diff
@@ -445,14 +620,14 @@ pub(crate) mod test {
...
-        let h = cfg_builder.finish_prelude_hugr()?;
+        let mut h = cfg_builder.finish_prelude_hugr()?;
 
         let (entry, exit) = (entry.node(), exit.node());
         let (split, merge, head, tail) = (split.node(), merge.node(), head.node(), tail.node());
-        // There's no need to use a FlatRegionView here but we do so just to check
-        // that we *can* (as we'll need to for "real" module Hugr's).
-        let v: SiblingGraph = SiblingGraph::try_new(&h, h.root()).unwrap();
-        let edge_classes = EdgeClassifier::get_edge_classes(&SimpleCfgView::new(&v));
+        // ALAN here's where we need a WholeHugrView...tho it's good to try SiblingMut at least somewhere
```
acl-cqc, line 627: This refers to the WholeHugrView in #548 .... however, even if that goes in, I'm not sure whether it'd be good to use it or not now, given the traversal algorithm in `transform_cfg_to_nested` has been specifically designed to work on a `SiblingMut`. So I should either use WholeHugrView; or stick with SiblingMut and drop this comment, and the other mentions.

cqc-alec (commented):
Looks good, just small comments. Would be good to have some more complicated test cases soon.

src/algorithm/half_node.rs
```diff
@@ -110,8 +119,13 @@ mod test {
         // N(entry) -> head -> split            > N(merge) -> N(tail) -> N(exit)
         //               |          \-> right -/                 |
         //               \---<---<---<---<---<---<---<---<---<---/
-        // Allowing to identity two nested regions (and fixing the problem with a SimpleCfgView on the same example)
-        let v = HalfNodeView::new(&h);
+        // Allowing to identity two nested regions (and fixing the problem with an IdentityCfgMap on the same example)
+
+        // ALAN and another case for WholeHugrView
```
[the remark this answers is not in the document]
acl-cqc: Yes, I think it does not belong here. I thought that, since the RootHandle tag-checking had been added to main *since* this PR started, I could roll into this PR updates to the existing structure-detection code to use said tag-checking....but that leads to wanting #548 which has ended up significantly more complicated/longer-lived than expected, so I think it's best to leave this for #548 or another PR. I don't have an issue but I can roll it into that PR easily if this goes in first.

src/algorithm/nest_cfgs.rs
```diff
@@ -42,22 +42,26 @@ use std::collections::{HashMap, HashSet, LinkedList, VecDeque};
...
-pub trait CfgView<T> {
+use crate::{Direction, Hugr, Node};
+
+/// A "view" of a CFG in a Hugr which allows basic blocks in the underlying CFG to be split into
+/// multiple blocks in the view (or merged together).
+/// `T` is the type of basic block; this can just be a BasicBlock (e.g. [`Node`]) in the Hugr,
+/// for an [IdentityCfgMap] if the extra level of indirection is not required. However, since
+/// SESE regions are bounded by edges between pairs of such `T`, such splitting may allow the
+/// algorithm to identify more regions than existed in the underlying CFG, without mutating the
+/// underlying CFG just for the analysis - the splitting /merging) can then be performed by
```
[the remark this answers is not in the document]
acl-cqc: I've gone with "splitting and/or merging"

src/algorithm/half_node.rs
```diff
@@ -87,14 +90,20 @@ impl<H: HugrView> CfgView<HalfNode> for HalfNodeView<'_, H> {
         };
         succs.into_iter()
     }
+
+    fn nest_sese_region(&mut self, _i: (HalfNode, HalfNode), _x: (HalfNode, HalfNode)) -> HalfNode {
+        todo!()
```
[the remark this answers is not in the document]
acl-cqc: There is now, #593. (Previously the issue was only on JIRA and there is more discussion linked from there.)

acl-cqc:
Ok, thanks for the prodding :), I've changed a few things around...
* Separate the `nest_sese_region` function out into a different trait. This allows doing the analysis (not the transformation) using a read-only HugrView, which seems useful.
   * This required IdentityCfgMap (and HalfNodeView) to take `H: HugrView`s rather than references thereof, as you can't parameterize over "whether a reference is &mut or &". Region views are cloneable, *references* to Hugrs are trivially too, so this seems ok.
   * The change to trait structure also leads to the HalfNode transform disappearing entirely (not even a `todo!()`), but there's now an issue, as you suggested
* Dropped all references to WholeHugrView, enabling root-node-tagging is a separate issue, doesn't need to be here

src/algorithm/half_node.rs
```diff
@@ -110,8 +114,10 @@ mod test {
         // N(entry) -> head -> split            > N(merge) -> N(tail) -> N(exit)
         //               |          \-> right -/                 |
         //               \---<---<---<---<---<---<---<---<---<---/
-        // Allowing to identity two nested regions (and fixing the problem with a SimpleCfgView on the same example)
+        // Allowing to identity two nested regions (and fixing the problem with an IdentityCfgMap on the same example)
```
cqc-alec, line 117:
```suggestion
        // Allowing to identify two nested regions (and fixing the problem with an IdentityCfgMap on the same example)
```
(Not strictly part of this PR but may as well fix here.)

cqc-alec (approved):

PR #249, opened 2023-07-06
Add coverage check.

cqc-alec:
On this run the check took 3m38s which makes it the slowest of the CI checks ("tests (nightly, true)" took 3m0s).

I took the simplest approach which was:
* run on PR only;
* compute coverage both for PR branch and for main;
* compare only the total line coverage (ignoring function coverage and region coverage whatever that is, and ignoring individual file reports).

This means we don't need to store anything persistently, but we do need to run the coverage checks _twice_, which increases the run time.

Obviously we could refine the checks in future, e.g. to include function coverage statistics. We could also look into running on push to main and storing the result persistently, but I'm not very happy with how tket does this (by pushing a file to the magic `gh-pages` branch).

### DIFF
--- a/.github/workflows/compare-coverage
+++ b/.github/workflows/compare-coverage
@@ -26,4 +26,4 @@ echo "line coverage here: ${LINECOVPERCENT}%"
 MAIN_LINECOVPERCENT=`line_cov_percent main`
 echo "line coverage on main: ${MAIN_LINECOVPERCENT}%"
 
-if (( $(echo "${MAIN_LINECOVPERCENT} > ${LINECOVPERCENT}" | bc -l) )); then exit 1 ; fi
+if (( $(echo "${MAIN_LINECOVPERCENT} > ${LINECOVPERCENT} + 0.01" | bc -l) )); then exit 1 ; fi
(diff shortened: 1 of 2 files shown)


aborgna-q (approved):
In the [CI job](https://github.com/CQCL-DEV/hugr/actions/runs/5474055492/jobs/9968331897?pr=249#step:5:1):
> Run ./.github/workflows/compare-coverage
line coverage here: 73.40%
line coverage on main: 73.39%

Shouldn't that be equal?

Would it make sense to add an epsilon to the comparison, to avoid any rounding nuisances?

cqc-alec:
> In the [CI job](https://github.com/CQCL-DEV/hugr/actions/runs/5474055492/jobs/9968331897?pr=249#step:5:1):
> 
> > Run ./.github/workflows/compare-coverage
> > line coverage here: 73.40%
> > line coverage on main: 73.39%
> 
> Shouldn't that be equal?
> 
> Would it make sense to add an epsilon to the comparison, to avoid any rounding nuisances?

Good spot. There must be some random variation due to sampling I guess. Yes, adding an epsilon sounds like the best way around this.

aborgna-q:
> Run ./.github/workflows/compare-coverage
line coverage here: 73.39%
line coverage on main: 73.42%

O.o

cqc-alec:
If this gets too annoying with false negatives we can revert.

Issue #248, opened 2023-07-05
Improve layout of CFG diagram with non-local edges


cqc-alec:

aborgna-q:
This one:
https://github.com/CQCL/hugr/blob/adfc0c28512a717f85ece4716cc850fded741fdc/specification/hugr.md#L661

Currently looks like this

```mermaid
flowchart
    subgraph CFG
        direction TB
        subgraph Entry
            direction TB
            EntryIn["Input"] -- "P" --> Entry_["P?1"]
            Entry_ -- "[()|(P)]" --> EntryOut["Output"]
            F
        end
        subgraph BB1
            direction TB
            BB1In["Input"]
            BB1_["Const"] -- "[()]" --> BB1Out["Output"]
            G -- "angle" --> BB1Out
        end
        subgraph BB2
            direction TB
            BB2In["Input"] -- "P" --> H -- "P" --> BB2_["P?2"]
            BB2_ -- "[()|()]" --> BB2Out["Output"]
        end
        subgraph BB3
            direction TB
            BB3In["Input"]
            BB3_["Const"] -- "[()]" --> BB3Out["Output"]
            C -- "angle" --> BB3Out
        end
        subgraph Exit
        end
        Entry -- "0" --> BB1
        Entry -- "1" --> BB2
        BB2 -- "0" --> BB1
        BB2 -- "1" --> BB3
        BB1 -- "0" --> Exit
        BB3 -- "0" --> Exit
    end
    A -- "P" --> CFG
    A -- "qubit" --> Rz_out["Rz"]
    CFG -- "angle" --> Rz_out
    Rz_out -- "qubit" --> B
    A == "angle" ==> F
    F == "angle" ==> G
    linkStyle 12,13,14,15,16,17 stroke:#ff3,stroke-width:4px;
```

PR #250, opened 2023-07-06
Convert remaining images to mermaid and latex

cqc-alec:
The image showing a non-local edge into a nested DFG came out pretty badly, similar to the one with the CFG. But I think it's logically correct and we can leave the task of better rendering to later: #248 .

I re-did the Latex formulae, without use of colour: not sure how to get colours, but I hope they are clear enough without?


specification/hugr.md
```diff
@@ -1821,7 +1847,13 @@ These operations allow this.
     in first order graphs as straightforward (albeit expensive)
     manipulations of Graph `struct`s/protobufs\!
 
-<img src="attachments/2647818241/2647818326.png" height="64px">
+$\displaystyle{
```
croyzor, line 1850: This equation isn't rendering. I can't see anything obviously wrong, but it's the only one making use of newlines within latex - maybe github doesn't like that?

cqc-alec: Ah hadn't spotted that, let me try and fix it.

cqc-alec: Your guess was right. Made it a single line; seems OK now.

specification/hugr.md
```diff
@@ -315,30 +315,29 @@ and an Output node, whose outputs and inputs respectively match the inputs and
 outputs of the containing DFG.
 
 ```mermaid
-stateDiagram-v2
-    [*] --> DFG0
-    [*] --> DFG0
-
-    state DFG0 {
+flowchart
```
croyzor, line 318: I guess the style here is changed for consistency, but it doesn't look as good to my eye. The `flowchart` style seems to have a tendency to cross wires which makes it a bit harder to read

specification/hugr.md
```diff
@@ -615,7 +614,34 @@ Specifically, these rules allow for edges where in a given execution of
 the HUGR the source of the edge executes once, but the target may
 execute \>=0 times.
 
-<img src="attachments/2647818241/2647818338.png" width="768px">
+The diagram below is equivalent to the diagram in the [Dataflow](#dataflow)
+section above, but the input edge to "op3" has been replaced with a non-local
+edge from the surrounding DFG (the thick arrow).
```
croyzor, line 619: 👍

croyzor: It's unfortunate that the two outputs from the nested graph also look like a double arrow! Maybe we should add another op which merges them into a single output for neatness?

croyzor (approved):

specification/hugr.md
cqc-alec: Yes, I found it was impossible to draw non-local edges with the `stateDiagram-v2` structure.

specification/hugr.md
cqc-alec: I'll try that.

PR #251, opened 2023-07-07
 Statically query the tag of an operation

aborgna-q:
Adds a `OpTagged` trait, implemented for all operation structs.
`OpTrait` still has a `tag(&self)` method for dynamic computation (e.g. for `OpType`).

Drive-by: Re-export `OpTag` from `crate::op`.

### DIFF
--- a/src/ops.rs
+++ b/src/ops.rs
@@ -15,7 +15,6 @@ use crate::{Direction, Port};
 use portgraph::NodeIndex;
 use smol_str::SmolStr;
 
-use self::tag::OpTag;
 use enum_dispatch::enum_dispatch;
 
 pub use constant::{Const, ConstValue};
@@ -23,6 +22,7 @@ pub use controlflow::{BasicBlock, Case, Conditional, TailLoop, CFG};
 pub use dataflow::{Call, CallIndirect, Input, LoadConstant, Output, DFG};
 pub use leaf::LeafOp;
 pub use module::{AliasDecl, AliasDefn, FuncDecl, FuncDefn, Module};
+pub use tag::OpTag;
 
 #[enum_dispatch(OpTrait, OpName, ValidateOp)]
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
@@ -139,13 +139,24 @@ pub trait OpName {
     fn name(&self) -> SmolStr;
 }
 
+/// Trait statically querying the tag of an operation.
+///
+/// This is implemented by all OpType variants, and always contains the dynamic
+/// tag returned by `OpType::tag(&self)`.
+pub trait OpTagged {
+    /// The name of the operation.
+    fn static_tag() -> OpTag;
+}
+
 #[enum_dispatch]
 /// Trait implemented by all OpType variants.
 pub trait OpTrait {
     /// A human-readable description of the operation.
     fn description(&self) -> &str;
+
     /// Tag identifying the operation.
     fn tag(&self) -> OpTag;
+
     /// The signature of the operation.
     ///
     /// Only dataflow operations have a non-empty signature.
--- a/src/ops/constant.rs
+++ b/src/ops/constant.rs
@@ -11,8 +11,8 @@ use crate::{
 use downcast_rs::{impl_downcast, Downcast};
 use smol_str::SmolStr;
 
-use super::tag::OpTag;
-use super::{OpName, OpTrait};
+use super::OpTag;
+use super::{OpName, OpTagged, OpTrait};
 
 /// A constant value definition.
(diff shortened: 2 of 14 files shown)


src/ops/tag.rs
```diff
@@ -65,13 +65,32 @@ pub enum OpTag {
 impl OpTag {
     /// Returns true if the tag is more general than the given tag.
     #[inline]
-    pub fn contains(self, other: OpTag) -> bool {
-        self == other || other.parent_tags().iter().any(|&tag| self.contains(tag))
+    pub const fn contains(self, other: OpTag) -> bool {
+        // We cannot call iter().any() or even do for loops in const fn yet.
+        // So we have to write this ugly code.
+        if self.eq(other) {
```
croyzor, line 71: I don't really understand this logic. The comment says we're looking for something more general, but the name implies we're querying whether a given optag can contain another as a child node. If it's the latter then, say, a leaf node can't contain another leaf node

src/ops.rs
```diff
@@ -139,13 +139,24 @@ pub trait OpName {
...
+}
+
 #[enum_dispatch]
 /// Trait implemented by all OpType variants.
 pub trait OpTrait {
     /// A human-readable description of the operation.
     fn description(&self) -> &str;
+
     /// Tag identifying the operation.
     fn tag(&self) -> OpTag;
```
croyzor, line 158: Can we make `OpTagged` a requirement for `OpTrait` so we don't have to keep writing the same logic for `tag`? Or even move `tag` into `OpTagged`

src/ops/dataflow.rs
```diff
@@ -183,14 +180,12 @@ pub struct CallIndirect {
 impl_op_name!(CallIndirect);
 
 impl DataflowOpTrait for CallIndirect {
+    const TAG: OpTag = OpTag::FnCall;
```
croyzor, line 183: Seems fishy that `Call` and `CallIndirect` both have the same tag, but I guess the latter operation is disappearing shortly anyway so I'm not too worried

src/ops.rs
aborgna-q: `enum_dispatch` derives the `OpTrait` trait for `OpType` by calling each variant's method.
Implementing `OpTagged` for `OpType` could only return `Tag::Any`, which isn't that useful (and I'd say calling that method is probably a bug in the logic).

Would it be better to rename `OpTagged` to `StaticTag` or something ?

src/ops/constant.rs
```diff
@@ -22,13 +22,16 @@ impl OpName for Const {
...
+    const TAG: OpTag = OpTag::Const;
+}
 impl OpTrait for Const {
     fn description(&self) -> &str {
         self.0.description()
     }
 
     fn tag(&self) -> OpTag {
-        OpTag::Const
+        <Self as OpTagged>::TAG
```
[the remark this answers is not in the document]
aborgna-q: 🦀

src/ops.rs
aborgna-q: Ah, maybe we can add a default implementation requiring `OpTagged`, so we avoid all the `<Self as OpTagged>::TAG` lines.

src/ops/dataflow.rs
aborgna-q: Yeah, tags are sets of operations so it's technically correct. But we've only added them as-needed.

src/ops/tag.rs
aborgna-q: Ah, we are not talking about hierarchies here.

Tags are sets of operations (hence `TAG::Any` and `TAG::None`). So we are looking at set inclusion with `contains`.

The method `parent_tags` returns the immediate successors in the set inclusion order, so we traverse the order using that (perhaps it should be renamed...).

src/ops/tag.rs
```diff
@@ -128,8 +147,14 @@ impl OpTag {
...
     #[inline]
-    pub fn is_empty(&self) -> bool {
-        self == &OpTag::None
+    pub const fn is_empty(&self) -> bool {
+        matches!(self, &OpTag::None)
+    }
+
+    /// Constant equality check.
+    #[inline]
+    pub const fn eq(self, other: OpTag) -> bool {
```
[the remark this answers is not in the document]
aborgna-q: Unfortunately `==` is not a constant operation in general (because it is a trait method that can be implemented using non-constant things).
We can only do that with integers and other "structurally equivalent" primitive types that the compiler knows about.

Enums without data should be in that category, but rustc doesn't support that yet.

Hence the cast to use the constant int equality check.

src/ops/tag.rs
aborgna-q: `contains` -> `is_superset`
`parent_tags` -> `immediate_supersets`

hopefully more readable now

croyzor (approved):

PR #252, opened 2023-07-07
Remove LinearType; ClassicTypes are the only subtype of SimpleTypes

acl-cqc:
First step in refactoring the type hierarchy! (I'm heading towards the creation of a Hashable subset of ClassicType but it's a long way to go)

There seemed some confusion in the type hierarchy here. Partly this stems from that TypeRow always contains SimpleType, so statically Container<T> can have tuples of SimpleType (regardless of T). However, if we imagine that TypeRow might be parameterized by the type contained within it (hint hint - that PR is upcoming, but I found there was a lot to sort out first) then the old LinearType should not have the option of `Container<LinearType>` but `Container<SimpleType>` as that's the only way we'd be able to have "mixed" types like Tuple(Qubit,Int).

After sorting out Linear/Simple confusion it didn't seem that LinearType really meant anything: SimpleTypes could be anything (classic/linear/both), so you have to treat them as linear as *they might be*; only ClassicTypes can definitely be copied/discarded. This is a straightforward subtyping relation, i.e., you can do anything to a ClassicType that you can to a SimpleType, *plus* you can copy/discard it.

Hence, it seemed that the best thing was to inline LinearType into SimpleType. So, there is now SimpleType::Qubit, SimpleType::Qpaque, and then it seemed consistent to call it SimpleType::Qontainer *ducks*.

There were some other fixes/tidies first too, see individual commits, <strike>including that predicate types can have linear components.</strike> (EDIT: the latter now removed, work on the aforementioned parameterization of TypeRow revealed that including linear bits in predicates is a bigger can of worms than I'd realized.)

Thoughts/questions: I am not convinced by the removal of LinearType (de40c363ac7764489e32a2b68fc4b8bcfd2e3096). It might be useful just to separate these out, but I think the more useful idea might be to remove all plays on `is_linear` (replacing only with `is_classic`), so maybe even consider NonClassicType?!

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -156,9 +156,9 @@ As well as the type, dataflow edges are also parametrized by a
     [Non-local Edges](#non-local-edges)
 
 
-`SimpleType` $\supset$ `ClassicType`
-
 ```
+SimpleType ⊃ ClassicType -- In the absence of unicode: "SimpleType is a superset of ClassicType"
+
 EdgeKind ::= Hierarchy | Value(Locality, SimpleType) | Static(Local | Ext, ClassicType) | Order | ControlFlow
 
 Locality ::= Local | Ext | Dom
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -139,8 +139,9 @@ of edge for different relationships.
 
 `Value` and `Static` edges are sometimes referred to as _dataflow_ edges.
 A `Value` edge can carry data of any `SimpleType`: either a `ClassicType`
-(ordinary classical data) or a `LinearType` (data which cannot be copied,
-including quantum data). A `Static` edge can only carry a `ClassicType`. For
+(ordinary classical data, which can be freely copied or discarded) or a
+`LinearType` (which cannot - e.g. anything including quantum data).
+A `Static` edge can only carry a `ClassicType`. For
 more details see the [Type System](#type-system) section.
 
 As well as the type, dataflow edges are also parametrized by a
@@ -1042,7 +1043,7 @@ ClassicType ::= int<N>
               | Container(ClassicType)
 LinearType ::= Qubit
               | QPaque(Name, #)
-              | Container(LinearType)
+              | Container(SimpleType)  -- Sometimes called 'Qontainer'
 ```
 
 SimpleTypes are the types of *values* which can be sent down wires,
@@ -1065,8 +1066,8 @@ Container types are defined in terms of statically-known element types.
 Besides `Array<N>`, `Sum` and `Tuple`, these also include variable-sized
 types: `Graph`, `Map` and
 `List` (TODO: can we leave those to the Tierkreis resource?). `NewType`
-allows named newtypes to be used. Containers are linear if any of their
-components are linear.
+allows named newtypes to be used. Containers are classic (copyable) only
+if all of their components are classic.
 
 
 
@@ -1099,7 +1100,7 @@ ports can have any number of connected edges (0 is equivalent to a discard).
 
 Our linear types behave like other values passed down a wire. Quantum
 gates behave just like other nodes on the graph with inputs and outputs,
-but there is only one edge leaving or entering each port. In fully
+but there is only one edge leaving (or entering) each port. In fully
 qubit-counted contexts programs take in a number of qubits as input and
 return the same number, with no discarding. See
 [quantum resource](#quantum-resource)
@@ -1135,20 +1136,12 @@ resource requirements.
 We will likely also want to add a fixed set of attributes to certain
 subsets of `TYPE`. In Tierkreis these are called “type constraints”. For
 example, the `Map` type can only be constructed when the type that we
-map from is `Hashable`. For the Hugr, we may need this `Hashable`
-constraint, as well as a `Nonlinear` constraint. Finally there may be a
-`const-able` or `serializable` constraint meaning that the value can be
-put into a `const`-node: this implies the type is `Nonlinear` (but not
-vice versa).
+map from is `Hashable`. For the Hugr, we will need this `Hashable`
+constraint, as well as a `Classic` constraint.
 
-**TODO**: is this set of constraints (nonlinear, const-able, hashable)
-fixed? Then Map is in the core HUGR spec.
-
-Or, can extensions (resources) add new constraints? This is probably too
-complex, but then both hashable and Map could be in the Tierkreis
-resource.
-
-(Or, can we do Map without hashable?)
+**TODO**: fix this set of constraints (classic/copyable, hashable);
+extensions/resources *cannot* add new constraints. Use hashable for
+static type parameters, put Map in Tierkreis resource not core spec.
 
 ### Resources
 
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1097,7 +1097,7 @@ arbitrary `SimpleType`s and the subset of `ClassicType`s. Only `ClassicType`s
 may be used more than once or discarded; the former must always be used
 exactly once. This leads to a constraint on the HUGR that outgoing ports
 must have exactly one edge leaving them unless they are `ClassicType`, where
-outgoing ports may have anynum number of connected edges (0 is equivalent
+outgoing ports may have any number of connected edges (0 is equivalent
 to a discard). All incoming ports must have exactly one edge connected to them.
 
 # Our linear types behave like other values passed down a wire.
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -138,9 +138,9 @@ of edge for different relationships.
   children.
 
 `Value` and `Static` edges are sometimes referred to as _dataflow_ edges.
-A `Value` edge can carry data of any `SimpleType`: either a `ClassicType`
-(ordinary classical data, which can be freely copied or discarded) or a
-`LinearType` (which cannot - e.g. anything including quantum data).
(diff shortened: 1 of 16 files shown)


acl-cqc:
@ss2165 I remember you had some thoughts about the classic/linear separation/subtyping?

src/ops/constant.rs
```diff
@@ -174,9 +174,11 @@ impl ConstValue {
 
     /// Constant Sum over Tuples, used as predicates.
     pub fn predicate(tag: usize, variant_rows: impl IntoIterator<Item = TypeRow>) -> Self {
+        let variants = TypeRow::predicate_variants_row(variant_rows);
+        assert!(variants.get(tag) == Some(&SimpleType::new_unit()));
```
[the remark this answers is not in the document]
acl-cqc: This method is a bit of a halfway house.

Note it produces constant *values*, not types.

If the idea were that it only produced simple predicates, then there'd be no need for an `IntoIterator<Item=TypeRow>` - a simple `usize` (greater than the first usize) would suffice.

If this were intended to produce arbitrary predicates, then it should take not just a `tag: usize` but also a `ConstValue` the *type* of which would have to match the `tag`th element of `variant_rows`.

As it is, we end up with a weird midpoint, where you can produce simple predicates by passing in an iterator of (your chosen number of) *empty* TypeRows, but you can also use it to produce instances of distinctly *non*-simple predicates such as `Sum< () | (Int,Float) | (List<Float>, Bool) >` - but only the 0th variant of that one, not the 1st or 2nd....

Happy to sort this out in either direction in a preliminary PR!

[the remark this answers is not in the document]
ss2165: Yes I think this is an incorrect implementation, we need to add a parameter to take the constant value for the variant.

See `simple_predicate` above for the other one

ss2165:
> What sort of updates were you thinking of?

I liked the "Classic as subtype of Simple" notion, thought it would make things cleaner to explain in the spec.

ss2165:
Following up from https://github.com/CQCL-DEV/tket2/issues/35#issuecomment-1633866815:


I wonder if we ought to add a variant `SimpleType::Qlassic(ClassicType)` to mean no-copy version of any classic type?

cqc-alec:
> Following up from [CQCL-DEV/tket2#35 (comment)](https://github.com/CQCL-DEV/tket2/issues/35#issuecomment-1633866815):
> 
> I wonder if we ought to add a variant `SimpleType::Qlassic(ClassicType)` to mean no-copy version of any classic type?

I wouldn't use the "Q" convention for this though, as there's nothing quantum about it. Quantization (the thing you do to `Bit` to get `Qubit`) is a different transformation (which we may also want to define somewhere, as we move into qudits, cat qubits and CVs -- not sure if this would belong in the core spec or the Quantum resource though).

ss2165:
agreed, the Q suggestion was somewhat tongue in cheek. 

I actually think our reliance on "Classic" as naming is also problematic, since what we actually mean is non-linear, or copyable. 

I suggest being clearer and taking this variant approach, then propagating (`is_copyable` etc.). Renaming `SimpleType`->`Type`:

```rust
pub enum Basic {
    F64,
    ...
}
pub enum Type {
    Qubit,
    Copyable(Basic),
    Linear(Basic),
    ...
}
```

src/builder/build_traits.rs
```diff
@@ -660,7 +660,7 @@ fn wire_up<T: Dataflow + ?Sized>(
 #[derive(Debug, Clone, PartialEq, Eq)]
 enum ValueKind {
     Classic,
-    Linear(LinearType),
+    Linear(SimpleType),
```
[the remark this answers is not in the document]
acl-cqc: Done in #267

src/ops/constant.rs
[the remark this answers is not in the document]
acl-cqc: Done in #257

acl-cqc:
> Following up from [CQCL-DEV/tket2#35 (comment)](https://github.com/CQCL-DEV/tket2/issues/35#issuecomment-1633866815):
> 
> I wonder if we ought to add a variant `SimpleType::Qlassic(ClassicType)` to mean no-copy version of any classic type?

I think I'm gonna say "not in this PR". If we really want it, a Resource could add it by defining a non-classic opaque type parameterized by a ClassicType, but I really don't think it's a good basis for optimization/rewriting, at the least. Happy for it to turn into a discussion on Hugr, but maybe the TKET2 discussion would be a better place for "why do we want this"...

acl-cqc:
> I think removing `is_linear` and using `is_classic` everywhere is a good idea.

Largely done now.
* I've kept the serialization spec the same, i.e. there is still an `l: bool` that's the opposite of is_classic
* I've also kept EdgeKind::is_linear - roughly, this means, must I have exactly one wire from an outport, or can I have any number (>=0). I think that's a good concept, and `is_linear` is a reasonable name, but we could rename to `must_be_linear` or `must_be_used_exactly_once` if we wanted, or invert the sense and call it `allows_multiple_uses` (inaccurate as true also implies it allows 0 uses which is hardly multiple), `any_number_uses`, `arbitrary_uses`, etc....suggestions?

I move that we keep 'classic(al)' in this PR but I could rename SimpleType=>(Any)Type and ClassicType => CopyableType in another PR.

acl-cqc:
> > What sort of updates were you thinking of?
> 
> I liked the "Classic as subtype of Simple" notion, thought it would make things cleaner to explain in the spec.

I've had a go - 2ee236995b22cd0c325499c7fd1f1ab8e6c81920 is a minimal set of updates for correctness but keeping the notion of LinearType (and updating the section on type constraints....that could be another PR but this seemed a good time to do it). f23d301e2bf2ddb8bff2418eb25f2257a926b212 goes further but I'm not convinced that's a good plan, thoughts?

ss2165:
A few response comments but not a review:

1. I like `EdgeKind::is_linear`, concise and correct.
2. " I could rename SimpleType=>(Any)Type and ClassicType => CopyableType in another PR" - yes, happy
3. "Note that Classic also implies Discardable" Is that by defining Classic as NotLinear? If so, I still prefer `NonLinear` to `Classic`. To me `Classic` means not "quantum", and that is not actually the useful concept here.

acl-cqc:
> 3. "Note that Classic also implies Discardable" Is that by defining Classic as NotLinear? If so, I still prefer `NonLinear` to `Classic`. To me `Classic` means not "quantum", and that is not actually the useful concept here. I am happy to introduce some other term to mean the opposite of linear, but not sure what is best. "UseMany" vs. "UseOnce" maybe?

`NonLinear` generally I would say, well if it's not linear, then what is it ;-). In this case it probably works, but only because we don't admit any other possibility besides copyable+discardable. (e.g. Affine - Network sockets might be discardable but not copyable, say, but we don't have those in Hugr.)

But, other ideas, hmmm. `Digital`? Maybe too close to classical, but if you can express it as bits, then you can copy or discard those. `Copy0able`? Not exactly *precise* but at least the 0 tells you to look it up rather than rely on your first impression of copyable....

specification/hugr.md
```diff
@@ -155,9 +156,9 @@ As well as the type, dataflow edges are also parametrized by a
     [Non-local Edges](#non-local-edges)
 
 
-```
-SimpleType ::= ClassicType | LinearType
+`SimpleType` $\supset$ `ClassicType`
```
ss2165, line 159: image


this looks to be free-floating, could just go in the ``` block below?

specification/hugr.md
```diff
@@ -1092,18 +1093,19 @@ i.e. this does not affect behaviour of the HUGR. Row types are used
...
 
 For expressing and rewriting quantum programs we draw a distinction between
-`ClassicType` and `LinearType`, the latter being values which must be used
+arbitrary `SimpleType`s and the subset of `ClassicType`s. Only `ClassicType`s
+may be used more than once or discarded; the former must always be used
 exactly once. This leads to a constraint on the HUGR that outgoing ports
-of `LinearType` must have exactly one edge leaving them. `ClassicType` outgoing
-ports can have any number of connected edges (0 is equivalent to a discard).
+must have exactly one edge leaving them unless they are `ClassicType`, where
+outgoing ports may have anynum number of connected edges (0 is equivalent
```
ss2165, line 1100:
```suggestion
outgoing ports may have any number of connected edges (0 is equivalent
```

src/types/simple.rs
```diff
@@ -364,17 +316,6 @@ impl TypeRow {
         }
     }
 
-    /// Create a new row from a Cow slice of types.
-    ///
-    /// See [`type_row!`] for a more ergonomic way to create a statically allocated rows.
-    ///
-    /// [`type_row!`]: crate::macros::type_row.
-    pub fn from(types: impl Into<Cow<'static, [SimpleType]>>) -> Self {
```
ss2165, line 372: why is this method gone?

ss2165 (approved):
Just some minor text things

specification/hugr.md
acl-cqc: Not if we want to keep that $\supset$...ok, I have used unicode. I'm mildly impressed that that actually renders in a monospaced font...so have included a comment for anyone viewing whose font doesn't.

src/types/simple.rs
acl-cqc: Because we already
```
impl<T> From<T> for TypeRow
where
    T: Into<Cow<'static, [SimpleType]>>,
```
and AFAICT this wasn't doing anything. (I removed it first in its own commit and everything was happy.)

PR #253, opened 2023-07-07
Allow linear elements in predicates

acl-cqc:
This is a followup to #252 and, ok, maybe not as hard as I thought after all. But I'll have more certainty after the parameterization of TypeRow so I'll come back to this then.


acl-cqc:
This'll look quite different following #256

PR #254, opened 2023-07-10
src/ops/validate.rs: Common-up some type row calculations

acl-cqc:
Various of these exist already as methods on the classes in `controlflow.rs`...

### DIFF
--- a/src/ops/validate.rs
+++ b/src/ops/validate.rs
@@ -194,10 +194,8 @@ impl ValidateOp for super::TailLoop {
         &self,
         children: impl DoubleEndedIterator<Item = (NodeIndex, &'a OpType)>,
     ) -> Result<(), ChildrenValidationError> {
-        let expected_output = SimpleType::new_sum(vec![
-            SimpleType::new_tuple(self.just_inputs.clone()),
-            SimpleType::new_tuple(self.just_outputs.clone()),
-        ]);
+        let expected_output =
+            SimpleType::new_predicate(vec![self.just_inputs.clone(), self.just_outputs.clone()]);
         let mut expected_output = vec![expected_output];
         expected_output.extend_from_slice(&self.rest);
         let expected_output: TypeRow = expected_output.into();
--- a/src/ops/validate.rs
+++ b/src/ops/validate.rs
@@ -162,11 +162,10 @@ impl ValidateOp for super::Conditional {
         for (i, (child, optype)) in children.into_iter().enumerate() {
             let OpType::Case(case_op) = optype else {panic!("Child check should have already checked valid ops.")};
             let sig = &case_op.signature;
-            let predicate_value = &self.predicate_inputs[i];
-            if sig.input[0..predicate_value.len()] != predicate_value[..]
-                || sig.input[predicate_value.len()..] != self.other_inputs[..]
-                || sig.output != self.outputs
-            {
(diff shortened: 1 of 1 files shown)


src/ops/validate.rs
```diff
@@ -162,11 +162,10 @@ impl ValidateOp for super::Conditional {
...
             let OpType::Case(case_op) = optype else {panic!("Child check should have already checked valid ops.")};
             let sig = &case_op.signature;
-            let predicate_value = &self.predicate_inputs[i];
-            if sig.input[0..predicate_value.len()] != predicate_value[..]
-                || sig.input[predicate_value.len()..] != self.other_inputs[..]
-                || sig.output != self.outputs
-            {
+            if let Some(expected_inputs) = self.case_input_row(i) {
+                if sig.input == expected_inputs && sig.output == self.outputs {
+                    continue;
```
acl-cqc, line 167: I had to do it this way (i.e. inverting the predicate) to use `let Some` as I couldn't find a way to compare an `Option<&TypeRow>` with an `Option<TypeRow>`. Seems like I needed the opposite of `as_ref`. Maybe you can see a way?

acl-cqc: (that is, I wanted to write something like `if Some(sig.input) != &self.case_input_row(i)` but with some `&`s or similar on the LHS but couldn't work out what)

src/ops/validate.rs
```diff
@@ -162,11 +162,10 @@ impl ValidateOp for super::Conditional {
         for (i, (child, optype)) in children.into_iter().enumerate() {
             let OpType::Case(case_op) = optype else {panic!("Child check should have already checked valid ops.")};
             let sig = &case_op.signature;
-            let predicate_value = &self.predicate_inputs[i];
-            if sig.input[0..predicate_value.len()] != predicate_value[..]
-                || sig.input[predicate_value.len()..] != self.other_inputs[..]
-                || sig.output != self.outputs
-            {
+            if let Some(expected_inputs) = self.case_input_row(i) {
```
aborgna-q, line 165: Should this be an `unwrap` instead? We know this is a Case op.

aborgna-q (approved):
👌

src/ops/validate.rs
acl-cqc: Hmmm, actually that would work (and solve the problem below too), as the length check above should rule out None

aborgna-q: Alternatively, I think you could have written
```rust
Some(&sig.input) != self.case_input_row(i).as_ref()
```

acl-cqc:
@cqc-alec I think this PR demonstrates the issue with the coverage check. The check fails, because I have shortened/compacted the code that is tested, but not done anything to the untested code...

cqc-alec:
> @cqc-alec I think this PR demonstrates the issue with the coverage check. The check fails, because I have shortened/compacted the code that is tested, but not done anything to the untested code...

That is an issue yes. I guess we just ignore it in that case.

src/ops/validate.rs
acl-cqc: (Actually the RHS needs to be `(&self.case_input_row(i)).as_ref()` but the key is to put the `as_ref` on the other side, which should have been obvious when I said I needed its inverse...d'oh)

PR #256, opened 2023-07-10
Parameterize TypeRow by element type

acl-cqc:
There is a mass of churn here but in the end few major changes. There are two main benefits:
* It makes assumptions clear, for example that (in this version, but largely inherited from main) predicates are classical-only.
* The possibility of building a `ClassicType::Container` of a tuple including a `Qubit` is now statically ruled out
* It avoids having to declare code to handle some cases that can never occur. For example, ConstTypeError::LinearTypeDisallowed can now be removed, and `Signature::get` for "other" ports no longer needs an error case

Note this does not follow #253; we could do #253 before or after this, and doing this first will involve significantly more churn, but also be more explicit and probably find more bugs...so I'm doing this first ;)

We retain that there are "kind of" two ways to declare the SimpleType tuple-of-two-ints:
* `SimpleType::Classic` of `ClassicType::Container` of `Container::Tuple` of TypeRow of two `ClassicType::i64()`s
* `SimpleType::Qontainer` of `Container::Tuple` of TypeRow of two `SimpleType::Classic(ClassicType::i64())`s

...the idea being that factory methods like `new_predicate`, `new_tuple` and `new_sum` will pick the correct one.

Drawbacks:
* I had to change the serialization code (`src/types/simple/serialize.rs`) so that, rather than SerSimpleType containing TypeRows containing SimpleTypes, SerSimpleType now contains only (TypeRows of) SerSimpleTypes; flattening is thus recursively and once only rather than iteratively one stage at a time.
* I've disabled some `pyo3` for the time being. We might bring it back via wrapper `SimpleRow` and `ClassicRow`, say.

Secondly there is a new macro - `type_row!` builds `TypeRow<SimpleType>` only, there is now also `classic_row!`. However, both of these are used (almost) only in tests, so I have to wonder whether we need such control over efficient static allocation for test code. There is a possibility of renaming `type_row!` to `simple_row!` (another mass of churn!) but I'm holding off on that for now because we might rename SimpleType to Type or AnyType in which case `type_row` may be a reasonable name to keep. Renaming can follow in a later PR.


src/types/simple.rs
```diff
@@ -335,65 +353,78 @@ impl TypeRow {
...
+    pub fn get_mut(&mut self, offset: usize) -> Option<&mut T> {
         self.types.to_mut().get_mut(offset)
     }
 
-    #[inline]
-    /// Return the type row of variants required to define a Sum of Tuples type
-    /// given the rows of each tuple
-    pub fn predicate_variants_row(variant_rows: impl IntoIterator<Item = TypeRow>) -> Self {
-        variant_rows
+    fn try_convert_elems<D: PrimType + TryFrom<T>>(self) -> Result<TypeRow<D>, D::Error> {
```
acl-cqc, line 382: I'd like to just have
```
impl <F: PrimType, T: PrimType+TryFrom<F>> TryFrom<TypeRow<F>> for TypeRow<T>
```
but get a conflicting instance declaration...

src/types/simple.rs
```diff
@@ -335,65 +353,78 @@ impl TypeRow {
...
-            .map(|row| SimpleType::Classic(ClassicType::Container(Container::Tuple(Box::new(row)))))
-            .collect_vec()
-            .into()
+            .map(D::try_from)
+            .collect::<Result<_, _>>()?;
+        Ok(TypeRow::from(elems))
+    }
+
+    /// Converts the elements of this TypeRow into some other type that they can `.into()`
+    pub fn map_into<T2: PrimType + From<T>>(self) -> TypeRow<T2> {
```
acl-cqc, line 392: Similarly, it'd be good for this just to be an Into/From impl, but again that seems to raise conflicting-instance problems

src/types/simple/serialize.rs
```diff
@@ -65,19 +65,27 @@ pub(crate) enum SerSimpleType {
     },
 }
 
-impl<T: PrimType + Into<SimpleType>> From<Container<T>> for SerSimpleType {
+impl PrimType for SerSimpleType {
+    const LINEAR: bool = true;
```
acl-cqc, line 69: This is one of the downsides of parameterizing `TypeRow<T: PrimType>` and then trying to instantiate with `SerSimpleType` :-(

Another possibility is to remove `const LINEAR` and require instance `.is_linear(&self) -> bool` which can be implemented by recursive traversal for SimpleType (being cautious! Or inspecting only one level if one is less cautious), by constant-false for ClassicType, and by panicking for SerSimpleType...

src/types/simple.rs
```diff
@@ -287,12 +287,12 @@ impl<'a> TryFrom<&'a SimpleType> for &'a ClassicType {
 #[cfg_attr(feature = "pyo3", pyclass)]
 #[non_exhaustive]
 #[serde(transparent)]
-pub struct TypeRow {
+pub struct TypeRow<T: PrimType> {
```
acl-cqc, line 290: We don't strictly need `PrimType` here - `Clone+'static+std::fmt::Debug` would do, but seemed a lot messier. Only when I hit [serialization](https://github.com/CQCL-DEV/hugr/pull/256/files#r1258318916) did I start to wonder.

acl-cqc: PrimType *is* now just `Clone + 'static + std::fmt::Debug`, with `const CLASSIC: bool` staying in `serialize.rs` :-)

src/types.rs
```diff
@@ -11,14 +11,16 @@ use std::ops::Index;
...
+pub use simple::{ClassicRow, ClassicType, Container, SimpleRow, SimpleType};
 
 use smol_str::SmolStr;
 
 use crate::hugr::{Direction, Port};
 use crate::utils::display_list;
 use crate::{resource::ResourceSet, type_row};
 
+use self::simple::{PrimType, TypeRow};
+
```
aborgna-q, line 23: Export `PrimType` and `TypeRow` too?

src/types/simple.rs
```diff
@@ -272,24 +274,30 @@ impl<'a> TryFrom<&'a SimpleType> for &'a ClassicType {
...
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.write_char('[')?;
         display_list(self.types.as_ref(), f)?;
         f.write_char(']')
     }
 }
 
-#[cfg_attr(feature = "pyo3", pymethods)]
-impl TypeRow {
+//#[cfg_attr(feature = "pyo3", pymethods)] // TODO: expose unparameterized versions
```
aborgna-q, line 299: We'll need a wrapper struct for pyo3, so may as well move `len` and `is_empty` to the impl block below.

src/macros.rs
```diff
@@ -28,54 +28,107 @@ macro_rules! impl_box_clone {
...
+/// let sig: Signature = Signature::new_df(static_row, dynamic_row);
 ///
-/// let repeated_row: TypeRow = type_row![B; 2];
-/// assert_eq!(repeated_row, static_row);
+/// let repeated_row: SimpleRow = type_row![B; 3];
+/// assert_eq!(repeated_row, sig.output);
 /// ```
 #[allow(unused_macros)]
 #[macro_export]
 macro_rules! type_row {
```
aborgna-q, line 58: I agree with the renaming... in a separate PR

aborgna-q (approved):
Quite a bit of renaming noise, but LGTM

src/types/simple.rs
acl-cqc: Fair! Done

src/types.rs
acl-cqc: So this is something I debate. Obviously Rust requires that these types have to be public, but I'd kinda like to encourage users to use the aliases. The case in which a user should be using TypeRow....almost certainly is in a function parameterized `<T: PrimType> ...TypeRow<T>....`

Which is not the common case, but it is a *reasonable* case ;-). So ok, exported (within `types` but no further, as nothing is exported out of `types`)

acl-cqc:
> > I have to wonder whether we need such control over efficient static allocation for test code
> 
> It's mostly useful when dealing with Signatures of leaf ops. `OpType::signature` on something like a `CNOT` does not need to allocate at all, so checking things on flat dataflow regions needs a lot less indirection.
> 
> It may be useless complexity, but we don't have any benchmarks at the moment so I'd wait for that before making any refactoring.

Just to clarify - I did not mean to suggest removal of `type_row!`. [You give a good example of which I was not aware, to boot!] However, I wonder whether there is really value in having `classic_row!` as well. Uses thereof outside tests amount to two, in `simple.rs`, both times empty i.e. `classic_row![]`. Maybe we can revisit when/if we make predicates simple rather than classic...

src/types.rs
acl-cqc: However @aborgna-q one thing we might do is make PrimType sealed (i.e. can only be implemented inside the crate)....?

Issue #231, opened 2023-06-30
Validate_intergraph_edge needs updating


acl-cqc:
(See below)
1. Current spec says Static edges should be either Ext or Dom (rather than "from/to anywhere"), just that if Ext they don't need Order edges.
2. I note we check the type of Static edges from Const nodes but not from function definitions or declarations - is this correct? (This seems an odd place to check types at all, but if we are so doing, shouldn't it apply regardless of source node)

https://github.com/CQCL-DEV/hugr/blob/161adaebb2fc5c24736ef7ccb2c97a7b4bfcfd2c/src/hugr/validate.rs#L433-L451

acl-cqc:
@ss2165 @aborgna-q any comments?

aborgna-q:
Yeah, I agree that it needs some updating, and the typecheck could be moved somewhere else.

Also, the docstring for that method still references copy nodes.

PR #257, opened 2023-07-10
Cleanup ConstValue::predicate and some ConstTypeErrors

acl-cqc:
* Wrong signature
* Add assert that the value is of the type we'll claim. Such errors should be caught by `Hugr::validate()` but are not, so this is a temporary measure/sanity check, mostly while I'm playing around with types, but until #231 is fixed
   * Improve `typecheck_const` errors, and test some of the cases
   * Also add a test of a "bad" constant, but this is `should_panic` for now (i.e. the assert fails)
* Also remove `SimpleType::new_unit`. If it was going to be used anywhere, `ConstValue::simple_predicate` seems the most likely place, and...it isn't. (But, if we want to keep it, I can revert the last commit but let's go for the form in 76b47232dc4cfd47ec8c9b5b2872dcec14e65c83)

Links from code to issues may not be the right way, depending on how much we are thinking about the code being forked perhaps. I will add links from the issue to the code but of course tracking these over time may not be so clear.

### DIFF
--- a/src/ops/constant.rs
+++ b/src/ops/constant.rs
@@ -235,3 +235,71 @@ pub trait CustomConst:
 
 impl_downcast!(CustomConst);
 impl_box_clone!(CustomConst, CustomConstBoxClone);
+
+#[cfg(test)]
+mod test {
+    use super::ConstValue;
+    use crate::{
+        builder::{BuildError, Container, DFGBuilder, Dataflow, DataflowHugr},
+        type_row,
+        types::{ClassicType, SimpleType, TypeRow},
+    };
+
+    #[test]
+    fn test_predicate() -> Result<(), BuildError> {
+        let pred_rows = vec![
+            type_row![
+                SimpleType::Classic(ClassicType::i64()),
+                SimpleType::Classic(ClassicType::F64)
+            ],
+            type_row![],
+        ];
+        let pred_ty = SimpleType::new_predicate(pred_rows.clone());
+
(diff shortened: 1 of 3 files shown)


acl-cqc:
Oooh, hang on. That method is public. Might want to do something better than an `assert`, therefore.

ss2165 (commented):
I think returning a Result is appropriate here, as opposed to assert

acl-cqc:
An alternative might be to wait for #258 to be ready!

src/ops/constant.rs
```diff
@@ -229,3 +235,71 @@ pub trait CustomConst:
...
+        let c = b
+            .add_constant(ConstValue::predicate(
+                0,
+                ConstValue::Tuple(vec![]),
+                pred_rows,
+            ))
+            .unwrap();
+        let w = b.load_const(&c).unwrap();
+        // When #231 is fixed, there should be a validation error here instead
+        b.finish_hugr_with_outputs([w]).unwrap_err();
```
ss2165, line 303: I think it would be better to instead match on the correct error here, since then the fix to #231 should make this test fail, forcing it to be updated correctly (i.e. the `#[should_panic]` being removed)

ss2165 (approved):

src/ops/constant.rs
acl-cqc: I'll claim the excuse that until #258 I wasn't sure what the correct error *was*....but now I know ;-). Good call, done.

PR #258, opened 2023-07-11
[bugfix] Non-local edges: fix validation, Consts inside CFG, no Static Dom edges

### LINKED ISSUE #231: Validate_intergraph_edge needs updating

acl-cqc:
* Check Static edges meet locality requirements rather than allowing "anywhere"
* Don't allow Static Dom edges; but, allow Consts in CFGs (they can always be lifted as they have no inputs)
* Update spec
* Fix an amusing test inversion which made many tests fails with "Ancestor is not a CFG! ancestor_op: CFG" :-D

Fixes #231


src/hugr/validate.rs
```diff
@@ -430,44 +428,45 @@ impl<'a> ValidationContext<'a> {
...
-        }
+        let local = Some(from_parent) == to_parent;
 
-        match from_optype.port_kind(from_offset).unwrap() {
+        let is_static = match from_optype.port_kind(from_offset).unwrap() {
             // Inter-graph constant wires do not have restrictions
             EdgeKind::Static(typ) => {
                 if let OpType::Const(ops::Const(val)) = from_optype {
-                    return typecheck_const(&typ, val).map_err(ValidationError::from);
+                    typecheck_const(&typ, val).map_err(ValidationError::from)?;
```
acl-cqc, line 437: I do wonder if this should go in `validate_operation`. Admittedly there is nothing else there like it (but nor anywhere else that I can see - am I missing somewhere?)

src/hugr/validate.rs
```diff
@@ -430,44 +428,45 @@ impl<'a> ValidationContext<'a> {
...
                 } else {
                     // If const edges aren't coming from const nodes, they're graph
                     // edges coming from FuncDecl or FuncDefn
-                    return if OpTag::Function.is_superset(from_optype.tag()) {
-                        Ok(())
-                    } else {
-                        Err(InterGraphEdgeError::InvalidConstSrc {
+                    if !OpTag::Function.is_superset(from_optype.tag()) {
+                        return Err(InterGraphEdgeError::InvalidConstSrc {
```
acl-cqc, line 442: I'm also not really sure about this check. If a node says its `other` port/edgekind is static, isn't that sufficient - do we really mean to restrict only const/function nodes to doing that here?

PR #262, opened 2023-07-12
CustomConst's must be of CustomType

acl-cqc:


src/ops/constant.rs
```diff
@@ -227,7 +227,8 @@ pub trait CustomConst:
     fn name(&self) -> SmolStr;
 
     /// Returns the type of the constant.
-    fn const_type(&self) -> ClassicType;
+    // TODO it would be good to ensure that this is a *classic* CustomType not a linear one!
+    fn const_type(&self) -> CustomType;
```
acl-cqc, line 231: Note this is a different method from the `pub fn const_type` (not part of any trait) on line 118 of this file

acl-cqc: Should we rename one or the other??

acl-cqc:
@aborgna-q this is also relevant to your #261 in that Const nodes and TypeArg's are similar (although the latter are not nodes in the graph so not quite the same I suppose)...

aborgna-q:
> Const nodes and TypeArg's are similar

Yes, but `TypeArg`s scope is a lot more limited, they have no runtime effects once the operations have been instantiated so we can get away with simpler representations. We could go the `Box<dyn ...>` way as here, but there is no immediate need to.

src/ops/constant.rs
croyzor: I think the `ConstValue` one could just be renamed to `type`, since we already know it's const

croyzor: This one could be `custom_type`

croyzor (approved):

src/ops/constant.rs
acl-cqc: I tried `type`....but of course that's a reserved word! So `custom_type` seems good, thanks for the suggestion :)

PR #267, opened 2023-07-12
[refactor] Remove ValueKind, error on nonlocal linear edges in Builder

acl-cqc:
Another chunk pulled out from a one-line comment in #252, specifically [here](https://github.com/CQCL-DEV/hugr/pull/252/files#r1258050562) - which lead to a bugfix/enhancement, in that we can can easily catch linear non-local edges at build time.

### DIFF
--- a/src/builder/dataflow.rs
+++ b/src/builder/dataflow.rs
@@ -194,6 +194,7 @@ mod test {
     use crate::builder::build_traits::DataflowHugr;
     use crate::builder::{DataflowSubContainer, ModuleBuilder};
     use crate::hugr::validate::InterGraphEdgeError;
+    use crate::ops::handle::NodeHandle;
     use crate::ops::OpTag;
     use crate::ops::OpTrait;
     use crate::{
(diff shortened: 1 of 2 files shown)


src/builder/dataflow.rs
```diff
@@ -330,6 +331,34 @@ mod test {
...
+            let [i1] = f_build.input_wires_arr();
+            let noop = f_build.add_dataflow_op(LeafOp::Noop { ty: QB }, [i1])?;
+            let i1 = noop.out_wire(0);
+
+            let mut nested =
+                f_build.dfg_builder(Signature::new_df(type_row![], type_row![QB]), [])?;
+
+            let id = nested.add_dataflow_op(LeafOp::Noop { ty: QB }, [i1])?;
+
+            let nested = nested.finish_with_outputs([id.out_wire(0)])?;
```
acl-cqc, line 349: In fact this test passes already because `finish_hugr_with_outputs` calls `validate` which raises the same error. Following this PR, the error actually comes from *this* line, so I guess I could drop the next one...

ss2165 (approved):
👍

PR #271, opened 2023-07-13
feat: Move `dot_string` to `HugrView`

aborgna-q:
It's the same code, but now it can be called on any view.

Updates to portgraph 0.7.1 to fix https://github.com/CQCL/portgraph/pull/91.


ss2165:
💪

cqc-alec (approved):

aborgna-q:
Coverage test is failing, but it's the same code so I guess it's because the 2 extra lines added in the move.

PR #272, opened 2023-07-13
Add Logic resource.

cqc-alec:


src/extensions/logic.rs
```diff
@@ -0,0 +1,113 @@
...
+        HashMap::default(),
+        |arg_values: &[TypeArg]| {
+            let a = arg_values.iter().exactly_one().unwrap();
+            let n: Result<&u128, SignatureError> = match a {
+                TypeArg::Int(n) => Ok(n),
+                _ => {
+                    return Err(SignatureError::TypeArgMismatch(TypeArgError::TypeMismatch(
+                        a.clone(),
+                        TypeParam::Int,
+                    )));
```
aborgna-q, line 82:
```suggestion
                    return Err(TypeArgError::TypeMismatch(a.clone(), TypeParam::Int).into());
```

src/extensions/logic.rs
```diff
@@ -0,0 +1,113 @@
...
```
aborgna-q, line 58:
```suggestion
                    return Err(TypeArgError::TypeMismatch(a.clone(), TypeParam::Int).into());
```

src/extensions/logic.rs
```diff
@@ -0,0 +1,113 @@
...
+    );
+
+    let and_op = OpDef::new_with_custom_sig(
+        "And".into(),
+        "logical 'and'".into(),
+        vec![TypeParam::Int],
+        HashMap::default(),
+        |arg_values: &[TypeArg]| {
+            let a = arg_values.iter().exactly_one().unwrap();
+            let n: Result<&u128, SignatureError> = match a {
```
aborgna-q, line 52: Why not just `u128` ?

aborgna-q (approved):
LGTM

I do wonder though why do we need runtime-defined `Resource` structs, instead of a trait and `&dyn Resource` as we do everywhere else 🤔

src/extensions/logic.rs
cqc-alec: > Ah, I meant replacing all the `Result<&u128, SignatureError>` with `u128` since it's immediately unwrapped.
> 
> Not that important though; it gets elided by the compiler.

Ah, whoops.

PR #274, opened 2023-07-13
[refactor] Signature resources

croyzor:
Distinguish between abstract signatures and concrete signatures.

The former, now `AbstractSignature` is what is specified by a node signature written in e.g. YAML. It says what resources it uses, `dR`, but it can be specialised to any set of input resources (`IR -> IR + dR`). `Signature`, now embeds an `AbstractSignature` along with the resource set it was specialised to (`IR`).

Correspondingly, distinguish `OpType`, which is aligned with `AbstractSignature`, is the generic form of an operation and has an `AbstractSignature`, whereas `NodeType` is a node which has been or will be added to the graph, and knows what input resource set it has been specialised to.

`NodeType` should then be used in place of `OpType` when building hugrs, so that input resources can be specified when they are known. Likewise, `NodeType` should also be the data structure used when replacing nodes. At the moment these are left for a future PR.

N.B. This change renames `Signature`, which is exposed by pyo3, to `AbstractSignature`, and adds resources to the `NodeSer` struct for serialisation


acl-cqc (commented):
So prior to this PR, I think we can summarize:
* A `Node` has an `OpType` which generates a `Signature`
* The `Node` computes its own `Signature` by asking its `OpType`

This is too simplistic to do what we want, so in this PR:
* A `Node` has a `NodeType` which contains an `OpType` and an `input_resources`
* The `OpType` generates an `AbstractSignature`
* The `NodeType` turns this into a (Concrete)`Signature`
* The Node computes it own `Signature` by asking its `NodeType`

How about an intermediate form with fewer levels:
* A `Node` contains an `OpType` and an `input_resources`
* The `OpType` generates a `Signature`: this is the AbstractSignature in the PR, i.e. the same `Signature` we have now
* The `Node` computes its `Signature` by asking its `OpType`
* There is no structure (apart from `Node`) that combines `input_resources` with `Signature` (if you really need to do this, you can always make a tuple, but I think there is probably one function that needs this i.e. the resource-checker.)

src/builder/circuit.rs
```diff
@@ -134,13 +141,13 @@ mod test {
...
         type_row,
-        types::Signature,
+        types::AbstractSignature,
     };
 
     #[test]
     fn simple_linear() {
         let build_res = build_main(
-            Signature::new_df(type_row![QB, QB], type_row![QB, QB]),
+            AbstractSignature::new_df(type_row![QB, QB], type_row![QB, QB]).pure(),
```
ss2165, line 150: looks like keeping the `new*` methods on `Signature` might be quite useful? Could revert all these changes by just keeping `new_df`, `new`, `new_linear` and delegating to `AbstractSignature`.

src/builder/build_traits.rs
```diff
@@ -556,18 +595,18 @@ pub trait Dataflow: Container {
...
     inputs: Vec<Wire>,
 ) -> Result<(Node, usize), BuildError> {
     let [inp, _] = data_builder.io();
 
-    let op: OpType = op.into();
-    let sig = op.signature();
-    let op_node = data_builder.add_child_op(op)?;
+    let op = op.into();
+    let op_node = data_builder.add_child_op(op.clone())?;
+    let sig = op.op.signature();
```
ss2165, line 605: can avoid the op clone by swapping these lines back

ss2165 (changes requested):
Looks good but I think adding a couple of utility methods will make this PR a _lot_ smaller

src/hugr.rs
```diff
@@ -38,15 +40,67 @@ pub struct Hugr {
...
+    }
+
+    /// Use the input resources to calculate the concrete signature of the node
+    pub fn signature(&self) -> Option<Signature> {
+        self.input_resources
+            .clone()
+            .map(|rs| self.op.signature().with_input_resources(rs))
+    }
+}
+
```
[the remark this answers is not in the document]
croyzor: I agree that it would reduce the size of this PR, but I didn't do it because it didn't seem right. Going this way requires forcing a semantics onto the `OpType` that seemed out of `into()`'s purview

I think the more neutral `Into` implementation might even be `NodeType::open_resources`?

src/builder/circuit.rs
croyzor: Sure, I expected that commit to be controversial. 

By removing the constructors for `Signature` I was trying to avoid being biased towards thinking of no resources as the default case. Just because it's so in our tests so far doesn't mean it should be modus operandi of the API going forward.

But I might've gone too far in the other direction!

ss2165: Looks like the `.pure`s are dotted around builder code. Does that mean builder code needs to be update to use `AbstractSignature` instead of `Signature` everywhere?

specification/hugr.md
```diff
@@ -1580,6 +1580,8 @@ struct HUGR {
 struct Node{
   // parent node index
   parent: Int,
+  // The input resources to the node
+  input_resources: Option<ResourceSet>
```
croyzor, line 1584: I thinki this is a bit neater in terms of the serialised format than embedding a Nodetype. I assume the serde serialisation is being deferred to the dispatchees of `OpType` and can vary in length a fair bit, so it's nicer to have it at the end imo

src/builder/build_traits.rs
```diff
@@ -43,10 +45,15 @@ pub trait Container {
     /// Immutable reference to HUGR being built
     fn hugr(&self) -> &Hugr;
     /// Add an [`OpType`] as the final child of the container.
-    fn add_child_op(&mut self, op: impl Into<NodeType>) -> Result<Node, BuildError> {
+    fn add_child_op(&mut self, op: impl Into<OpType>) -> Result<Node, BuildError> {
         let parent = self.container_node();
         Ok(self.hugr_mut().add_op_with_parent(parent, op)?)
     }
+    /// Add an [`OpType`] as the final child of the container.
```
ss2165, line 52:
```suggestion
    /// Add a [`NodeType`] as the final child of the container.
```

src/builder/build_traits.rs
```diff
@@ -163,10 +176,24 @@ pub trait Dataflow: Container {
...
     fn add_dataflow_op(
         &mut self,
-        op: impl Into<NodeType>,
+        op: impl Into<OpType>,
+        input_wires: impl IntoIterator<Item = Wire>,
+    ) -> Result<BuildHandle<DataflowOpID>, BuildError> {
+        self.add_dataflow_node(NodeType::pure(op), input_wires)
+    }
+
+    /// Add a dataflow op to the sibling graph, wiring up the `input_wires` to the
```
ss2165, line 185:
```suggestion
    /// Add a dataflow [`NodeType`] to the sibling graph, wiring up the `input_wires` to the
```

ss2165 (commented):
Couple of docs changes

ss2165 (approved):

PR #279, opened 2023-07-17
More Constant tidies

acl-cqc:
* ConstValue::Opaque cached the type, but we never used it - so do so in `typecheck_const`
* Whereas `ConstTypeError::TypeMismatch` gives both expected and actual type, `ConstTypeError::Failed` gave only the expected type. However, we always have the actual type available via `const_type()`, so we can alway use the more informative TypeMismatch error
* The code for `typecheck_const` contains a catch-all match case `(ty, _) => ...TypeMismatch`. Since the `match` is on a pair, it would be easy to land here when what you actually meant was `Unimplemented`. Hence, rewrite the `match` into two steps so as to avoid the need for this case (that is: the Rust type checker confirms we have a separate case for each variety of ClassicType; within each of those the Rust type checker confirms we handle all constant values, which is actually pretty easy with the new, magic, `let...else` construct :-) )
* This also makes it very explicit which constants are unimplemented - most containers (except Sum+Tuple), Graph and String


src/hugr/typecheck.rs
```diff
@@ -114,41 +118,35 @@ pub fn typecheck_const(typ: &ClassicType, val: &ConstValue) -> Result<(), ConstT
...
-                    ty_act.into(),
-                ));
+        }
+        ClassicType::Container(_) => Err(ConstTypeError::Unimplemented(typ.clone())),
+        ClassicType::Graph(_) => Err(ConstTypeError::Unimplemented(typ.clone())),
+        ClassicType::String => Err(ConstTypeError::Unimplemented(typ.clone())),
+        ClassicType::Variable(_) => Err(ConstTypeError::ConstCantBeVar),
+        ClassicType::Opaque(ty) => {
+            let ConstValue::Opaque(ty_act, _val) = val else {return failure};
+            if ty_act != ty {
```
croyzor, line 144: This logic is the wrong way round

src/hugr/typecheck.rs
```diff
@@ -79,8 +75,10 @@ fn check_valid_width(width: HugrIntWidthStore) -> Result<(), ConstTypeError> {
 
 /// Typecheck a constant value
 pub fn typecheck_const(typ: &ClassicType, val: &ConstValue) -> Result<(), ConstTypeError> {
-    match (typ, val) {
-        (ClassicType::Int(exp_width), ConstValue::Int { value, width }) => {
+    let failure = Err(ConstTypeError::TypeMismatch(typ.clone(), val.const_type()));
+    match typ {
```
croyzor, line 79: I think it would be neater to keep matching on the pair `(typ, val)`, rather than starting every branch with a `let ... else ...`

croyzor: I think falling through to returning `failure` is also neater than returning it in the branches

acl-cqc: Ok, there was a goal besides neatness (that it's easy to miss a case when you match on a pair), but never mind - I did write `[subjective]` in the title of that commit :-) :-)

I've removed that commit so now am just dropping `Failure` in favour of only TypeMismatch, how's that?

croyzor (approved):

PR #280, opened 2023-07-18
[1 of 4] Rewrites use HugrMut methods rather than .hierarchy/.graph

acl-cqc:
This is tidier; never longer, and often much less long; and avoids many calls converting between Port/PortIndex/PortOffset, Node/NodeIndex, etc. etc.


acl-cqc:
This is failing the coverage check because it makes tested code more compact without affecting (or adding tests of) any untested code!

ss2165 (approved):
👍

acl-cqc:
@cqc-alec just FYI as you wrote some of this as well as I did, and I'm not sure either of us realized the power of HugrMut (TM :-) )

PR #283, opened 2023-07-24
Add HashableType as subset of ClassicType

acl-cqc:
[Note there are a few small/preliminary commits here before the bulk starts.]

This was surprisingly painless...
* Add a new TypeTag enum, replacing `is_classical()` bool with a ternary; `is_classical` and `is_hashable` move here.
* Keep `purely_classical` and `purely_hashable` on TypeRow - it's silly to fall back to having to handle TypeTag::Any when you have a row of types already known to be Classic.
* PrimType: add a `.tag()` method. However, keep SerializableType with its `const CLASS` as for well-formed instances we do not need to check the runtime value.
* There is some fun with `new_sum` and `new_tuple` requiring adding these methods to ClassicType *as well as* SimpleType

Unresolved issues:
* What about type variables. These are not really sorted before, specifically that a type variable is always considered classic (!). But I'm leaving this here. _(However, note: I guess the thorough thing would be to have variables that are simple; variables that are only classic; variables that are only hashable. But really any discussion of variables has to mention_
   * _where do variables occur? I think "in YAML type schemes" is the primary place, which doesn't need to even appear in the SimpleType hierarchy, but also probably in graphs. Note that Graphs are opaque and will be defined only in the Tierkreis resource, if that helps...._ 
   * _where are they bound? In Graph types (which will be an opaque type), and const values thereof.)_
   * _**What about function declarations?** Do these have to be monomorphic?_

The next step will be to remove parts of TypeParam and incorporate Hashable there, that can be another PR.

A final note is this leaves us with a lot of different containers: `SimpleType::Qontainer`, and `SimpleType::Classic(ClassicType::Container)`, as before, but now also `SimpleType::Classic(ClassicType::Hashable(HashableType::Container))`. And there will soon be `TypeParam::Container` too! I do have an idea here, as hinted in [this comment](https://github.com/CQCL-DEV/hugr/pull/283#discussion_r1272247479), but I think that can be another PR.

### DIFF
--- a/src/hugr/typecheck.rs
+++ b/src/hugr/typecheck.rs
@@ -9,7 +9,7 @@ use crate::hugr::*;
 
 // For static typechecking
 use crate::ops::ConstValue;
-use crate::types::simple::ClassicRow;
+use crate::types::simple::{ClassicRow, HashableType};
 use crate::types::{ClassicType, Container};
 
 use crate::ops::constant::{HugrIntValueStore, HugrIntWidthStore, HUGR_MAX_INT_WIDTH};
@@ -77,7 +77,7 @@ fn check_valid_width(width: HugrIntWidthStore) -> Result<(), ConstTypeError> {
 /// Typecheck a constant value
 pub fn typecheck_const(typ: &ClassicType, val: &ConstValue) -> Result<(), ConstTypeError> {
     match (typ, val) {
-        (ClassicType::Int(exp_width), ConstValue::Int { value, width }) => {
+        (ClassicType::Hashable(HashableType::Int(exp_width)), ConstValue::Int { value, width }) => {
             // Check that the types make sense
             check_valid_width(*exp_width)?;
             check_valid_width(*width)?;
@@ -126,8 +126,12 @@ pub fn typecheck_const(typ: &ClassicType, val: &ConstValue) -> Result<(), ConstT
             _ => Err(ConstTypeError::Unimplemented(ty.clone())),
         },
         (ty @ ClassicType::Graph(_), _) => Err(ConstTypeError::Unimplemented(ty.clone())),
-        (ty @ ClassicType::String, _) => Err(ConstTypeError::Unimplemented(ty.clone())),
-        (ClassicType::Variable(_), _) => Err(ConstTypeError::ConstCantBeVar),
+        (ty @ ClassicType::Hashable(HashableType::String), _) => {
+            Err(ConstTypeError::Unimplemented(ty.clone()))
+        }
+        (ClassicType::Hashable(HashableType::Variable(_)), _) => {
+            Err(ConstTypeError::ConstCantBeVar)
+        }
         (ClassicType::Opaque(ty), ConstValue::Opaque(_tm, ty2)) => {
             // The type we're checking against
             let ty_act = ty2.custom_type();
@@ -153,10 +157,10 @@ mod test {
 
     #[test]
     fn test_typecheck_const() {
-        const INT: ClassicType = ClassicType::Int(64);
+        const INT: ClassicType = ClassicType::int::<64>();
         typecheck_const(&INT, &ConstValue::i64(3)).unwrap();
         assert_eq!(
-            typecheck_const(&ClassicType::Int(32), &ConstValue::i64(3)),
+            typecheck_const(&HashableType::Int(32).into(), &ConstValue::i64(3)),
             Err(ConstTypeError::IntWidthMismatch(32, 64))
         );
         typecheck_const(&ClassicType::F64, &ConstValue::F64(17.4)).unwrap();
--- a/src/hugr/typecheck.rs
+++ b/src/hugr/typecheck.rs
@@ -125,19 +125,14 @@ pub fn typecheck_const(typ: &ClassicType, val: &ConstValue) -> Result<(), ConstT
             (Container::Sum(_), _) => Err(ConstTypeError::Failed(ty.clone())),
             _ => Err(ConstTypeError::Unimplemented(ty.clone())),
         },
-        (ty @ ClassicType::Hashable(HashableType::Container(c)), tm) => match c {
+        (ClassicType::Hashable(HashableType::Container(c)), tm) => {
             // Here we deliberately build malformed Container-of-Hashable types
             // (rather than Hashable-of-Container) in order to reuse logic above
-            Container::Tuple(row) => typecheck_const(
-                &ClassicType::Container(Container::Tuple(Box::new(row.clone().map_into()))),
+            typecheck_const(
+                &ClassicType::Container(c.clone().map_vals(&ClassicType::Hashable)),
                 tm,
-            ),
-            Container::Sum(row) => typecheck_const(
-                &ClassicType::Container(Container::Sum(Box::new(row.clone().map_into()))),
-                tm,
-            ),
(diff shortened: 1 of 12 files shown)


src/hugr/typecheck.rs
```diff
@@ -125,9 +125,26 @@ pub fn typecheck_const(typ: &ClassicType, val: &ConstValue) -> Result<(), ConstT
             (Container::Sum(_), _) => Err(ConstTypeError::Failed(ty.clone())),
             _ => Err(ConstTypeError::Unimplemented(ty.clone())),
         },
+        (ty @ ClassicType::Hashable(HashableType::Container(c)), tm) => match c {
+            // Here we deliberately build malformed Container-of-Hashable types
+            // (rather than Hashable-of-Container) in order to reuse logic above
```
acl-cqc, line 130: This is a bit of a code smell :-(, and hints that maybe there is a better way to do this. I think there is, but it's quite a big refactor, which I'll try to put up after this, as it's not urgent.

src/hugr/validate.rs
```diff
@@ -1141,7 +1141,7 @@ mod test {
         let lcst = h.add_op_with_parent(
             h.root(),
             ops::LoadConstant {
-                datatype: ClassicType::Int(1),
+                datatype: ClassicType::int::<1>(),
```
acl-cqc, line 1144: This does not have to be const, so could also be `HashableType::Int(1).into()`

ss2165 (commented):
Looks good, I support adding `is_classical` and `is_hashable` methods to `PrimType` instead of consts

src/types/simple/serialize.rs
```diff
@@ -32,51 +34,61 @@ pub(crate) enum SerSimpleType {
...
     },
     Var {
         name: SmolStr,
     },
 }
 
-impl PrimType for SerSimpleType {}
+impl PrimType for SerSimpleType {
+    fn class(&self) -> TypeClass {
+        unimplemented!()
```
acl-cqc, line 72: I don't think this should ever be called (we actually only use CLASS in this file, and instances of SerSimpleType should not be seen outside this file)

src/types/simple/serialize.rs
```diff
@@ -169,64 +191,120 @@ impl From<SerSimpleType> for SimpleType {
...
+                c: TypeClass::Any,
             } => Container::<SimpleType>::Tuple(Box::new(inner.map_into())).into(),
             SerSimpleType::Tuple {
                 row: inner,
-                l: false,
+                c: TypeClass::Classic,
             } => {
                 Container::<ClassicType>::Tuple(Box::new(inner.try_convert_elems().unwrap())).into()
             }
+            SerSimpleType::Tuple {
```
acl-cqc, line 211: Yeah, every SerSimpleType container has to appear here three ways. I think the only way round this is a macro, so I'll look into that

src/types/simple.rs
```diff
@@ -47,11 +48,55 @@ impl Display for SimpleType {
     }
 }
 
+/// Categorizes types into three classes according to basic operations supported.
+#[derive(
+    Copy, Clone, PartialEq, Eq, Hash, Debug, derive_more::Display, Serialize_repr, Deserialize_repr,
+)]
+#[repr(u8)]
+pub enum TypeClass {
+    /// Any value, including linear and quantum values;
```
acl-cqc, line 57: I'm nervous of calling this `Any` because it will not be the least-upper-bound / union once we have the fourth TypeClass corresponding to TypeArg/TypeParam (that is: Any is *not* a superset of TypeParam). We dislike Linear ;-) but maybe "AnyValue" or "Value" ??

src/types/simple.rs
```diff
@@ -47,11 +48,55 @@ impl Display for SimpleType {
     }
 }
 
+/// Categorizes types into three classes according to basic operations supported.
+#[derive(
+    Copy, Clone, PartialEq, Eq, Hash, Debug, derive_more::Display, Serialize_repr, Deserialize_repr,
+)]
+#[repr(u8)]
+pub enum TypeClass {
```
acl-cqc, line 56: TODO: rename to TypeTag (?)

acl-cqc:
Finally @ss2165....what tests do we have of serialization? I'd like to add some here...

src/types/simple/serialize.rs
acl-cqc: But this was one reason for keeping CLASS - we can (statically) avoid the need for `SerSimpleType::CLASS` and then duck out of this one dynamically

src/types/simple.rs
```diff
@@ -344,6 +531,23 @@ impl<T: PrimType> TypeRow<T> {
...
+    #[inline(always)]
+    pub fn purely_hashable(&self) -> bool {
+        self.types
+            .iter()
+            .map(PrimType::tag)
+            .all(TypeTag::is_hashable)
+    }
+
+    /// Returns the smallest [TypeTag] that contains all elements of the row
+    pub fn common_class(&self) -> TypeTag {
```
acl-cqc, line 544: This isn't actually used (we could use it for `purely_classical` and `purely_hashable` but that would be marginally less efficient). But it seems handy. If keeping it, we need a better name though. "common_tag" I guess, but that's not great. "union_tag"? "lub" (least upper bound)??

src/types/simple/serialize.rs
ss2165: yes a macro would make this much more readable

src/types/simple.rs
```diff
@@ -529,6 +541,14 @@ impl<T: PrimType> TypeRow<T> {
             .all(TypeClass::is_hashable)
     }
 
+    /// Returns the smallest [TypeClass] that contains all elements of the row
+    pub fn common_class(&self) -> TypeClass {
+        self.types
+            .iter()
+            .map(PrimType::class)
+            .fold(TypeClass::Hashable, TypeClass::union)
```
ss2165: neat!

src/types/simple.rs
ss2165: `shared_tag`?

ss2165 (commented):
looks good

src/types/simple.rs
acl-cqc: (renamed)

ss2165:
> Finally @ss2165....what tests do we have of serialization? I'd like to add some here...

https://github.com/CQCL-DEV/hugr/blob/b43b78bfb3dc449ffb050c4a9a445f166bd64ae9/src/hugr/serialize.rs#L240

src/types/simple/serialize.rs
acl-cqc: Et, voila! (My first macro!)
I particularly like that I was worried I'd get some of the Simple/Classic/Hashable stuff the wrong way round if I had to write it that many times, but now I only have to write that logic out once :)

acl-cqc:
> > Finally @ss2165....what tests do we have of serialization? I'd like to add some here...
> 
> https://github.com/CQCL-DEV/hugr/blob/b43b78bfb3dc449ffb050c4a9a445f166bd64ae9/src/hugr/serialize.rs#L240

Ah, ta, I keep forgetting that there are two serialize.rs's. I've used the roundtrip utility function there to add a couple of tests here. (Notably, including one of current behaviour, not necessarily desired behaviour)

ss2165 (approved):
👍

Issue #286, opened 2023-07-26
Opaque types can be hashable/classic/linear


acl-cqc:
1. TypeDef needs to store a function as to whether the type is classic or hashable (either "no it isn't" or "yes it is *iff* the following type arguments are classic-or-hashable-respectively); and then have a function `is_classic(&self, List<TypeArg>) -> bool` and `is_hashable` similarly
2. Along the way, it probably makes sense to move Opaque into Container (alongside Alias)

PR #297, opened 2023-07-26
move `Opaque` to `Container`

ss2165:
mentioned in #286

### DIFF
--- a/src/types/simple.rs
+++ b/src/types/simple.rs
@@ -29,8 +29,6 @@ pub enum SimpleType {
     Classic(ClassicType),
     /// A qubit.
     Qubit,
-    /// A linear opaque type that can be downcasted by the extensions that define it.
-    Qpaque(CustomType),
     /// A nested definition containing other linear types (possibly as well as classical ones)
     Qontainer(Container<SimpleType>),
 }
@@ -42,7 +40,6 @@ impl Display for SimpleType {
         match self {
             SimpleType::Classic(ty) => ty.fmt(f),
             SimpleType::Qubit => f.write_str("Qubit"),
-            SimpleType::Qpaque(custom) => custom.fmt(f),
             SimpleType::Qontainer(c) => c.fmt(f),
         }
     }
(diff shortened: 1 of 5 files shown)


src/types/simple.rs
```diff
@@ -125,6 +122,8 @@ pub enum Container<T: PrimType> {
     Array(Box<T>, usize),
     /// Alias defined in AliasDefn or AliasDecl nodes.
     Alias(SmolStr),
+    /// An opaque operation that can be downcasted by the extensions that define it.
```
acl-cqc, line 125:
```suggestion
    /// An opaque type that can be downcasted by the extensions that define it.
```

src/types/simple.rs
```diff
@@ -148,6 +147,7 @@ impl<T: PrimType> Container<T> {
             Self::Sum(variants) => Container::Sum(map_row(*variants, f)),
             Self::Array(elem, sz) => Container::Array(Box::new(f(*elem)), sz),
             Self::Alias(s) => Container::Alias(s),
+            Self::Opaque(custom) => Container::Opaque(custom),
```
acl-cqc, line 150: This is the tricky one. So if we have `c: Container<ClassicType>` we can make a `Container<SimpleType>` by calling `c.map_vals(&SimpleType::Classic)`. I can't see what else we could do here apart from just return another `Container::Opaque`...

src/types/simple.rs
acl-cqc: Let's also add "this will always have a TypeTag contained within that of the Container" (so a `Container<SimpleType>` could have an Alias whose tag was Classic, say). Or words to that effect - CustomType's don't *have* a TypeTag yet until the next PR...

acl-cqc: The same concern does apply to Alias, of course! (But there we have only a name, which doesn't tell us much on its own, whereas here we have a CustomType)

acl-cqc (approved):
Looks good, though there is a slight concern with map_vals. (Maybe we can reduce visibility of `map_vals` a bit?)

ss2165:
> Maybe we can reduce visibility of map_vals a bit?

Made it a private function in the only place it's used

PR #304, opened 2023-07-27
Allow any Hashable type as a TypeParam

acl-cqc:
* Add an option to parameterize by a hashable type - this should have gone in #283 really
* And, replace TypeParam::Int, TypeParam::String, etc., by "a value of any (given) Hashable Type"

The latter is a bit more complex (e.g. we now have to specify widths of ints). And, Hashable types include sums/tuples/lists/etc. - however, TypeArg (rather like ConstValue) doesn't let one provide values of most of these yet.

I've commoned up checking of int widths with `typecheck_const`, I think the next step might be to common up the representation of *values* of hashable type, extending both to include more containers. So yeah, either I should add tests, or go straight to that. I'm gonna see how hard a full-blown "ContainerValue" is...

### DIFF
--- a/src/types/type_param.rs
+++ b/src/types/type_param.rs
@@ -6,9 +6,10 @@
 
 use thiserror::Error;
 
+use crate::hugr::typecheck::{check_int_fits_in_width, ConstIntError};
 use crate::ops::constant::HugrIntValueStore;
 
-use super::{ClassicType, SimpleType};
+use super::{simple::Container, ClassicType, HashableType, SimpleType};
 
 /// A parameter declared by an OpDef. Specifies a value
 /// that must be provided by each operation node.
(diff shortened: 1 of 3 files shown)


src/types/type_param.rs
```diff
@@ -19,55 +20,84 @@ use super::{ClassicType, SimpleType};
...
-            return Err(TypeArgError::TypeMismatch(arg.clone(), param.clone()));
+        (TypeArg::Int(v), TypeParam::Value(HashableType::Int(width))) => {
+            check_int_fits_in_width(*v, *width).map_err(TypeArgError::Int)
         }
-    };
-    Ok(())
+        (TypeArg::String(_), TypeParam::Value(HashableType::String)) => Ok(()),
+        (arg, TypeParam::Value(HashableType::Container(ctr))) => match ctr {
+            Container::Opaque(_) => match arg {
+                TypeArg::CustomValue(_) => Ok(()), // Are there more checks we should do here?
```
acl-cqc, line 77: I don't think there's anything yet, but we might want to allow `TypeDef` providers to give code that checks a value is correct (e.g. deserializes to something appropriate)

src/types/type_param.rs
```diff
@@ -19,55 +20,84 @@ use super::{ClassicType, SimpleType};
...
+                _ => Err(TypeArgError::TypeMismatch(arg.clone(), param.clone())),
+            },
+            Container::List(elem) => check_type_arg(
+                arg,
+                &TypeParam::List(Box::new(TypeParam::Value((**elem).clone()))),
+            ),
+            Container::Map(_) => unimplemented!(),
+            Container::Tuple(_) => unimplemented!(),
```
acl-cqc, line 85: We could allow a TypeArg::List to provide the value of a Tuple type too - we'd have to check arity (list for an Array type) and then zip the elements. But I'm not sure this is a good plan, rather we should go for a fuller implementation of a ContainerValue

src/types/type_param.rs
```diff
@@ -19,55 +20,84 @@ use super::{ClassicType, SimpleType};
...
+    /// this is the value.
+    HashableType(HashableType),
+    /// Where the (Type/Op)Def declares a [TypeParam::Value] of type [HashableType::Int], a constant value thereof
     Int(HugrIntValueStore),
-    /// Where an argument has type [TypeParam::List]`<T>` - all elements will implicitly
-    /// be of the same variety of TypeArg, representing a `T`.
+    /// Where the (Type/Op)Def declares a [TypeParam::Value] of type [HashableType::String], here it is
+    String(String),
+    /// Where the (Type/Op)Def declares a [TypeParam::List]`<T>` - all elements will implicitly
+    /// be of the same variety of TypeArg, i.e. `T`s.
```
ss2165, line 53: So in a future PR replace this with something like `Value(HashableValue)`?

src/types/type_param.rs
```diff
@@ -19,55 +20,84 @@ use super::{ClassicType, SimpleType};
...
+            },
+            Container::List(elem) => check_type_arg(
+                arg,
+                &TypeParam::List(Box::new(TypeParam::Value((**elem).clone()))),
+            ),
+            Container::Map(_) => unimplemented!(),
+            Container::Tuple(_) => unimplemented!(),
+            Container::Sum(_) => unimplemented!(),
+            Container::Array(elem, sz) => {
+                let TypeArg::List(items) = arg else {return Err(TypeArgError::TypeMismatch(arg.clone(), param.clone()))};
```
ss2165, line 88: Do we need both list and array for typearg or is this looking ahead to list being removed from core spec?

ss2165 (approved):

src/types/type_param.rs
acl-cqc: We'll need both, if values have to distinguish (they might not) - or as you say we might remove one from the spec!

PR #305, opened 2023-07-27
Enable TypeDef reporting of type tag.

### LINKED ISSUE #286: Opaque types can be hashable/classic/linear

ss2165:
Closes #286

### DIFF
--- a/src/types/type_param.rs
+++ b/src/types/type_param.rs
@@ -9,7 +9,7 @@ use thiserror::Error;
 use crate::hugr::typecheck::{check_int_fits_in_width, ConstIntError};
 use crate::ops::constant::HugrIntValueStore;
 
-use super::{simple::Container, ClassicType, HashableType, SimpleType};
+use super::{simple::Container, ClassicType, HashableType, PrimType, SimpleType, TypeTag};
 
 /// A parameter declared by an OpDef. Specifies a value
 /// that must be provided by each operation node.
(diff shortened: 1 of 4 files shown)


src/types/type_param.rs
```diff
@@ -51,6 +51,19 @@ pub enum TypeArg {
     Value(serde_yaml::Value),
 }
 
+impl TypeArg {
+    /// Report [`TypeArg`] if param is a type
```
acl-cqc, line 60: Think you mean to link to TypeTag here rather than TypeArg?

src/types/type_param.rs
```diff
@@ -56,6 +56,19 @@ pub enum TypeArg {
     CustomValue(serde_yaml::Value),
 }
 
+impl TypeArg {
+    /// Report [`TypeArg`] if param is a type
+    pub fn tag(&self) -> Option<TypeTag> {
```
acl-cqc, line 61: Might be better to rename this, it's a bit different to `PrimType::tag`. Maybe `tag_if_type` (or `tag_of_type`) ?

src/types/type_param.rs
```diff
@@ -56,6 +56,19 @@ pub enum TypeArg {
...
+        match self {
+            TypeArg::Type(s) => Some(s.tag()),
+            TypeArg::ClassicType(c) => Some(c.tag()),
+            // assume list is well formed - all elements of same type
```
acl-cqc, line 65: I *think* this is a "kinding error" (Conor's term), i.e. this is operating at the wrong level of types vs values (vs kinds etc.)

E.g., if the TypeParam was `List<ClassicType>`, then two valid instantiations might be
1. `[ClassicType::Float, ClassicType::Hashable(HashableType::Int), .....]`
2. `[ClassicType::Hashable(HashableType::String), ClassicType::Hashable(HashableType::Int)].

So the "right" thing to do is probably to take the `union` of all the elements, which would declare Classic for the first example, but hashable for the second.

However, I'd also be happy to drop this case and just return None

acl-cqc (approved):
Neat - really like TypeDefTag - just a couple thoughts about the `tag` method

src/types/type_param.rs
ss2165: yes agreed - I've gone with None since I'm not convinced treating List[TypeArg] implicitly as a Type is the right thing to do.

PR #307, opened 2023-07-27
[Refactor] Derive PartialEq for ConstValue

acl-cqc:
AFAICT our manual implementation does exactly what the derive one would, except for one case - ConstValue::Opaque. But with a bit of "wizardry" (?!) we can get provide that ourselves and get Rust to derive the rest.

The extra level of tupling inside `ConstValue::Opaque` was because without that the `derive(serde::Serialize)` on ConstValue produced a complex error about moving out of a Box<dyn CustomConst>. A `struct OpaqueValue(CustomType, Box<dyn CustomConst>)` is one way to do that but just the tuple works too.

### DIFF
--- a/src/hugr/typecheck.rs
+++ b/src/hugr/typecheck.rs
@@ -155,7 +155,7 @@ pub fn typecheck_const(typ: &ClassicType, val: &ConstValue) -> Result<(), ConstT
             (Container::Sum(_), _) => {
                 Err(ConstTypeError::TypeMismatch(ty.clone(), tm.const_type()))
             }
-            (Container::Opaque(ty), ConstValue::Opaque(ty_act, _val)) => {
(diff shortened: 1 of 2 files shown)


aborgna-q:
Apparently the error that prevents us from deriving `PartialEq` with the barebones box is a known bug :/ 
[rust-lang/rust#78808](https://www.github.com/rust-lang/rust/issues/78808)

Here's a minimal case. Hopefully it will be fixed some day... [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3c3531a3d8bcb2cc10b041bb60320e95)

aborgna-q (approved):

aborgna-q (approved):
👍

PR #308, opened 2023-07-28
Correct field names for op/type args/params

ss2165:
Fly-by refactor out arg checking for future reuse

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -169,7 +169,7 @@ impl OpDef {
     pub fn new_with_yaml_types(
         name: SmolStr,
         description: String,
-        args: Vec<TypeParam>,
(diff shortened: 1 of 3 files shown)


src/resource.rs
```diff
@@ -198,7 +198,7 @@ impl OpDef {
             resource: Default::default(), // Currently overwritten when OpDef added to Resource
             name,
             description,
-            args,
+            params: args,
```
aborgna-q, line 201: Rename the fn parameter. (and same thing for `new_with_yaml_types`)

aborgna-q (approved):

src/resource.rs
ss2165: yeah just did that

PR #309, opened 2023-07-28
add resource and description fields for TypeDef

ss2165:
almost parity with `OpDef`

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -143,8 +143,8 @@ impl Debug for LowerFunc {
 /// TODO: Define a way to construct new OpDef's from a serialized definition.
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub struct OpDef {
-    /// The unique Resource owning this OpDef (of which this OpDef is a member)
-    pub resource: ResourceId,
+    /// The unique Resource, if any, owning this OpDef (of which this OpDef is a member)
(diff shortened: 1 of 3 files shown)


src/resource.rs
```diff
@@ -344,18 +348,24 @@ impl Resource {
     }
 
     /// Add an exported type to the resource.
-    pub fn add_type(&mut self, ty: TypeDef) {
+    pub fn add_type(&mut self, ty: TypeDef) -> Result<(), String> {
+        if ty.resource != ResourceId::default() {
```
aborgna-q, line 352: This never updates the empty `ty.resource`?

Maybe we should use an `Option<SmolStr>` to flag when the definition hasn't been attached to a resource, and overwrite it before inserting it to `self.types`? (and do the same for OpDef)

I just checked, and `SmolStr` works well with NPO (it takes 24 bytes both with and without the `Option`).

ss2165: sorry yeah not updating the resource was an oversight, took your suggestion

aborgna-q (approved):
We could do the same Option<> thing for OpDef to keep the symmetry.
But feel free to requeue the merge, we can do it later too.

ss2165:
> We could do the same Option<> thing for OpDef to keep the symmetry. 
I have done already?

PR #310, opened 2023-07-28
spec rename OpDef args -> params

ss2165:
matches implementation in #308

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -837,7 +837,7 @@ resources:
   types:
   - name: QubitVector
     # Opaque types can take type arguments, with specified names
-    args: [["size", Int]]
+    params: [["size", Int]]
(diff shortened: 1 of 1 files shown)


specification/hugr.md
```diff
@@ -893,20 +893,20 @@ resources:
     # Hence, no signature block => will look up a compute_signature in registry.
   - name: GraphOp
     description: "Involves running an argument Graph. E.g. run it some variable number of times."
-    args:
+    params:
       - r: ResourceSet
```
croyzor, line 897: Slightly off-topic, but I don't think `ResourceSet` as a param is valid

ss2165: raise issue?

croyzor (approved):

PR #312, opened 2023-07-28
[fix] Spec formatting

croyzor:
I'm assuming this wasn't done on purpose?


ss2165:
Looking at the blame, I think it was done on purpose in #252 because we stopped referring to linear types explicitly, and the rest is covered in the quantum resource

croyzor:
Regardless, I think it looks terrible:

image

Messes up our ToC:

image



and don't think it's worth keeping this way

ss2165 (changes requested):
lines can be removed

ss2165 (approved):

PR #315, opened 2023-07-28
Interactions between resource definitions and instances

ss2165:
Currently build on #305  but not strictly dependent.

Bring TypeDef and CustomType to parity with OpDef and OpaqueOp (most importantly by making them point to the resource they belong to).

Then generalise behaviour across the definition structs and the instance structs using sealed traits.

Then implement methods in to generate instances from definitions by providing TypeArgs, and then to check those instances are valid against the definitions.

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -147,8 +147,43 @@ impl Debug for LowerFunc {
     }
 }
 
-/// Sealed trait for shared type parametrised functionality between [`TypeDef`] and [`OpDef`].
-pub trait TypeParametrisedInternal: sealed::SealedDef {
+/// Concrete instantiations of types and operations defined in resources.
+trait CustomConcrete {
+    fn def_name(&self) -> &SmolStr;
+    fn type_args(&self) -> &[TypeArg];
+    fn parent_resource(&self) -> &ResourceId;
+}
+
+impl CustomConcrete for OpaqueOp {
+    fn def_name(&self) -> &SmolStr {
+        self.name()
+    }
+
+    fn type_args(&self) -> &[TypeArg] {
+        self.args()
+    }
+
+    fn parent_resource(&self) -> &ResourceId {
+        self.resource()
+    }
+}
+
+impl CustomConcrete for CustomType {
+    fn def_name(&self) -> &SmolStr {
+        self.name()
+    }
+
+    fn type_args(&self) -> &[TypeArg] {
+        self.args()
+    }
+
+    fn parent_resource(&self) -> &ResourceId {
+        self.resource()
+    }
+}
+
+/// Type-parametrised functionality shared between [`TypeDef`] and [`OpDef`].
+trait TypeParametrised: sealed::SealedDef {
     /// The concrete object built by binding type arguments to parameters
     type Concrete: CustomConcrete;
     /// The resource-unique name.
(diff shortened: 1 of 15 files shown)


src/types/type_param.rs
```diff
@@ -67,6 +67,19 @@ impl TypeArg {
     }
 }
 
+impl TypeArg {
```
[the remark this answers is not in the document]
ss2165: yes will do now sorry

src/resource.rs
```diff
@@ -323,6 +417,42 @@ pub struct TypeDef {
     pub tag: TypeDefTag,
 }
 
+/// Concrete instantiations of types and operations defined in resources.
+pub trait CustomConcrete: sealed::SealedConcrete {
```
acl-cqc, line 424: This is fine, but I wonder how much benefit there really is in having a trait to generalize over....things you might get by calling `to_custom` on one of the only-two-possible instances of `TypeParametrisedInternal`?

acl-cqc: What happens if you combine this with SealedConcrete i.e. hide the whole thing - can external folk still call its methods ?

src/resource.rs
```diff
@@ -323,6 +417,42 @@ pub struct TypeDef {
...
+    /// Name of the definition.
+    fn name(&self) -> &SmolStr;
+    /// Type arguments.
+    fn args(&self) -> &[TypeArg];
+    /// Parent resource.
+    fn resource(&self) -> &ResourceId;
```
acl-cqc, line 430: Note that there is a difference in meaning here - for ops, the resource means, you'll need access to that resource in order to do anything much to the op; as per resource checking. But for types, it doesn't really matter what the resource is - using a type from a resource doesn't mean you need the resource definition, or anything else, really; resource checking does not infer that an operation uses the resource just if it accepts/returns a type defined therein.

So again fine, but if you didn't have the trait, I don't think you need to pay as much attention to there being `CustomType::resource()` as `OpaqueOp::resource()`

src/resource.rs
```diff
@@ -139,6 +144,72 @@ impl Debug for LowerFunc {
     }
 }
 
+/// Sealed trait for shared type parametrised functionality between [`TypeDef`] and [`OpDef`].
+pub trait TypeParametrisedInternal: sealed::SealedDef {
```
acl-cqc, line 151: This is not `Internal` - it's `pub` ! (Maybe you should merge this into `SealedDef` tho I think it's more useful to keep than `CustomConcrete`). The use case for parameterising over types `<T: TypeParametrisedInternal>` is *small* but I admit non-zero.

Also, the correct spelling is `parameterised`

src/resource.rs
```diff
@@ -139,6 +144,72 @@ impl Debug for LowerFunc {
     }
 }
 
+/// Sealed trait for shared type parametrised functionality between [`TypeDef`] and [`OpDef`].
```
acl-cqc, line 150:
```suggestion
/// Sealed trait for type-parameterised functionality shared between [`TypeDef`] and [`OpDef`].
```

src/resource.rs
```diff
@@ -139,6 +147,72 @@ impl Debug for LowerFunc {
...
+        Ok(())
+    }
+
+    /// Instantiate a concrete instance by providing type arguments.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the provided arguments are not
+    /// valid instances of the TypeDef parameters.
+    fn to_custom(&self, args: impl Into<Vec<TypeArg>>) -> Result<Self::Concrete, SignatureError>;
```
acl-cqc, line 174: Let's rename this - how about `instantiate`? Or `to_concrete` if you must (we're calling instantiations concrete, which I'm happy with, but we could be consistent)

src/resource.rs
```diff
@@ -139,6 +147,72 @@ impl Debug for LowerFunc {
...
+
+    /// Check custom instance is a valid instantiation of this definition.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the type of the instance does not
+    /// match the definition.
+    fn check_custom(&self, custom: &Self::Concrete) -> Result<(), SignatureError> {
```
acl-cqc, line 182: `check_concrete` ?

acl-cqc (approved):
It's more type/trait-heavy than I was expecting so perhaps think about whether we need so many, and a few thoughts on naming etc., but the essence of it seems right :-)

ss2165:
I've made both new traits private and added public methods to each struct for the functionality

PR #316, opened 2023-07-31
keep rust on nixpkgs version in devenv

ss2165:
otherwise clippy has a different version and pre-commit fails


aborgna-q:
Can you link a failing job?
What version is it using instead?

ss2165:
default value for channel is "nixpkgs" which tracks the version on the version of nixpkgs being used, in this case it is 1.70

pre-commit isn't used on CI so I can't link, only fails locally

aborgna-q (approved):

PR #319, opened 2023-07-31
Check Const's are well-formed using a constructor

acl-cqc:
`typecheck_const` is currently called from `validate_edge` - that is, once per outgoing edge. This is (a) repetitive, (b) doesn't complain about malformed constants with no edges, (c) confusing and hard to find ;-).

Instead, make Const constructible only via a new `new()` method (returning a Result), and validate only there.

### DIFF
--- a/src/hugr/validate.rs
+++ b/src/hugr/validate.rs
@@ -12,10 +12,9 @@ use thiserror::Error;
 #[cfg(feature = "pyo3")]
 use pyo3::prelude::*;
 
-use crate::hugr::typecheck::{typecheck_const, ConstTypeError};
 use crate::ops::validate::{ChildrenEdgeData, ChildrenValidationError, EdgeValidationError};
 use crate::ops::OpTag;
-use crate::ops::{self, OpTrait, OpType, ValidateOp};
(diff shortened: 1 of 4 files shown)


src/hugr/validate.rs
```diff
@@ -434,21 +433,16 @@ impl<'a> ValidationContext<'a> {
...
-                    // edges coming from FuncDecl or FuncDefn
-                    if !OpTag::Function.is_superset(from_optype.tag()) {
-                        return Err(InterGraphEdgeError::InvalidConstSrc {
-                            from,
-                            from_offset,
-                            typ,
-                        }
-                        .into());
-                    };
+                if !(OpTag::Const.is_superset(from_optype.tag())
```
acl-cqc, line 437: I could add a new `OpTag::ConstSrc` here, but alternatively I could read that statement in the spec as a note that these are the only kinds of nodes which actually have a `Static` other-edgetype and remove this check and the corresponding element of `InterGraphEdgeError`....?

ss2165 (approved):
👍 

In my follow-up I will build on this to make 

`Const::new(value: ConstValue, typ: ClassicType) -> Result<....`

PR #320, opened 2023-07-31
Run coverage check daily on main and generate html report

### LINKED ISSUE #299: Simplify/shorten coverage checks?
* Installing LLVM takes a long time
* Refactoring code to be shorter with the same tests can fail coverage

Maybe weekly runs? or just on main?

cqc-alec:
Closes #299 .

This generates a detailed daily coverage report that can be downloaded from github (as an artifact) and then viewed in a browser: see example [here](https://github.com/CQCL-DEV/hugr/actions/runs/5715419042).


cqc-alec:
> 👍
> 
> It'd be nice to have some automation to look at this and report changes. Do you have any ideas for that?

Doable, we'd probably want to just look at a few numbers from the summary (as in the existing check); and then upload them somewhere. (We could also upload the whole report.) Will look into this.

Issue #318, opened 2023-07-31
Make ConstValue type checkable rather than synthesisable


ss2165:
Values can currently report their types - which requires storing a lot of type information (e.g. sum storing every variant). Instead, implement

```rust
ConstValue::check_type(&self, type: ClassicType) -> Result<(), ConstTypeError>
```

and also make that a requirement of `CustomConst`

PR #321, opened 2023-07-31
Make ConstValue checkable not synethesisable

ss2165:
Partially addresses #318 , still need to add `CustomConst` checking

* new error variant where value doesn't check
* new constructor for Const
* specify const type when building

### DIFF
--- a/src/algorithm/nest_cfgs.rs
+++ b/src/algorithm/nest_cfgs.rs
@@ -428,8 +428,14 @@ pub(crate) mod test {
         //               \-> right -/             \-<--<-/
         let mut cfg_builder = CFGBuilder::new(type_row![NAT], type_row![NAT])?;
 
-        let pred_const = cfg_builder.add_constant(ConstValue::simple_predicate(0, 2))?; // Nothing here cares which
-        let const_unit = cfg_builder.add_constant(ConstValue::simple_unary_predicate())?;
+        let pred_const = cfg_builder.add_constant(
+            ConstValue::simple_predicate(0, 2),
+            ClassicType::new_simple_predicate(2),
+        )?; // Nothing here cares which
+        let const_unit = cfg_builder.add_constant(
+            ConstValue::simple_unary_predicate(),
+            ClassicType::new_simple_predicate(1),
+        )?;
 
         let entry = n_identity(
             cfg_builder.simple_entry_builder(type_row![NAT], 1)?,
@@ -647,8 +653,14 @@ pub(crate) mod test {
         separate: bool,
     ) -> Result<(Hugr, BasicBlockID, BasicBlockID), BuildError> {
         let mut cfg_builder = CFGBuilder::new(type_row![NAT], type_row![NAT])?;
-        let pred_const = cfg_builder.add_constant(ConstValue::simple_predicate(0, 2))?; // Nothing here cares which
-        let const_unit = cfg_builder.add_constant(ConstValue::simple_unary_predicate())?;
+        let pred_const = cfg_builder.add_constant(
+            ConstValue::simple_predicate(0, 2),
+            ClassicType::new_simple_predicate(2),
+        )?; // Nothing here cares which
+        let const_unit = cfg_builder.add_constant(
+            ConstValue::simple_unary_predicate(),
+            ClassicType::new_simple_predicate(1),
+        )?;
 
         let entry = n_identity(
             cfg_builder.simple_entry_builder(type_row![NAT], 2)?,
@@ -682,8 +694,14 @@ pub(crate) mod test {
 
         let mut cfg_builder = CFGBuilder::new(type_row![NAT], type_row![NAT])?;
 
-        let pred_const = cfg_builder.add_constant(ConstValue::simple_predicate(0, 2))?; // Nothing here cares which
-        let const_unit = cfg_builder.add_constant(ConstValue::simple_unary_predicate())?;
+        let pred_const = cfg_builder.add_constant(
+            ConstValue::simple_predicate(0, 2),
+            ClassicType::new_simple_predicate(2),
+        )?; // Nothing here cares which
+        let const_unit = cfg_builder.add_constant(
+            ConstValue::simple_unary_predicate(),
+            ClassicType::new_simple_predicate(1),
+        )?;
 
         let entry = n_identity(
             cfg_builder.simple_entry_builder(type_row![NAT], 1)?,
--- a/src/algorithm/nest_cfgs.rs
+++ b/src/algorithm/nest_cfgs.rs
@@ -399,10 +399,8 @@ pub(crate) mod test {
     use super::*;
     use crate::builder::{BuildError, CFGBuilder, Container, DataflowSubContainer, HugrBuilder};
     use crate::hugr::region::{FlatRegionView, Region};
-    use crate::ops::{
-        handle::{BasicBlockID, ConstID, NodeHandle},
-        ConstValue,
-    };
+    use crate::ops::handle::{BasicBlockID, ConstID, NodeHandle};
+    use crate::ops::Const;
     use crate::types::{ClassicType, SimpleType};
     use crate::{type_row, Hugr};
     const NAT: SimpleType = SimpleType::Classic(ClassicType::i64());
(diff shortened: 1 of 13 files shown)


src/algorithm/nest_cfgs.rs
```diff
@@ -428,8 +428,14 @@ pub(crate) mod test {
         //               \-> right -/             \-<--<-/
         let mut cfg_builder = CFGBuilder::new(type_row![NAT], type_row![NAT])?;
 
-        let pred_const = cfg_builder.add_constant(ConstValue::simple_predicate(0, 2))?; // Nothing here cares which
-        let const_unit = cfg_builder.add_constant(ConstValue::simple_unary_predicate())?;
+        let pred_const = cfg_builder.add_constant(
```
acl-cqc, line 431: Is it worth making `add_constant` take a tuple, so that we can call one factory method (thinking of the case of predicates here) that returns both value and type together?

src/builder/build_traits.rs
```diff
@@ -61,11 +61,10 @@ pub trait Container {
     ///
     /// This function will return an error if there is an error in adding the
     /// [`OpType::Const`] node.
-    fn add_constant(&mut self, val: ConstValue) -> Result<ConstID, BuildError> {
-        let typ = val.const_type();
-        let const_n = self.add_child_op(ops::Const(val))?;
+    fn add_constant(&mut self, value: ConstValue, typ: ClassicType) -> Result<ConstID, BuildError> {
+        let const_n = self.add_child_op(ops::Const::new(value, typ))?;
```
acl-cqc, line 65: note `Const::new(value,typ)` returns a Result here so there should be a `?` inside the outer brackets

src/algorithm/nest_cfgs.rs
ss2165: this suggests there is utility in a `TypedValue` that holds both.... Could make it a builder-only concept

acl-cqc: `add_load_const` too

src/builder/build_traits.rs
```diff
@@ -61,11 +61,10 @@ pub trait Container {
     ///
     /// This function will return an error if there is an error in adding the
     /// [`OpType::Const`] node.
-    fn add_constant(&mut self, val: ConstValue) -> Result<ConstID, BuildError> {
-        let typ = val.const_type();
-        let const_n = self.add_child_op(ops::Const::new(val).map_err(BuildError::BadConstant)?)?;
+    fn add_constant(&mut self, constant: ops::Const) -> Result<ConstID, BuildError> {
```
acl-cqc, line 64: So my first thought was - "ah-ha! do we even need `add_constant`?" but now I see we kinda need it to control ConstID.

A possible future refactor, then. Imagine
```
trait NodeTemplate { // or BuildableNode, or something
   const op: OpTag;
   type op_ty: OpType; // probably only one of these
   ....whatever else we need....
}

impl NodeTemplate for Const {
  type op_ty = ops::Constant;
}

struct/?? NodeID<T> { // or maybe T is a "type", not sure
   fn get_node(&self) -> T {...}
}
```
and then `add_op<T: NodeTemplate>(n: T) -> NodeID<T>`

if you see what I mean
```

File: src/builder/build_traits.rs [hunk 4]
```diff
@@ -301,10 +300,16 @@ pub trait Dataflow: Container {
     /// This function will return an error if there is an error when adding the node.
     fn load_const(&mut self, cid: &ConstID) -> Result<Wire, BuildError> {
         let const_node = cid.node();
+        let op @ ops::Const { .. } = self
```
acl-cqc, line 303: (this is the NodeID::get_node from above)

acl-cqc: Why `op @ ops::Const { .. }` rather than `op: ops::Const` ?

src/hugr/validate.rs
```diff
@@ -808,7 +808,11 @@ mod test {
         parent: Node,
         predicate_size: usize,
     ) -> (Node, Node, Node, Node) {
-        let const_op = ops::Const::new(ConstValue::simple_predicate(0, predicate_size)).unwrap();
+        let const_op = ops::Const::new(
```
acl-cqc, line 811: not `ops::Const::simple_predicate(0, predicate_size)`? I guess that's the same thing, but let's test more prod code with less test code ;-)

src/ops/constant.rs
```diff
@@ -3,49 +3,117 @@
...
+    }
+
+    /// Fixed width integer
+    pub fn int<const N: u8>(value: HugrIntValueStore) -> Result<Self, ConstTypeError> {
+        Self::new(ConstValue::Int(value), ClassicType::int::<N>())
+    }
+
+    /// 64-bit integer
+    pub fn i64(value: i64) -> Result<Self, ConstTypeError> {
+        Self::new(ConstValue::i64(value), ClassicType::i64())
```
acl-cqc, line 94:
```suggestion
        Self::int<64>(value)
```

src/ops/constant.rs
```diff
@@ -3,49 +3,117 @@
...
+        Ok(Self { value, typ })
+    }
+
+    /// Returns a reference to the value of this [`Const`].
+    pub fn value(&self) -> &ConstValue {
+        &self.value
+    }
+
+    /// Returns a reference to the type of this [`Const`].
+    pub fn get_type(&self) -> &ClassicType {
```
acl-cqc, line 39: Possibly call this `const_type` as that previously-existing method has gone?

src/ops/constant.rs
```diff
@@ -63,20 +131,11 @@ pub(crate) const HUGR_MAX_INT_WIDTH: HugrIntWidthStore =
...
     F64(f64),
     /// A constant specifying a variant of a Sum type.
-    Sum {
-        tag: usize,
-        // We require the type to be entirely Classic (i.e. we don't allow
-        // a classic variant of a Sum with other variants that are linear)
-        variants: ClassicRow,
-        val: Box<ConstValue>,
-    },
+    Sum { tag: usize, val: Box<ConstValue> },
```
acl-cqc, line 138: I'd be tempted to make this a tuple. The types will ensure you don't get them the wrong way around and it's a lot easier on the eyes/fingers. Up to you!

src/ops/constant.rs
```diff
@@ -141,34 +188,19 @@ impl ConstValue {
...
-    }
-
-    /// Constant "false" value, i.e. the first variant of Sum((), ()).
-    pub fn false_val() -> Self {
-        Self::simple_predicate(0, 2)
-    }
-
     /// Constant Sum over units, used as predicates.
-    pub fn simple_predicate(tag: usize, size: usize) -> Self {
+    pub fn simple_predicate(tag: usize, _size: usize) -> Self {
```
acl-cqc, line 192: I think it's fine to drop the `_size` and the commented-out line beneath

src/ops/constant.rs
```diff
@@ -180,10 +212,7 @@ impl ConstValue {
 
     /// New 64 bit integer constant
     pub fn i64(value: i64) -> Self {
-        Self::Int {
-            value: value as HugrIntValueStore,
-            width: 64,
-        }
+        Self::Int(value as HugrIntValueStore)
```
acl-cqc, line 215: Is it worth keeping this? There's only thing about it that's `i64` is the argument type. You'll get a Rust type error rather than a constructor Error so I suppose that's slightly better but I'm not sure that's worth the implication of this method that the value is inherently 64-bit, e.g. if we made a Const out of a `ConstValue::i64` and `HashableType::Int(32)`....

src/ops/constant/typecheck.rs
```diff
@@ -136,25 +131,16 @@ pub fn typecheck_const(typ: &ClassicType, val: &ConstValue) -> Result<(), ConstT
...
                 }
-                if **row != *variants {
-                    return Err(ConstTypeError::TypeRowMismatch(
-                        *row.clone(),
-                        variants.clone(),
-                    ));
-                }
-                let ty = variants.get(*tag).unwrap();
+
+                let ty = row.get(*tag).unwrap();
```
acl-cqc, line 140: nit - slight preference to `match row.get(*tag)` to combine the check+unwrap into one

src/ops/constant.rs
```diff
@@ -264,17 +290,8 @@ mod test {
...
-            0,
-            ConstValue::Tuple(vec![]),
-            pred_rows,
-        ));
-        assert_eq!(
-            res,
-            Err(BuildError::BadConstant(ConstTypeError::TupleWrongLength))
-        );
+        let res = Const::predicate(0, ConstValue::Tuple(vec![]), pred_rows);
+        assert_matches!(res, Err(ConstTypeError::TupleWrongLength));
```
acl-cqc, line 295: Hehe - so we never get as far as calling `add_constant`. Nice. I think you can remove BuildError::BadConstant, thus.

acl-cqc (approved):
Fab. Most of this was extremely boring ;-), which is great :). A good number of nits, nothing major (ok a wild suggestion to think about for some point in the future but not now)

src/builder/build_traits.rs
ss2165: vestigial, thanks

src/ops/constant.rs
ss2165: I still think it's useful as a BuildError (it alllows use of `?` inside builder
contexts on `Const::new`)

src/ops/constant.rs
ss2165: yeah good point, removed

PR #322, opened 2023-07-31
Add HashableValue as part of ConstValue, and ContainerValue

acl-cqc:
* No additional sharing of code with TypeParam/TypeArg yet, but the idea is that we will be able to
* Still have both `serde_yaml::Value` and `Box<dyn CustomConst>` in different places
* `typecheck.rs` remains, as the code that's used by both the new and the old, not-yet-updated (TypeParam/TypeArg) code. When we update the latter, the classes/routines in `typecheck.rs` should only have one caller (in values.rs) and then they can be moved into values.rs. [So, this is kinda in the wrong place now, if we want to move it.]

### DIFF
--- a/src/ops/constant.rs
+++ b/src/ops/constant.rs
@@ -4,13 +4,15 @@ use std::any::Any;
 
 use crate::{
     macros::impl_box_clone,
-    types::{ClassicRow, ClassicType, CustomType, EdgeKind},
+    types::{simple::Container, ClassicRow, ClassicType, CustomType, EdgeKind, HashableType},
+    values::{
+        map_container_type, ConstTypeError, ContainerValue, CustomCheckFail, HashableValue,
+        ValueOfType,
+    },
 };
 
 use downcast_rs::{impl_downcast, Downcast};
 use smol_str::SmolStr;
-
-use self::typecheck::{typecheck_const, ConstTypeError, CustomCheckFail};
 
 use super::OpTag;
 use super::{OpName, OpTrait, StaticTag};
@@ -48,8 +50,7 @@ impl Const {
         variant_rows: impl IntoIterator<Item = ClassicRow>,
     ) -> Result<Self, ConstTypeError> {
         let typ = ClassicType::new_predicate(variant_rows);
-
-        Self::new(ConstValue::Sum(tag, Box::new(value)), typ)
+        Self::new(ConstValue::sum(tag, value), typ)
     }
 
     /// Constant Sum over units, used as predicates.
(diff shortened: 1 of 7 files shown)


src/ops/constant/typecheck.rs
```diff
@@ -87,132 +54,3 @@ pub(crate) fn check_int_fits_in_width(
...
-            Err(ConstTypeError::Unimplemented(ty.clone()))
-        }
-        (ClassicType::Hashable(HashableType::Variable(_)), _) => {
-            Err(ConstTypeError::ConstCantBeVar)
-        }
-        (ty, _) => Err(ConstTypeError::ValueCheckFail(ty.clone(), val.clone())),
-    }
-}
-
-#[cfg(test)]
```
acl-cqc, line 173: These have gone to ops/constant.rs. I could write some HashableValue-specific ones in `values.rs`...

src/ops/constant/typecheck.rs
```diff
@@ -87,132 +54,3 @@ pub(crate) fn check_int_fits_in_width(
         Err(ConstIntError::IntWidthInvalid(width))
     }
 }
-
-fn map_vals<T: PrimType, T2: PrimType>(
```
acl-cqc, line 91: This has moved to values.rs

src/ops/constant/typecheck.rs
```diff
@@ -87,132 +54,3 @@ pub(crate) fn check_int_fits_in_width(
...
-        Container::Tuple(elems) => Container::Tuple(map_row(*elems, f)),
-        Container::Sum(variants) => Container::Sum(map_row(*variants, f)),
-        Container::Array(elem, sz) => Container::Array(Box::new(f(*elem)), sz),
-        Container::Alias(s) => Container::Alias(s),
-        Container::Opaque(custom) => Container::Opaque(custom),
-    }
-}
-
-/// Typecheck a constant value
-pub(super) fn typecheck_const(typ: &ClassicType, val: &ConstValue) -> Result<(), ConstTypeError> {
```
acl-cqc, line 118: This has been split into values.rs and constant.rs parts

src/ops/constant.rs
```diff
@@ -116,23 +126,17 @@ pub(crate) type HugrIntWidthStore = u8;
...
-/// TODO: bigger/smaller integers.
+/// Value constants. (This could be "ClassicValue" to parallel [HashableValue])
 #[derive(Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize)]
 #[non_exhaustive]
 #[allow(missing_docs)]
 pub enum ConstValue {
-    /// An arbitrary length integer constant.
-    Int(HugrIntValueStore),
+    Hashable(HashableValue),
+    /// A collection of constant values (at least some of which are not [ConstValue::Hashable])
```
acl-cqc, line 135: "not **known to be**", perhaps

src/values.rs
```diff
@@ -0,0 +1,242 @@
...
+impl<Elem: ValueOfType> ContainerValue<Elem> {
+    pub(crate) fn desc(&self) -> String {
+        match self {
+            ContainerValue::Sequence(vals) => {
+                let names: Vec<_> = vals.iter().map(ValueOfType::name).collect();
+                format!("const:seq:{{{}}}", names.join(", "))
+            }
+            ContainerValue::Map(_) => "a map".to_string(),
+            ContainerValue::Sum(tag, val) => format!("const:sum:{{tag:{tag}, val:{}}}", val.name()),
+            ContainerValue::Opaque(c) => format!("const:{:?}", c),
```
ss2165, line 123: worth saying opaque here?

src/values.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// Errors that arise from typechecking constants
+#[derive(Clone, Debug, PartialEq, Error)]
+pub enum ConstTypeError {
+    /// There was some problem fitting a const int into its declared size
+    #[error("Error with int constant")]
+    Int(#[from] ConstIntError),
+    /*/// Expected width (packed with const int) doesn't match type
+    #[error("Type mismatch for int: expected I{0}, but found I{1}")]
+    IntWidthMismatch(HugrIntWidthStore, HugrIntWidthStore),*/
```
ss2165, line 222: delete

ss2165 (approved):

src/values.rs
acl-cqc: Used `const:custom:` and `const:yaml:` for the two different forms that we still have ATM

PR #323, opened 2023-07-31
Add `DataflowParent` OpTag

lmondada:
I propose adding an `OpTag` to gather all `OpType` whose children form a dataflow sibling graph. I've called this an `IOBlock` to refer to the property that they are guaranteed to have an `Input` and `Output` child.

Other possible names could be `DataflowWrapper`, `DataflowParent`, `DataBlock`... Opinions?


cqc-alec:
Good idea. Not too keen on the name `IOBlock`, since it only really makes sense with detailed knowledge of the spec (one might assume that a CFG is also an I/O block). Maybe `DataflowParent` or simply `Dataflow`?

lmondada:
I'm not sure how what the definition of `DataflowChild` should be exactly. Right now both `OpTag::Dfg` and `OpTag::TailLoop` are both `DataflowChild` and `DataflowParent`.

cqc-alec (approved):

PR #324, opened 2023-07-31
Allow all Containers in TypeParam

acl-cqc:
This'll follow #322 (which itself will follow #321), as until that I've had to comment out the check_arg_fits_param method, but you can see the changes.

Serialization is now better separated from other concerns (it no longer uses/requires `PrimType` :-)). Is my `Deserialized` what you meant or is there a way to take that further still, e.g. were you thinking
```
enum SerContainer {
   Simple(Container<SimpleType>),
   Classic(Container<ClassicType>),
   Hashable(Container<HashableType>),
   TypeParam(Container<TypeParam>)
}

enum SerSimpleType {
   Ctr(SerContainer),
   // remove SerSimpleType::List,Array,Tuple,...
}
```
? (I'm not convinced this'll work with the `Container`->`Type` change but there may be something I'm missing)

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -23,6 +23,8 @@ use crate::hugr::{Direction, Port};
 use crate::utils::display_list;
 use crate::{resource::ResourceSet, type_row};
 
+mod serialize;
+
 /// The kinds of edges in a HUGR, excluding Hierarchy.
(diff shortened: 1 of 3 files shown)


acl-cqc:
This has gotten mangled somehow and lost most of the serialization changes.

PR #328, opened 2023-08-01
CustomType's are Simple not necessarily Classic, and report their TypeTag

acl-cqc:
* Hopefully this will fix problems you were seeing @ss2165

### DIFF
--- a/src/types/custom.rs
+++ b/src/types/custom.rs
@@ -6,7 +6,7 @@ use std::fmt::{self, Display};
 
 use crate::resource::ResourceId;
 
-use super::{type_param::TypeArg, ClassicType, Container};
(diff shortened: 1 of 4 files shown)


src/types/custom.rs
```diff
@@ -66,8 +65,13 @@ impl Display for CustomType {
     }
 }
 
-impl From<CustomType> for ClassicType {
-    fn from(ty: CustomType) -> Self {
-        ty.classic_type()
+impl From<CustomType> for SimpleType {
```
acl-cqc, line 68: So this is the API also known as `SimpleType::new_opaque` equivalent to `SimpleType::new_tuple`

ss2165 (approved):
👍

PR #329, opened 2023-08-01
Move TypeRow into own file, introduce TypeRowElem

acl-cqc:
Increase modularity - TypeRow is just a row of *things* really, it doesn't require them to be "types" so there's no need to require elements to support `.tag()` or even to limit it to sealed things.

The additional trait TypeRowElem has the advantage that we no longer need to implement `.tag()` for `SerSimpleType`.

This is pulled out of #324 but can stand alone, and shouldn't conflict with any of the other stuff, so can go in first.

Update: I added a "blanket" implementation of TypeRowElem in commit 66f25691. I'm not certain about this, but don't really see why not.

---> Should we rename "TypeRow" (and others) to just Row? It is, basically, just a list-like data structure...

### DIFF
--- a/src/types/simple.rs
+++ b/src/types/simple.rs
@@ -1,17 +1,13 @@
 //! Dataflow types
 
-use std::{
-    borrow::Cow,
-    fmt::{self, Display, Formatter, Write},
-    ops::{Deref, DerefMut},
-};
+use std::fmt::{self, Display, Formatter, Write};
 
+use super::type_row::{TypeRow, TypeRowElem};
+use super::{custom::CustomType, AbstractSignature};
+use crate::{classic_row, ops::constant::HugrIntWidthStore};
 use itertools::Itertools;
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use smol_str::SmolStr;
-
-use super::{custom::CustomType, AbstractSignature};
-use crate::{classic_row, ops::constant::HugrIntWidthStore, utils::display_list};
 
 /// A type that represents concrete data. Can include both linear and classical parts.
 ///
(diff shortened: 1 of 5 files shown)


src/types/simple.rs
```diff
@@ -478,121 +457,19 @@ impl TypeRow<ClassicType> {
...
-
-    /// Returns `true` if the row contains no types.
-    #[inline(always)]
-    pub fn is_empty(&self) -> bool {
-        self.types.len() == 0
-    }
-
     /// Returns whether the row contains only hashable classic data.
     #[inline(always)]
     pub fn purely_hashable(&self) -> bool {
```
aborgna-q, line 459: Why isn't this one (and `containing_tag`) in `type_row.rs`?

src/types/type_row.rs
```diff
@@ -0,0 +1,134 @@
...
+    fmt::{self, Display, Write},
+    ops::{Deref, DerefMut},
+};
+
+use crate::utils::display_list;
+
+/// Base trait for anything that can be put in a [TypeRow]
+pub trait TypeRowElem: std::fmt::Debug + Clone + 'static {}
+
+impl<T: Clone + std::fmt::Debug + 'static> TypeRowElem for T {}
```
aborgna-q, line 15: I think we can do without the `Debug` constraint here.
(Clone is needed for `[T]: ToOwned` in the `Cow`)

```suggestion
pub trait TypeRowElem: Clone + 'static {}

impl<T: Clone + 'static> TypeRowElem for T {}
```

aborgna-q (approved):
LGTM

src/types/simple.rs
acl-cqc: Because it only applies to TypeRow's of PrimType, not (for example) to TypeRow's of SerSimpleType

src/types/type_row.rs
acl-cqc: Neat! I didn't realize that. I presume `#[derive(Debug)]` automagically inserts the appropriate constraint i.e. `impl<T: Debug + ...> Debug for TypeRow<T>`

PR #330, opened 2023-08-02
[new] Infer extension requirements

croyzor:
Add a module `extension::infer`, which given a Hugr attempts to infer the resource requirements where they are left open.

Addresses #425

### DIFF
--- a/devenv.lock
+++ b/devenv.lock
@@ -3,11 +3,11 @@
     "devenv": {
       "locked": {
         "dir": "src/modules",
-        "lastModified": 1690534632,
-        "narHash": "sha256-kOXS9x5y17VKliC7wZxyszAYrWdRl1JzggbQl0gyo94=",
+        "lastModified": 1692700763,
+        "narHash": "sha256-pWbQEVSjlDfrrBoHv5XFe+IdYVpJXY8C9Wz7rO98yqU=",
         "owner": "cachix",
         "repo": "devenv",
-        "rev": "6568e7e485a46bbf32051e4d6347fa1fed8b2f25",
+        "rev": "6b84a85dd5cde41cbf992b6b2445bf25fe0abb21",
         "type": "github"
       },
       "original": {
@@ -25,11 +25,11 @@
         "rust-analyzer-src": "rust-analyzer-src"
       },
       "locked": {
-        "lastModified": 1690525255,
-        "narHash": "sha256-jWmEJsV0mbWhUTCmkVg0URTVaQxzX9SPx33DPJPOmWc=",
+        "lastModified": 1692944460,
+        "narHash": "sha256-9SWRcFh/TvpZ7hc84SvlwP5QPkahWMRB2mMZmkBhkBQ=",
         "owner": "nix-community",
         "repo": "fenix",
-        "rev": "6d22d0989ebc9e8fc7e1d15f973613fae928ebb2",
+        "rev": "dd160229ca7895895462291cded1ea313c1e306f",
         "type": "github"
       },
       "original": {
@@ -136,11 +136,11 @@
         "nixpkgs-stable": "nixpkgs-stable"
       },
       "locked": {
-        "lastModified": 1690464206,
-        "narHash": "sha256-38V4kmOh6ikpfGiAS+Kt2H/TA2DubSqE66veP/jmB4Q=",
+        "lastModified": 1692274144,
+        "narHash": "sha256-BxTQuRUANQ81u8DJznQyPmRsg63t4Yc+0kcyq6OLz8s=",
         "owner": "cachix",
         "repo": "pre-commit-hooks.nix",
-        "rev": "9289996dcac62fd45836db7c07b87d2521eb526d",
+        "rev": "7e3517c03d46159fdbf8c0e5c97f82d5d4b0c8fa",
         "type": "github"
       },
       "original": {
@@ -160,11 +160,11 @@
     "rust-analyzer-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1690216158,
-        "narHash": "sha256-TUF0YoWweQj0hNHWykC1vtBQVUncARlLLPmQP9hUeME=",
+        "lastModified": 1692775770,
+        "narHash": "sha256-LwoR5N1JHykSte2Ak+Pj/HjJ9fKy9zMJNEftfBJQkLs=",
         "owner": "rust-lang",
         "repo": "rust-analyzer",
-        "rev": "b64e5b3919b24bc784f36248e6e1f921ee7bb71b",
+        "rev": "f5b7c60ff7a79bfb3e10f3e98c81b7bb4cb53c68",
         "type": "github"
       },
       "original": {


src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+        // Seperate loop so that we can assume that a metavariable has been
+        // added for every (Node, Direction) in the graph already.
+        for tgt_node in hugr.nodes() {
+            let sig: &OpType = hugr.get_nodetype(tgt_node).into();
+            // Incoming ports with a dataflow edge
+            for port in hugr
+                .node_inputs(tgt_node)
+                .filter(|src_port| matches!(sig.port_kind(*src_port), Some(EdgeKind::Value(_))))
+            {
+                for (src_node, _) in hugr.linked_ports(tgt_node, port) {
```
ss2165, line 284: Can this nested for loop be replaced with a call to `hugr.all_neighbours(tgt_node)`

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+                    [(node1, rs1.clone()), (node2, rs2.clone())]
+                } else {
+                    [(node2, rs2.clone()), (node1, rs1.clone())]
+                };
+
+                if src_rs.is_subset(&tgt_rs) {
+                    Some(InferExtensionError::EdgeMismatch(
+                        ExtensionError::TgtExceedsSrcExtensions {
+                            from: *src,
+                            from_offset: Port::new(Direction::Outgoing, 0),
```
ss2165, line 339: how do you know the ports are 0?

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+                    continue;
+                }
+
+                if let Some(cs) = self.constraints.get(m).cloned() {
+                    for c in cs
+                        .into_iter()
+                        .filter(|c| !matches!(c, Constraint::Equal(_)))
+                    {
+                        self.add_constraint(combined_meta, c.clone());
+                    }
```
ss2165, line 391:
```suggestion
                if let Some(cs) = self.constraints.get(m) {
                    for c in cs
                        .iter()
                        .filter(|c| !matches!(c, Constraint::Equal(_)))
                        .cloned()
                        .collect::<Vec<_>>()
                        .into_iter()
                    {
                        self.add_constraint(combined_meta, c);
                    }
```

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+                                ));
+                            }
+                        }
+                        None => {
+                            self.solved.insert(combined_meta, solution.clone());
+                        }
+                    }
+                }
+                if self.variables.contains(m) {
+                    self.variables.insert(combined_meta);
```
ss2165, line 417: does m need removing from variables?

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+                }
+                self.shunted.insert(*m, combined_meta);
+            }
+            // This doesn't do anything, but if it did it would just be
+            // implementing the same mechanism as provided by `resolve`
+            let mut updates = HashMap::<(Node, Direction), Meta>::new();
+            for (loc, meta) in self.extensions.iter() {
+                // If there's any node in this CC which equals `meta`, they are
+                // all equal to meta
+                if cc.iter().filter(|cc_meta| meta == *cc_meta).count() > 0 {
```
ss2165, line 427:
```suggestion
                if cc.iter().any(|cc_meta| meta == cc_meta) {
```

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+                            self.solved.insert(combined_meta, solution.clone());
+                        }
+                    }
+                }
+                if self.variables.contains(m) {
+                    self.variables.insert(combined_meta);
+                }
+                self.shunted.insert(*m, combined_meta);
+            }
+            // This doesn't do anything, but if it did it would just be
```
ss2165, line 421: why is there code that doesn't do anything?

src/extension/infer.rs
ss2165: You need an allocation to avoid borrow checker getting mad but you don't have to
clone the whole thing

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+                                        rs.clone(),
+                                        rrs,
+                                    ));
+                                }
+                            }
+                            None => self.add_solution(meta, rrs),
+                        };
+                        solved = true;
+                    } else {
+                        // TODO: Try and go backwards with a `Minus` constraint
```
ss2165, line 466: are we likely to run in to this case often?

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+                    if self.live_var(meta).is_some() {
+                        Err(InferExtensionError::Unsolved { location: *loc })
+                    } else {
+                        continue;
+                    }
+                }
+            }?;
+            results.insert(*loc, rs);
+        }
+        assert!(self.live_metas().is_empty());
```
ss2165, line 504:
```suggestion
        debug_assert!(self.live_metas().is_empty());
```

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+    // TODO: This should really be a list
+    fn live_var(&self, m: &Meta) -> Option<Meta> {
+        if self.variables.contains(m) || self.variables.contains(&self.resolve(*m)) {
+            return None;
+        }
+
+        // TODO: We should be doing something to ensure that these are the same check...
+        if self.get_solution(m).is_none() {
+            if self.get_constraints(m).is_some() {
+                for c in self.get_constraints(m).unwrap().iter() {
```
ss2165, line 518:
```suggestion
            if let Some(cs) = self.get_constraints(m) {
                for c in cs {
```

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+            // Try to solve metas with the information we have now. This may
+            // register new equalities on the EqGraph
+            let to_delete = self.solve_constraints(&remaining)?;
+            // Merge metas based on the equalities we just registered
+            let (new, merged) = self.merge_equal_metas()?;
+            // All of the metas for which we've made progress
+            let delta: HashSet<Meta> = HashSet::from_iter(to_delete.union(&merged).cloned());
+
+            for m in delta.iter() {
+                if !merged.contains(m) {
```
ss2165, line 581: you could do this step on `to_delete` before the union right?

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+                // Remove solved and merged metas from remaining "to solve" list
+                remaining.remove(m);
+            }
+
+            // If we made no progress, we're done!
+            if delta.is_empty() && new.is_empty() {
+                break;
+            }
+
+            // Add all of the new nodes we made (by merging old ones) to the pile
```
ss2165, line 595:
```suggestion
            remaining.extend(new);
```

ss2165 (changes requested):
Looks good overall (though could do with more gnarly tests); bunch of queries/small edits/comments

src/hugr/views.rs
```diff
@@ -308,7 +308,7 @@ where
     #[inline]
     fn get_io(&self, node: Node) -> Option<[Node; 2]> {
         let op = self.get_nodetype(node);
-        if op.tag().is_superset(OpTag::DataflowParent) {
+        if OpTag::DataflowParent.is_superset(op.tag()) {
```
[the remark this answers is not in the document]
croyzor: Yes, I'll factor it out into a seperate PR (#442)

src/extension/infer.rs
croyzor: I don't think it's so uncommon that we can ignore it. If, in the `dangling_src` test, the `src` node was wired into a node that adds extension requirements rather than leaving them be, we'd hit this case and fail to solve its requirements

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+/// A graph of metavariables which we've found equality constraints for. Edges
+/// between nodes represent equality constraints.
+struct EqGraph {
+    equalities: pg::Graph<Meta, (), petgraph::Undirected>,
+    node_map: HashMap<Meta, pg::NodeIndex>,
+}
+
+impl EqGraph {
+    /// Create a new `EqGraph`
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
croyzor: Apart from `add_or_retreive` I think they do? It could be refactored, but this is the API used by `UnificationContext`

src/extension/infer.rs
croyzor: Fixed it

src/extension/infer.rs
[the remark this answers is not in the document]
ss2165: which is also in this module so can access these methods without them being pub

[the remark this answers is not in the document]
croyzor: 🤯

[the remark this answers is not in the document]
ss2165: `pub` should only be used for things we want users to call (other code in the crate not in the same module could use `pub(super), pub(crate)` etc.)

src/extension/infer.rs
croyzor: No, because we specifically want to find nodes connected by _value edges_, not just neighbours

src/extension/infer.rs
croyzor: It's harmless to leave it in, but we could probably delete it

src/extension/infer.rs
croyzor: The `0`s are just dummy ports to indicate the connection between two nodes. This isn't doing the work to know which ports connect the nodes

ss2165: makes the error quite misleading no? If the ports aren't important to report maybe this needs a new error that just reports the nodes?

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+        self.constraints.insert(fresh, HashSet::new());
+        fresh
+    }
+
+    /// Declare a constraint on the metavariable
+    fn add_constraint(&mut self, m: Meta, c: Constraint) {
+        self.constraints
+            .entry(m)
+            .and_modify(|cs| {
+                cs.insert(c.clone());
```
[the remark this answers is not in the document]
croyzor: I'd be interested to hear how! As far as I can see this is equivalent to the
```
if ... {
    c
} else {
    c
}
```
situation which rust complains about to me all the time

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
+//! Inference for extension requirements on nodes of a hugr.
+//!
+//! Checks if the extensions requirements are sane, and comes up with concrete
```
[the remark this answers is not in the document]
croyzor: Elaborated in comment

src/extension/infer.rs
croyzor: Deleting it now

src/extension/infer.rs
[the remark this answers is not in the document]
ss2165:
```rust
        self.constraints
            .entry(m)
            .or_insert(HashSet::new())
            .insert(c);
```

src/extension/infer.rs
```diff
@@ -0,0 +1,855 @@
...
+        let mut last_meta = input;
+        // Create fresh metavariables with `Plus` constraints for
+        // each extension that should be added by the node
+        // Hence a extension delta [A, B] would lead to
+        // > ma = fresh_meta()
+        // > add_constraint(ma, Plus(a, input)
+        // > mb = fresh_meta()
+        // > add_constraint(mb, Plus(b, ma)
+        // > add_constraint(output, Equal(mb))
+        for r in delta.iter() {
```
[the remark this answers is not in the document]
ss2165: this doesn't look like it has been done

src/extension/infer.rs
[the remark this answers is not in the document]
croyzor: 🙇

ss2165 (approved):
Just one outstanding clone thing but otherwise looks good

PR #333, opened 2023-08-02
remove unused sealed traits

ss2165:


aborgna-q (approved):
`SealedDef` is not unused.
But I agree that `TypeParameter` doesn't really need to be sealed, it's only used to avoid duplicated code.

ss2165:
yes "unused" isn't the correct word, unnecessary is. The traits used to be `pub` but are now module private, so sealing is not necessary.

Issue #290, opened 2023-07-26
[Big Refactor]Make Container contain single elements (?)


acl-cqc:
Rather than `SimpleType::Qontainer`, `ClassicType::Container` and `HashableType::Container`, instead make SimpleType/ClassicType/HashableType be aliases for Container<SimpleElem/ClassicElem/HashableElem> and add to Container an enum variant `Single(T)` (a single value).

This may not be a win, but will probably make more sense after #286 which takes us partway there, and solves the hack in `typecheck_const` that there are currently two ways to express the SimpleType that's (any container of) a ClassicType:
* `SimpleType::Classic(ClassicType::Container(Container::List(clas_ty)))`
* `SimpleType::Qontainer(Container::List(SimpleType::Classic(clas_ty)))`

Instead there would be only one way: `Container::List(Container::Single(SimpleElem::Classic(clas_ty)))`

ss2165:
Done as part of #378

PR #335, opened 2023-08-02
Allow all Containers in TypeParam, reuse ContainerValue + ValueOfType

acl-cqc:
The first commit here might be broken off as a preliminary. Basically everything in the second (major) commit seemed to need to go in all at once...perhaps the biggest thing being the rewrite of serialization using a `Deserialized` struct. ISTR @ss2165 had an idea here so I'm not sure if this is the same idea or not...

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -11,9 +11,9 @@ use smol_str::SmolStr;
 use thiserror::Error;
 
 use crate::ops::custom::OpaqueOp;
-use crate::types::type_param::{check_type_arg, TypeArgError};
 use crate::types::type_param::{TypeArg, TypeParam};
 use crate::types::CustomType;
+use crate::values::{ConstTypeError, ValueOfType};
 
 mod opdef;
 pub use opdef::{CustomSignatureFunc, OpDef};
@@ -22,7 +22,7 @@ pub use type_def::{TypeDef, TypeDefTag};
 
 /// An error that can occur in computing the signature of a node.
 /// TODO: decide on failure modes
-#[derive(Debug, Clone, Error, PartialEq, Eq)]
+#[derive(Debug, Clone, Error, PartialEq)]
 pub enum SignatureError {
     /// Name mismatch
     #[error("Definition name ({0}) and instantiation name ({1}) do not match.")]
@@ -30,9 +30,12 @@ pub enum SignatureError {
     /// Resource mismatch
     #[error("Definition resource ({0:?}) and instantiation resource ({1:?}) do not match.")]
     ResourceMismatch(ResourceId, ResourceId),
+    /// When the wrong number of type arguments were supplied
+    #[error("Received {0} type arguments but expected {1}")]
+    WrongNumberTypeArgs(usize, usize),
     /// When the type arguments of the node did not match the params declared by the OpDef
     #[error("Type arguments of node did not match params declared by definition: {0}")]
-    TypeArgMismatch(#[from] TypeArgError),
+    TypeArgMismatch(#[from] ConstTypeError),
 }
 
 /// Concrete instantiations of types and operations defined in resources.
@@ -83,13 +86,13 @@ trait TypeParametrised {
     /// Check provided type arguments are valid against parameters.
     fn check_args_impl(&self, args: &[TypeArg]) -> Result<(), SignatureError> {
         if args.len() != self.params().len() {
-            return Err(SignatureError::TypeArgMismatch(TypeArgError::WrongNumber(
+            return Err(SignatureError::WrongNumberTypeArgs(
                 args.len(),
                 self.params().len(),
-            )));
+            ));
         }
         for (a, p) in args.iter().zip(self.params().iter()) {
-            check_type_arg(a, p).map_err(SignatureError::TypeArgMismatch)?;
+            a.check_type(p).map_err(SignatureError::TypeArgMismatch)?;
         }
         Ok(())
     }
--- a/src/resource/type_def.rs
+++ b/src/resource/type_def.rs
@@ -149,8 +149,9 @@ impl Resource {
 #[cfg(test)]
 mod test {
     use crate::resource::SignatureError;
-    use crate::types::type_param::{TypeArg, TypeArgError, TypeParam};
+    use crate::types::type_param::{TypeArg, TypeParam};
     use crate::types::{ClassicType, HashableType, TypeTag};
+    use crate::values::ConstTypeError;
 
     use super::{TypeDef, TypeDefTag};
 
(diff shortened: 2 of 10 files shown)


src/resource.rs
```diff
@@ -22,17 +22,20 @@ pub use type_def::{TypeDef, TypeDefTag};
 
 /// An error that can occur in computing the signature of a node.
 /// TODO: decide on failure modes
-#[derive(Debug, Clone, Error, PartialEq, Eq)]
+#[derive(Debug, Clone, Error, PartialEq)]
```
acl-cqc, line 25: I had to drop the `Eq` here because we are reusing `ConstTypeError`, and according to its definition a `ConstTypeError` could contain a ValueCheckFail with an arbitrary (non-Eq) ConstValue. *BUT* of course in practice here we are checking only TypeParams-vs-TypeArgs, so we'll never see such a ValueCheckFail (we'll only get TypeArgCheckFails).

So, the right thing to do would be to *parameterize* `ConstTypeError` by a `ValueOfType` and then combine `ValueCheckFail` with `TypeArgCheckFail` (using the `T: ValueOfType`). I haven't experimented with this, but will have a go...

src/resource/type_def.rs
```diff
@@ -163,31 +164,31 @@ mod test {
             description: "Some parameterised type".into(),
             tag: TypeDefTag::FromParams(vec![0]),
         };
-        let typ: SimpleType = def
+        let typ = def
```
acl-cqc, line 167: This is from the first commit, which might stand alone

src/types/type_param.rs
```diff
@@ -24,15 +32,61 @@ pub enum TypeParam {
...
+                })
+                .collect::<Vec<_>>()
+                .into())
+        } else {
+            Err(typarams)
+        }
+    }
+
+    /// New Tuple TypeParam, elements defined by TypeRow
+    pub fn new_tuple(elems: impl Into<TypeRow<TypeParam>>) -> Self {
```
acl-cqc, line 64: These parallel `SimpleType::new_tuple`...but should go when we do the big `Container`->`Type` refactor (probably next)

src/types/type_param.rs
```diff
@@ -24,15 +32,61 @@ pub enum TypeParam {
...
+
+    /// New Tuple typeparam, elements defined by TypeRow
+    pub fn new_sum(elems: impl Into<TypeRow<TypeParam>>) -> Self {
+        match TypeParam::value_types(elems.into()) {
+            Ok(h_tys) => TypeParam::Value(HashableType::Container(Container::Sum(Box::new(h_tys)))),
+            Err(ty_params) => Self::Container(Container::Sum(Box::new(ty_params))),
+        }
+    }
+
+    /// Creates a new TypeParam accepting values of a specified CustomType, which must be hashable.
```
acl-cqc, line 81: Hopefully the mere presence of this will prevent anyone creating a `TypeParam::Container(Container::Opaque(...))` directly

src/types/type_param.rs
```diff
@@ -62,72 +114,103 @@ impl TypeArg {
...
-pub fn check_type_arg(arg: &TypeArg, param: &TypeParam) -> Result<(), TypeArgError> {
-    match (arg, param) {
-        (TypeArg::Type(_), TypeParam::Type) => Ok(()),
-        (TypeArg::ClassicType(_), TypeParam::ClassicType) => Ok(()),
-        (TypeArg::HashableType(_), TypeParam::HashableType) => Ok(()),
-        (TypeArg::List(items), TypeParam::List(ty)) => {
-            for item in items {
-                check_type_arg(item, ty.as_ref())?;
+/// A serialized representation of a value of a [CustomType]
+/// restricted to Hashable types.
```
acl-cqc, line 124: This would just *be* `CustomSerialized` were it not for enforcing this, which I think is an important sanity check (no `MyType<3.51f>` even by putting that in a CustomType, but `MyType<3>` or custom versions of are ok)

src/types/type_param.rs
```diff
@@ -62,72 +114,103 @@ impl TypeArg {
...
+impl ValueOfType for TypeArg {
+    type T = TypeParam;
+
+    fn check_type(&self, ty: &TypeParam) -> Result<(), ConstTypeError> {
+        match (self, ty) {
+            (TypeArg::Type(_), TypeParam::Type) => Ok(()),
+            (TypeArg::ClassicType(_), TypeParam::ClassicType) => Ok(()),
+            (TypeArg::HashableType(_), TypeParam::HashableType) => Ok(()),
+            (TypeArg::Container(vals), TypeParam::Container(c_ty)) => vals.check_container(c_ty),
+            // We might have an argument *value* that is a TypeArg (but not a HashableValue)
```
acl-cqc, line 154: This is the same case as https://github.com/CQCL-DEV/hugr/blob/729e96fc7815f565b5be4d2246ffbb2a0e00f272/src/ops/constant.rs#L181

src/types/type_param.rs
```diff
@@ -62,72 +114,103 @@ impl TypeArg {
...
+            // that fits a Hashable type because the argument contains an [TypeArg::Opaque]:
+            (TypeArg::Container(vals), TypeParam::Value(HashableType::Container(c_ty))) => {
+                vals.check_container(&map_container_type(c_ty, &TypeParam::Value))
+            }
+            (TypeArg::Value(hv), TypeParam::Value(ht)) => hv.check_type(ht),
+            // A "hashable" value might be argument to a non-hashable TypeParam:
```
acl-cqc, line 160: And this corresponds to https://github.com/CQCL-DEV/hugr/blob/729e96fc7815f565b5be4d2246ffbb2a0e00f272/src/ops/constant.rs#L181

src/types/type_param.rs
acl-cqc: No, this one's bogus, can't happen, because Opaques must be hashable

acl-cqc: "fail to deduce" where I linked isn't really right - "be unable to represent a container value as hashable, because the value contains opaque constants, and HashableValue deliberately has no way to represent anything opaque"

src/types/type_param.rs
```diff
@@ -62,72 +114,103 @@ impl TypeArg {
...
+
+    fn check_type(&self, ty: &TypeParam) -> Result<(), ConstTypeError> {
+        match (self, ty) {
+            (TypeArg::Type(_), TypeParam::Type) => Ok(()),
+            (TypeArg::ClassicType(_), TypeParam::ClassicType) => Ok(()),
+            (TypeArg::HashableType(_), TypeParam::HashableType) => Ok(()),
+            (TypeArg::Container(vals), TypeParam::Container(c_ty)) => vals.check_container(c_ty),
+            // We might have an argument *value* that is a TypeArg (but not a HashableValue)
+            // that fits a Hashable type because the argument contains an [TypeArg::Opaque]:
+            (TypeArg::Container(vals), TypeParam::Value(HashableType::Container(c_ty))) => {
```
acl-cqc, line 156: This is the same case as https://github.com/CQCL-DEV/hugr/blob/729e96fc7815f565b5be4d2246ffbb2a0e00f272/src/ops/constant.rs#L193 (but should go in the big `Container`->`Type` refactor)

acl-cqc:
@ss2165 I managed to pull the serialization changes out into #351 so this just `#[derive(serde::Serialize, serde::Deserialize)]` for TypeParam (and also Container).   (When I said there was a problem with TypeRow, I was mistaken, as SerSimpleType contains TypeRow's; the problem was with Container, and quite fixable.)

acl-cqc:
Closing in favour of new approach in #290 and #353

PR #336, opened 2023-08-02
Add SubHugr view

lmondada:
- remove unused sealed traits (#333)
- Add IOBlock OpTag
- Add SubHugr
- REVERTME use portgraph HEAD

### DIFF
--- a/src/hugr/rewrite/simple_replace.rs
+++ b/src/hugr/rewrite/simple_replace.rs
@@ -1,5 +1,6 @@
 //! Implementation of the `SimpleReplace` operation.
 
+use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
 
 use itertools::Itertools;
@@ -58,7 +59,7 @@ impl Rewrite for SimpleReplacement {
 
     fn apply(self, h: &mut Hugr) -> Result<(), SimpleReplacementError> {
         // 1. Check the parent node exists and is a DFG node.
-        if h.get_optype(self.parent).tag() != OpTag::Dfg {
+        if h.get_optype(self.parent).tag().partial_cmp(&OpTag::IOBlock) != Some(Ordering::Less) {
             return Err(SimpleReplacementError::InvalidParentNode());
         }
         // 2. Check that all the to-be-removed nodes are children of it and are leaves.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ path = "src/lib.rs"
 
 [dependencies]
 thiserror = "1.0.28"
-portgraph = { version = "0.7.1", features = ["serde", "petgraph"] }
+portgraph = { version = "0.7.2", features = ["serde", "petgraph"] }
 pyo3 = { version = "0.19.0", optional = true, features = [
     "multiple-pymethods",
 ] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ path = "src/lib.rs"
 
 [dependencies]
 thiserror = "1.0.28"
-portgraph = { version = "0.7.2", features = ["serde", "petgraph"] }
+portgraph = { git = "https://github.com/CQCL/portgraph.git", rev = "e5bd1ca", features = ["serde", "petgraph"] }
 pyo3 = { version = "0.19.0", optional = true, features = [
     "multiple-pymethods",
 ] }
--- a/src/hugr.rs
+++ b/src/hugr.rs
@@ -5,6 +5,7 @@ mod hugrmut;
 pub mod region;
 pub mod rewrite;
 pub mod serialize;
+pub mod subhugr;
 pub mod validate;
 pub mod view;
 
--- a/src/hugr/subhugr.rs
+++ b/src/hugr/subhugr.rs
@@ -1,0 +1,312 @@
+use derive_more::Into;
+use itertools::Itertools;
+
+use crate::{
+    ops::{OpTag, OpTrait},
+    Direction, Hugr, Node, Port, SimpleReplacement,
+};
+
+use super::{region::RegionView, HugrView};
+
+/// A boundary edge of a subhugr.
+///
+/// It is always uniquely identified (any Hugr edge is) by the node and port
+/// of the target of the edge. Source edges are not unique as they can involve
+/// copies.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Into)]
+struct BoundaryEdge(Node, Port);
+
+impl BoundaryEdge {
+    fn get_boundary_edges<H: HugrView>(
+        node: Node,
+        port: Port,
+        hugr: &H,
+    ) -> impl Iterator<Item = Self> {
+        if port.direction() == Direction::Incoming {
+            [BoundaryEdge(node, port)].to_vec().into_iter()
+        } else {
+            hugr.linked_ports(node, port)
+                .map(|(n, p)| BoundaryEdge(n, p))
+                .collect_vec()
+                .into_iter()
+        }
+    }
+
+    /// The DFG input ports of a `node`
+    fn get_boundary_incoming<H: HugrView>(node: Node, hugr: &H) -> impl Iterator<Item = Self> + '_ {
+        hugr.node_inputs(node)
+            .filter(move |&p| hugr.get_optype(node).signature().get(p).is_some())
+            .map(move |p| BoundaryEdge(node, p))
+    }
+
+    /// The linked ports of DFG output ports of a `node`
+    ///
+    /// We always consider target ports, as they are unique
+    /// (source ports can be copied)
+    fn get_boundary_outgoing<H: HugrView>(node: Node, hugr: &H) -> impl Iterator<Item = Self> + '_ {
+        hugr.node_outputs(node)
+            .filter(move |&p| hugr.get_optype(node).signature().get(p).is_some())
+            .flat_map(move |p| hugr.linked_ports(node, p))
+            .map(|(n, p)| BoundaryEdge(n, p))
+    }
+}
+
+/// A subgraph of a HUGR.
+///
+/// The subgraph is described by the following data:
+///  - a HugrView to the underlying HUGR,
+///  - a root node, of which all nodes in the [`SubHugr`] must be descendants,
+///  - incoming/outgoing boundary edges, pointing into/out of the subgraph.
+/// The incoming/outgoing edges must link direct children of the
+/// root node. Their ordering matters when using the [`SubHugr`] for replacements,
+/// as it will match the ordering of the ports in the replacement.
+/// The list of incoming and/or outgoing ports may be empty.
+///
+/// A subhugr is well-defined if the incoming/outgoing edges partition the
+/// edges between the children of the root node into three sets:
+///  - boundary edges: either incoming boundary or outgoing boundary edges,
+///  - interior edges: edges such that all the successor edges are either
+///    outgoing boundary edges or interior edges AND all the predecessor edges
+///    are either incoming boundary edges or interior edges,
+///  - exterior edges: edges such that all the successor edges are either
+///    incoming boundary edges or exterior edges AND all the predecessor edges
+///    are either outgoing boundary edges or exterior edges.
+///
+/// Then the subhugr contains all children nodes of the root that are
+///  - adjacent to an interior edge, or
+///  - in-adjacent to an incoming edge and out-adjacent to an outgoing edge.
+///
+/// The root node itself is not included in the subhugr. Any node that is a
+/// descendant of a node in the subhugr is itself also in the subhugr.
+/// If both incoming and outgoing ports are empty, the subhugr is taken to be
+/// all children of the root.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SubHugr<Base> {
+    hugr: Base,
+    root: Node,
+    incoming: Vec<BoundaryEdge>,
+    outgoing: Vec<BoundaryEdge>,
+}
+
+impl<'g, Base: HugrView> SubHugr<RegionView<'g, Base>> {
+    /// Creates a new subhugr from a region.
+    pub fn from_region(region: RegionView<'g, Base>) -> Self {
+        let root = region.root();
+        Self {
+            hugr: region,
+            root,
+            incoming: Vec::new(),
+            outgoing: Vec::new(),
+        }
+    }
+}
+
+impl<Base> SubHugr<Base> {
+    /// Creates a new subhugr from a root node
+    pub fn from_root(hugr: Base, root: Node) -> Self {
+        Self {
+            hugr,
+            root,
+            incoming: Vec::new(),
+            outgoing: Vec::new(),
+        }
+    }
+
+    /// Create a new subhugr from a DFG-rooted hugr
+    pub fn from_dfg(hugr: Base) -> Self
+    where
+        Base: HugrView,
+    {
+        let root = hugr.root();
+        Self::from_dfg_root(hugr, root)
+    }
+
+    /// Create a new subhugr from a DFG node in a Hugr
+    /// 
+    /// This also works for e.g. FuncDefn nodes, i.e. roots where the first
+    /// two children correspond to inputs and outputs.
+    /// 
+    /// Panics if it could not find an input and an output node.
+    pub fn from_dfg_root(hugr: Base, root: Node) -> Self
+    where
+        Base: HugrView,
+    {
+        let (inp, out) = hugr.children(root).take(2).collect_tuple().unwrap();
+        let incoming = BoundaryEdge::get_boundary_outgoing(inp, &hugr).collect();
+        let outgoing = BoundaryEdge::get_boundary_incoming(out, &hugr).collect();
+        Self {
+            hugr,
+            root,
+            incoming,
+            outgoing,
+        }
+    }
+
+    /// Creates a new subhugr.
+    ///
+    /// The incoming and outgoing edges can be expressed as either source or
+    /// target ports. If source ports are given, they will be converted to
+    /// target ports, and, in the presence of copies, the signature will be
+    /// expanded accordingly.
+    pub fn new(
+        hugr: Base,
+        root: Node,
+        incoming: impl IntoIterator<Item = (Node, Port)>,
+        outgoing: impl IntoIterator<Item = (Node, Port)>,
+    ) -> Self
+    where
+        Base: HugrView,
+    {
+        let incoming = incoming
+            .into_iter()
+            .flat_map(|(n, p)| BoundaryEdge::get_boundary_edges(n, p, &hugr))
+            .collect();
+        let outgoing = outgoing
+            .into_iter()
+            .flat_map(|(n, p)| BoundaryEdge::get_boundary_edges(n, p, &hugr))
+            .collect();
+        Self {
+            hugr,
+            root,
+            incoming,
+            outgoing,
+        }
+    }
+
+    /// An iterator over the nodes in the subhugr.
+    pub fn nodes(&self) -> impl Iterator<Item = Node> {
+        [].into_iter()
+    }
+
+    /// The number of incoming and outgoing wires of the subhugr.
+    pub fn boundary_signature(&self) -> (usize, usize)
+    where
+        Base: HugrView,
+    {
+        let incoming = self.incoming.len();
+        let outgoing = self.outgoing.len();
+        (incoming, outgoing)
+    }
+
+    /// Construct a [`SimpleReplacement`] to replace `self` with `replacement`.
+    ///
+    /// `replacement` must be a hugr with DFG root and its signature must
+    /// match the signature of the subhugr.
+    ///
+    /// Panics if
+    ///  - `replacement` is not a hugr with DFG root, or
+    ///  - the DFG root does not have an input and output node, or
+    ///  - the number of incoming and outgoing ports in replacement does not
+    ///    match the subhugr boundary signature.
+    pub fn create_simple_replacement(&self, replacement: Hugr) -> SimpleReplacement
+    where
+        Base: AsRef<Hugr>,
+    {
+        let removal = self.nodes().collect();
+
+        let rep_root = replacement.root();
+        if replacement.get_optype(rep_root).tag() != OpTag::Dfg {
+            panic!("Replacement must have DFG root");
+        }
+        let Some((rep_input, rep_output)) = replacement
+            .children(rep_root)
+            .take(2)
+            .collect_tuple() else { panic!("Invalid DFG node") };
+        let rep_inputs = BoundaryEdge::get_boundary_outgoing(rep_input, &replacement);
+        let rep_outputs = BoundaryEdge::get_boundary_incoming(rep_output, &replacement);
+        let nu_inp = rep_inputs
+            .map_into()
+            .zip_eq(self.incoming.iter().copied().map_into())
+            .collect();
+        let nu_out = self
+            .outgoing
+            .iter()
+            .copied()
+            .map_into()
+            .zip_eq(rep_outputs.map(|BoundaryEdge(_, p)| p))
+            .collect();
+
+        SimpleReplacement::new(self.root, removal, replacement, nu_inp, nu_out)
+    }
+
+    /// Whether the subhugr is convex.
+    pub fn is_convex(&self) {
+        todo!()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        builder::{
+            BuildError, DFGBuilder, Dataflow, DataflowHugr, DataflowSubContainer, HugrBuilder,
+            ModuleBuilder,
+        },
+        hugr::region::Region,
+        ops::{handle::NodeHandle, LeafOp},
+        type_row,
+        types::{Signature, SimpleType},
+    };
+
+    use super::*;
+
+    const QB: SimpleType = SimpleType::Qubit;
+
+    fn build_hugr() -> Result<(Hugr, Node), BuildError> {
+        let mut mod_builder = ModuleBuilder::new();
+        let func = mod_builder.declare("test", Signature::new_linear(type_row![QB, QB]))?;
+        let func_id = {
+            let mut dfg = mod_builder.define_declaration(&func)?;
+            let outs = dfg.add_dataflow_op(LeafOp::CX, dfg.input_wires())?;
+            dfg.finish_with_outputs(outs.outputs())?
+        };
+        let hugr = mod_builder
+            .finish_hugr()
+            .map_err(|e| -> BuildError { e.into() })?;
+        Ok((hugr, func_id.node()))
+    }
+
+    #[test]
+    fn construct_subhugr() {
+        let (hugr, func_root) = build_hugr().unwrap();
+        let from_root = SubHugr::from_root(&hugr, func_root);
+        let region = RegionView::new(&hugr, func_root);
+        let from_region = SubHugr::from_region(region);
+        assert_eq!(from_root.root, from_region.root);
+        assert_eq!(from_root.incoming, from_region.incoming);
+        assert_eq!(from_root.outgoing, from_region.outgoing);
+    }
+
+    #[test]
+    fn construct_simple_replacement() {
+        let (mut hugr, func_root) = build_hugr().unwrap();
+        let sub = SubHugr::from_dfg_root(&hugr, func_root);
+
+        let empty_dfg = {
+            let builder = DFGBuilder::new(type_row![QB, QB], type_row![QB, QB]).unwrap();
+            let inputs = builder.input_wires();
+            builder.finish_hugr_with_outputs(inputs).unwrap()
+        };
+
+        let rep = sub.create_simple_replacement(empty_dfg);
+
+        hugr.apply_rewrite(rep).unwrap();
+
+        assert_eq!(hugr.node_count(), 5); // Module + Def + DFG + In + Out
+    }
+
+    #[test]
+    #[should_panic(expected = "zip_eq")]
+    fn construct_simple_replacement_signature_panics() {
+        let (hugr, dfg) = build_hugr().unwrap();
+        let sub = SubHugr::from_dfg_root(&hugr, dfg);
+
+        let empty_dfg = {
+            let builder = DFGBuilder::new(type_row![QB], type_row![QB]).unwrap();
+            let inputs = builder.input_wires();
+            builder.finish_hugr_with_outputs(inputs).unwrap()
+        };
+
+        sub.create_simple_replacement(empty_dfg);
+    }
+}
--- a/src/hugr/subhugr.rs
+++ b/src/hugr/subhugr.rs
@@ -1,12 +1,28 @@
+//! Views into subgraphs of HUGRs.
+//!
+//! Subgraphs are a generalisation of a [`Region`], in which there can be
+//! more than one root node. However, all root nodes must belong to the same
+//! sibling graph.
+//!
+//! TODO:
+//! Subgraphs implement [`HugrView`] as well as petgraph's _visit_ traits.
+
+use std::cell::OnceCell;
+
 use derive_more::Into;
 use itertools::Itertools;
+use portgraph::{view::Subgraph, LinkView, PortIndex, PortView};
 
 use crate::{
     ops::{OpTag, OpTrait},
     Direction, Hugr, Node, Port, SimpleReplacement,
 };
 
-use super::{region::RegionView, HugrView};
+use super::{
+    region::{Region, RegionView},
+    view::sealed::HugrInternals,
+    HugrView,
+};
 
 /// A boundary edge of a subhugr.
 ///
@@ -17,13 +33,24 @@ use super::{region::RegionView, HugrView};
 struct BoundaryEdge(Node, Port);
 
 impl BoundaryEdge {
+    fn target_port_index<G: PortView>(&self, g: &G) -> PortIndex {
+        let Node { index: node } = self.0;
+        let Port { offset: port } = self.1;
+        g.port_index(node, port).unwrap()
+    }
+
+    fn source_port_index<G: LinkView>(&self, g: &G) -> PortIndex {
+        let tgt = self.target_port_index(g);
+        g.port_link(tgt).unwrap().into()
+    }
+
     fn get_boundary_edges<H: HugrView>(
         node: Node,
         port: Port,
         hugr: &H,
     ) -> impl Iterator<Item = Self> {
         if port.direction() == Direction::Incoming {
-            [BoundaryEdge(node, port)].to_vec().into_iter()
+            vec![BoundaryEdge(node, port)].into_iter()
         } else {
             hugr.linked_ports(node, port)
                 .map(|(n, p)| BoundaryEdge(n, p))
@@ -80,40 +107,40 @@ impl BoundaryEdge {
 /// descendant of a node in the subhugr is itself also in the subhugr.
 /// If both incoming and outgoing ports are empty, the subhugr is taken to be
 /// all children of the root.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct SubHugr<Base> {
-    hugr: Base,
+///
+/// This does not implement Sync as we use a `OnceCell` to cache the sibling
+/// graph.
+#[derive(Clone, Debug)]
+pub struct SubHugr<'g, Base: HugrInternals> {
+    hugr: &'g Base,
     root: Node,
     incoming: Vec<BoundaryEdge>,
     outgoing: Vec<BoundaryEdge>,
+    sibling_graph: OnceCell<Subgraph<'g, Base::Portgraph>>,
 }
 
-impl<'g, Base: HugrView> SubHugr<RegionView<'g, Base>> {
+impl<'g, Base: HugrView> SubHugr<'g, RegionView<'g, Base>> {
     /// Creates a new subhugr from a region.
-    pub fn from_region(region: RegionView<'g, Base>) -> Self {
+    pub fn from_region(region: &'g RegionView<'g, Base>) -> Self {
         let root = region.root();
-        Self {
-            hugr: region,
-            root,
-            incoming: Vec::new(),
-            outgoing: Vec::new(),
-        }
+        Self::new(region, root, [], [])
     }
 }
 
-impl<Base> SubHugr<Base> {
+impl<'g, Base: HugrInternals> SubHugr<'g, Base> {
     /// Creates a new subhugr from a root node
-    pub fn from_root(hugr: Base, root: Node) -> Self {
+    pub fn from_root(hugr: &'g Base, root: Node) -> Self {
         Self {
             hugr,
             root,
             incoming: Vec::new(),
             outgoing: Vec::new(),
+            sibling_graph: OnceCell::new(),
         }
     }
 
     /// Create a new subhugr from a DFG-rooted hugr
-    pub fn from_dfg(hugr: Base) -> Self
+    pub fn from_dfg(hugr: &'g Base) -> Self
     where
         Base: HugrView,
     {
@@ -122,23 +149,24 @@ impl<Base> SubHugr<Base> {
     }
 
     /// Create a new subhugr from a DFG node in a Hugr
-    /// 
+    ///
     /// This also works for e.g. FuncDefn nodes, i.e. roots where the first
     /// two children correspond to inputs and outputs.
-    /// 
+    ///
     /// Panics if it could not find an input and an output node.
-    pub fn from_dfg_root(hugr: Base, root: Node) -> Self
+    pub fn from_dfg_root(hugr: &'g Base, root: Node) -> Self
     where
         Base: HugrView,
     {
         let (inp, out) = hugr.children(root).take(2).collect_tuple().unwrap();
-        let incoming = BoundaryEdge::get_boundary_outgoing(inp, &hugr).collect();
-        let outgoing = BoundaryEdge::get_boundary_incoming(out, &hugr).collect();
+        let incoming = BoundaryEdge::get_boundary_outgoing(inp, hugr).collect();
+        let outgoing = BoundaryEdge::get_boundary_incoming(out, hugr).collect();
         Self {
             hugr,
             root,
             incoming,
             outgoing,
+            sibling_graph: OnceCell::new(),
         }
     }
 
@@ -149,7 +177,7 @@ impl<Base> SubHugr<Base> {
     /// target ports, and, in the presence of copies, the signature will be
     /// expanded accordingly.
     pub fn new(
-        hugr: Base,
+        hugr: &'g Base,
         root: Node,
         incoming: impl IntoIterator<Item = (Node, Port)>,
         outgoing: impl IntoIterator<Item = (Node, Port)>,
@@ -159,23 +187,47 @@ impl<Base> SubHugr<Base> {
     {
         let incoming = incoming
             .into_iter()
-            .flat_map(|(n, p)| BoundaryEdge::get_boundary_edges(n, p, &hugr))
+            .flat_map(|(n, p)| BoundaryEdge::get_boundary_edges(n, p, hugr))
             .collect();
         let outgoing = outgoing
             .into_iter()
-            .flat_map(|(n, p)| BoundaryEdge::get_boundary_edges(n, p, &hugr))
+            .flat_map(|(n, p)| BoundaryEdge::get_boundary_edges(n, p, hugr))
             .collect();
         Self {
             hugr,
             root,
             incoming,
             outgoing,
+            sibling_graph: OnceCell::new(),
         }
     }
 
+    fn get_sibling_graph(&self) -> &Subgraph<'g, Base::Portgraph> {
+        self.sibling_graph.get_or_init(|| {
+            let graph = self.hugr.portgraph();
+            let incoming = self
+                .incoming
+                .iter()
+                .copied()
+                .map(|e| e.target_port_index(graph));
+            let outgoing = self
+                .outgoing
+                .iter()
+                .copied()
+                .map(|e| e.source_port_index(graph));
+            Subgraph::new_subgraph(graph, incoming.chain(outgoing))
+        })
+    }
+
     /// An iterator over the nodes in the subhugr.
-    pub fn nodes(&self) -> impl Iterator<Item = Node> {
-        [].into_iter()
+    pub fn nodes(&self) -> impl Iterator<Item = Node> + '_
+    where
+        Base: HugrView,
+    {
+        self.get_sibling_graph().nodes_iter().flat_map(|index| {
+            let region = RegionView::new(self.hugr, Node { index });
+            region.nodes().collect_vec()
+        })
     }
 
     /// The number of incoming and outgoing wires of the subhugr.
@@ -230,8 +282,8 @@ impl<Base> SubHugr<Base> {
     }
 
     /// Whether the subhugr is convex.
-    pub fn is_convex(&self) {
-        todo!()
+    pub fn is_convex(&self) -> bool {
+        self.get_sibling_graph().is_convex()
     }
 }
 
@@ -271,7 +323,7 @@ mod tests {
         let (hugr, func_root) = build_hugr().unwrap();
         let from_root = SubHugr::from_root(&hugr, func_root);
         let region = RegionView::new(&hugr, func_root);
-        let from_region = SubHugr::from_region(region);
+        let from_region = SubHugr::from_region(&region);
         assert_eq!(from_root.root, from_region.root);
         assert_eq!(from_root.incoming, from_region.incoming);
         assert_eq!(from_root.outgoing, from_region.outgoing);
@@ -290,9 +342,11 @@ mod tests {
 
         let rep = sub.create_simple_replacement(empty_dfg);
 
+        assert_eq!(rep.removal.len(), 1);
+
         hugr.apply_rewrite(rep).unwrap();
 
-        assert_eq!(hugr.node_count(), 5); // Module + Def + DFG + In + Out
+        assert_eq!(hugr.node_count(), 4); // Module + Def + In + Out
     }
 
     #[test]
@@ -308,5 +362,25 @@ mod tests {
         };
 
         sub.create_simple_replacement(empty_dfg);
+    }
+
+    #[test]
+    fn convex_subhugr() {
+        let (hugr, func_root) = build_hugr().unwrap();
+        let sub = SubHugr::from_dfg_root(&hugr, func_root);
+        assert!(sub.is_convex());
+    }
+
+    #[test]
+    fn non_convex_subhugr() {
+        let (hugr, func_root) = build_hugr().unwrap();
+        let (inp, out) = hugr.children(func_root).take(2).collect_tuple().unwrap();
+        let sub = SubHugr::new(
+            &hugr,
+            func_root,
+            hugr.node_outputs(inp).map(|p| (inp, p)),
+            hugr.node_inputs(out).map(|p| (out, p)),
+        );
+        assert!(!sub.is_convex());
     }
 }
--- a/src/ops/tag.rs
+++ b/src/ops/tag.rs
@@ -62,6 +62,9 @@ pub enum OpTag {
     BasicBlock,
     /// A control flow exit node.
     BasicBlockExit,
+
+    /// A node with input and output children
+    IOBlock,
 }
 
 impl OpTag {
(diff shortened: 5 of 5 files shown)


lmondada:
I've been trying to implement the `HugrView` and `HugrInternals` traits but it's a nightmare in two ways:
- getting the iterator types right,
- exposing a `PortView` that represents the underlying portgraph -- probably need to create yet another type in portgraph that combines `Region`s and `Subgraph`s.

For now, I've added a todo flag and will leave this for later.

src/hugr/rewrite/simple_replace.rs
```diff
@@ -58,7 +59,7 @@ impl Rewrite for SimpleReplacement {
 
     fn apply(self, h: &mut Hugr) -> Result<(), SimpleReplacementError> {
         // 1. Check the parent node exists and is a DFG node.
-        if h.get_optype(self.parent).tag() != OpTag::Dfg {
+        if h.get_optype(self.parent).tag().partial_cmp(&OpTag::IOBlock) != Some(Ordering::Less) {
```
acl-cqc, line 62: Something like `OpTag::IOBlock.is_superset(h.get_optype(self.parent).tag())` I think?

acl-cqc:
Hmmm. `Region` is of course a bit of a misnomer - really, `FlatRegion` should be something like `SiblingGraph` and `Region` something like `Descendants`. Then you could call this region, or similar....

But if we want a general term, how about Area?

acl-cqc:
> really, `FlatRegion` should be something like `SiblingGraph`

Or maybe `FlatRegionView` should be `ChildrenView`? (Or `FamilyView` :-) :-) )

src/hugr/rewrite/simple_replace.rs
acl-cqc: Oh, `OpTag::DataflowParent` after merging

doug-q:
I've implemented HugrView here: https://github.com/CQCL-DEV/hugr/pull/355

lmondada:
Hi @acl-cqc, 

I've made a few changes that build on top of #347, #323 and @doug-q 's #355, I think the PR is now in much better shape:
- `SugHugr` was renamed to `SiblingSubgraph` and its semantics were changed to only include the subgraph within the sibling graph and not its further descendants. This is all that I need and will make the `HugrView` impl much simpler.
- `HugrView` and the hierarchy-related views were moved into a common `hugr/views` folder, along with the new sibling subgraph code.

Regarding change 2: I find the file organisation much cleaner this way, apologies to @ss2165 for changing this twice in a row... I hope you will agree with me, but if you dislike the additional nesting or anything else, happy to revert the change of course!

src/hugr/views/hierarchy.rs
```diff
@@ -50,6 +50,15 @@ where
     hugr: &'g Base,
 }
 
+impl<'g, Base> SiblingGraph<'g, Base>
+where
+    Base: HugrInternals,
+{
+    pub(super) fn base(&self) -> &Base {
```
acl-cqc, line 57: Does this actually give you anything over just `pub(super)`-ing the `hugr` field? I haven't grokked what's going on with lifetimes yet but I think this just narrows the lifetime of the `self.hugr` reference, right?

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
+//! Views for HUGR sibling subgraphs.
+//!
+//! Views into subgraphs of HUGRs within a single level of the
+//! hierarchy, i.e. within a sibling graph. Such a subgraph is
+//! given by a root node, of which all nodes in the sibling subgraph are
+//! children, as well as a subgraph boundary, defining the separation between
+//! inside and outside the subgraph. The boundary is fully contained within the
```
acl-cqc, line 7: The *subgraph* is fully contained within the *graph* too, where contained = is-subset-of.

We don't know what a boundary is exactly yet, so the notion of containment of a boundary is necessarily somewhat vague...

src/hugr/views/hugr.rs
```diff
@@ -9,11 +9,11 @@ use itertools::{Itertools, MapInto};
 use portgraph::dot::{DotFormat, EdgeStyle, NodeStyle, PortStyle};
 use portgraph::{multiportgraph, LinkView, MultiPortGraph, PortView};
 
-use super::{Hugr, NodeMetadata, NodeType};
-use super::{Node, Port};
+use crate::hugr::{Hugr, NodeMetadata, NodeType};
```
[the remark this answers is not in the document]
lmondada: Yes, why not, very good idea!

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
+//! Views for HUGR sibling subgraphs.
```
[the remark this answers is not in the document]
lmondada: Yes, will do. The reason is historical: I had just made another renaming PR, and here I changed my mind. I thought making to PRs in a row that rename the same thing was a bit outrageous. But it's probably the right thing to do.

I'll split this in a separate PR!

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+        let outgoing = self.outgoing.len();
+        (incoming, outgoing)
+    }
+
+    /// Construct a [`SimpleReplacement`] to replace `self` with `replacement`.
+    ///
+    /// `replacement` must be a hugr with DFG root and its signature must
+    /// match the signature of the subgraph.
+    ///
+    /// Panics if
```
acl-cqc, line 265: Note that `specification/hugr.md` defines SimpleReplace as requiring the replaced nodes to be convex. That check isn't implemented there, but you could check that here. (That might reduce the utility of allowing non-convex SiblingSubgraphs, too...)

acl-cqc (commented):
Ok, thanks Luca, there is some fairly serious stuff here, although the really hard work of convexity checking is in Portgraph so I'm not looking at that :-). I've quite a few comments on restructuring though and TBH I'm not certain I've grokked all of what's here (hopefully I will after I've clarified the bits I'm commenting on). So, sorry for the long review, but I reckon we can make this significantly easier to understand and reason about, here goes....

src/hugr/views/hugr.rs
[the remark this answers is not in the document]
lmondada: resolved in 386

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
lmondada: resolved in #386

src/hugr/views/hierarchy.rs
lmondada: Done!

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
+//! Views for HUGR sibling subgraphs.
+//!
+//! Views into subgraphs of HUGRs within a single level of the
+//! hierarchy, i.e. within a sibling graph. Such a subgraph is
+//! given by a root node, of which all nodes in the sibling subgraph are
```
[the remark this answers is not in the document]
lmondada: Done

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+
+use super::{sealed::HugrInternals, HierarchyView, HugrView, SiblingGraph};
+
+/// A boundary edge of a sibling subgraph.
+///
+/// It is always uniquely identified (any Hugr edge is) by the node and port
+/// of the target of the edge. Source edges are not unique as they can involve
+/// copies.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Into)]
+struct BoundaryEdge(Node, Port);
```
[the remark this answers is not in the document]
lmondada: Yes, I agree with you. I've opted instead to implement an _actual_ boundary edge: now this struct also includes whether it's an incoming or outgoing boundary edge, and it panics if the edge does not exist.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+
+impl BoundaryEdge {
+    fn target_port_index<G: PortView>(&self, g: &G) -> PortIndex {
+        let Node { index: node } = self.0;
+        let Port { offset: port } = self.1;
+        g.port_index(node, port).unwrap()
```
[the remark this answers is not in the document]
lmondada: Took your suggestion + some rewriting.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    fn source_port_index<G: LinkView>(&self, g: &G) -> Option<PortIndex> {
```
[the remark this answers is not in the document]
lmondada: Yes, I've done this. Now `BoundaryEdge` really is a boundary edge, and dangling ports are no longer allowed. Thanks for the suggestion!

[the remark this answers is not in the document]
lmondada: As for your last comment: I'll line up another PR that will make `SimpleReplacement` use `SiblingSubgraph`.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+        port: Port,
+        hugr: &H,
+    ) -> impl Iterator<Item = Self> {
+        if port.direction() == Direction::Incoming {
+            vec![BoundaryEdge(node, port)].into_iter()
+        } else {
+            hugr.linked_ports(node, port)
+                .map(|(n, p)| BoundaryEdge(n, p))
+                .collect_vec()
+                .into_iter()
```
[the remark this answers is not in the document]
lmondada: This was also rewritten, returning `Vec<_>` as you suggest.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+        }
+    }
+
+    /// The DFG input ports of a `node`
+    fn get_boundary_incoming<H: HugrView>(node: Node, hugr: &H) -> impl Iterator<Item = Self> + '_ {
```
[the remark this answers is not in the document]
lmondada: Yes that's better. I've gone with `from_node_inputs` and `from_node_outputs`. I thought this worked better with the other new names

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+            .flat_map(move |p| hugr.linked_ports(node, p))
+            .map(|(n, p)| BoundaryEdge(n, p))
+    }
+}
+
+/// A subgraph of a HUGR sibling graph.
+///
+/// The subgraph is described by the following data:
+///  - a HugrView to the underlying HUGR,
+///  - a root node, of which all nodes in the [`SiblingSubgraph`] must be
```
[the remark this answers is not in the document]
lmondada: Absolutely!

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+        Self {
+            hugr,
+            root,
+            incoming,
+            outgoing,
+            sibling_graph: OnceCell::new(),
+        }
+    }
+
+    fn get_sibling_graph(&self) -> &Subgraph<'g, Base::Portgraph> {
```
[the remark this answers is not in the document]
lmondada: It's now `underlying_portgraph`. Apologies for the bad name, this is also a relict of earlier code semantics.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+    ops::{OpTag, OpTrait},
+    Direction, Hugr, Node, Port, SimpleReplacement,
+};
+
+use super::{sealed::HugrInternals, HierarchyView, HugrView, SiblingGraph};
+
+/// A boundary edge of a sibling subgraph.
+///
+/// It is always uniquely identified (any Hugr edge is) by the node and port
+/// of the target of the edge. Source edges are not unique as they can involve
```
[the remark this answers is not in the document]
lmondada: That's much better, thanks!

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
lmondada: In fact, it's now totally gone as I've removed the `OnceCell`.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+/// are
+///  - adjacent to an interior edge, or
+///  - the target of an incoming boundary edge and the source of an outgoing
+///    boundary edge.
+///
+/// The root node itself is not included in the subgraph. If both incoming and
+/// outgoing ports are empty, the subgraph is taken to be all children of the
+/// root and is equivalent to a [`SiblingGraph`].
+///
+/// This does not implement Sync as we use a `OnceCell` to cache the sibling
```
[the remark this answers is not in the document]
lmondada: Done!

src/hugr/views/sibling.rs
lmondada: I've changed `create_simple_replacement` to return a `Result` and to check that the subgraph is convex.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+///  - a HugrView to the underlying HUGR,
+///  - a root node, of which all nodes in the [`SiblingSubgraph`] must be
+///    children,
+///  - incoming/outgoing boundary edges, pointing into/out of the subgraph.
+/// The incoming/outgoing edges must be contained within the sibling graph of the
+/// root node. Their ordering matters when using the [`SiblingSubgraph`] for
+/// replacements, as it will match the ordering of the ports in the replacement.
+/// The list of incoming and/or outgoing ports may be empty.
+///
+/// A subgraph is well-defined if the incoming/outgoing edges partition the
```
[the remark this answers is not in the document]
lmondada: Sure, let's try and simplify. Your proposal doesn't quite work, for example in the following

```mermaid
graph TD
  0-->|incoming|1
  subgraph "subgraph"
  2-->1
  end
  1-->|outgoing|3
```

Your definition would not include node 2 unfortunately. Furthermore, we need to be talking about edges and not vertices for the general case (although talking about vertices would be enough for the convex case, my issue with that is that we are pushing half of the convexity check onto the client, as they have to check that the set of edges they have is indeed an induced subgraph).

An alternative criterion could be something along the lines of:

Let $G = (V, E)$ be the graph and $B$ the set of boundary edges. Edges in $B$ can be an incoming boundary edge, an outgoing boundary edge, or both. Then the subgraph defined by $B$ is the graph
given by the connected components of $G' = (V, E \setminus B)$ that contain a node that is either the target of an incoming boundary edge, or the source of an outgoing boundary edge.

A subgraph is well-formed if every edge in $B$ into the subgraph is an incoming boundary edge and every edge in $B$ pointing out of the subgraph is an outgoing boundary edge.

I will happily implement a well-formedness check, but that would belong in `portgraph` and should be IMO a separate PR.

Is that clearer?

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+}
+
+/// A subgraph of a HUGR sibling graph.
+///
+/// The subgraph is described by the following data:
+///  - a HugrView to the underlying HUGR,
+///  - a root node, of which all nodes in the [`SiblingSubgraph`] must be
+///    children,
+///  - incoming/outgoing boundary edges, pointing into/out of the subgraph.
```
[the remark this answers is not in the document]
lmondada: Rewrote this.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,427 @@
...
+/// The incoming/outgoing edges must be contained within the sibling graph of the
+/// root node. Their ordering matters when using the [`SiblingSubgraph`] for
+/// replacements, as it will match the ordering of the ports in the replacement.
+/// The list of incoming and/or outgoing ports may be empty.
```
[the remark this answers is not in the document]
lmondada: Rewrote this.

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
lmondada: See https://github.com/CQCL-DEV/hugr/discussions/406 and my proposed commits.

src/hugr/views.rs
```diff
@@ -1,6 +1,7 @@
 //! Read-only access into HUGR graphs and subgraphs.
 
 pub mod hierarchy;
+pub mod sibling;
 
 pub use hierarchy::{DescendantsGraph, HierarchyView, SiblingGraph};
```
aborgna-q, line 6:
```suggestion
pub use hierarchy::{DescendantsGraph, HierarchyView, SiblingGraph};
pub use sibling::SiblingSubgraph
```

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+//! Sibling subgraphs complement [`super::HierarchyView`]s in the sense that the
+//! latter provide views for subgraphs defined by hierarchical relationships,
+//! while the former provide views for subgraphs within a single level of the
+//! hierarchy.
+//!
+
+// TODO:
+// //! This module exposes the [`SiblingView`] trait, which is currently
+// //! implemented by the [`SiblingSubgraph`] struct in this module, as well as
+// //! the [`SiblingGraph`] hierarchical view.
```
aborgna-q, line 18: Why the TODO? What'd be the goal of the trait?

(If not needed, remove the commented-out stub below too)

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// This panics if the sibling graph is empty.
+    pub fn from_sibling_graph(sibling_graph: &'g Base) -> Self
+    where
+        Base: HugrView,
+    {
+        let root = sibling_graph.root();
+        let nodes = sibling_graph.children(root).collect_vec();
+        if nodes.is_empty() {
+            panic!("Empty graph");
+        }
```
aborgna-q, line 72: I'd rather return an `InvalidSubgraph` for an user's error. (Same for `from_dataflow_graph`)

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// A sibling subgraph from a [`crate::ops::OpTag::DataflowParent`]-rooted HUGR.
+    ///
+    /// The subgraph is given by the nodes between the input and output
+    /// children nodes of the parent node. If you wish to create a subgraph
+    /// from another root, wrap the `region` argument in a [`super::SiblingGraph`].
+    ///
+    /// This panics if the DFG graph is empty.
+    pub fn from_dataflow_graph(dfg_graph: &'g Base) -> Self
+    where
+        Base: HugrView,
```
aborgna-q, line 88: You can restrict the root node to dfgs here.
```suggestion
        Base: HugrView<RootHandle = DfgID>,
```
You may need to edit the siblinggraph defs in your tests;
`SiblingGraph<'_>` → `SiblingGraph<'_, _>`

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+        if !checker.is_node_convex(nodes.iter().map(|n| n.index)) {
+            return Err(InvalidSubgraph::NotConvex);
+        }
+        if nodes.is_empty() {
+            return Err(InvalidSubgraph::EmptySubgraph);
+        }
+        let Some(parent) = base.get_parent(nodes[0]) else {
+            return Err(InvalidSubgraph::NoSharedParent);
+        };
+        if !nodes.iter().all(|&n| base.get_parent(n) == Some(parent)) {
```
aborgna-q, line 228:
```suggestion
        if !nodes.iter().map(|&n| base.get_parent(n)).all_equal() {
```

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// Intuitively, the sibling subgraph is all the edges and nodes between
+    /// an edge of $B_I$ and an edge of $B_O$.
+    ///
+    /// The `incoming` and `outgoing` arguments give $B_I$ and $B_O$ respectively.
+    /// They can be either source or target ports. We currently assume that if
+    /// the source port of an outgoing boundary edge is linked to multiple
+    /// target ports, then all edges from the same source port are outgoing
+    /// boundary edges.
+    ///
+    /// More fomally, the sibling subgraph defined by $B_I$ and $B_O$ is the
```
aborgna-q, line 114:
```suggestion
    /// More formally, the sibling subgraph defined by $B_I$ and $B_O$ is the
```

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+    ///  - the target of an incoming boundary edge, or
+    ///  - the source of an outgoing boundary edge.
+    ///
+    /// A subgraph is well-formed if every edge in B into the subgraph is an
+    /// incoming boundary edge and every edge in B pointing out of the subgraph
+    /// is an outgoing boundary edge.
+    ///
+    /// This function fails if the subgraph if it is not convex, if the nodes
+    /// do not share a common parent or if the subgraph is empty.
+    /// TODO: fail if not well-formed.
```
aborgna-q, line 126:
```suggestion
```
?

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+        if nodes.is_empty() {
+            panic!("Empty DFG graph");
+        }
+        Self {
+            base: dfg_graph,
+            nodes,
+        }
+    }
+
+    /// Create a new sibling subgraph from some boundary edges.
```
aborgna-q, line 101: Shouldn't this be called `from_boundary_ports` instead?

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+            return Err(InvalidSubgraph::NotConvex);
+        }
+        let nodes = subpg.nodes_iter().map_into().collect_vec();
+        if nodes.is_empty() {
+            return Err(InvalidSubgraph::EmptySubgraph);
+        }
+        let Some(parent) = base.get_parent(nodes[0]) else {
+            return Err(InvalidSubgraph::NoSharedParent);
+        };
+        if !nodes.iter().all(|&n| base.get_parent(n) == Some(parent)) {
```
aborgna-q, line 186:
```suggestion
        if !nodes.iter().map(|&n| base.get_parent(n)).all_equal() {
```

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+        replacement: Hugr,
+    ) -> Result<SimpleReplacement, InvalidReplacement>
+    where
+        Base: HugrView,
+    {
+        let removal = self.nodes().iter().copied().collect();
+
+        let rep_root = replacement.root();
+        let dfg_optype = replacement.get_optype(rep_root);
+        if dfg_optype.tag() != OpTag::Dfg {
```
aborgna-q, line 347:
```suggestion
        if !OpTag::Dfg.is_superset(dfg_optype.tag())  {
```
It's currently the same, but for future-proofing.

aborgna-q (changes requested):
Nice!

src/hugr/views/sibling.rs
acl-cqc: That would mean an empty sibling subgraph, i.e. without even Input/Output nodes right? Isn't that an invalid Hugr? In which case panicking is acceptable??

acl-cqc: AFAICS this method is *identical* to `from_dataflow_graph` so we can probably remove it?

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// target ports, then all edges from the same source port are outgoing
+    /// boundary edges.
+    ///
+    /// More fomally, the sibling subgraph defined by $B_I$ and $B_O$ is the
+    /// graph given by the connected components of the graph with edges
+    /// $E\B_I\B_O$ which contain at least one node that is either
+    ///  - the target of an incoming boundary edge, or
+    ///  - the source of an outgoing boundary edge.
+    ///
+    /// A subgraph is well-formed if every edge in B into the subgraph is an
```
acl-cqc, line 120: This paragraph needs rephrasing given that incoming/outgoing are already separated. Something like "there are no incoming edges other than those in `incoming` and similarly for `outgoing`....no, hang on, that's guaranteed. IOW I'm not sure what the well-formedness criterion is here...

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// incoming boundary edge and every edge in B pointing out of the subgraph
+    /// is an outgoing boundary edge.
+    ///
+    /// This function fails if the subgraph if it is not convex, if the nodes
```
acl-cqc, line 124: Is there an additional requirement that `incoming` and `outgoing` are disjoint?

acl-cqc: ok, it'd be slightly nastier than that - you can now identify a single edge by *either* endpoint...

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+        outgoing: impl IntoIterator<Item = (Node, Port)>,
+        checker: &mut ConvexChecker<&'g Base::Portgraph>,
+    ) -> Result<Self, InvalidSubgraph>
+    where
+        Base: HugrView,
+    {
+        let pg = base.portgraph();
+        let to_pg = |(n, p): (Node, Port)| pg.port_index(n.index, p.offset).expect("invalid port");
+        let incoming = incoming.into_iter().flat_map(|(n, p)| {
+            if p.direction() == Direction::Outgoing {
```
acl-cqc, line 162: nit: consider `match` here

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+        });
+        let outgoing = outgoing.into_iter().flat_map(|(n, p)| {
+            if p.direction() == Direction::Incoming {
+                base.linked_ports(n, p).map(to_pg).collect()
+            } else {
+                vec![to_pg((n, p))]
+            }
+        });
+        let subpg = Subgraph::new_subgraph(pg, incoming.chain(outgoing));
+        if !subpg.is_convex_with_checker(checker) {
```
acl-cqc, line 176: (probably for another PR) it'd be great if these returned the node(s) demonstrating non-convexity (i.e. reached by the outgoing edges and reaching the incoming edges)

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+        nodes: Vec<Node>,
+        checker: &mut ConvexChecker<&'g Base::Portgraph>,
+    ) -> Result<Self, InvalidSubgraph>
+    where
+        Base: HugrView,
+    {
+        if !checker.is_node_convex(nodes.iter().map(|n| n.index)) {
+            return Err(InvalidSubgraph::NotConvex);
+        }
+        if nodes.is_empty() {
```
acl-cqc, line 222: Do we need to rule this out? Since `SiblingSubgraph` is now represented as base+nodes, and the latter can be an empty vec, can we just proceed?

An empty SiblingSubgraph seems to follow naturally if you pass in an empty set of boundary edges; the case of a SiblingSubgraph being *the whole SiblingGraph* to me seems less natural. And I'm not sure I really see a SimpleReplacement replacing the input/output nodes either (@cqc-alec ?) - we'd need a special case as the replacement part thereof specifically does *not* copy its input/output nodes in.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+    pub fn nodes(&self) -> &[Node] {
+        &self.nodes
+    }
+
+    /// Whether a port is at the subgraph boundary.
+    fn is_boundary_port(&self, n: Node, p: Port) -> bool
+    where
+        Base: HugrView,
+    {
+        self.base
```
acl-cqc, line 244:
```suggestion
        self.nodes.contains(n) && self.base
```

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+            .collect_vec();
+        AbstractSignature::new(input, output, input_static)
+    }
+
+    /// The parent of the sibling subgraph.
+    pub fn get_parent(&self) -> Node
+    where
+        Base: HugrView,
+    {
+        self.base
```
acl-cqc, line 312: Don't we have `self.base.root()`?

src/hugr/views/sibling.rs
lmondada: I've removed the TODO. @doug-q had a PR for this #355 so I will leave it up to him to see if he wants to implement this.

src/hugr/views/sibling.rs
lmondada: I've gone with @aborgna-q's suggestion. The difference between `from_sibling_graph` and `from_dataflow_graph` is that the latter takes the input/output nodes as inputs and outputs, whereas the former takes the whole graph as-is. As such, an empty graph isn't invalid.

src/hugr/views/sibling.rs
lmondada: Mmm, to me boundary _ports_ would refer to the ports of the nodes within the subgraph, whereas in this case we allow any ports for a given edge (i.e. you can specify outgoing ports for an incoming boundary edge). Really, you are using ports as a way to specify an edge, not the port itself.

Maybe this problem arises from the fact that HUGR doesnt have an official way to refer to an edge

src/hugr/views/sibling.rs
lmondada: I've gone with 

> A subgraph is well-formed if for every edge in the HUGR
>  - it is in $B_I$ if and only if it has a source outside of the subgraph and a target inside of it, and
 > - it is in $B_O$ if and only if it has a source inside of the subgraph and a target outside of it.

src/hugr/views/sibling.rs
lmondada: That's not an additional requirement: that incoming and outgoing edges are disjoint is guaranteed by convexity.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+        let incoming = incoming.into_iter().flat_map(|(n, p)| {
+            if p.direction() == Direction::Outgoing {
+                base.linked_ports(n, p).map(to_pg).collect()
+            } else {
+                vec![to_pg((n, p))]
+            }
+        });
+        let outgoing = outgoing.into_iter().flat_map(|(n, p)| {
+            if p.direction() == Direction::Incoming {
+                base.linked_ports(n, p).map(to_pg).collect()
```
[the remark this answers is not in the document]
lmondada: Yes true, done!

src/hugr/views/sibling.rs
lmondada: Done, this _did_ make the code clearer + conciser

src/hugr/views/sibling.rs
lmondada: Yes true, but this would require a PR in `portgraph` first and it isn't immediately clear to me if this could be useful when pattern matching.

src/hugr/views/sibling.rs
lmondada: The problem with the empty case is that there is no way to specify _where_ this subgraph lives when using nodes to define it. It could be on any set of edges... Furthermore, from a pattern matching perspective, empty subgraphs match everywhere, which blows up the number of matches and makes them useless.

If you wanted to be able to refer to empty subgraphs, you'd have to revert to a more general representation of subgraph using boundary edges as I was suggesting initially. 

Finally, replacing entire graphs is useful for testing and could also apply to non-DFG graphs. I cannot tell you whether there is a use case for them presently, but I see no harm in allowing it.

src/hugr/views/sibling.rs
lmondada: A subgraph could have any node as its parent. As currently implemented, this is not necessarily the root of the Hugr view.

lmondada:
@aborgna-q and @acl-cqc thank you so much for your precious comments! I should have addressed all of them now!

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+                replacement
+                    .get_optype(rep_input)
+                    .signature()
+                    .get(p)
+                    .is_some()
+            })
+            .map(|p| {
+                replacement
+                    .linked_ports(rep_input, p)
+                    .exactly_one()
```
acl-cqc, line 358: If the replacement copies one of its inputs, won't this fire?

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+                .get(p)
+                .is_some()
+        });
+        let incoming = self.incoming_ports();
+        let outgoing = self
+            .outgoing_ports()
+            .map(|(n, p)| self.base.linked_ports(n, p));
+        let nu_inp = rep_inputs.into_iter().zip_eq(incoming).collect();
+        let nu_out = outgoing
+            .zip_eq(rep_outputs)
```
acl-cqc, line 375: I thought about this for awhile and agreed it was right :). However, it might be clearer if we reordered:
```
let nu_out = self
            .outgoing_ports()
            .zip_eq(rep_outputs)
            .flat_map(|((srcn, srcp), new_src)| {
                self.base
                    .linked_ports(srcn, srcp)
                    .map(move |target| (target, new_src))
            })
            .collect();
```
which although longer has one fewer `map` and I thought was a bit easier to follow. YMMV!

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,572 @@
...
+            .children(rep_root)
+            .take(2)
+            .collect_tuple()
+        else { return Err(InvalidReplacement::InvalidDataflowParent) };
+        if dfg_optype.signature() != self.signature() {
+            return Err(InvalidReplacement::InvalidSignature);
+        }
+        let rep_inputs = replacement
+            .node_outputs(rep_input)
+            .filter(|&p| {
```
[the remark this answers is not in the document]
lmondada: Are they called state order edges? I'm honestly not too sure, just know there are extra ports that for the purposes of flat subgraphs I can ignore.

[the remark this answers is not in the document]
lmondada: Or is this not a thing on input/output nodes?

src/hugr/views/sibling.rs
acl-cqc: Ah I see that's exactly what `InvalidCopy` is about. Hmmm. @cqc-alec do we need to extend/tweak SimpleReplace a bit here, to allow the replacement to copy its inputs?

lmondada: Sorry @acl-cqc , I'm confused here. Yes, it breaks if the replacement copies one of its inputs, but that is exactly what line 361 handles, and returns a `InvalidReplacement::InvalidCopy` error. At the output boundary, this problem does not exist because two edges cannot connect to the same port on the output port. If there was a mismatch in the output boundary it would get caught by the signature check on line 345.

Am I missing your point?

src/hugr/views/sibling.rs
lmondada: Thanks, went with your suggestion with a few variable name changes. Hope this is clearer!

lmondada:
I still need to fix the changes from updates to HUGR

lmondada:
Hey @acl-cqc sorry to bother you _again_. For two of your three comments I'd need some clarification, I'm not sure I've resolved your issues.

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
acl-cqc: Ah! Yes indeed, good point.

How about testing against `replacement.get_optype(rep_input).other_port_index(Direction::Outgoing)`? You'd have to do `Some(p) != ...`(that) I guess, so you may not feel that's any better.

src/hugr/views/sibling.rs
acl-cqc: Yes, indeed, you are quite right. (Crossed in post?)

src/hugr/views/sibling.rs
lmondada: Oh I see, I misunderstood you. The problem is that `self.base` might not be a SiblingGraph, it could be a generic HUGR.

What is true is that `self.get_parent() == self.root()` *IF* `SiblingSubgraph` was implementing `HugrView`. At the moment it is not. I've added this as a future TODO for another PR.

What do you suggest I do?

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
lmondada: I'll go with the current code + a comment explaining it. 

The docs of `other_port_index` state
> Returns None if there is no such port, **or if the operation defines multiple non-dataflow ports**

which doesnt inspire me.

src/hugr/views/sibling.rs
acl-cqc: Right, chatted to Alec, think SimpleReplacement is fine (i.e. capable) as it stands. What I meant to say was, you can handle this case, and it's no more difficult than what you do for the outputs.
I think you want:
```
let rep_inputs = replacement // as before
    ... // filter as before
    .map(|p| replacement.linked_ports(rep_input, p).collect()) // in place of the map+exactly_one+error you had before
   .collect::<Vec<Vec<_>>>();
...
let nu_inp = rep_inputs
   .into_iter()
   .zip_eq(incoming)
   .flat_map(|(rep_inp, inc)| {
      rep_inp
         .into_iter()
         .map(|tgt| (tgt, inc))
         .collect::<Vec<_>>()
      })
   .collect();
```

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
acl-cqc: Yes, good stuff. Love the connected-components definition, fwiw! :)

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
acl-cqc: Yes, comment is good.

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
acl-cqc: Yeah, it's more defensive, but the `unwrap` looks scarier, fair enough

src/hugr/views/sibling.rs
lmondada: A more grounded objection (and the actual reason this check was added):

This constructor with `(base, nodes)` signature cannot infer the shared parent if `nodes` is empty. What shall we do in that case then? A `SiblingSubgraph` should always have a parent.

src/hugr/views/sibling.rs
acl-cqc: What I meant was:
```
fn get_parent(&self) -> Node {
  self.base.root()
}
```

src/hugr/views/sibling.rs
acl-cqc: There may be shorter ways but that seems good to me!

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,575 @@
...
+    /// [`super::SiblingGraph`].
+    ///
+    /// This will return an [`InvalidSubgraph::EmptySubgraph`] error if the
+    /// subgraph is empty.
+    pub fn from_sibling_graph(sibling_graph: &'g Base) -> Result<Self, InvalidSubgraph>
+    where
+        Base: HugrView,
+    {
+        let root = sibling_graph.root();
+        let nodes = sibling_graph.children(root).collect_vec();
```
acl-cqc, line 68: Ah, I had missed that `from_dataflow_graph` below skipped the input/output nodes. I'm gonna question whether this should really be a public method though, or something that should live inside `mod tests`. You think whole-subgraph is useful?

src/hugr/views/sibling.rs
acl-cqc: Yes, true. I guess we might leave that PR to someone else ;-)

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,584 @@
+//! Views for HUGR sibling subgraphs.
+//!
+//! Views into subgraphs of HUGRs within a single level of the
+//! hierarchy, i.e. within a sibling graph. Such a subgraph is
+//! represented by a parent node, of which all nodes in the sibling subgraph are
```
acl-cqc, line 5: nit: this `represented by` isn't true anymore - "can be created from" if you want, but probably just represented by or *induced* by a subset of nodes

src/hugr/views/sibling.rs
lmondada: Oh very nice!!

src/hugr/views/sibling.rs
acl-cqc: Yes, ok

acl-cqc: Sorry for noise!

src/hugr/views/sibling.rs
lmondada: You're right, will move it to tests.

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,575 @@
...
+        let to_pg = |(n, p): (Node, Port)| pg.port_index(n.index, p.offset).expect("invalid port");
+        let incoming = incoming.into_iter().flat_map(|(n, p)| match p.direction() {
+            Direction::Outgoing => base.linked_ports(n, p).map(to_pg).collect(),
+            Direction::Incoming => vec![to_pg((n, p))],
+        });
+        let outgoing = outgoing.into_iter().flat_map(|(n, p)| match p.direction() {
+            Direction::Incoming => base.linked_ports(n, p).map(to_pg).collect(),
+            Direction::Outgoing => vec![to_pg((n, p))],
+        });
+        let subpg = Subgraph::new_subgraph(pg, incoming.chain(outgoing));
```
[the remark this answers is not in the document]
lmondada: Done :)

src/hugr/views/sibling.rs
lmondada: Oh good spot! I missed this one when I rewrote the code.

src/hugr/views/sibling.rs
acl-cqc: So the concept here seems a bit ill-defined (convexity is a porperty of a subgraph, so first you have to define what nodes/edges are in the subgraph, which seems to rule out incoming+outgoing having edges in common). However, if the code here will detect that using ConvexityChecker, that's good enough for me! :)

lmondada:
@acl-cqc Do you think we are ready to merge? As far as I can see there is only this one question that is still open-ish

https://github.com/CQCL-DEV/hugr/pull/336#discussion_r1300176824

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,575 @@
...
+    /// An iterator of the outgoing boundary ports.
+    fn outgoing_ports(&self) -> impl Iterator<Item = (Node, Port)> + '_
+    where
+        Base: HugrView,
+    {
+        self.boundary_ports(Direction::Outgoing)
+    }
+
+    /// An iterator of the boundary ports, either incoming or outgoing.
+    fn boundary_ports(&self, dir: Direction) -> impl Iterator<Item = (Node, Port)> + '_
```
acl-cqc, line 233: You might want to make some of these public. (If you keep them internal, you might want to just inline incoming/outgoing, but I think making those public is reasonable!)

lmondada: Man, you've got so many good points!

Thanks a lot for your thorough reviewing, an @acl-cqc reviewed code is just so much better than what I started with

src/hugr/views/sibling.rs
acl-cqc: Hmmmm, ok. I could argue that `SiblingGraph` is the thing that selects a group of siblings, and then this should be a subset of that - so you should embed a `SiblingGraph` struct directly inside this. But I can see that that'd be annoying as much as anything :-(.... so ok, yeah. I don't see anything better really

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
acl-cqc: super-nit - `.filter_map` avoids the `is_some`

src/hugr/views/sibling.rs
acl-cqc: grammar nit tho - "if the subgraph if it is" - drop the "if it"

src/hugr/views/sibling.rs
acl-cqc: (but ok, fine as it is)

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,574 @@
...
+    InvalidDataflowGraph,
+    /// Malformed DataflowParent in replacement graph.
+    #[error("Malformed DataflowParent in replacement graph.")]
+    InvalidDataflowParent,
+    /// Replacement graph boundary size mismatch.
+    #[error("Replacement graph boundary size mismatch.")]
+    InvalidSignature,
+    /// SiblingSubgraph is not convex.
+    #[error("SiblingSubgraph is not convex.")]
+    NonConvexSubgrah,
```
acl-cqc, line 383:
```suggestion
    NonConvexSubgraph,
```

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,574 @@
...
+        let pg = base.portgraph();
+        let incoming = incoming.into_iter().flat_map(|(n, p)| match p.direction() {
+            Direction::Outgoing => base.linked_ports(n, p).collect(),
+            Direction::Incoming => vec![(n, p)],
+        });
+        let outgoing = outgoing.into_iter().flat_map(|(n, p)| match p.direction() {
+            Direction::Incoming => base.linked_ports(n, p).collect(),
+            Direction::Outgoing => vec![(n, p)],
+        });
+        let to_pg = |(n, p): (Node, Port)| pg.port_index(n.index, p.offset).expect("invalid port");
```
acl-cqc, line 145: You might add a comment here that this preserves the order in which the boundary edges were provided in order to give the correct signature. (A neat trick ;-) )

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,574 @@
...
+        let subpg = Subgraph::new_subgraph(pg, incoming.chain(outgoing).map(to_pg));
+        if !subpg.is_convex_with_checker(checker) {
```
acl-cqc, line 147: presumably it's more efficient to do the convexity check this way, rather than by `checker.is_node_convex' on `nodes` below ? If the latter is ok, you could replace all these last checks by returning `try_new_with_checker`

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,574 @@
...
+            .children(rep_root)
+            .take(2)
+            .collect_tuple()
+        else { return Err(InvalidReplacement::InvalidDataflowParent) };
+        if dfg_optype.signature() != self.signature() {
+            return Err(InvalidReplacement::InvalidSignature);
+        }
+        let rep_inputs = replacement
+            .node_outputs(rep_input)
+            // filter out any non-dataflow ports
```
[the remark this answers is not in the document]
acl-cqc: at present it sounds like you'd need to include order edges specifically in the maps passed to SimpleReplace, so I'm not sure whether you should do that or maybe raise an issue to handle it some other way ;)....

acl-cqc (commented):
Thanks @lmondada for running the marathon :) and some of my misreadings. (Maybe this wasn't a good first-task-after-a-holiday "gearchange"!)

I'm pretty happy with this now except that I think there is an issue with state edges and I'm not sure what we should do about that here.

acl-cqc:
> @acl-cqc Do you think we are ready to merge? As far as I can see there is only this one question that is still open-ish
> 
> [#336 (comment)](https://github.com/CQCL-DEV/hugr/pull/336#discussion_r1300176824)

Happy to drop that one, sorry for barking up the wrong tree. I've posted 4 more comments above, 3 of which I think are pretty small/straightforward. That leaves the state edges - it's possible that the best thing to do there is to leave that as a TODO, I'm not sure.

acl-cqc:
(i.e. relating to #432)

src/hugr/views/sibling.rs
lmondada: Done :)

src/hugr/views/hierarchy.rs
```diff
@@ -48,7 +48,7 @@ where
     graph: FlatRegionGraph<'g, Base>,
 
     /// The rest of the HUGR.
-    hugr: &'g Base,
+    pub(super) hugr: &'g Base,
```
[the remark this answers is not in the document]
lmondada: indeed!

src/hugr/views/sibling.rs
lmondada: It's not more efficient, however doing the convexity on nodes only will miss some edge cases. Calling `try_new_with_checker` after the convexity check with the nodes would do the convexity check twice, so I think this is the only way.

lmondada:
Not at all! All your remarks and questions were very pertinent @acl-cqc. Thank you so much! This review process was very necessary and more than worth it! If anything I'm sorry it took so much of your time!

For the time being, I suggest leaving the order edges question open and merging this in. I've added a TODO for this.

We will come back once we have a clear answer, or @aborgna-q and me figure out that we need order edges after all 😛

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
lmondada: I do not have a satisfactory answer, other than that for the moment we can get away without order edges... I suggest we come back to this when we have made a spec decisions and will amend simple replace at the same time.

For the time being I've added a TODO.

acl-cqc:
> For the time being, I suggest leaving the order edges question open and merging this in. I've added a TODO for this.

What I'd suggest for the time being is to panic if we find any order edges. I think there are two cases:
1. Edges in/out of the subgraph. I think these would have to be part of the boundary (if you created a SiblingSubgraph from just the value edges in/out, you'd end up getting more of the graph than you intended, and probably fail the convexity/well-formedness check, right). But I think at the moment if such edges were part of the boundary, you'd filter them out, or else look for corresponding edges in the replacement?? So maybe panic rather than filter out.
2. Edges in/out of the replacement (i.e. from Input node to the rest, or to the Output node). You filter these out, but again, perhaps best to panic.

Other than that, I think all good, I'll have a final check through in the morning!

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,562 @@
...
+///
+/// A HUGR region in which all nodes share the same parent. Unlike
+/// [`super::SiblingGraph`],  not all nodes of the sibling graph must be
+/// included. A convex subgraph is always an induced subgraph, i.e. it is defined
+/// by a set of nodes and all edges between them.
+
+/// The incoming boundary (resp. outgoing boundary) is given by the input (resp.
+/// output) ports of the subgraph that are linked to nodes outside of the subgraph.
+/// The signature of the subgraph is then given by the types of the incoming
+/// and outgoing boundary edges. Given a replacement with the same signature,
```
acl-cqc, line 37: nit: `edges` -> `ports`? If I understand correctly and you include a node with a port that has many edges to outside the subgraph, the port gets included once, not once per edge?

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,562 @@
...
+    ///
+    /// The `incoming` and `outgoing` arguments give $B_I$ and $B_O$ respectively.
+    /// They can be either source or target ports. We currently assume that if
+    /// the source port of an outgoing boundary edge is linked to multiple
+    /// target ports, then all edges from the same source port are outgoing
+    /// boundary edges.
+    ///
+    /// More formally, the sibling subgraph defined by $B_I$ and $B_O$ is the
+    /// graph given by the connected components of the graph with edges
+    /// $E\B_I\B_O$ which contain at least one node that is either
```
acl-cqc, line 92: nit: note we haven't defined `$E$` here! I quite liked the "connected components of the graph G'=(V, E \ B_I \ B_O)` too as it's then clearer that "which contain at least one node..." refers to the connected components rather than the edges. (But it's not really ambiguous as edges don't contain nodes so ok.)

src/hugr/views/sibling.rs
acl-cqc:
```suggestion
    /// This function fails if the subgraph is not convex, if the nodes
```

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,562 @@
...
+    {
+        let input = self
+            .incoming_ports()
+            .filter_map(|(n, p)| self.base.get_optype(n).signature().get(p).cloned())
+            .collect_vec();
+        let output = self
+            .outgoing_ports()
+            .filter_map(|(n, p)| self.base.get_optype(n).signature().get(p).cloned())
+            .collect_vec();
+        FunctionType::new(input, output)
```
acl-cqc, line 259: Oh yes, you can use FunctionType - nice :)

src/hugr/views/sibling.rs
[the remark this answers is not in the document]
acl-cqc: I'd probably panic here rather than filter out (then we'll know when the TODO comes into play) but ok

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,562 @@
...
+    NotConvex,
+    /// Not all nodes have the same parent.
+    #[error("Not a sibling subgraph.")]
+    NoSharedParent,
+    /// Empty subgraphs are not supported.
+    #[error("Empty subgraphs are not supported.")]
+    EmptySubgraph,
+}
+
+// /// A common trait for views of a HUGR sibling subgraph.
```
acl-cqc, line 390: Probably just remove this? Unless you want to `impl` it for SiblingGraph and SiblingSubgraph as a no-methods marker trait

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,562 @@
...
+    impl<'g, Base: HugrInternals> SiblingSubgraph<'g, Base> {
+        /// A sibling subgraph from a HUGR.
+        ///
+        /// The subgraph is given by the sibling graph of the root. If you wish to
+        /// create a subgraph from another root, wrap the argument `region` in a
+        /// [`super::SiblingGraph`].
+        ///
+        /// This will return an [`InvalidSubgraph::EmptySubgraph`] error if the
+        /// subgraph is empty.
+        pub fn from_sibling_graph(sibling_graph: &'g Base) -> Result<Self, InvalidSubgraph>
```
acl-cqc, line 420: nit: no need/point in making this `pub`, as `tests` is hidden. (For now!)

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,562 @@
...
+            builder.finish_hugr_with_outputs(inputs).unwrap()
+        };
+
+        let rep = sub.create_simple_replacement(empty_dfg).unwrap();
+
+        assert_eq!(rep.removal.len(), 1);
+
+        hugr.apply_rewrite(rep).unwrap();
+
+        assert_eq!(hugr.node_count(), 4); // Module + Def + In + Out
```
acl-cqc, line 484: Would be good to have a contrasting assert before you apply the rewrite

src/hugr/views/sibling.rs
```diff
@@ -0,0 +1,562 @@
...
+        let pg = base.portgraph();
+        let incoming = incoming.into_iter().flat_map(|(n, p)| match p.direction() {
+            Direction::Outgoing => base.linked_ports(n, p).collect(),
+            Direction::Incoming => vec![(n, p)],
+        });
+        let outgoing = outgoing.into_iter().flat_map(|(n, p)| match p.direction() {
+            Direction::Incoming => base.linked_ports(n, p).collect(),
+            Direction::Outgoing => vec![(n, p)],
+        });
+        let to_pg = |(n, p): (Node, Port)| pg.port_index(n.index, p.offset).expect("invalid port");
```
acl-cqc, line 145: this might be a place to look for state-order edges and panic(?) if you find them. Else you might comment here that we probably need to deal with such here too (?)

acl-cqc (approved):
Looks great, thanks Luca!

PR #337, opened 2023-08-03
[test] Add unit tests for const typechecking

croyzor:
Improve coverage with some low hanging fruit.
Is this overkill? I'm not sure


aborgna-q (approved):
If they improve the coverage they are not overkill

cqc-alec (approved):

PR #338, opened 2023-08-03
[test] Add unit tests for ConditionalBuilder failure modes

croyzor:


src/builder/conditional.rs
```diff
@@ -266,4 +266,28 @@ mod test {
...
+    fn test_not_all_cases() -> Result<(), BuildError> {
+        let predicate_inputs = vec![type_row![]; 2];
+        let mut builder = ConditionalBuilder::new(predicate_inputs, type_row![], type_row![])?;
+        n_identity(builder.case_builder(0)?)?;
+        match builder.finish_sub_container() {
+            Err(BuildError::ConditionalError(ConditionalBuildError::NotAllCasesBuilt {
+                ..
+            })) => Ok(()),
+            _ => panic!(),
+        }
```
ss2165, line 280: you can use `assert_matches!` from `cool_asserts`

ss2165 (commented):
looks good, can just avoid some boiler plate

src/builder/conditional.rs
croyzor: I tried, but `assert_matches` requires that the `BuildHandle` implements Debug.
I could just call `map(|_| ())` on the success case though...

ss2165 (approved):
👍

PR #340, opened 2023-08-03
A few more Eq fixes

acl-cqc:
* HashableValue's are Eq, that's the whole point of them ;-)
* ContainerValue's can also be Eq if they contain things that are
* CustomSerialized probably should not be
* But, provide missing impl of `equal_const`

### DIFF
--- a/src/ops/constant.rs
+++ b/src/ops/constant.rs
@@ -450,5 +450,7 @@ mod test {
         let t: SimpleType = typ_float.clone().into();
         assert_matches!(val.check_type(&t.try_into().unwrap()),
             Err(ConstTypeError::CustomCheckFail(CustomCheckFail::TypeMismatch(a, b))) => a == typ_int && b == typ_float);
+
+        assert_eq!(val, val);
     }
 }
--- a/src/ops/constant.rs
+++ b/src/ops/constant.rs
@@ -306,7 +306,9 @@ pub trait CustomConst:
 impl_downcast!(CustomConst);
 impl_box_clone!(CustomConst, CustomConstBoxClone);
 
-#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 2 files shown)


src/ops/constant.rs
```diff
@@ -444,5 +450,7 @@ mod test {
         let t: SimpleType = typ_float.clone().into();
         assert_matches!(val.check_type(&t.try_into().unwrap()),
             Err(ConstTypeError::CustomCheckFail(CustomCheckFail::TypeMismatch(a, b))) => a == typ_int && b == typ_float);
+
+        assert_eq!(val, val);
```
acl-cqc, line 454: Yes, this was failing before, because equality of `Box<dyn CustomConst>` does not fall back to `PartialEq`....

src/values.rs
```diff
@@ -16,7 +16,7 @@ use crate::{
 
 /// A constant value/instance of a [HashableType]. Note there is no
 /// equivalent of [HashableType::Variable]; we can't have instances of that.
-#[derive(Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
```
ss2165, line 19: should it derive `Hash` as well?

src/values.rs
```diff
@@ -108,6 +108,8 @@ pub enum ContainerValue<T> {
     Sum(usize, Box<T>), // Tag and value
 }
 
+impl<Elem: Eq> Eq for ContainerValue<Elem> {}
```
ss2165, line 111: 👍

acl-cqc: In fact we can just `#[derive(Eq)]` - the `derive` macro is smart enough to condition the generated `impl` on the appropriate condition of the arguments, which is pretty neat.

PR #342, opened 2023-08-03
[bugfix] Check #TypeArgs == #TypeParams for TypeDefs as well as OpDefs

acl-cqc:
The code was in there, but seemingly in the wrong place. Not sure if my comment on `TypeArgError::WrongNumber` had mislead so I've corrected that ;)

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -194,6 +194,12 @@ trait TypeParametrised {
     fn resource(&self) -> Option<&ResourceId>;
     /// Check provided type arguments are valid against parameters.
     fn check_args_impl(&self, args: &[TypeArg]) -> Result<(), SignatureError> {
+        if args.len() != self.params().len() {
+            return Err(SignatureError::TypeArgMismatch(TypeArgError::WrongNumber(
+                args.len(),
+                self.params().len(),
+            )));
+        }
         for (a, p) in args.iter().zip(self.params().iter()) {
             check_type_arg(a, p).map_err(SignatureError::TypeArgMismatch)?;
         }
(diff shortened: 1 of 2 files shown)


src/resource.rs
```diff
@@ -688,3 +688,48 @@ impl FromIterator<ResourceId> for ResourceSet {
...
+            ))])
+            .unwrap()
+            .into();
+        assert_eq!(typ2.tag(), TypeTag::Hashable);
+
+        // And some bad arguments...firstly, wrong kind of TypeArg:
+        def.instantiate_concrete([TypeArg::HashableType(HashableType::String)])
+            .unwrap_err();
+        // Too few arguments:
+        def.instantiate_concrete([]).unwrap_err();
```
acl-cqc, line 727: This one was previously panicking with an index-out-of-bounds in computing the tag

src/resource.rs
```diff
@@ -688,3 +688,48 @@ impl FromIterator<ResourceId> for ResourceSet {
...
+        // Too many arguments:
+        def.instantiate_concrete([
+            TypeArg::ClassicType(ClassicType::F64),
+            TypeArg::ClassicType(ClassicType::F64),
+        ])
+        .unwrap_err();
```
ss2165, line 733: can we assert_matches on some known errors?

src/resource.rs
acl-cqc: (If I'd provided a TypeDefTag::Explicit it would have just returned a CustomType)

src/resource.rs
acl-cqc: We can. They'll change in an upcoming PR but yeah, ok

ss2165: ah, in which case some TODOs should be enough

acl-cqc: They'll be simple updates, and it'll make the change obvious and explicit, so done.

ss2165 (approved):
👍  thanks

PR #346, opened 2023-08-03
Rename `type_def` to `typedef` for consistency with `opdef`; docstring fixes

cqc-alec:
(Wondered if `typedef` might be a reserved token but it doesn't seem to be.)


ss2165 (approved):
would it not be better to go `opdef` -> `op_def` instead?

PR #347, opened 2023-08-03
Rename `Region` views

lmondada:
I propose the following renaming
 - `FlatRegionView` -> `SiblingGraph`
 - `RegionView` -> `DescendantsGraph`
 - `Region` -> `HierarchyView`

I've reserved the term `View` for the trait rather than the `struct`s as that is closer to our use of the word elsewhere. Happy to discuss other opinions.

In addition to being more descriptive, this will also allow us to make the complementarity with #336 more obvious.


lmondada:
Credits to @acl-cqc for the suggestions!

src/hugr/hierarchy.rs
```diff
@@ -1,14 +1,16 @@
-//! Region-level views of a HUGR.
```
ss2165, line 1: maybe the file/module should be called `views.rs`?

ss2165 (commented):
I like the names, and the reserving of `View` for traits. Just a suggestion on the module name.

src/hugr/hierarchy.rs
lmondada: I would like that, but there is already the `view.rs` file in the same folder that contains `HugrView` etc.

I could make a `views` folder with
- `views/hugr.rs`
- `views/hierarchy.rs`
in it?

ss2165: or just go with `hierarchical_views`

ss2165 (approved):
👍 thanks

PR #349, opened 2023-08-03
Alternative #335 with simpler Type(Param/Arg) - no containers

ss2165:
How much TypeParam expressivity do we need?

### DIFF
--- a/src/extensions/logic.rs
+++ b/src/extensions/logic.rs
@@ -6,12 +6,13 @@ use itertools::Itertools;
 use smol_str::SmolStr;
 
 use crate::{
+    ops::constant::HugrIntValueStore,
     resource::ResourceSet,
     types::{
-        type_param::{TypeArg, TypeParam},
+        type_param::{ArgValue, TypeArg, TypeParam},
         HashableType, SimpleType,
     },
-    values::{ConstTypeError, HashableValue},
+    values::HashableValue,
     Resource,
 };
 
@@ -56,12 +57,7 @@ pub fn resource() -> Resource {
             Vec::new(),
             |arg_values: &[TypeArg]| {
                 let a = arg_values.iter().exactly_one().unwrap();
-                let n: u128 = match a {
-                    TypeArg::Value(HashableValue::Int(n)) => *n,
-                    _ => {
-                        return Err(ConstTypeError::TypeArgCheckFail(H_INT, a.clone()).into());
-                    }
-                };
+                let n = get_n(a);
                 Ok((
                     vec![bool_type(); n as usize].into(),
                     vec![bool_type()].into(),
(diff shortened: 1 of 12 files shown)


acl-cqc:
Ok, so this is a good question, and this post might be best turned into a discussion. (Note proposal for how to proceed at the end though!).

There are things I like about this (e.g. that `TypeParam::Type(TypeTag)`) as well as things that I think are just wrong - e.g. `ValueList`: what type of values in that list?

So, first example. consider the type(def) for tensors. It takes a *type* of elements, and a list of integer dimensions. E.g. `Tensor<Float,[3]>` is a xyz co-ordinate, whereas `Tensor<Int,[5,7,9]>` is a 3-dimensional array with a total of 315 integers. If we just say that that's a list of values then Tensor<Int,["foo", "bar"]> is a valid type. (Note we need string/opaque arguments for parameterizing ops, as is being done for TKET1, and I don't think we want 3 systems - we should have the same for parameterizing ops and types). Moreover *there is no check on the validity of TypeArgs other than that they fit the TypeParams*; you can't provide custom code (and you shouldn't: methinks it's important for the TypeDef, specifying that validity check with its TypeParams, to be YAML-serializable so it can always be passed about).

However, this can be fixed easily enough by giving ValueList an element type, for which the obvious thing is HashableType. So we could have `enum TypeParam { Type(TypeTag), Value(HashableType), ValueList(HashableType), TypeList(TypeTag) }`. Note that already we have a confusion for matrix - is that a Value of the-hashable-type-"List-of-int", or a ValueList of the-hashable-type-"int" ?

I note also you are in the process of removing Map from Container (hurrah, not having maps as type args seems a definite improvement :-) ), but also List, which earns a :-( . That "List requires dynamic allocation" seems a good reason to remove List from *core* runtime values, but a poor reason to remove from *static* constants (because their static, so no runtime allocation required). So this does indeed suggest *we might want a different list of container types for TypeParam vs. HashableValue*.

So, next example. We mentioned Tensors above. Now consider the type(def) of tensors with *headers* on each row/column/etc. For example, a 2d table of population by country by year would have string headers (country name) on one dimension, and int headers (year) in the other. That could be a `Table<Int,[(String, #countries), (Int, #years)]>`. (Maybe "Int" elements are too narrow - you might want a range with uncertainty estimates, or an Option with None=no data.) So I think there's a reasonable argument for having Tuple in there too. Which leads me to something like `enum TypeParam {Type(TypeTag), Value(HashableType), List(TypeParam), Tuple(TypeRow<TypeParam>) }`. So we've dropped the runtime/Container `Array` (only ever a synonym for tuple, supporting a runtime operation `get(int)` - well runtime operations aren't relevant here). And we've added List, as per above. I've also dropped `Sum`, which I haven't got a good use for at the moment, but maybe someone will come up with one. Note that after merging the Types using TypeTag, that's more variants of TypeParam (and less code reuse) than Container...

Note we still have the `TypeParam::Value(HashableType::Container(Container::Tuple(hashtyp1, hashtyp2)))` vs `TypeParam::Tuple(TypeParam::Value(hashtyp1), TypeParam::Value(hashtyp2))` question, that has plagued the Simple/Classic/Hashable hierarchy. I am pondering whether our ability to solve that dilemma by the `Container`->`Type` refactor is relevant here; it might mean that a lot of the extra boilerplate (and repetition of SimpleType stuff) here just goes away, and depending on how well it works that might make reuse of Container too compelling to resist.

That said, I am starting to find the arguments about List and Array persuasive.

(You might consider also `TypeParam::Value(HashableType::Container(Container::Opaque("list", elemtype)))` versus `TypeParam::List(TypeParam::Value(elemtype))`, as given we have opaque typeparams/args for e.g. TKET1 ops, we can't stop people using the extension list, but at least these are two very different types.)

Perhaps the best way forward is to apply the `Container`->`Type` refactor to the (Simple/Classic/Hashable)Type hierarchy first, and then come back to this?

ss2165:
Closing to revisit types more holistically

PR #352, opened 2023-08-04
[doc] Expand '"' in hierarchy table

croyzor:
My granpaw used to say: "If it's worth saying once, it's worth saying again"


ss2165 (approved):
a wise man

PR #353, opened 2023-08-04
[WIP] Start on making "Single" be an element of Container, SimpleType=Container<SimpleElem>

acl-cqc:
* Think I've got the biggest five files compiling (simple, constant, values, type_param, serialize - i.e. the implementation). The rest should mostly just be clients, which might be best solved by adding utility methods like `SimpleType::int` (=> `Container::Single(SimpleElem::Classic(ClassicElem::Hashable(HashableElem::Int(w))))`)
* Not at all sure about deserialization, nor about the TAG's written out in serialization - see comments
* Need to standardize naming, I'm using "Elem" for things implementing PrimType (they should probably be "Prim"s) and "Leaf" for corresponding Values (they might want to be "Elem"!)
* Not sharing much code between type_param and simple, but type_param is currently using HashableType, it should probably only use HashableLeaf (which will be, what, "Int" ?)
* Parameterized ConstTypeError and TypeArgError as aliases of combined ValueError, this works well
* `map_vals` etc. preserved but used only for mapping error types (!). I've generally tried to make these as `map_into` and I'd like to just have `impl<T, T2: Into<T>> From<Container<T2>> for Container<T>` but every type is `Into` of itself (a blanket impl in the standard library), so that allows instantiation with `T` == `T2` which is therefore the identity `From` which conflicts with said blanket impl! I don't know how to turn this off, maybe you / @aborgna-q can see a better way
* I took a slightly different approach to `TypeRowElem` vs `PrimType` - rather than a blanket impl of `TypeRowElem` for anything `Clone + 'static` I've explicitly implemented it only for things I want to put in a TypeRow, i.e. `SerSimpleType` and `Container<T>` (where `T: PrimType` - but `PrimType` no longer requires `TypeRowElem`). That is, `PrimType` is sealed and just `ClassicElem`, `SimpleElem` and `HashableElem`; whereas `TypeRowElem` *could* be implemented by everything but I'm using it as a marker for the things that I actually want to make rows of. (This spotted a few erroneous `TypeRow<ClassicElem>` etc. that I had missed.)

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -6,9 +6,9 @@ use thiserror::Error;
 use pyo3::prelude::*;
 
 use crate::hugr::{HugrError, Node, ValidationError, Wire};
+use crate::ops::constant::ConstTypeError;
 use crate::ops::handle::{BasicBlockID, CfgID, ConditionalID, DfgID, FuncID, TailLoopID};
 use crate::types::SimpleType;
-use crate::values::ConstTypeError;
 
 pub mod handle;
 pub use handle::BuildHandle;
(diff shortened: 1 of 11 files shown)


acl-cqc:
Closing as superceded by #378, and who knows how many further generations of reductionism ;-)

PR #355, opened 2023-08-07
Add impl HugrView for SubHugr

doug-q:
No tests as of yet

### DIFF
--- a/src/hugr/subhugr.rs
+++ b/src/hugr/subhugr.rs
@@ -4,16 +4,19 @@
 //! more than one root node. However, all root nodes must belong to the same
 //! sibling graph.
 //!
-//! TODO: Subgraphs implement [`HugrView`] as well as petgraph's _visit_ traits.
+//! Subgraphs implement [`HugrView`].
+//! TODO:  as well as petgraph's _visit_ traits.
 
 use std::cell::OnceCell;
 
+use context_iterators::{ContextIterator, IntoContextIterator, MapWithCtx};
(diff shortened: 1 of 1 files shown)


doug-q:
@zrho Can you review?

lmondada:
Oh this is awsome!!! Thank you so much @doug-q !! Will have a look this afternoon

lmondada (commented):
Hi @doug-q, thanks a lot for the work, it's great!!

We've got an issue whereby the datastructure that we would need from `portgraph`, i.e. something that traverses both the sibling subgraph + descendants nodes within does not exist.

Your code makes me think that probably the best solution for the moment would be to change the semantics of `SubHugr` to only cover the sibling graph itself and not its descendants. In that case this code would work.

This is good enough for my current needs.

I see two options to proceed:
- You merge this in as is right now, and I take care of changing `SubHugr` in #336
- You keep these changes and apply them as a separate PR once I've merged in #336.

Choose the one that you prefer, both work for me. I personally think option 2 will give you more credit for your work, but it will be slightly more work for you to resolve the merge conflicts -- the work should be minimal though.

Thanks a lot again for everything!

doug-q:
Thanks for the review. If you are happy changing semantics of SubHugr, go ahead. I'm ignorant as to whether that's a good idea. I'll post another PR once your work is merged

lmondada:
Ok, perfect. Should be merged in later today hopefully!

zrho (approved):

PR #364, opened 2023-08-07
separate TypeParam hierarchy

### LINKED ISSUE #359: Separate typeparam and simpletype hierarchies

ss2165:
Closes #359

### DIFF
--- a/src/types/type_param.rs
+++ b/src/types/type_param.rs
@@ -7,7 +7,7 @@
 use thiserror::Error;
 
 use super::CustomType;
-use super::{ClassicType, PrimType, SimpleType, TypeTag};
+use super::{PrimType, SimpleType, TypeTag};
 
 /// A parameter declared by an OpDef. Specifies a value
 /// that must be provided by each operation node.
@@ -76,12 +76,7 @@ impl TypeArg {
 /// Checks a [TypeArg] is as expected for a [TypeParam]
 pub fn check_type_arg(arg: &TypeArg, param: &TypeParam) -> Result<(), TypeArgError> {
     match (arg, param) {
-        (
-            TypeArg::Type(SimpleType::Classic(ClassicType::Hashable(_))),
-            TypeParam::Type(TypeTag::Hashable),
-        ) => Ok(()),
-        (TypeArg::Type(SimpleType::Classic(_)), TypeParam::Type(TypeTag::Classic)) => Ok(()),
-        (TypeArg::Type(_), TypeParam::Type(TypeTag::Simple)) => Ok(()),
+        (TypeArg::Type(t), TypeParam::Type(tag)) if tag.contains(t.tag()) => Ok(()),
         (TypeArg::Sequence(items), TypeParam::List(param)) => {
             items.iter().try_for_each(|arg| check_type_arg(arg, param))
         }
--- a/src/types/simple.rs
+++ b/src/types/simple.rs
@@ -58,12 +58,11 @@ impl TypeTag {
     /// Returns the smallest TypeTag containing both the receiver and argument.
     /// (This will be one of the receiver or the argument.)
     pub fn union(self, other: Self) -> Self {
-        if self == Self::Simple || other == Self::Simple {
-            Self::Simple
-        } else if self == Self::Classic || other == Self::Classic {
-            Self::Classic
+        if self.contains(other) {
+            self
         } else {
-            Self::Hashable
+            debug_assert!(other.contains(self));
+            other
         }
     }
 
@@ -77,6 +76,15 @@ impl TypeTag {
     /// (with a strong notion of equality, i.e. [HashableType]s)
     pub fn is_hashable(self) -> bool {
         self == Self::Hashable
+    }
+
+    /// Report if this tag contains another.
+    pub fn contains(&self, other: TypeTag) -> bool {
+        use TypeTag::*;
+        matches!(
+            (self, other),
+            (Simple, _) | (_, Hashable) | (Classic, Classic)
+        )
     }
 }
 


src/ops/constant.rs
```diff
@@ -450,15 +446,15 @@ mod test {
         val.check_type(&ClassicType::Container(Container::Opaque(typ_int.clone())))
             .unwrap();
 
-        let typ_float = CustomType::new(
+        let typ_qb = CustomType::new(
             "mytype",
-            vec![TypeArg::ClassicType(ClassicType::F64)],
+            vec![TypeArg::Type(SimpleType::Qubit)],
```
acl-cqc, line 451: There is something suspicious here in that the tag passed below is `TypeTag::Hashable`....

src/resource.rs
```diff
@@ -22,7 +22,7 @@ pub use type_def::{TypeDef, TypeDefTag};
 
 /// An error that can occur in computing the signature of a node.
 /// TODO: decide on failure modes
-#[derive(Debug, Clone, Error, PartialEq, Eq)]
+#[derive(Debug, Clone, Error, PartialEq)]
```
acl-cqc, line 25: +1

src/types/type_param.rs
```diff
@@ -6,31 +6,26 @@
...
-    Type,
-    /// Argument is a [TypeArg::ClassicType] - hashable or otherwise
-    ClassicType,
-    /// Argument is a [TypeArg::HashableType]
-    HashableType,
-    /// Node must provide a [TypeArg::List] (of whatever length)
-    /// TODO it'd be better to use [`Container`] here.
-    ///
-    /// [`Container`]: crate::types::simple::Container
+    /// A type as parameter, with given type tag.
```
acl-cqc, line 19: "with" -> "within"? (A `Type(Tag::Classic)` is happy to be given a type whose tag is Hashable)

src/types/type_param.rs
```diff
@@ -6,31 +6,26 @@
...
+    Type(TypeTag),
+    /// 64-bit unsigned integer
```
acl-cqc, line 21: I liked the cross-linking with relevant elements of `TypeArg`...

src/types/type_param.rs
```diff
@@ -39,29 +34,21 @@ pub enum TypeParam {
...
-    /// be of the same variety of TypeArg, i.e. `T`s.
+    /// 64-bit unsigned integer
+    Usize(u64),
+    /// An opaque value, stored as serialized blob.
+    Opaque(CustomSerialized),
+    /// List of arguments
     List(Vec<TypeArg>),
-    /// Where the TypeDef declares a [TypeParam::Value] of [Container::Opaque]
-    CustomValue(serde_yaml::Value),
+    /// Tuple of arguments
```
acl-cqc, line 43: Note that given we are only checkable not synthesizing here, we can combine TypeArg::List and TypeArg::Tuple into a single TypeArg (e.g. Sequence). We do this for constant values and I think there's an argument for this.

src/types/type_param.rs
```diff
@@ -70,49 +57,38 @@ impl TypeArg {
...
+        ) => Ok(()),
+        (TypeArg::Type(SimpleType::Classic(_)), TypeParam::Type(TypeTag::Classic)) => Ok(()),
+        (TypeArg::Type(_), TypeParam::Type(TypeTag::Simple)) => Ok(()),
+        (TypeArg::List(items), TypeParam::List(param)) => {
+            items.iter().try_for_each(|arg| check_type_arg(arg, param))
+        }
+        (TypeArg::Tuple(items), TypeParam::Tuple(types)) => {
+            let items_len = items.len();
+            let types_len = types.len();
+            if items.len() != types.len() {
```
acl-cqc, line 72: Given you've just pulled out `items_len` and `types_len`, use them here!

fwiw I don't tend to worry much about repeated constants in error-handling code, it's exceptional so I prefer to do whatever's best for the non-exceptional case

src/ops/constant.rs
ss2165: any reason the custom type can't be hashable with a qubit as type argument?

src/resource.rs
ss2165: is this a thumbs up?

src/types/type_param.rs
```diff
@@ -70,49 +57,38 @@ impl TypeArg {
...
-            }
-            Container::Alias(n) => Err(TypeArgError::NoAliases(n.to_string())),
-        },
         _ => Err(TypeArgError::TypeMismatch(arg.clone(), param.clone())),
     }
 }
 
 /// Errors that can occur fitting a [TypeArg] into a [TypeParam]
-#[derive(Clone, Debug, PartialEq, Eq, Error)]
+#[derive(Clone, Debug, PartialEq, Error)]
```
acl-cqc, line 91: Hang on...no floats here. This is because of, what, CustomCheckFail ?

src/types/type_param.rs
ss2165: have reworded for clarity

src/types/type_param.rs
```diff
@@ -6,31 +6,26 @@
...
-    /// Argument is a [TypeArg::HashableType]
-    HashableType,
-    /// Node must provide a [TypeArg::List] (of whatever length)
-    /// TODO it'd be better to use [`Container`] here.
-    ///
-    /// [`Container`]: crate::types::simple::Container
+    /// A type as parameter, with given type tag.
+    Type(TypeTag),
+    /// 64-bit unsigned integer
+    Usize,
```
acl-cqc, line 22: Ok, this'll be the killer....`Usize` or `USize`? I feel that "usize" is not one word; it's an abbreviation for "unsigned size" and as such should be "USize". Or "Size"...

src/types/type_param.rs
ss2165: yes, we don't seem to need to it so I removed.

acl-cqc (approved):
Good stuff. I hadn't realized `check_int_fits_in_width` would become no-longer shared, but I guess that's even better. Do you want to recombine `typecheck.rs` with `values.rs` or leave that for another PR?

src/resource.rs
acl-cqc: Yes, tho originally I was misremembering that it had to contain arbitrary values - but that was only in a PR where I was including a `ConstTypeError` that could be from checking an arbitrary constant. Actually I'm not sure that this shouldn't be `Eq` after all, it goes back to CustomCheckFail.

ss2165: have reverted

src/types/type_param.rs
ss2165: added back

src/ops/constant.rs
```diff
@@ -308,8 +307,9 @@ impl_box_clone!(CustomConst, CustomConstBoxClone);
 
 // Don't derive Eq here - the yaml could contain floats etc.
 // (Perhaps we could derive Eq if-and-only-if "typ.tag() == TypeTag::Hashable"!)
-#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
-struct CustomSerialized {
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
```
acl-cqc, line 310: Whoa - do we need *this* `Eq`? See comment just above

src/types/type_param.rs
```diff
@@ -39,29 +34,19 @@ pub enum TypeParam {
...
-    /// Where the (Type/Op)Def declares a [TypeParam::List]`<T>` - all elements will implicitly
-    /// be of the same variety of TypeArg, i.e. `T`s.
-    List(Vec<TypeArg>),
-    /// Where the TypeDef declares a [TypeParam::Value] of [Container::Opaque]
-    CustomValue(serde_yaml::Value),
+    /// Instance of [TypeParam::USize]. 64-bit unsigned integer.
+    USize(u64),
+    ///Instance of [TypeParam::Opaque] An opaque value, stored as serialized blob.
+    Opaque(CustomSerialized),
+    /// Instance of [TypeParam::List] or [TypeParam::Tuple] List of arguments
```
acl-cqc, line 41: nit: not quite grammatically correct here

src/types/type_param.rs
```diff
@@ -70,43 +55,30 @@ impl TypeArg {
...
 pub fn check_type_arg(arg: &TypeArg, param: &TypeParam) -> Result<(), TypeArgError> {
     match (arg, param) {
-        (TypeArg::Type(_), TypeParam::Type) => Ok(()),
-        (TypeArg::ClassicType(_), TypeParam::ClassicType) => Ok(()),
-        (TypeArg::HashableType(_), TypeParam::HashableType) => Ok(()),
-        (TypeArg::List(items), TypeParam::List(ty)) => {
-            for item in items {
-                check_type_arg(item, ty.as_ref())?;
+        (
+            TypeArg::Type(SimpleType::Classic(ClassicType::Hashable(_))),
```
acl-cqc, line 59: Ah, just realized - I think you can do
```
(TypeArg::Type(t), TypeParam::Type(tag)) if tag.contains(t.tag()) => Ok(())
```
(You'll need `TypeTag::contains` but we should have that)

src/ops/constant.rs
ss2165: CustomSerialized is now in TypeArg -> TypeArg is in TypeArgError -> TypeArgError can't be Eq  , which is probably why I originally removed it

src/types/type_param.rs
```diff
@@ -123,11 +95,12 @@ pub enum TypeArgError {
...
     // However in the future it may be applicable to e.g. contents of Tuples too.
     #[error("Wrong number of type arguments: {0} vs expected {1} declared type parameters")]
-    WrongNumber(usize, usize),
-    /// The type declared for a TypeParam was an alias that was not resolved to an actual type
-    #[error("TypeParam required an unidentified alias type {0}")]
-    NoAliases(String),
-    /// There was some problem fitting a const int into its declared size
-    #[error("Error with int constant")]
-    Int(#[from] ConstIntError),
+    WrongNumberArgs(usize, usize),
```
acl-cqc, line 98: Following from previous suggestion - I actually thought you could move this out of TypeArgError into SignatureError. But I'm also quite happy not to, it might be nicer to keep it here

Issue #363, opened 2023-08-07
Types Redux: tracking issue


ss2165:
Tracking issue for various tasks aimed at refactoring and simplifying the type system.
Some require refining and issues made

- [x] #361 
- [x] #360 
- [x] #359
- [x] #370
- [x] #366 
- [x] #275
- [x] Include in prelude: float, qubit, array (for existing tests)
- [x] #409
- [x] #408
- [x] Decide whether to move Graph out of core (update: graph stays)
- [x] #371

ss2165:
Closing as tracked issues are done.

PR #369, opened 2023-08-07
[RFC] Allow linear Constants

acl-cqc:
There are two things I dislike about this.

Firstly, that the maze of `SimpleType::Qontainer` of `SimpleType::Classic` of `ClassicType::Container` of .... just got worse. That is what much of #363 is aiming to fix, so we might want to wait for that.

Secondly, I've changed `static_input` from ClassicType to SimpleType too (as required), but wtf happens with EdgeKind::Static now. Do we enforce linearity of linear-typed *static* edges? See discussion #368


acl-cqc:
`load_constant` is not `produce_value` ;)

PR #375, opened 2023-08-08
[new] Validate IO nodes have resources matching parent signatures

croyzor:
Add a validation check that dataflow graph's input and output nodes have the same resource requirements as the incoming/outgoing wires of the parent.

### DIFF
--- a/src/hugr/validate.rs
+++ b/src/hugr/validate.rs
@@ -70,6 +70,48 @@ impl<'a> ValidationContext<'a> {
             self.validate_node(node)?;
         }
 
+        // Check that io nodes match their parents' signatures
+        self.validate_io_resources(self.hugr, self.hugr.root())?;
+
+        Ok(())
+    }
+
+    /// Check that input and output nodes match up with the signature of their parents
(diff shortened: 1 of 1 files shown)


croyzor:
I'm open to refactoring suggestions, because the code seems quite ugly

src/hugr/validate.rs
```diff
@@ -70,6 +70,48 @@ impl<'a> ValidationContext<'a> {
             self.validate_node(node)?;
         }
 
+        // Check that io nodes match their parents' signatures
+        self.validate_io_resources(self.hugr, self.hugr.root())?;
```
aborgna-q, line 74: This should rather go in `ops::validate::validate_io_nodes`.

We check the I/O signatures there, it _should_ be straightforward to also check the resources.

aborgna-q (approved):
This should do for now.

If there was an specialized version of `get_io` (s/get_io/get_io_nodes/ and add `get_io(parent: NodeIndex, dir: Direction)`?) then we could remove the repeated code 🤔

PR #376, opened 2023-08-08
Use a 64-bit `USize` as the sole integer type in the core

### LINKED ISSUE #356: Use size type for integers in core

cqc-alec:
Closes #356 .

### DIFF
--- a/src/values.rs
+++ b/src/values.rs
@@ -5,12 +5,8 @@
 
 use thiserror::Error;
 
-use crate::ops::constant::{HugrIntWidthStore, HUGR_MAX_INT_WIDTH};
 use crate::types::{ClassicType, Container, CustomType, HashableType, PrimType};
-use crate::{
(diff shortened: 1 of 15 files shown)


src/values.rs
```diff
@@ -180,55 +176,6 @@ pub(crate) fn map_container_type<T: PrimType, T2: PrimType>(
...
-    if width > HUGR_MAX_INT_WIDTH {
-        return Err(ConstIntError::IntWidthTooLarge(width));
-    }
-
-    if VALID_WIDTHS.contains(&width) {
-        let max_value = if width == HUGR_MAX_INT_WIDTH {
-            HugrIntValueStore::MAX
-        } else {
-            HugrIntValueStore::pow(2, width as u32) - 1
-        };
```
ss2165, line 221: should this code go in the int type resource or is some version already there?

cqc-alec: I'll add a test in `int_types.rs`.

Issue #366, opened 2023-08-07
Remove `ClassicType::F64`


cqc-alec:
This is provided in the `float.types` resource.
It's a bit difficult to do this yet because some of the quantum operations depend on it, so blocked by #365 .

PR #377, opened 2023-08-08
Remove f64

acl-cqc:
Spec was updated already in #358. I took out `LeafOp::Rzf64` too (the only LeafOp that actually used it), also `SerSimpleType::F`. Added a couple of other ClassicType instances for use in tests because we need some (specifically, `ClassicType::Graph`s are a pain to construct, cannot be `const` as they contain Boxes, and do not yet have ConstValue instances)


aborgna-q (approved):
LGTM

ss2165:
Closes #366

PR #378, opened 2023-08-08
Simplified HUGR Type and TypeBound system

### LINKED ISSUE #408: Refactor type hierarchy in to "Type tree"

### LINKED ISSUE #409: HashableType -> EqType

ss2165:
See tracking issue #363 
Closes #408 
Closes #409 

See also individual commits  from sub-PRs


src/types/basic.rs
```diff
@@ -0,0 +1,160 @@
...
+
+mod sealed {
+    use super::{Any, Classic, Eq};
+    pub trait Sealed {}
+    impl Sealed for Any {}
+    impl Sealed for Classic {}
+    impl Sealed for Eq {}
+}
+pub trait TypeClass: sealed::Sealed {
+    const TAG: TypeTag;
```
acl-cqc, line 28: I think I might call this STATIC_TAG just to drive that home. (It's what Rust'll guarantee for us, but the actual runtime tag could be narrower.)

We might actually define `fn tag(&self)` (or `dynamic_tag`) here as well, although I suspect this isn't quite the right place for that

src/types/basic.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub const fn tag(&self) -> TypeTag {
+        T::TAG
+    }
+
+    pub fn new_tuple(types: impl IntoIterator<Item = Type<T>>) -> Self {
+        Self::Tuple(types.into_iter().collect())
+    }
+
+    pub fn new_opaque(opaque: CustomType) -> Self {
+        assert!(opaque.tag() == T::TAG, "tag ");
```
acl-cqc, line 71: `T::TAG.contains(opaque.tag())`, I think?

src/types/basic.rs
```diff
@@ -0,0 +1,160 @@
...
+impl<T: TypeClass> Type<T> {
+    pub const fn tag(&self) -> TypeTag {
+        T::TAG
+    }
+
+    pub fn new_tuple(types: impl IntoIterator<Item = Type<T>>) -> Self {
+        Self::Tuple(types.into_iter().collect())
+    }
+
+    pub fn new_opaque(opaque: CustomType) -> Self {
```
acl-cqc, line 70: So we'd need an additional level of wrapping to prevent people getting around this (as `Type` is a public enum), right?

src/types/basic.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+impl NewClassic for Type<Classic> {
+    fn graph(signature: AbstractSignature) -> Self {
+        Self::Prim(Classic(ClassicTypeImpl::Graph(Box::new(signature))))
+    }
+}
+
+impl NewEq for Type<Any> {
+    const USIZE: Self = Self::Prim(Any(AnyTypeImpl::C(ClassicTypeImpl::E(EqTypeImpl::USize))));
```
acl-cqc, line 103: Ok, ouch. So we have `Any` separate from `AnyTypeImpl` because....we want to hide AnyTypeImpl as it's an `enum`?

src/types/basic.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub fn new_opaque(opaque: CustomType) -> Self {
+        assert!(opaque.tag() == T::TAG, "tag ");
+        Self::Extension(opaque)
+    }
+}
+
+/*
+Public traits for construction
+*/
+pub trait NewEq {
```
acl-cqc, line 79: Unless you're somewhere gonna return a `dyn NewEq`, can't we just say `pub type AnyType = Type<Any>` and put this `const USIZE` in `impl Type<Any>` ?

src/types/basic.rs
acl-cqc: (And similarly we'd want a new_alias that makes the same check on the provided `TaggedWrapper<String>`)

acl-cqc (commented):
Interesting. I note you haven't gone for the Leaf<T> approach (I admit I am unsure about this). I think we probably need a struct (constructible only via `new`) to prevent bypassing of `new_opaque`, and I'm not sure we can't drop one or two other structs - it seems like it may be a case of "too many traits" ?

acl-cqc:
Ok, so I think the main question could be - supposing I want to find all the aliases (to check they are defined by resources that are loaded), or to verify all the opaque types (similarly), something like that - a function that wants to inspect the *leaves* of the tree. (IOW: property holds of a array/sum/tuple iff it holds for its components). I could write a recursive function that `match`es on the `Type`, but a *nice* way would to be have a function something like `iter_leaves(&Type) -> Iterator<Item=Alias|Opaque|Prim>`. However I think as this stands there is no Rust type that I can write for `Alias|Opaque|Prim`. Do we want one or how else can we handle this situation? Or are we willing to write recursive matches? (I reckon there'll be maybe 3 or 4)

src/values.rs
```diff
@@ -3,213 +3,198 @@
...
+    }
+
+    /// Constant Sum over units, used as predicates.
+    pub fn simple_predicate(tag: usize) -> Self {
+        Self::sum(tag, Self::unit())
+    }
+
+    /// Constant Sum over Tuples with just one variant of unit type
+    pub fn simple_unary_predicate() -> Self {
+        Self::simple_predicate(0)
```
cqc-alec, line 78:
```suggestion
        Self::simple_predicate(1)
```
?

src/resource/prelude.rs
```diff
@@ -0,0 +1,87 @@
...
+
+lazy_static! {
+    /// Prelude resource
+    pub static ref PRELUDE: Resource = {
+        let mut prelude = Resource::new(SmolStr::new_inline("prelude"));
+        prelude
+            .add_type(
+                SmolStr::new_inline("float64"),
+                vec![],
+                "float64".into(),
```
cqc-alec, line 23: Are we removing the `arithmetic.float.types` resource (which also defines a "float64" type)? Or should that (and maybe all of the arithmetic resources) move into the prelude?

src/types.rs
```diff
@@ -32,8 +42,227 @@ impl EdgeKind {
...
+
+    /// Initialize a new sum type by providing the possible variant types.
+    #[inline(always)]
+    pub fn new_sum(types: impl Into<TypeRow>) -> Self {
+        Self::new(TypeEnum::Sum(types.into()))
+    }
+
+    /// Initialize a new custom type.
+    // TODO remove? Resources/TypeDefs should just provide `Type` directly
+    pub const fn new_extension(opaque: CustomType) -> Self {
```
cqc-alec, line 182: Would `new_opaque` be a better name?

src/types.rs
```diff
@@ -32,8 +43,280 @@ impl EdgeKind {
...
+            (Copyable, Eq) => true,
+            (Eq, Copyable) => false,
+            _ => true,
+        }
+    }
+}
+
+/// Check containment of optional type bound (where None implies no bound -
+/// which contains bounded types).
+pub fn optional_bound_contains(bound: Option<TypeBound>, other: Option<TypeBound>) -> bool {
```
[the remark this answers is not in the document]
ss2165: yes - done

src/types/serialize.rs
```diff
@@ -0,0 +1,82 @@
...
+use super::AbstractSignature;
+
+use crate::ops::AliasDecl;
+use crate::types::primitive::PrimType;
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug)]
+#[serde(tag = "t")]
+pub(crate) enum SerSimpleType {
+    I,
+    G(Box<AbstractSignature>),
```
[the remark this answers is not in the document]
ss2165: These are the serialized variants so we've kept them small. We could expand and
use serde-rename, but I think outside the scope of this PR

src/values.rs
ss2165: don't think so - this is a value, and the tag is zero indexed.

src/resource/prelude.rs
ss2165: None of the tests actually use it so I'm just removing it - what exactly the
prelude should contain can wait (for the purposes of this PR it is types
required for existing tests)

src/types.rs
ss2165: I'm not sure I like referring to things as opaque any more - quite often they
are not actually opaque, just not core.

src/types/check.rs
```diff
@@ -0,0 +1,89 @@
...
+
+use crate::values::{PrimValue, Value};
+
+use super::{primitive::PrimType, CustomType, Type, TypeEnum};
+
+/// Struct for custom type check fails.
+#[derive(Clone, Debug, PartialEq, Eq, Error)]
+pub enum CustomCheckFail {
+    /// The value had a specific type that was not what was expected
+    #[error("Expected type: {0} but value was of type: {1}")]
```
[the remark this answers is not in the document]
ss2165: made them both explicit

src/types.rs
ss2165: not that I like `new_extension` much either

cqc-alec (approved):
Nice.

PR #379, opened 2023-08-08
Tag the regions with their root node handle

aborgna-q:
Add's a `Root: NodeHandle` parameter to the views to restrict the hugr type. The root tag is dynamically checked at construction time.

The next step would be to tag the `Hugr`s too, and add dynamic casting between handle types.


ss2165 (approved):
nice

PR #381, opened 2023-08-08
SerSimpleType: use Vec not TypeRow

acl-cqc:
* Also inlining singly-used methods `box_convert` and `box_convert_try`
* And switch TypeRowElem from a blanket impl to "elective" - we elect to allow only TypeRows of SimpleType/ClassicType/HashableType


ss2165 (approved):
:+1 looks good

PR #384, opened 2023-08-09
Tag wrapper for self-reporting leaves

ss2165:
check compatibility on construction

### DIFF
--- a/src/types/basic.rs
+++ b/src/types/basic.rs
@@ -1,5 +1,7 @@
 #![allow(missing_docs)]
 
+use std::marker::PhantomData;
+
 use super::{AbstractSignature, CustomType, TypeTag};
 
 pub enum EqLeaf {
@@ -49,23 +51,32 @@ impl TypeClass for ClassicLeaf {
 impl TypeClass for AnyLeaf {
     const TAG: TypeTag = TypeTag::Simple;
 }
-pub struct TaggedWrapper<T>(TypeTag, T);
+pub struct Tagged<I, T>(I, PhantomData<T>);
 
-impl<T> TaggedWrapper<T> {
-    pub fn tag(&self) -> TypeTag {
-        self.0
+pub trait GetTag {
+    fn tag(&self) -> TypeTag;
+}
+
+impl GetTag for CustomType {
+    fn tag(&self) -> TypeTag {
+        self.tag()
     }
+}
 
+impl<T: GetTag, C: TypeClass> Tagged<T, C> {
+    pub fn new(inner: T) -> Self {
+        assert!(C::TAG.contains(inner.tag()), "tag ");
+        Self(inner, PhantomData)
+    }
     pub fn inner(&self) -> &T {
-        &self.1
+        &self.0
     }
 }
 
 pub struct OpaqueType(CustomType);
 pub enum Type<T> {
     Prim(T),
-    Extension(OpaqueType),
-    Alias(TaggedWrapper<String>),
+    Extension(Tagged<CustomType, T>),
     Array(Box<Type<T>>, usize),
     Tuple(Vec<Type<T>>),
     Sum(Vec<Type<T>>),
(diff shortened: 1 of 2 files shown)


src/types/basic.rs
```diff
@@ -26,16 +34,13 @@ impl<T: Into<ClassicLeaf>> From<T> for AnyLeaf {
...
-    pub trait SealedType {}
-    impl<T: SealedLeaf> SealedType for Type<T> {}
+    use super::{AnyLeaf, ClassicLeaf, EqLeaf};
+    pub trait Sealed {}
+    impl Sealed for AnyLeaf {}
+    impl Sealed for ClassicLeaf {}
+    impl Sealed for EqLeaf {}
 }
-pub trait TypeClass: sealed::SealedLeaf {
+pub trait TypeClass: sealed::Sealed {
```
acl-cqc, line 43: Ok, this is reverting the change in the previous, right.

We might want to think about merging TypeClass and Sealed?

src/types/basic.rs
```diff
@@ -104,8 +161,8 @@ impl<T> Type<T> {
     fn upcast<T2: From<T>>(self) -> Type<T2> {
         match self {
             Type::Prim(t) => Type::Prim(t.into()),
-            Type::Extension(t) => Type::Extension(t),
-            Type::Alias(_) => todo!(),
+            Type::Extension(Tagged(t, _)) => Type::Extension(Tagged(t, PhantomData)),
```
acl-cqc, line 164: fwiw If we had the old Upcast, we could call that here, as proof we know this is OK, rather than bypassing the constructor check (so this code here moves inside the "trust" boundary whereas before it was outside)

src/types/basic.rs
```diff
@@ -49,41 +54,93 @@ impl TypeClass for ClassicLeaf {
...
 }
-pub struct TaggedWrapper<T>(TypeTag, T);
 
-impl<T> TaggedWrapper<T> {
-    pub fn tag(&self) -> TypeTag {
-        self.0
+#[derive(Clone, PartialEq, Debug, Eq)]
+pub struct Tagged<I, T>(I, PhantomData<T>);
+
+pub trait GetTag {
```
acl-cqc, line 61: I'd call this Tagged or DynamicTag

acl-cqc: RuntimeTag? Maybe just Tagged, the statically-known one is called type_tag_bound() after all

src/types/basic.rs
ss2165: you can't - that's how sealed traits have to work

src/types/basic.rs
```diff
@@ -49,41 +54,93 @@ impl TypeClass for ClassicLeaf {
 impl TypeClass for AnyLeaf {
     const TAG: TypeTag = TypeTag::Simple;
 }
-pub struct TaggedWrapper<T>(TypeTag, T);
 
-impl<T> TaggedWrapper<T> {
-    pub fn tag(&self) -> TypeTag {
-        self.0
+#[derive(Clone, PartialEq, Debug, Eq)]
+pub struct Tagged<I, T>(I, PhantomData<T>);
```
acl-cqc, line 59: I was sure I had a version of this without `PhantomData` but think I must have just been imagining it because indeed I can't make it work either! So ok :), yes this is right :-(

src/types/basic.rs
ss2165: i think that's ok  - but if it wasn't we could call Tagged::new here and implement try_into

src/types/basic.rs
acl-cqc: Rust being Rust, if this is GetTag then the method should be `get_tag`. I am *not* saying I like this....

but that'd work for RuntimeTag or DynamicTag, tho, if you can think of a better adjective than Runtime/Dynamic

src/types/basic.rs
acl-cqc: Yah I'm not seriously suggesting we shouldn't trust this code. One should be careful about this boundary creeping but I think it should be stable enough soon.

acl-cqc (approved):
👍

PR #386, opened 2023-08-09
Reorganise HUGR views

lmondada:
I have moved the file reorganisation out of #336. Please ignore the slightly
messy history, this will get squashed away.

### DIFF
--- a/src/hugr/views/hierarchy.rs
+++ b/src/hugr/views/hierarchy.rs
@@ -25,8 +25,7 @@ use portgraph::{LinkView, PortIndex, PortView};
 
 use crate::{hugr::NodeType, hugr::OpType, Direction, Hugr, Node, Port};
 
-use super::view::sealed::HugrInternals;
-use super::{HugrView, NodeMetadata};
+use super::{sealed::HugrInternals, HugrView, NodeMetadata};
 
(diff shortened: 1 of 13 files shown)


lmondada:
As per your wise comment in #336 @acl-cqc :)

lmondada:
True, I've got no problem switching back to the singular `view`. What is your opinion @acl-cqc ?

NB The more "interesting" change is to move hierarchical views within `view(s)`, so that #336 can add further views there.

src/hugr/views/hierarchy.rs
```diff
@@ -345,7 +344,7 @@ where
     }
 }
 
-/// A common trait for views of a hugr region.
+/// A common trait for views of a HUGR hierarchical subgraph.
```
acl-cqc, line 347: nit: "for views of subgraphs of a Hugr that" and then something like "include all descendants/subtrees" or "are closed over children"

acl-cqc:
I'm neutral on `views` vs `view`, the main bit is to move `hierarchical_views` :)

acl-cqc:
Also, should `HierarchyView` be `HierarchicalView`? It's not a view *of* the hierarchy, so we want the adjective

PR #389, opened 2023-08-09
fix: Automatically add resource req for custom ops

### LINKED ISSUE #388: Custom operations implicitly require their extension

aborgna-q:
Closes #388

This fix is blocked by extension inference.
The failing test in `builder::circuit::test::with_nonlinear_and_outputs` constructs this diagram:

image

Where `MissingRsrc.MyOp` now requires `MissingRsrc`. The output node has that extension in `input_resources`, but the rest of the operations in the region need to define it too. (And I think we'll require a lift node for qubit 1?)

### DIFF
--- a/src/builder/circuit.rs
+++ b/src/builder/circuit.rs
@@ -132,6 +132,7 @@ mod test {
     use super::*;
     use cool_asserts::assert_matches;
 
+    use crate::extension::ExtensionSet;
     use crate::{
         builder::{
             test::{build_main, NAT, QB},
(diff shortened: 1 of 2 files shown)


acl-cqc:
Hmmm, see here - how does this relate, does this "fix" it?

https://github.com/CQCL-DEV/hugr/blob/377dc57ebe7a87b99c9b86ae2473da426d380f44/src/extension/op_def.rs#L227-L229

aborgna-q:
Updated to the latest HEAD.
This change still causes either a `TgtExceedsSrcExtensions` or a `MismatchedConcreteWithLocations` error in 24 tests.

acl-cqc:
> Updated to the latest HEAD. This change still causes either a `TgtExceedsSrcExtensions` or a `MismatchedConcreteWithLocations` error in 24/154 tests.

Sounds like the same as changing https://github.com/CQCL/hugr/blob/b256c2bc42e8499dac89c6f8bff7a0d478a1ff59/src/extension/op_def.rs#L193-L195 to add the new element rather than assert - which I think is probably an easier place to do it (that catches all cases in one go).

But yeah, the bulk of the work here is figuring out why the tests are failing. In most cases it is probably that the tests need updating....

acl-cqc:
I think we can close this as #734 is an up-to-date version that includes all the necessary test fixes ?

PR #391, opened 2023-08-09
Redux/serialize

acl-cqc:
Add src/types/type_enum/serialize.rs - we might want to think about moving this when we remove the old simple/serialize.rs.

Note I haven't tried the possibility of serializing Type<T> as a `Type<FlattenedLeaf>` or anything like that, that might also work (?).

### DIFF
--- a/src/types/type_enum.rs
+++ b/src/types/type_enum.rs
@@ -12,9 +12,16 @@ use super::{
 
 mod serialize;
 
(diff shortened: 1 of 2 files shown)


ss2165 (approved):
I've pushed the fix for this in #392 

otherwise I'm happy, certainly much nicer serialisation code than before!

PR #394, opened 2023-08-10
Put LeafOp::CustomOp inside a box.

aborgna-q:
Low-hanging fruit to reduce the OpType footprint.

Reduces LeafOp footprint from 216b to 88b, and OpType from 216b to 152b


ss2165 (approved):
some weird diffs but seems good 👍

PR #399, opened 2023-08-11
Move from static type bounds to dynamic

ss2165:
also move some non-core types to a prelude resourece
rename TypeTag -> TypeBound

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -288,19 +288,6 @@ lazy_static! {
     /// Prelude resource
     pub static ref PRELUDE: Resource = {
         let mut prelude = Resource::new(SmolStr::new_inline("prelude"));
-
-        prelude
-            .add_type(
-                SmolStr::new_inline("graph"),
-                vec![
-                    TypeParam::List(Box::new(TypeParam::Type(crate::types::TypeTag::Simple))),
-                    TypeParam::List(Box::new(TypeParam::Type(crate::types::TypeTag::Simple))),
-                ],
-                "graph".into(),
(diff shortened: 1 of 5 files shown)


src/resource.rs
```diff
@@ -281,3 +281,40 @@ impl FromIterator<ResourceId> for ResourceSet {
...
+                TypeDefTag::Explicit(crate::types::TypeTag::Hashable),
+            )
+            .unwrap();
+
+
+            prelude
+            .add_type(
+                SmolStr::new_inline("array"),
+                vec![TypeParam::Type(TypeTag::Simple), TypeParam::USize],
+                "usize".into(),
```
aborgna-q, line 314:
```suggestion
                "array".into(),
```

src/types/type_enum.rs
```diff
@@ -1,82 +1,104 @@
...
+    Tuple(Vec<Type>),
     #[display(fmt = "Sum({})", "DisplayRow(_0)")]
-    Sum(Vec<Type<T>>),
+    Sum(Vec<Type>),
+}
+impl TypeEnum {
+    fn least_upper_bound(&self) -> Option<TypeBound> {
+        match self {
+            TypeEnum::Prim(p) => p.bound(),
+            TypeEnum::Tuple(ts) => least_upper_bound(ts.iter().map(Type::least_upper_bound)),
```
aborgna-q, line 28: Using the same names for different things is a bit confusing.

aborgna-q (approved):
Quite noisy diff, but LGTM

PR #400, opened 2023-08-11
refactor signature in to own private module

ss2165:
looking ahead to #363 

this will allow more core types operations to live in the top level file


src/types/simple/serialize.rs
```diff
@@ -15,7 +15,7 @@ use super::TypeRow;
 
 use super::SimpleType;
 
-use super::super::AbstractSignature;
+use super::super::signature::AbstractSignature;
```
acl-cqc, line 18: surely this change is not necessary as it was re-exported?

acl-cqc (approved):
One nit, LGTM

cqc-alec (approved):

Issue #284, opened 2023-07-26
Simplify trait structure in the project


aborgna-q:
`HugrView`, `HugrMut`, `AsMut`, `AsRef`, `Buildable`, ...
https://github.com/CQCL-DEV/hugr/pull/281#issuecomment-1650224814

Steps:
- [x] #402
- [ ] Add views/references for mutable regions (implementing `HugrMut`) #293
- [ ] Move builder primitives to the mut trait
- [ ] Use the mutable view on the builder (statically checking the `RootHandle`s).

aborgna-q:
Related: #293

PR #402, opened 2023-08-11
`s/HugrMut/HugrInternalsMut/` as a sealed trait, and expose a minimal HugrMut

aborgna-q:
First step of #284 

Moves the mutating methods to a sealed trait, and makes `HugrMut` `pub`.
It also relates it to `HugrView` via a borrowing method.


ss2165 (approved):
👍

PR #403, opened 2023-08-11
Move Type, TypeBound, PrimType to better locations

ss2165:
namely Type, TypeBound -> types.rs
PrimType -> primitive.rs

also add docstrings


aborgna-q (approved):

src/types.rs
```diff
@@ -90,9 +90,191 @@ impl TypeTag {
...
+        }
+    }
+}
+
+#[derive(
+    Clone, PartialEq, Debug, Eq, derive_more::Display, serde::Serialize, serde::Deserialize,
+)]
+#[display(fmt = "{}", "_0")]
+#[serde(into = "serialize::SerSimpleType", from = "serialize::SerSimpleType")]
+/// A HUGR type.
```
aborgna-q, line 177: This is a key thing in the public API, so it will need a more extensive explanation in the future.

PR #407, opened 2023-08-14
Propagate new type system to code base

ss2165:
* requires adding some constant types to replace those removed
* move all value related code to values.rs
* move value type checking code to the type rather than the value
* Should pass CI as-is


src/ops/constant.rs
```diff
@@ -115,220 +107,85 @@ impl OpTrait for Const {
...
-        false
-    }
-}
-
-impl_downcast!(CustomConst);
-impl_box_clone!(CustomConst, CustomConstBoxClone);
-
-#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
-/// A value stored as a serialized blob that can report its own type.
-struct CustomSerialized {
```
ss2165, line 292: moved to values.rs

src/types.rs
```diff
@@ -270,11 +270,14 @@ where
 }
 
 pub(crate) const ERROR_TYPE: Type = Type(
-    TypeEnum::Prim(primitive::PrimType::E(CustomType::new_simple(
-        smol_str::SmolStr::new_inline("error"),
-        smol_str::SmolStr::new_inline("MyRsrc"),
-        Some(TypeBound::Eq),
-    ))),
+    TypeEnum::Prim(primitive::PrimType::E(
```
ss2165: just format I think

src/ops/leaf.rs
```diff
@@ -78,7 +78,9 @@ pub enum LeafOp {
 
 impl Default for LeafOp {
     fn default() -> Self {
-        Self::Noop { ty: Type::Qubit }
+        Self::Noop {
+            ty: Type::graph(AbstractSignature::new_linear(vec![])),
```
ss2165, line 81: this should maybe be Type::unit()?

src/ops/constant.rs
```diff
@@ -213,33 +156,33 @@ mod test {
...
         assert_matches!(res, Err(ConstTypeError::TupleWrongLength));
     }
 
     #[test]
     fn test_constant_values() {
         const T_INT: Type = Type::usize();
-        const V_INT: Value = Value::Hashable(HashableValue::Int(257));
-        V_INT.check_type(&T_INT).unwrap();
-        custom_value(17.4).check_type(&CLASSIC_T).unwrap();
+        const V_INT: Value = CustomTestValue(Some(TypeBound::Eq)).into();
```
ss2165: yes these are not actually constants so won't compile, left for a future commit!

src/types/type_row.rs
```diff
@@ -32,7 +31,7 @@ impl<T: Display + TypeRowElem> Display for TypeRow<T> {
 
 // TODO some of these, but not all, will probably want exposing via
 // pyo3 wrappers eventually.
-impl<T: TypeRowElem> TypeRow<T> {
+impl TypeRow {
```
aborgna-q, line 34: P-low issue: replace these with a `delegate!` call (so we don't miss the #inline directives :) )

src/ops/leaf.rs
aborgna-q: Yeah. This shouldn't normally be used but `unit` is a good default.

src/hugr/validate.rs
```diff
@@ -393,7 +393,7 @@ impl<'a> ValidationContext<'a> {
                 true
             }
             ty => {
-                if !local && !matches!(ty, EdgeKind::Value(SimpleType::Classic(_))) {
+                if !local && !matches!(&ty, EdgeKind::Value(t) if t.least_upper_bound().is_some()) {
```
aborgna-q, line 396: At this point it may be clearer to separate the `EdgeKind::Value` match branch (even if it needs repeating the return statement).
And maybe add a comment that `least_upper_Bound = None ⟹ non-classical`.

aborgna-q (approved):
Nice!

I probably missed things, but LGTM

aborgna-q:
I hope the intertwined definitions for the prelude ops & consts can be moved out soon.

src/resource/type_def.rs
```diff
@@ -3,28 +3,28 @@ use std::collections::hash_map::Entry;
...
+use crate::types::TypeBound;
 
-/// The type tag of a [`TypeDef`]
+/// The type bound of a [`TypeDef`]
 #[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
-pub enum TypeDefTag {
+pub enum TypeDefBound {
     /// Defined by an explicit tag.
-    Explicit(TypeTag),
+    Explicit(Option<TypeBound>),
```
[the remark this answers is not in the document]
ss2165: i think the third variant makes the most sense

src/types.rs
```diff
@@ -265,30 +218,77 @@ impl Type {
...
 {
-    variant_rows.into_iter().map(Type::new_tuple).collect()
+    variant_rows
+        .into_iter()
+        .map(Type::new_tuple)
+        .collect_vec()
+        .into()
 }
 
+pub(crate) const ERROR_TYPE: Type = Type(
```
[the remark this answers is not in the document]
ss2165: It makes use of types/ private constructors to do a constant definition.

src/types/type_row.rs
```diff
@@ -39,36 +39,29 @@ impl TypeRow {
         }
     }
 
-    /// Iterator over the types in the row.
-    pub fn iter(&self) -> impl Iterator<Item = &Type> {
-        self.types.iter()
-    }
+    delegate! {
+        to self.types {
```
aborgna-q, line 43: I forgot that docs are not inherited 😓 
Maybe I'll make an issue for that in delegate's repo.

You can drop the #inlines, they are automatically added.

PR #411, opened 2023-08-15
Const usize, qubit, and ERROR type definitions in the prelude

ss2165:
Isolate PRELUDE "stringy" codes to the `prelude` module - can be improved with statics etc. in future.

### DIFF
--- a/src/ops/leaf.rs
+++ b/src/ops/leaf.rs
@@ -4,6 +4,7 @@ use smol_str::SmolStr;
 
 use super::custom::ExternalOp;
 use super::{OpName, OpTag, OpTrait, StaticTag};
+use crate::resource::prelude::{QB_T, USIZE_T};
 use crate::{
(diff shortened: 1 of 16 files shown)


src/ops/leaf.rs
```diff
@@ -150,14 +151,8 @@ impl OpTrait for LeafOp {
...
-        let qb_type: Type = Type::new_extension(
-            crate::resource::PRELUDE
-                .get_type("qubit")
-                .unwrap()
-                .instantiate_concrete(vec![])
-                .unwrap(),
-        );
-        let bit_type: Type = Type::new_usize();
+        let qb_type: Type = QB_T;
+        let bit_type: Type = USIZE_T;
```
aborgna-q: Can we solve this TODO now?

aborgna-q (approved):
👍

PR #414, opened 2023-08-16
Add static values to resources

ss2165:
* Addition and retrieval of typed ops::Const to resources
* simplify constant definitions with `KnownConstType`
* move `ConstUsize` to prelude

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -3,6 +3,7 @@
 //! TODO: YAML declaration and parsing. This should be similar to a plugin
 //! system (outside the `types` module), which also parses nested [`OpDef`]s.
 
+use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::fmt::{Debug, Display, Formatter};
 use std::sync::Arc;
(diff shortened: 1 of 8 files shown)


src/resource.rs
```diff
@@ -126,6 +128,32 @@ trait TypeParametrised {
     }
 }
 
+/// A constant value provided by a resource.
+/// Must be an instance of a type available to the resource.
```
ss2165, line 132: maybe this should be enforced as a check on creation?

aborgna-q (approved):
LGTM

src/values.rs
```diff
@@ -115,12 +116,44 @@ pub trait CustomConst:
...
 }
 
+/// Simpler trait for constant structs that have a known custom type to check against.
+pub trait KnownTypeConst: CustomConst {
+    /// The type of the constants.
+    const TYPE: CustomType;
+
+    /// Fixed implementation of [CustomConst::check_custom_type] that checks
+    /// against known correct type.
+    fn check_custom_type(&self, typ: &CustomType) -> Result<(), CustomCheckFailure> {
```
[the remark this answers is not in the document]
ss2165: good idea yes

src/resource.rs
ss2165: never mind - requires the resource dependency system first

PR #417, opened 2023-08-17
Enable static simple predicates

### LINKED ISSUE #397: More efficient representation (memory and serialised) for "simple predicate" types
Simple Predicates types := Sum of 0-tuples used all over the place, potentially even as the index on arrays. Would be good to optimise for them.

e.g. 
```rust 
enum Sum {
    Pred(usize),
    Other(TypeRow),
}
```

ss2165:
Since these are very commonly used types, optimise them a bit by adding a dedicated internal representation.

Fortunately the check for constructing them is easy.

Closes #397


src/ops/constant.rs
```diff
@@ -46,7 +46,7 @@ impl Const {
     }
 
     /// Constant Sum over units, used as predicates.
-    pub fn simple_predicate(tag: usize, size: usize) -> Self {
+    pub fn simple_predicate(tag: usize, size: u8) -> Self {
```
cqc-alec, line 49: Why restrict to `u8`?

src/types.rs
```diff
@@ -93,22 +93,65 @@ pub(crate) fn least_upper_bound(mut tags: impl Iterator<Item = TypeBound>) -> Ty
     .into_inner()
 }
 
+#[derive(Clone, PartialEq, Debug, Eq, derive_more::Display, Serialize, Deserialize)]
+/// Representation of a Sum type.
+/// Either store the types of the variants, or in the special (but common) case
+/// of a "simple predicate" (sum over empty tuples), store only the size of the predicate.
+enum SumType {
+    #[display(fmt = "SimplePredicate({})", "_0")]
+    Simple(u8),
```
cqc-alec, line 102: Why not use `usize` here?

ss2165: The thinking being that simple predicates branching factor >255 is not worth optimising for, and u8 will allow memory and serialisation optimisations.

cqc-alec (approved):

PR #418, opened 2023-08-17
Allow custom types defined by static references

ss2165:
Most of the time can avoid string comparison using pointer equality short circuiting.

Might be a bit sketch.

### DIFF
--- a/src/extension/prelude.rs
+++ b/src/extension/prelude.rs
@@ -48,20 +48,23 @@ lazy_static! {
     };
 }
 
-pub(crate) const USIZE_CUSTOM_T: CustomType = CustomType::new_simple(
+const USIZE_CUSTOM_T: CustomType = CustomType::new_simple(
     SmolStr::new_inline("usize"),
     SmolStr::new_inline("prelude"),
     TypeBound::Eq,
 );
 
-pub(crate) const QB_CUSTOM_T: CustomType = CustomType::new_simple(
+const QB_CUSTOM_T: CustomType = CustomType::new_simple(
     SmolStr::new_inline("qubit"),
     SmolStr::new_inline("prelude"),
     TypeBound::Any,
 );
 
-pub(crate) const QB_T: Type = Type::new_extension(QB_CUSTOM_T);
-pub(crate) const USIZE_T: Type = Type::new_extension(USIZE_CUSTOM_T);
+pub(crate) const USIZE_CUSTOM_REF: &CustomType = &USIZE_CUSTOM_T;
+pub(crate) const QB_CUSTOM_REF: &CustomType = &QB_CUSTOM_T;
+
+pub(crate) const QB_T: Type = Type::new_static_extension(QB_CUSTOM_REF);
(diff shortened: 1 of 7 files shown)


src/extension/prelude.rs
```diff
@@ -48,20 +48,23 @@ lazy_static! {
...
-pub(crate) const QB_CUSTOM_T: CustomType = CustomType::new_simple(
+const QB_CUSTOM_T: CustomType = CustomType::new_simple(
     SmolStr::new_inline("qubit"),
     SmolStr::new_inline("prelude"),
     TypeBound::Any,
 );
 
-pub(crate) const QB_T: Type = Type::new_extension(QB_CUSTOM_T);
-pub(crate) const USIZE_T: Type = Type::new_extension(USIZE_CUSTOM_T);
+pub(crate) const USIZE_CUSTOM_REF: &CustomType = &USIZE_CUSTOM_T;
```
aborgna-q, line 63: Why do we need the explicit `const` references?

src/utils.rs
```diff
@@ -39,3 +39,51 @@ pub(crate) mod test {
...
+        }
+    }
+}
+
+// can use pointer equality to compare static instances
+impl<T: PartialEq> PartialEq for MaybeStatic<T> {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            // pointer equality can give false-negative
+            (Self::Static(l0), Self::Static(r0)) => std::ptr::eq(*l0, *r0) || l0 == r0,
```
aborgna-q: This is only valid for `T: Eq`, otherwise you don't have reflexivity.

src/extension/prelude.rs
ss2165: I think because it is not temporary?
```
Checking quantinuum-hugr v0.1.0 (/Users/seyon/ng/hugr)
error[E0493]: destructor of `CustomType` cannot be evaluated at compile-time
  --> src/extension/prelude.rs:66:59
   |
66 | pub(crate) const QB_T: Type = Type::new_static_extension(&QB_CUSTOM_T);
   |                                                           ^^^^^^^^^^^- value is dropped here
   |                                                           |
   |                                                           the destructor for this type cannot be evaluated in constants

error[E0716]: temporary value dropped while borrowed
  --> src/extension/prelude.rs:66:59
   |
66 | pub(crate) const QB_T: Type = Type::new_static_extension(&QB_CUSTOM_T);
   |                               ----------------------------^^^^^^^^^^^-
   |                               |                           |          |
   |                               |                           |          temporary value is freed at the end of this statement
   |                               |                           creates a temporary value which is freed while still in use
   |                               argument requires that borrow lasts for `'static`

Some errors have detailed explanations: E0493, E0716.
For more information about an error, try `rustc --explain E0493`.
error: could not compile `quantinuum-hugr` (lib) due to 2 previous errors
```

aborgna-q: You'd need to use `static` instead of `const` so they are included in the compiled binary... But then you can't use it for the `KnownTypeConst::TYPE` const >.<

PR #421, opened 2023-08-18
add Collections extension with list type and operations

ss2165:
map is left for future (tierkreis)

### DIFF
--- a/src/std_extensions/collections.rs
+++ b/src/std_extensions/collections.rs
@@ -30,11 +30,7 @@ pub struct ListValue(Vec<Value>);
 #[typetag::serde]
 impl CustomConst for ListValue {
     fn name(&self) -> SmolStr {
-        if self.0.is_empty() {
-            "list".into()
-        } else {
-            format!("list:<{}>", self.0[0].name()).into()
-        }
+        SmolStr::new_inline("list")
     }
 
     fn check_custom_type(&self, typ: &CustomType) -> Result<(), CustomCheckFailure> {
--- a/src/std_extensions/collections.rs
+++ b/src/std_extensions/collections.rs
@@ -1,0 +1,209 @@
+//! Basic logical operations.
+
+use lazy_static::lazy_static;
+use serde::{Deserialize, Serialize};
+use smol_str::SmolStr;
+
+use crate::{
+    extension::{ExtensionSet, SignatureError, TypeDef, TypeDefBound},
+    types::{
+        type_param::{TypeArg, TypeParam},
+        CustomCheckFailure, CustomType, Type, TypeBound, TypeRow,
+    },
+    values::{CustomConst, Value},
+    Extension,
+};
+
+/// Reported unique name of the list type.
+pub const LIST_TYPENAME: SmolStr = SmolStr::new_inline("List");
+/// Pop operation name.
+pub const POP_NAME: SmolStr = SmolStr::new_inline("pop");
+/// Push operation name.
+pub const PUSH_NAME: SmolStr = SmolStr::new_inline("push");
+/// Reported unique name of the extension
+pub const RESOURCE_NAME: SmolStr = SmolStr::new_inline("Collections");
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+/// Dynamically sized list of values, all of the same type.
+pub struct ListValue(Vec<Value>);
+
+#[typetag::serde]
+impl CustomConst for ListValue {
+    fn name(&self) -> SmolStr {
+        if self.0.is_empty() {
+            "list".into()
+        } else {
+            format!("list:<{}>", self.0[0].name()).into()
+        }
+    }
+
+    fn check_custom_type(&self, typ: &CustomType) -> Result<(), CustomCheckFailure> {
+        let error = || {
+            // TODO more bespoke errors
+            CustomCheckFailure::Message("List type check fail.".to_string())
+        };
+
+        get_type(&LIST_TYPENAME)
+            .check_custom(typ)
+            .map_err(|_| error())?;
+
+        // constant can only hold classic type.
+        let [TypeArg::Type(t)] =  typ.args() else {
+            return Err(error());
+        };
+
+        // check all values are instances of the element type
+        for val in &self.0 {
+            t.check_type(val).map_err(|_| error())?;
+        }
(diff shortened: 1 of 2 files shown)


src/std_extensions/collections.rs
```diff
@@ -0,0 +1,209 @@
+//! Basic logical operations.
```
doug-q, line 1:
```suggestion
//! List type and operations.
```

src/std_extensions/collections.rs
```diff
@@ -0,0 +1,209 @@
...
+};
+
+/// Reported unique name of the list type.
+pub const LIST_TYPENAME: SmolStr = SmolStr::new_inline("List");
+/// Pop operation name.
+pub const POP_NAME: SmolStr = SmolStr::new_inline("pop");
+/// Push operation name.
+pub const PUSH_NAME: SmolStr = SmolStr::new_inline("push");
+/// Reported unique name of the extension
+pub const RESOURCE_NAME: SmolStr = SmolStr::new_inline("Collections");
```
doug-q, line 24:
```suggestion
pub const EXTENSION_NAME: SmolStr = SmolStr::new_inline("Collections");
```

src/std_extensions/collections.rs
```diff
@@ -0,0 +1,209 @@
...
+        Ok(())
+    }
+
+    fn equal_consts(&self, other: &dyn CustomConst) -> bool {
+        crate::values::downcast_equal_consts(self, other)
+    }
+}
+
+fn extension() -> Extension {
+    let mut extension = Extension::new(RESOURCE_NAME);
```
doug-q, line 68:
```suggestion
    let mut extension = Extension::new(EXTENSION_NAME);
```

src/std_extensions/collections.rs
```diff
@@ -0,0 +1,209 @@
...
+            "Pop from back of list".into(),
+            vec![TypeParam::Type(TypeBound::Any)],
+            Default::default(),
+            vec![],
+            move |args: &[TypeArg]| {
+                let (list_type, element_type) = list_types(args)?;
+
+                let inputs = TypeRow::from(vec![list_type.clone()]);
+                let outputs = TypeRow::from(vec![list_type, element_type]);
+                let resource_set = ExtensionSet::singleton(&RESOURCE_NAME);
```
doug-q, line 90:
```suggestion
                let resource_set = ExtensionSet::singleton(&EXTENSION_NAME);
```

src/std_extensions/collections.rs
```diff
@@ -0,0 +1,209 @@
...
+            "Push to back of list".into(),
+            vec![TypeParam::Type(TypeBound::Any)],
+            Default::default(),
+            vec![],
+            move |args: &[TypeArg]| {
+                let (list_type, element_type) = list_types(args)?;
+
+                let outputs = TypeRow::from(vec![list_type.clone()]);
+                let inputs = TypeRow::from(vec![list_type, element_type]);
+                let resource_set = ExtensionSet::singleton(&RESOURCE_NAME);
```
doug-q, line 107:
```suggestion
                let resource_set = ExtensionSet::singleton(&EXTENSION_NAME);
```

src/std_extensions/collections.rs
```diff
@@ -0,0 +1,209 @@
...
+        types::{type_param::TypeArg, Type},
+        Extension,
+    };
+
+    use super::*;
+    fn get_op(name: &str) -> &OpDef {
+        EXTENSION.get_op(name).unwrap()
+    }
+    #[test]
+    fn test_resource() {
```
doug-q, line 148:
```suggestion
    fn test_extension() {
```

src/std_extensions/collections.rs
```diff
@@ -0,0 +1,209 @@
...
+        assert_eq!(pop_sig.output(), &both_row);
+
+        let push_sig = get_op(&PUSH_NAME)
+            .compute_signature(&[TypeArg::Type(FLOAT64_TYPE)])
+            .unwrap();
+
+        let list_type = Type::new_extension(CustomType::new(
+            LIST_TYPENAME,
+            vec![TypeArg::Type(FLOAT64_TYPE)],
+            RESOURCE_NAME,
```
doug-q, line 200:
```suggestion
            EXTENSION_NAME,
```

src/std_extensions/collections.rs
```diff
@@ -0,0 +1,209 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+/// Dynamically sized list of values, all of the same type.
+pub struct ListValue(Vec<Value>);
+
+#[typetag::serde]
+impl CustomConst for ListValue {
+    fn name(&self) -> SmolStr {
+        if self.0.is_empty() {
+            "list".into()
```
doug-q, line 34: This is annoying, is it possible to add a type to `ListValue`?

doug-q (changes requested):

src/std_extensions/collections.rs
ss2165: Not sure what you mean, you mean add a concrete `CustomType` as a field of
`ListValue`? How would you use that for `name()`?

doug-q: change `ListValue` to be defined like
```rust
pub struct ListValue(Vec<Value>,Type); // not sure what the right type is, substitute that for Type
```
replace this `"list".into()` line with
```rust
format!("list:<{}>",self.1.name.into())
```
of course this will then coalesce somewhat with the else branch

ss2165: ah, got it

ss2165: elsewhere in the code we've tried to keep the value-holders minimal in terms of just storing data, keeping all the type information in the type structures. I think the right move here is just to have a simple "list" here for all cases and not worry about the element type in the constant name. What do you think?

doug-q: I'm fine with your approach. 

It's not the name that's that I'm worried about (though of course that's what this code is), it's that I would expect a `CustomConst` to have a `get_type` method. But, clearly it doesn't!

ss2165: yeah we had a whole discussion about synthesisable vs. checkable types and we went for checkable across the board. #318

PR #429, opened 2023-08-21
Remove `static_input` from `AbstractSignature`

### LINKED ISSUE #419: Make `static_input` just special fields on `LoadConst` and `Call` ops, and remove them everywhere else.

ss2165:
Closes #419
Need to add some logic to report the port for the static input if it exists.

Matches existing templated logic for "other_input" etc.

Boilerplate methods on Signature that ignore static_input can therefore be removed.

### DIFF
--- a/src/ops/dataflow.rs
+++ b/src/ops/dataflow.rs
@@ -10,6 +10,10 @@ pub(super) trait DataflowOpTrait {
     const TAG: OpTag;
     fn description(&self) -> &str;
     fn signature(&self) -> AbstractSignature;
+
+    fn static_input(&self) -> bool {
+        false
+    }
     /// The edge kind for the non-dataflow or constant inputs of the operation,
     /// not described by the signature.
     ///
@@ -120,6 +124,10 @@ impl<T: DataflowOpTrait> OpTrait for T {
     fn other_output(&self) -> Option<EdgeKind> {
         DataflowOpTrait::other_output(self)
     }
+
+    fn static_input(&self) -> bool {
+        DataflowOpTrait::static_input(self)
+    }
 }
 impl<T: DataflowOpTrait> StaticTag for T {
     const TAG: OpTag = T::TAG;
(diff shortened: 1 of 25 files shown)


src/ops/validate.rs
```diff
@@ -81,6 +81,24 @@ impl ValidateOp for super::Module {
...
+
+// impl ValidateOp for Call {
+//     fn validity_flags(&self) -> OpValidityFlags {
+//         OpValidityFlags {
+//             non_df_ports: (Some(1), None),
+//             ..Default::default()
+//         }
+//     }
+// }
+
```
aborgna-q, line 101:
```suggestion
```
?

src/ops/dataflow.rs
```diff
@@ -10,6 +10,10 @@ pub(super) trait DataflowOpTrait {
     const TAG: OpTag;
     fn description(&self) -> &str;
     fn signature(&self) -> AbstractSignature;
+
+    fn static_input(&self) -> Option<Type> {
```
aborgna-q, line 14:
```suggestion
    /// Get the static input type of this operation if it has one.
    fn static_input(&self) -> Option<Type> {
```

src/types/signature.rs
```diff
@@ -290,22 +229,15 @@ impl Signature {
...
-            pub fn static_input(&self) -> &TypeRow;
         }
     }
 }
 
 impl Display for AbstractSignature {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let has_inputs = !(self.static_input.is_empty() && self.input.is_empty());
+        let has_inputs = !self.input.is_empty();
         if has_inputs {
```
aborgna-q, line 239: It should be readable now
```suggestion
        if !self.input.is_empty() {
```

aborgna-q (approved):
LGTM

It makes the logic in `OpType` a bit messier, but AbstractSignature is a lot cleaner 👍

PR #431, opened 2023-08-21
add HUGR payload to PrimValue::Function

### LINKED ISSUE #374: Add PrimValue::Graph
Possibly unless we drop "Graph" types from core, which we might do according to outcome of #285.

Again according to #285, this might be as simple as
```
struct Graph(Hugr);

impl Graph {

    fn new(h: Hugr) -> Self {
        assert!(h.is_dfg());
        Self(h)
    }
}
```
(if monomorphic) whereas if polymorphic it might be a bit more complex

ss2165:
add `get_function_type` method to HugrView to report the function type
of the root if valid.

Closes #374

### DIFF
--- a/src/hugr/views.rs
+++ b/src/hugr/views.rs
@@ -154,7 +154,7 @@ pub trait HugrView: sealed::HugrInternals {
 
     /// For function-like HUGRs (DFG, FuncDefn, FuncDecl), report the function
     /// type. Otherwise return None.
-    fn get_function_type(&self, node: Node) -> Option<&FunctionType>;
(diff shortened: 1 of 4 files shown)


src/hugr/views.rs
```diff
@@ -315,6 +319,15 @@ where
         }
     }
 
+    fn get_function_type(&self, node: Node) -> Option<&FunctionType> {
```
croyzor, line 322: I think for the purposes this PR is working towards, this should always return the function type for the root. For purposes other than getting the function type _of a hugr_, people can use `get_nodetype` as you have

croyzor (approved):

PR #433, opened 2023-08-21
Spec out polymorphism "extension"

acl-cqc:
When we allow for type variables that could be, say, sizes of arrays, this works out worse than I thought.


acl-cqc:
@ss2165 I think #285 now has more clarity and precision where it's useful and less detail where it isn't, compared to this - OK to just close this?

PR #434, opened 2023-08-22
chore: Format let-else with nightly rustfmt

aborgna-q:
Preventively fixes CI errors before rust 1.72 is released, and rustfmt starts having opinions on let-else formatting.

https://blog.rust-lang.org/2023/07/01/rustfmt-supports-let-else-statements.html

### DIFF
--- a/src/algorithm/half_node.rs
+++ b/src/algorithm/half_node.rs
@@ -126,7 +126,14 @@ mod test {
             .map(|(s, _)| s)
             .exactly_one()
             .unwrap();
-        let [&left,&right] = edge_classes.keys().filter(|(s,_)| *s == split).map(|(_,t)|t).collect::<Vec<_>>()[..] else {panic!("Split node should have two successors");};
+        let [&left, &right] = edge_classes
+            .keys()
+            .filter(|(s, _)| *s == split)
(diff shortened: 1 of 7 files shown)


src/algorithm/half_node.rs
```diff
@@ -126,7 +126,14 @@ mod test {
...
             .exactly_one()
             .unwrap();
-        let [&left,&right] = edge_classes.keys().filter(|(s,_)| *s == split).map(|(_,t)|t).collect::<Vec<_>>()[..] else {panic!("Split node should have two successors");};
+        let [&left, &right] = edge_classes
+            .keys()
+            .filter(|(s, _)| *s == split)
+            .map(|(_, t)| t)
+            .collect::<Vec<_>>()[..]
+        else {
+            panic!("Split node should have two successors");
```
acl-cqc, line 135: The frequency of this suggests we should have a helper method in `nest_cfgs::test` to do this. Or perhaps these would be better written using `collect_tuple` or some such. Probably not for this PR, but I could happily come up with something before/after/as an extra commit here ??

acl-cqc (approved):
Looks good. I think the word you want in the description is "pre-emptively" (rather than preventatively) ;-)

PR #438, opened 2023-08-22
Change compute_signature to return FunctionType (et al)

acl-cqc:
Addressing a longstanding TODO.

I used `FunctionType {input: ..., output:...,  extension_reqs:... }` but alternatively one could use `FunctionType::new(..., ...).with_extension_delta(...)`.  (I missed this at first because it looks like other similar methods returning a Signature, so it might be preferable)

### DIFF
--- a/src/extension/op_def.rs
+++ b/src/extension/op_def.rs
@@ -7,7 +7,7 @@ use super::{
     Extension, ExtensionBuildError, ExtensionId, ExtensionSet, SignatureError, TypeParametrised,
 };
 
-use crate::types::{SignatureDescription, TypeRow};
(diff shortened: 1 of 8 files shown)


acl-cqc:
> I think that is much preferable - makes the extension definitions more readable.

Heh, I came to the same (or nearly) conclusion independently, and have switched to `FunctionType::new` in cases where the ExtensionSet is empty (new/default) - the vast majority.

There does still appear to be some confusion/disagreement in the code as to whether the ExtensionSet should include the owning Resource, but let's leave that for #389 .

ss2165 (approved):
👍

PR #441, opened 2023-08-22
feat: Split methods between `HugrMut` and `HugrMutInternals`

aborgna-q:
Splits the mutable methods between `HugrMutInternals` & `HugrMut`.

  When implemented by a (future) non-`AsMut<Hugr>`, the default impl checks whether the related nodes are valid in the view before calling the internal hugrs mutable primitives.
  
Other small changes:

- Moved `canonicalize_nodes` from `HugrMut` to `Hugr`, since it doesn't make sense to call in on a view.
- Added a root parameter to `Hugr::canonical_order` so we can get the canonical order for regions.
- Added an assert to validate the root type when querying a hugrview
- ~Add a lifetime to `HugrInternals::Portgraph` so the views can return clones instead of double references.~
  I had to roll it back due to the convex check requiring `Copy` on the portgraph view.


src/hugr.rs
```diff
@@ -238,6 +242,46 @@ impl Hugr {
...
     }
+
+    /// Compact the nodes indices of the hugr to be contiguous, and order them as a breadth-first
+    /// traversal of the hierarchy.
+    ///
+    /// The rekey function is called for each moved node with the old and new indices.
+    ///
+    /// After this operation, a serialization and deserialization of the Hugr is guaranteed to
+    /// preserve the indices.
+    pub fn canonicalize_nodes(&mut self, mut rekey: impl FnMut(Node, Node)) {
```
ss2165, line 253: this is just a direct move yes?

aborgna-q:
I'll re-add the rolled-back lifetime parameters in `HugrViewInternals::Portgraph` once the fix is released on portgraph.
https://github.com/CQCL/portgraph/pull/104

src/hugr/hugrmut.rs
```diff
@@ -8,277 +8,435 @@ use portgraph::{LinkMut, NodeIndex, PortMut, PortView, SecondaryMap};
...
+    #[inline]
+    fn add_op_with_parent(
+        &mut self,
+        parent: Node,
+        op: impl Into<OpType>,
+    ) -> Result<Node, HugrError> {
+        match self.contains_node(parent) {
+            true => self.hugr_mut().add_op_with_parent(parent, op),
+            false => Err(HugrError::InvalidNode { node: parent }),
+        }
```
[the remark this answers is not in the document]
aborgna-q: Yep, that makes more sense :)

src/hugr.rs
aborgna-q: Yes, it's a direct copy

src/hugr/hugrmut.rs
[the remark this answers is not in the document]
ss2165: probably looks nicer as an internal impl method

src/hugr/views.rs
```diff
@@ -52,14 +52,23 @@ pub trait HugrView: sealed::HugrInternals {
     where
         Self: 'a;
 
-    /// Return index of HUGR root node.
-    fn root(&self) -> Node;
+    /// Return the root node of this view.
+    #[inline]
+    fn root(&self) -> Node {
```
acl-cqc, line 57: This is just a way of exposing the hidden-in-HugrInternals root_node, right?
(Might one call that `root` too and then do `(self as HugrInternals).root()` here?)

acl-cqc: Yeah, I don't seriously suggest that, think I've got it now

src/hugr.rs
```diff
@@ -340,6 +384,12 @@ pub enum HugrError {
     /// An error occurred while manipulating the hierarchy.
     #[error("An error occurred while manipulating the hierarchy.")]
     HierarchyError(#[from] portgraph::hierarchy::AttachError),
+    /// The node doesn't exist.
```
acl-cqc, line 387: This doesn't seem to be used? (Also I wonder about the utility of naming the single field within, but there.)

src/hugr/views/hierarchy.rs
```diff
@@ -54,6 +53,20 @@ where
     _phantom: std::marker::PhantomData<Root>,
 }
 
+impl<'g, Root, Base> SiblingGraph<'g, Root, Base>
+where
+    Base: HugrInternals,
+{
+    /// Helper function to initialize the internal portgraph view.
+    fn init_graph(hugr: &Base, root: Node) -> FlatRegionGraph<'_> {
```
acl-cqc, line 61: AFAICS these two are used exactly once - in quite small calling methods - if the goal is to reuse this elsewhere (in a later PR), might be better to do this pulling out in that PR?

acl-cqc:
I'd also mention in the description - remove `add_node` and `add_op` (keeping only the variants thereof that specify a position in the hierarchy) - I think that's quite significant

src/hugr/hugrmut.rs
```diff
@@ -8,277 +8,413 @@ use portgraph::{LinkMut, NodeIndex, PortMut, PortView, SecondaryMap};
...
+    ///
+    /// The sibling node's parent becomes the new node's parent.
+    ///
+    /// # Errors
+    ///
+    ///  - If the sibling node does not have a parent.
+    ///  - If the attachment would introduce a cycle.
+    #[inline]
+    fn add_op_before(&mut self, sibling: Node, op: impl Into<OpType>) -> Result<Node, HugrError> {
+        self.valid_node(sibling)?;
```
acl-cqc, line 59: Ok, so the question here is - if "sibling" was the root (parent) of a SiblingGraph, would that be accepted by "valid_node"? By most definitions, that lets you add something outside the region, so we might need *two* versions of `valid_node`. (I *think* it's reasonable to add something as a child of one of the siblings in a SiblingGraph, even though the something-added would then not strictly be in the SiblingGraph...)

acl-cqc (commented):
Nice!! :)

src/hugr.rs
aborgna-q: I prefer to use named fields when the order is not relevant (what does `x.0` mean?). But I have no strong opinion on this.

src/hugr/hugrmut.rs
aborgna-q: Good catch!

src/hugr/views.rs
aborgna-q: The different name is to avoid ambiguity at local call sites. Users shouldn't see the `HugrMutInternals`, so there shouldn't be any confusion.

src/hugr/views/hierarchy.rs
aborgna-q: They where used in the rolled-back changes (that will come in a following PR once portgraph is updated).
I'll clean this up.

src/hugr.rs
aborgna-q: I changed it into a tuple struct for consistency with the other errors.

PR #443, opened 2023-08-22
[RFC] demo of tag char parametirsed handles and hugrs

ss2165:


src/hugr.rs
```diff
@@ -25,15 +25,16 @@ use pyo3::prelude::*;
...
 use crate::types::{FunctionType, Signature};
 
 use delegate::delegate;
-
+const ROOT_REPR: TagRepr = OpTag::ModuleRoot.repr();
 /// The Hugr data structure.
 #[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(feature = "pyo3", pyclass)]
-pub struct Hugr {
+pub struct Hugr<const C: TagRepr = ROOT_REPR> {
```
aborgna-q, line 37: Assuming we have a custom Debug impl here (and in every other thing with tags as const params), we will see the right tag names on error messages.

The main friction point is in type-hints from `rust-analyzer`, e.g.
image

I don't think there's a way around that...

ss2165: ah yeah I guess for that the char solution is better

acl-cqc: Hang on, what am I missing here. `C` isn't used inside the struct, or anywhere else; and is fixed (AFAICS?) at ModuleRoot. What about non-module-rooted Hugrs?

ss2165: shouldn't be fixed, I've just made ModuleRoot the default to avoid changes all over the place

acl-cqc: Ok, so I suspect I haven't understood this yet - it might become clearer when/if you do the "changes all over the place". But if you parameterize Hugr's (as well as views) we'll need to check that we don't change the root node's optype (via e.g. `replace_op`) in a way that invalidates the guarantee.

Hmmm, which criticism could also be levelled at the `replace_op` method of `HugrMutInternals` when used on a SiblingMut, I think. At least it's not public, oooops....

PR #445, opened 2023-08-22
feat: Change `TypeParam::USize` to `TypeParam::BoundedNat` and use in int extensions

### LINKED ISSUE #437: Replace integer width with a Sum
this should avoid invalid width errors

ss2165:
Closes #437 

BREAKING CHANGE: `USize` type param and type arg changed to `BoundedNat` with optional maximum. For previous behaviour use `TypeParam::max_nat()`

### DIFF
--- a/src/std_extensions/arithmetic/int_ops.rs
+++ b/src/std_extensions/arithmetic/int_ops.rs
@@ -2,7 +2,7 @@
 
 use smol_str::SmolStr;
 
-use super::int_types::{get_width, int_type};
+use super::int_types::{get_width_power, int_type};
 use crate::extension::prelude::{BOOL_T, ERROR_TYPE};
 use crate::type_row;
 use crate::types::type_param::TypeParam;
@@ -18,8 +18,8 @@ pub const EXTENSION_ID: SmolStr = SmolStr::new_inline("arithmetic.int");
 
 fn iwiden_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let m: u8 = get_width(arg0)?;
-    let n: u8 = get_width(arg1)?;
+    let m: u8 = get_width_power(arg0);
+    let n: u8 = get_width_power(arg1);
     if m > n {
         return Err(SignatureError::InvalidTypeArgs);
     }
@@ -32,8 +32,8 @@ fn iwiden_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet)
 
 fn inarrow_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let m: u8 = get_width(arg0)?;
-    let n: u8 = get_width(arg1)?;
+    let m: u8 = get_width_power(arg0);
+    let n: u8 = get_width_power(arg1);
     if m < n {
         return Err(SignatureError::InvalidTypeArgs);
     }
@@ -62,7 +62,7 @@ fn btoi_sig(_arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet),
 
 fn icmp_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg] = collect_array(arg_values);
-    let n: u8 = get_width(arg)?;
+    let n: u8 = get_width_power(arg);
     Ok((
         vec![int_type(n); 2].into(),
         type_row![BOOL_T],
@@ -72,7 +72,7 @@ fn icmp_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), 
 
 fn ibinop_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg] = collect_array(arg_values);
-    let n: u8 = get_width(arg)?;
+    let n: u8 = get_width_power(arg);
     Ok((
         vec![int_type(n); 2].into(),
         vec![int_type(n)].into(),
@@ -82,7 +82,7 @@ fn ibinop_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet)
 
 fn iunop_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg] = collect_array(arg_values);
-    let n: u8 = get_width(arg)?;
+    let n: u8 = get_width_power(arg);
     Ok((
         vec![int_type(n)].into(),
         vec![int_type(n)].into(),
@@ -92,8 +92,8 @@ fn iunop_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet),
 
 fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let n: u8 = get_width(arg0)?;
-    let m: u8 = get_width(arg1)?;
+    let n: u8 = get_width_power(arg0);
+    let m: u8 = get_width_power(arg1);
     let intpair: TypeRow = vec![int_type(n), int_type(m)].into();
     Ok((
         intpair.clone(),
@@ -104,8 +104,8 @@ fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet
 
 fn idiv_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let n: u8 = get_width(arg0)?;
-    let m: u8 = get_width(arg1)?;
+    let n: u8 = get_width_power(arg0);
+    let m: u8 = get_width_power(arg1);
     Ok((
         vec![int_type(n), int_type(m)].into(),
         vec![Type::new_sum(vec![int_type(n), ERROR_TYPE])].into(),
@@ -115,8 +115,8 @@ fn idiv_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), 
 
 fn imod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let n: u8 = get_width(arg0)?;
-    let m: u8 = get_width(arg1)?;
+    let n: u8 = get_width_power(arg0);
+    let m: u8 = get_width_power(arg1);
     Ok((
         vec![int_type(n), int_type(m)].into(),
         vec![Type::new_sum(vec![int_type(m), ERROR_TYPE])].into(),
@@ -126,8 +126,8 @@ fn imod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), 
 
 fn ish_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let n: u8 = get_width(arg0)?;
-    let m: u8 = get_width(arg1)?;
+    let n: u8 = get_width_power(arg0);
+    let m: u8 = get_width_power(arg1);
     Ok((
         vec![int_type(n), int_type(m)].into(),
         vec![int_type(n)].into(),
--- a/src/std_extensions/arithmetic/int_ops.rs
+++ b/src/std_extensions/arithmetic/int_ops.rs
@@ -89,7 +89,7 @@ fn iunop_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet),
 
 fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let intpair: TypeRow = vec![int_type(arg1.clone()), int_type(arg0.clone())].into();
+    let intpair: TypeRow = vec![int_type(arg0.clone()), int_type(arg1.clone())].into();
     Ok((
         intpair.clone(),
         vec![Type::new_sum(vec![Type::new_tuple(intpair), ERROR_TYPE])].into(),
@@ -100,8 +100,8 @@ fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet
 fn idiv_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
     Ok((
-        vec![int_type(arg1.clone()), int_type(arg0.clone())].into(),
-        vec![Type::new_sum(vec![int_type(arg1.clone()), ERROR_TYPE])].into(),
+        vec![int_type(arg0.clone()), int_type(arg1.clone())].into(),
+        vec![Type::new_sum(vec![int_type(arg0.clone()), ERROR_TYPE])].into(),
         ExtensionSet::default(),
     ))
 }
@@ -109,8 +109,8 @@ fn idiv_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), 
 fn imod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
     Ok((
-        vec![int_type(arg1.clone()), int_type(arg0.clone())].into(),
-        vec![Type::new_sum(vec![int_type(arg0.clone()), ERROR_TYPE])].into(),
+        vec![int_type(arg0.clone()), int_type(arg1.clone())].into(),
+        vec![Type::new_sum(vec![int_type(arg1.clone()), ERROR_TYPE])].into(),
         ExtensionSet::default(),
     ))
 }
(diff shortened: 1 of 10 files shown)


ss2165 (commented):
Tests pass but I think logic like
 typ.clone() == int_custom_type(self.width)

is incorrect because now the parameter should be the log width (requires updating the field stored in the contents to log width)

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -30,37 +29,22 @@ fn int_custom_type(n: u8) -> CustomType {
...
-            return Err(TypeArgError::TypeMismatch {
-                arg: arg.clone(),
-                param: TypeParam::USize,
-            }
-            .into());
-        }
-    };
-    if !is_valid_width(n) {
-        return Err(TypeArgError::InvalidValue(arg.clone()).into());
+pub fn get_width_power(arg: &TypeArg) -> u8 {
```
acl-cqc, line 43: I am a bit confused by this `width_power` nomenclature. What this returns is an integer width, right? A widely-understood concept being the number of bits in the int representation. It can be seen as a power of the tag of the SimplePredicate, but I don't think that really helps - indeed that's a description of the *implementation* of this method, but the purpose of this method is to isolate the details of this encoding from the rest of the method, right? Once I've got it, it's a width, I'm not expecting to take logarithm or do anything fancy with it, right?

So, alternative suggestions for naming:
* `int_width_as_power_of_simple_predicate` clearly too long but maximally explicit
* `int_width_as_power` shortened form of previous
* `int_width_from_typearg`, `int_width_from_tag`, `int_width_from_simp_pred` / `int_width_from_sum` etc.
* or maybe just `int_width` or `get_width`...

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -30,37 +29,22 @@ fn int_custom_type(n: u8) -> CustomType {
 /// Integer type of a given bit width.
 /// Depending on the operation, the semantic interpretation may be unsigned integer, signed integer
 /// or bit string.
-pub fn int_type(n: u8) -> Type {
-    Type::new_extension(int_custom_type(n))
+pub fn int_type(width_power: u8) -> Type {
```
acl-cqc, line 32: Comment says this is "a given bit width". It certainly looks like it's just a `width` rather than a `width_power` whatever one of those is. (Same confusion as other comment, IOW)

src/std_extensions/arithmetic/int_types.rs
ss2165: it should be `get_log_width`, I would like to use the log of the width everywhere

src/std_extensions/arithmetic/int_types.rs
ss2165: see  my comment above - I have not updated docs (or even the consts for that matter), just getting initial feedback on the idea for now

src/std_extensions/arithmetic/int_types.rs
acl-cqc: Yes, actually that was my initial suspicion!! Lol.

However, I see that this is only used in tests; in prod code. it is *always* followed immediately by passing the returned value to `int_type`. So you could just have one function `int_type_from_type_arg` (or similar) that does both together and I think that would completely avoid the opportunity for confusion

acl-cqc (commented):
Ok, so apart from confusion about the terminology of `width_power` this looks great to me :-)

src/types/type_param.rs
```diff
@@ -21,6 +21,9 @@ pub enum TypeParam {
     Type(TypeBound),
     /// Argument is a [TypeArg::USize].
     USize,
+    /// Argument is a [TypeArg::SimplePredicate] which is bounded by the size
+    /// specified here.
+    SimplePredicate(usize),
```
cqc-alec, line 26: Would it be possible to call this something like `BoundedUsize` which is more descriptive, even if it is implemented as a sum type?

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -30,37 +29,22 @@ fn int_custom_type(n: u8) -> CustomType {
...
-            }
-            .into());
-        }
-    };
-    if !is_valid_width(n) {
-        return Err(TypeArgError::InvalidValue(arg.clone()).into());
+pub fn get_width_power(arg: &TypeArg) -> u8 {
+    match arg {
+        TypeArg::SimplePredicate(n) if *n < POWERS_OF_TWO => *n as u8,
+        _ => panic!("type check should prevent this."),
```
cqc-alec, line 46: Should this really be a panic? The tests demonstrate that it's possible to invoke via the public API, I think?

ss2165: yes - I agree with alan that this function shouldn't be in the public API

src/types/type_param.rs
ss2165: I'd like to maintain the analogy with predicates in the main type system, but I agree a more descriptive name would be better. Maybe `UnitPredicate` (and update simple predicate to that in the main code base too?)

acl-cqc: I was thinking the same (BoundedUSize) but I take your argument. Really I think "predicate" is the wrong word to start. We're calling it that because we use it to control branches, but IMHO predicate is inherently binary - true or false - Google gives me this:

> 2.
> LOGIC
> something which is affirmed or denied concerning an argument of a proposition

So perhaps go with predicate for now but a better word might be "selector" or "choice"? or wilder ideas like "Nway"

cqc-alec: I like `Choice`.

ss2165: Yeah I can make a new issue for that rename, and that can cover this too when it is done.

ss2165: Yeah that works - the yaml syntax could offer "bool" which maps to Usize(bound: 2)

ss2165: though I prefer then `BoundedU` as the only one, with the default being the maximum (`BoundedU(u64::MAX)`)

src/std_extensions/arithmetic/int_types.rs
acl-cqc: hmmm, if it's called `get_log_width` I think that'd be clear, ok. Note the checks in `int_ops.rs` (that `n>m` etc.) would be equally valid on the actual width, if that was easier (probably not). Moreover, `int_type` is actually performing the inverse operation to this, so this is just validating - i.e. panicking....

src/types/type_param.rs
acl-cqc: Yes, I was meaning the Option to express the idea of defaulting to None, too

cqc-alec: I'd still approve of renaming `Predicate` to `Choice`, but if that can be made to work it sounds quite appealing.

ss2165: yeah I think that is a separate task in the main codebase

src/std_extensions/arithmetic/int_types.rs
ss2165: changed so int_type is just constructed from the arg directly, and the checks are performed on the log width where necessary

ss2165: now fewer lines overall!

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -15,113 +14,103 @@ use crate::{
...
-        || (n == 8)
-        || (n == 16)
-        || (n == 32)
-        || (n == 64)
-        || (n == 128)
+const fn is_valid_log_width(n: u8) -> bool {
+    n <= POWERS_OF_TWO
 }
 
+const POWERS_OF_TWO: u8 = 7;
```
cqc-alec, line 35: Maybe `MAX_LOG_WIDTH` would be better?

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -15,113 +14,103 @@ use crate::{
...
-            return Err(TypeArgError::TypeMismatch {
-                arg: arg.clone(),
-                param: TypeParam::USize,
-            }
-            .into());
-        }
-    };
-    if !is_valid_width(n) {
-        return Err(TypeArgError::InvalidValue(arg.clone()).into());
+pub(super) fn get_width_power(arg: &TypeArg) -> Result<u8, TypeArgError> {
```
cqc-alec, line 40: Call this `get_log_width()`?

src/std_extensions/arithmetic/int_ops.rs
```diff
@@ -55,46 +55,41 @@ fn itob_sig(_arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet),
...
         ExtensionSet::default(),
     ))
 }
 
 fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let n: u8 = get_width(arg0)?;
-    let m: u8 = get_width(arg1)?;
-    let intpair: TypeRow = vec![int_type(n), int_type(m)].into();
+    let intpair: TypeRow = vec![int_type(arg1.clone()), int_type(arg0.clone())].into();
```
cqc-alec, line 92:
```suggestion
    let intpair: TypeRow = vec![int_type(arg0.clone()), int_type(arg1.clone())].into();
```

src/std_extensions/arithmetic/int_ops.rs
```diff
@@ -104,33 +99,27 @@ fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet
 
 fn idiv_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let n: u8 = get_width(arg0)?;
-    let m: u8 = get_width(arg1)?;
     Ok((
-        vec![int_type(n), int_type(m)].into(),
-        vec![Type::new_sum(vec![int_type(n), ERROR_TYPE])].into(),
+        vec![int_type(arg1.clone()), int_type(arg0.clone())].into(),
```
cqc-alec, line 103:
```suggestion
        vec![int_type(arg0.clone()), int_type(arg1.clone())].into(),
```

src/std_extensions/arithmetic/int_ops.rs
```diff
@@ -104,33 +99,27 @@ fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet
...
+        vec![Type::new_sum(vec![int_type(arg1.clone()), ERROR_TYPE])].into(),
```
cqc-alec, line 104:
```suggestion
        vec![Type::new_sum(vec![int_type(arg0.clone()), ERROR_TYPE])].into(),
```

src/std_extensions/arithmetic/int_ops.rs
```diff
@@ -104,33 +99,27 @@ fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet
...
 }
 
 fn imod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let n: u8 = get_width(arg0)?;
-    let m: u8 = get_width(arg1)?;
     Ok((
-        vec![int_type(n), int_type(m)].into(),
-        vec![Type::new_sum(vec![int_type(m), ERROR_TYPE])].into(),
+        vec![int_type(arg1.clone()), int_type(arg0.clone())].into(),
```
cqc-alec, line 112:
```suggestion
        vec![int_type(arg0.clone()), int_type(arg1.clone())].into(),
```

src/std_extensions/arithmetic/int_ops.rs
```diff
@@ -104,33 +99,27 @@ fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet
...
+        vec![Type::new_sum(vec![int_type(arg0.clone()), ERROR_TYPE])].into(),
```
cqc-alec, line 113:
```suggestion
        vec![Type::new_sum(vec![int_type(arg1.clone()), ERROR_TYPE])].into(),
```

src/std_extensions/arithmetic/int_ops.rs
```diff
@@ -104,33 +99,27 @@ fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet
...
 }
 
 fn ish_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet), SignatureError> {
     let [arg0, arg1] = collect_array(arg_values);
-    let n: u8 = get_width(arg0)?;
-    let m: u8 = get_width(arg1)?;
     Ok((
-        vec![int_type(n), int_type(m)].into(),
-        vec![int_type(n)].into(),
+        vec![int_type(arg1.clone()), int_type(arg0.clone())].into(),
```
cqc-alec, line 121:
```suggestion
        vec![int_type(arg0.clone()), int_type(arg1.clone())].into(),
```

src/std_extensions/arithmetic/int_ops.rs
```diff
@@ -104,33 +99,27 @@ fn idivmod_sig(arg_values: &[TypeArg]) -> Result<(TypeRow, TypeRow, ExtensionSet
...
+        vec![int_type(arg1.clone())].into(),
```
cqc-alec, line 122:
```suggestion
        vec![int_type(arg0.clone())].into(),
```

cqc-alec (changes requested):

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -15,113 +14,103 @@ use crate::{
...
+    CustomType::new(INT_TYPE_ID, [width_arg], EXTENSION_ID, TypeBound::Copyable)
 }
 
 /// Integer type of a given bit width.
 /// Depending on the operation, the semantic interpretation may be unsigned integer, signed integer
 /// or bit string.
-pub fn int_type(n: u8) -> Type {
-    Type::new_extension(int_custom_type(n))
+pub fn int_type(width_arg: TypeArg) -> Type {
+    Type::new_extension(int_custom_type(width_arg))
```
acl-cqc, line 28: Given this function is `pub` (should it be??), you should probably do some validation here. Or are we leaving that to when we have the extension registry available in `Hugr::validate`?

ss2165: So lots of things:

- The function was only used by the operation definitions, which should have their arguments validated also, so I have just made this method pub(super).
- However, the ops all listed their type params as TypeParam::max_usize(), which is incorrect! Now they all list the correct bounded parameter, so the check should happen. 
- Noting that direct access to valid integer types might be useful, I've added a lazy static array of the valid types. So all pub access should be correctly validated.


All that is to say...I wish we had more tests of this.

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -21,23 +21,34 @@ fn int_custom_type(width_arg: TypeArg) -> CustomType {
...
+/// The largest allowed log width.
+pub const MAX_LOG_WIDTH: u8 = 7;
 /// Type parameter for the log width of the integer.
-pub const LOG_WIDTH_TYPE_PARAM: TypeParam = TypeParam::BoundedUSize(POWERS_OF_TWO as u64);
+pub const LOG_WIDTH_TYPE_PARAM: TypeParam = TypeParam::BoundedUSize(MAX_LOG_WIDTH as u64);
 
-/// Get the bit width of the specified integer type, or error if the width is not supported.
-pub(super) fn get_width_power(arg: &TypeArg) -> Result<u8, TypeArgError> {
+/// Get the log width  of the specified type argument or error if the argument
+/// is invalid..
```
cqc-alec, line 50:
```suggestion
/// is invalid.
```

cqc-alec (approved):

src/types/type_param.rs
```diff
@@ -33,14 +33,21 @@ pub enum TypeParam {
...
 /// A statically-known argument value to an operation.
 #[derive(Clone, Debug, PartialEq, Eq, serde::Deserialize, serde::Serialize)]
 #[non_exhaustive]
 pub enum TypeArg {
     /// Where the (Type/Op)Def declares that an argument is a [TypeParam::Type]
     Type(Type),
-    /// Instance of [TypeParam::USize]. 64-bit unsigned integer.
-    USize(u64),
+    /// Instance of [TypeParam::BoundedUSize]. 64-bit unsigned integer.
+    BoundedUSize(u64),
```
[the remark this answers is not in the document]
cqc-alec: I'd be OK with that.

The other thing that slightly grates is that the parameter is the maximum allowed value rather than the minimum excluded value, which would be more consistent with range conventions and also allow us to define an instance with 0 values, but if we made that change we'd either have to make the parameter a `u128` or not allow 2^64-1 as a value; not sure I like either of those options though both probably harmless.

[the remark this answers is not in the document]
ss2165: yes I only made that choice for that reason. I don't like not allowing 2^64 -1, that could lead to surprising failures. making the parameter a u128 is probably fine though.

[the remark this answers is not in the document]
ss2165: going the whole hog in commit [d13a4c1](https://github.com/CQCL-DEV/hugr/pull/445/commits/d13a4c112165f2a65e2ed93099e3dbe9d82f25dc) to use an Option<NonZeroU64> to encode the bound in the same size as a u64.

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -1,127 +1,132 @@
...
-    };
-    if !is_valid_width(n) {
-        return Err(TypeArgError::InvalidValue(arg.clone()).into());
+/// The largest allowed log width.
+pub const MAX_LOG_WIDTH: u8 = 7;
+
+/// Type parameter for the log width of the integer.
+// SAFETY: unsafe block should be ok as the value is definitely not zero.
+pub const LOG_WIDTH_TYPE_PARAM: TypeParam =
+    TypeParam::bounded_nat(unsafe { NonZeroU64::new_unchecked(MAX_LOG_WIDTH as u64 + 1) });
```
cqc-alec, line 52: Would this be avoidable by using a macro definition for `MAX_LOG_WIDTH`, so that the compiler can see that it's a safe value? @aborgna-q ?

ss2165: The safe alternative is to use `new()` but that produces an option (None if you gave it zero). We could happily unwrap it...but unwrap is not const yet, in the works.

ss2165: our [miri](https://github.com/rust-lang/miri) CI check should check this is fine

PR #449, opened 2023-08-23
[WIP] Substitution into type schemes to produce FunctionTypes

acl-cqc:
No tests and I expect bugs (especially true/false inversions) but I hope this demonstrates what the code will look like. In particular if you `validate` the definition and `check_args` the args then substitution should be guaranteed to be ok.

Note the passing in of the `HashMap<SmolStr, Extension>` requested more widely in validation and needed here. Also note how this repeats bits of code from type checking, custom type bound computation, etc. etc. - if we integrated type variables properly into Hugr we would not need two copies of the code (one which deals with variables and one which does not), the version that knows about variables would be the only one we need.

[ ] Needs proper error reporting not just `Err(())` / "invalid"
[ ] tests!!
[ ] Possibly we can avoid duplicating struct/enum defs by representing type variables as opaque types (but unlikely to avoid duplicating code, these opaque types would have to be treated very differently to standard)
[ ] Also you can represent a type variable as a `TypeArgTemplate::TypeVar` or a `TypeArgTemplate::Type(TypeTemplate::TypeVar)`....reminiscent of the old `SimpleType::Classic(ClassicType::Container(...` situation. Here I think all representations just work, but they will be falsely declared as different (by PartialEq), I have an idea for a representation that avoids this

### DIFF
--- a/src/extension/type_template.rs
+++ b/src/extension/type_template.rs
@@ -1,0 +1,220 @@
+use std::collections::HashMap;
+
+use smol_str::SmolStr;
+
+use super::{ExtensionId, ExtensionSet, TypeParametrised};
+use crate::{ops::AliasDecl, Extension};
+
+use crate::types::{
+    type_param::{CustomTypeArg, TypeArg, TypeParam},
+    CustomType, FunctionType, Type, TypeRow,
+};
+
+/// deBruijn-indexed Variables
+#[derive(Clone, Debug, PartialEq, Eq, serde::Deserialize, serde::Serialize)]
+struct VariableRef(usize);
+
+#[derive(Clone, Debug, PartialEq, Eq, serde::Deserialize, serde::Serialize)]
+enum ExtensionSetTemplate {
+    Concrete(ExtensionSet),
+    TypeVar(VariableRef),
+}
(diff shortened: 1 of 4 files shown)


src/extension/type_template.rs
```diff
@@ -0,0 +1,353 @@
...
+    USize(u64),
+    Opaque(CustomTypeArg),
+    Sequence(Vec<TypeArgTemplate>),
+    Extensions(ExtensionSet),
+    TypeVar(VariableRef),
+}
+
+/// Representation of the type scheme for an Op with declared TypeParams.
+/// This should be the result of YAML parsing.
+pub struct SignatureTemplate(pub Vec<TypeParam>, FunctionTypeTemplate);
```
ss2165, line 68: anything stopping this being 
```suggestion
pub struct SignatureTemplate(pub Vec<TypeParam>, TypeTemplate);
```
even if we don't provide syntax for all types and keep it to function types, seems the logic is generally applicable

acl-cqc:
Closing in favour of #484 which saves a lot of code by using the existing Type(Arg)etc. structures

PR #452, opened 2023-08-24
[REVERTME] turn off extension validation

ss2165:
shouldn't be necessary after #330 
for tket2 tests to pass


ss2165:
Closing as tket2 is using a simpler workaround: https://github.com/CQCL-DEV/tket2/pull/71

PR #456, opened 2023-08-24
feat: Extension validation can take a closure of solutions

### LINKED ISSUE #454: Validation check needs to work with open extension sets
Resource inference succeeds even when not all locations on the hugr can have a concrete resource associated with them. If we succeed in this manner, a subsequent validation check should also succeed

croyzor:
Fixes #454.
Return from extension inference an extra solution set, which instantiates all of the inference variables to the empty set, meaning metavariables which depend on them can be solved

### DIFF
--- a/src/hugr/validate.rs
+++ b/src/hugr/validate.rs
@@ -38,7 +38,10 @@ struct ValidationContext<'a> {
 }
 
 impl Hugr {
-    /// Check the validity of the HUGR.
+    /// Check the validity of the HUGR, assuming that it has no open extension
+    /// variables.
+    /// TODO: Add a version of validation which allows for open extension
+    /// variables (see github issue #457)
     pub fn validate(&self) -> Result<(), ValidationError> {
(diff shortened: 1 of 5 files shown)


src/hugr/validate.rs
```diff
@@ -40,18 +40,30 @@ struct ValidationContext<'a> {
 impl Hugr {
     /// Check the validity of the HUGR.
     pub fn validate(&self) -> Result<(), ValidationError> {
-        let mut validator = ValidationContext::new(self);
+        self.validate_with_extension_closure(HashMap::new())
```
ss2165, line 43: please add a todo for validating hugrs with input extension set variables

croyzor: I've added issue #457 after our discussion this morning

ss2165: cool, the comment here can link to that then

ss2165 (approved):

PR #458, opened 2023-08-25
chore: devenv update

ss2165:
including update to rust stable 1.72


aborgna-q (approved):
Can't review much about SHA hashes, but 👍

PR #459, opened 2023-08-25
chore: undo nixpkgs update

ss2165:
on my mac aarch64 system at least I was getting link errors due to some other update...
I believe something to do with mktemp


aborgna-q (approved):

ss2165:
perhaps a `devenv ci` run in CI when devenv files are modified would be useful

PR #460, opened 2023-08-25
feat: add `HugrView::node_connections` to get all links between nodes

ss2165:
Uses portgraph `get_connections`


aborgna-q (approved):
LGTM

PR #461, opened 2023-08-29
feat: Instantiate inferred extensions

### LINKED ISSUE #446: [new] Add inferred extension requirements to the Hugr
After running extension inference, instantiate the solution

croyzor:
After running extension inference on a hugr, add the inferred extensions to the `op_type` field of the hugr.
Resolves #446


src/hugr.rs
```diff
@@ -202,9 +202,22 @@ impl Hugr {
...
+        // We only care about inferred _input_ extensions, because `NodeType`
+        // uses those to infer the output extensions
+        for ((node, _), input_extensions) in solution
+            .iter()
+            .filter(|((_, dir), _)| *dir == Direction::Incoming)
+        {
+            let nodetype = self.op_types.try_get_mut(node.index).unwrap();
+            match nodetype.signature() {
+                None => nodetype.input_extensions = Some(input_extensions.clone()),
+                Some(existing_ext_reqs) => {
```
acl-cqc, line 216: This confused me - let's rename `existing_ext_reqs`. Maybe `sig` or `sig_with_reqs` or something like that?

acl-cqc: And, `nodetype.input_extensions` still exists. Yet here we neither assign to it, nor assert about it, nor anything else. Should it be equal to the input_extensions of the signature, something like that? This `match` would make a lot more sense if we dealt with (e.g. `debug_assert`ed about) `nodetype.input_extensions` in both cases, rather than just one...

src/hugr.rs
```diff
@@ -447,4 +467,55 @@ mod test {
         let hugr = simple_dfg_hugr();
         assert_matches!(hugr.get_io(hugr.root()), Some(_));
     }
+
+    #[test]
+    fn extension_instantiation() -> Result<(), Box<dyn Error>> {
+        const BIT: Type = crate::extension::prelude::USIZE_T;
+        let r = ExtensionSet::singleton(&"R".into());
+
+        let root = NodeType::pure(ops::DFG {
```
acl-cqc, line 476: driveby nit: `NodeType::pure` seems a bit misnamed. The so-called "purity" is not about the nodetype - it's about what's already present on the inputs before we even get to that node. A "pure" nodetype would be one with an empty delta; `pure` should be called `new_no_extensions` or something...."no extensions" == unextended(??) with synonyms "brief",  "truncated", "compact" and "concise". I admit none of these are great. "Basic"? "Retracted" (!)?? Maybe just `NodeType::no_extensions` paralleling `open_extensions` ?

acl-cqc (commented):
Mostly this looks fine, and the tests look great, which is always a good indicator :). I could hit approve, but would like to see what you come up with re. `nodetype.input_extensions` first...

src/hugr.rs
croyzor: `nodetype.input_extensions` should be the same as `existing_ext_reqs.input_extensions`, but I'll refactor to make it more clear

src/hugr.rs
```diff
@@ -202,9 +202,22 @@ impl Hugr {
...
     }
 
-    /// TODO: Write this
-    fn instantiate_extensions(&mut self, _solution: ExtensionSolution) {
-        //todo!()
+    /// Add extension requirement information to the hugr in place.
+    fn instantiate_extensions(&mut self, solution: ExtensionSolution) {
+        // We only care about inferred _input_ extensions, because `NodeType`
+        // uses those to infer the output extensions
+        for ((node, _), input_extensions) in solution
```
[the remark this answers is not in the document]
croyzor: I think it's worth only returning the input extensions. As you say, there's no use for the rest

src/extension/infer.rs
```diff
@@ -857,21 +857,28 @@ mod test {
...
 
         builder.set_outputs([w])?;
-        let hugr = builder.base;
-        // TODO: when we put new extensions onto the graph after inference, we
-        // can call `finish_hugr` and just look at the graph
-        let (solution, extra) = infer_extensions(&hugr)?;
-        assert!(extra.is_empty());
+        let mut hugr = builder.base;
+        let closure = hugr.infer_extensions()?;
+        assert!(closure.is_empty());
```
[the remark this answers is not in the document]
croyzor: The closure is non-empty when the graph has open extension variables on it's inputs. In which case the closure provides an extra solution which assumes those variables are instantiated to the empty seat
(See the comment on `infer::infer_extensions`)

src/hugr.rs
croyzor: I've pushed `cc04a84` which addresses this

src/hugr.rs
croyzor: That's fair, but I'm not sure what the better name is. Maybe `open_extensions => open` and `pure => closed`?

src/hugr.rs
```diff
@@ -193,7 +193,7 @@ impl Hugr {
         rw.apply(self)
     }
 
-    /// Infer extension requirements
+    /// Infer extension requirements and add new information to `op_types` field
```
acl-cqc, line 196: Add a doclink to `infer::infer_extensions` here?
And, should the return type here be `ExtensionSolution`? (A typedef to the same type as here)

src/hugr.rs
acl-cqc: Yes, this is clearer, thanks :) (I see now what `signature()` does and yes equivalent, but now we don't need to know that)

src/hugr.rs
croyzor: I think it should be, yes. I was trying to distinguish the genuine solution (as `ExtensionSolution`) from the closure, but they really should be the same type regardless

src/hugr.rs
acl-cqc: Yes - let's leave renaming to another PR, but I see two reasonable schemes: `open_extensions` and `no_extensions` or `open` and `closed` (where currently we have `open_extensions` and `pure`)

src/hugr.rs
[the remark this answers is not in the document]
acl-cqc: In which case - this PR looks ok as it stands, if that's a followup; or do you want to roll that in here?

src/hugr.rs
croyzor: I've now addressed both of these in `8e55229` and `1802ccd`, respectively

PR #462, opened 2023-08-29
[fix] Use internal tag for SumType enum serialisation

mark-koch:
This makes it easier to handle the serialised format with pydantic.


ss2165:
for PR title and description pls use conventional commit + BREAKING footer 
https://www.conventionalcommits.org/en/v1.0.0/

ss2165 (approved):

PR #463, opened 2023-08-29
refactor: Add extension info to `Conditional` and `Case`

croyzor:
N.B. this does nothing with `input_extensions`
`Conditional` ops can now have a set of extension reqs which are added by their child `Case` ops. It's expected that the difference is the same set for the `Conditional` and all of its children

### DIFF
--- a/src/ops/controlflow.rs
+++ b/src/ops/controlflow.rs
@@ -2,6 +2,7 @@
 
 use smol_str::SmolStr;
 
+use crate::extension::ExtensionSet;
 use crate::types::{EdgeKind, FunctionType, Type, TypeRow};
 
 use super::dataflow::DataflowOpTrait;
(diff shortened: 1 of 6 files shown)


src/ops/controlflow.rs
```diff
@@ -209,6 +212,10 @@ impl OpTrait for Case {
     fn tag(&self) -> OpTag {
         <Self as StaticTag>::TAG
     }
+
+    fn signature(&self) -> FunctionType {
```
ss2165, line 216: could this return a reference (and clone on use)

ss2165 (commented):
wouldn't the natural thing be to allow each case to have its own delta and inference assigns the union of those to the conditional?

src/ops/controlflow.rs
croyzor: This is just following the OpTrait interface, but I think they could all just as easily be implemented that way

ss2165 (approved):

PR #464, opened 2023-08-29
feat: Location information in extension inference error

croyzor:


src/extension/infer.rs
```diff
@@ -375,10 +387,19 @@ impl UnificationContext {
         } else {
             None
         };
-        err.unwrap_or(InferExtensionError::MismatchedConcrete {
-            expected: rs1,
-            actual: rs2,
-        })
+        if let (Some(loc1), Some(loc2)) = (loc1, loc2) {
```
aborgna-q, line 390: I think the doc for this function needs updating?

> Try to turn mismatches into `ExtensionError` when possible

aborgna-q (approved):

PR #465, opened 2023-08-29
feat: Extension inference for conditional nodes

croyzor:
This builds on from #463 (merged) and #461.
Generate constraints saying that `Case` nodes' interfaces should match those of their `Conditional` and add a test

### DIFF
--- a/src/hugr.rs
+++ b/src/hugr.rs
@@ -1,6 +1,6 @@
 //! The Hugr data structure, and its basic component handles.
 
-mod hugrmut;
+pub mod hugrmut;
 
 pub mod rewrite;
 pub mod serialize;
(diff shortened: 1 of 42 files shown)


src/hugr.rs
```diff
@@ -447,4 +467,55 @@ mod test {
         let hugr = simple_dfg_hugr();
         assert_matches!(hugr.get_io(hugr.root()), Some(_));
     }
+
+    #[test]
+    fn extension_instantiation() -> Result<(), Box<dyn Error>> {
+        const BIT: Type = crate::extension::prelude::USIZE_T;
```
ss2165, line 473:
```suggestion
        const NAT: Type = crate::extension::prelude::USIZE_T;
```

to avoid confusion

src/hugr.rs
```diff
@@ -447,4 +467,55 @@ mod test {
...
+
+        assert_eq!(
+            hugr.op_types
+                .get(lift.index)
+                .signature()
+                .unwrap()
+                .input_extensions,
+            ExtensionSet::new()
+        );
+        assert_eq!(
```
ss2165, line 511: this test doesn't look related to the rest of the PR, worth adding a drive-by note

ss2165 (approved):
👍

src/hugr.rs
croyzor: I think this is a result of being based on #461. Now that that's merged I'll merge from `main` and expect this part of the diff to go away

croyzor:
I've refactored the test to avoid using builder methods, since we want the parent `Conditional` to have the empty resource set and in the near future the builder will default to making it open. Unfortunately, this means losing access to the `CaseBuilder` too

ss2165 (approved):

PR #466, opened 2023-08-29
feat: add T, Tdg, X gates

lmondada:
I need these gates to benchmark against Quartz.


lmondada:
I wasn't sure how verbose the gate descriptions should be, happy to add more if that is useful.

ss2165 (approved):

PR #468, opened 2023-08-30
Check for acyclicity; remove boundedness requirement

### LINKED ISSUE #435: Detect cyclic DAGs in validation
As per #228, this test creates a cyclic DAG but validation is "happy":
```
   #[test]
    fn test_cyclic() -> Result<(), HugrError> {
        let mut h = Hugr::new(NodeType::pure(ops::DFG {
            signature: FunctionType::new(type_row![QB_T], type_row![QB_T]),
        }));
        let input = h.add_op_with_parent(h.root(), ops::Input::new(type_row![QB_T]))?;
        let output = h.add_op_with_parent(h.root(), ops::Output::new(type_row![QB_T]))?;
        let cx = h.add_op_with_parent(h.root(), cx_gate())?;
        h.connect(input, 0, cx, 0)?;
        h.connect(cx, 0, output, 0)?;
        h.connect(cx, 1, cx, 1)?;
        h.validate().unwrap(); // TODO this should raise an error
        Ok(())
    }
```
Instead we should probably get a `ValidationError::NotABoundedDag` (or perhaps a new error, but anything cyclic clearly isn't any sort of DAG bounded or otherwise)

### LINKED ISSUE #447: Either document (in spec) or remove (from implementation) single-source-single-sink assumption
See https://github.com/CQCL-DEV/hugr/discussions/432

Not sure how much we've actually baked this assumption into the code.

### LINKED ISSUE #455: `validate_children_dag()` doesn't check for acyclicity
It uses `DfsPostOrder` from petgraph; it should probably use `Topo`.

cqc-alec:
Fixes #455 .
Fixes #447 .
Fixes #435 .

### DIFF
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -210,9 +210,8 @@ pub trait Dataflow: Container {
         let num_outputs = hugr.get_optype(hugr.root()).signature().output_count();
         let node = self.add_hugr(hugr)?;
 
-        let [inp, _] = self.io();
         let inputs = input_wires.into_iter().collect();
-        wire_up_inputs(inputs, node, self, inp)?;
+        wire_up_inputs(inputs, node, self)?;
 
         Ok((node, num_outputs).into())
     }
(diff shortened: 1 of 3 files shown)


cqc-alec:
I want to remove the single-source-single-sink requirement on the grounds that it adds unnecessary edges, and edges only reduce, never increase, opportunities for optimization.

aborgna-q (approved):
I agree this is better for optimizations and I cannot find a reason not to allow it, so LGTM

PR #473, opened 2023-08-31
feat: Default DFG builders to open extension sets

croyzor:
This does NOT provide a full solution to #424 (even just for DFGs) yet, but is a first step towards doing so.
- In inference tests, when we want concrete resources we need to avoid using builder methods.
- We should do inference on every kind of graph (instead of assuming some kinds will be closed).
- Default _some_ DFG builder methods to instantiating nodes with open resources (but not enough)


aborgna-q (approved):
LGTM

PR #478, opened 2023-08-31
fix: Allow for variables to get solved in inference

croyzor:
Variables are _allowed_ to be unsolved in an extension inference solution, but it's okay if they _are_ solved.


aborgna-q (approved):
LGTM

PR #480, opened 2023-08-31
refactor: `ExtensionSolution` only consists of input extensions

croyzor:
Only input extensions need to be inferred. From the values of the input extensions, the outputs can always be calculated

### DIFF
--- a/src/extension/validate.rs
+++ b/src/extension/validate.rs
@@ -5,10 +5,9 @@ use std::collections::HashMap;
 
 use thiserror::Error;
 
+use super::{ExtensionSet, ExtensionSolution};
 use crate::hugr::NodeType;
 use crate::{Direction, Hugr, HugrView, Node, Port};
-
-use super::ExtensionSet;
 
 /// Context for validating the extension requirements defined in a Hugr.
 #[derive(Debug, Clone, Default)]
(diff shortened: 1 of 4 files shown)


src/extension/validate.rs
```diff
@@ -23,10 +22,17 @@ impl ExtensionValidator {
     ///
     /// The `closure` argument is a set of extensions which doesn't actually
     /// live on the graph, but is used to close the graph for validation
-    pub fn new(hugr: &Hugr, closure: HashMap<(Node, Direction), ExtensionSet>) -> Self {
-        let mut validator = ExtensionValidator {
-            extensions: closure,
-        };
+    pub fn new(hugr: &Hugr, closure: ExtensionSolution) -> Self {
+        let mut extensions: HashMap<(Node, Direction), ExtensionSet> = HashMap::new();
```
croyzor, line 26: N.B. Here we reconstruct the original form of `ExtensionSolution`, because that's what validation works with. In light of this I'm not really sure whether this PR actually succeeds in making things more ergonomic, or reducing the size of the `ExtensionSolution` (since it'll be immediately expanded upon when `validate` is called.)

It might make sense as a halfway point towards refactoring validation in the same way? But that doesn't really seem necessary at the moment

acl-cqc: Ok....so....I think this is still an improvement as it stands - a good number of other places get cleaner and simpler at the cost of a bit of extra complexity here (in one place).

But what I think we could do....
* Make `output_extensions` be a method of OpSignature (not Signature), taking `input_extensions` as a parameter. (You can keep `Signature::output_extensions() {self.signature.output_extensions(self.input_extensions)}` if you want, but it'll only have one call site, in `infer.rs`, so you might decide not to bother with it)
* In validation, build a hashmap just of input extensions (so that's closure + any input_extensions in the hugr); then whenever you want the output-extensions, use OpSignature::output_extensions passing the value from the hashmap
* Then you don't need `Signature::get_extension(Direction)` either (and `ExtensionValidator::gather_extensions` starts to look trivial enough to inline and do on the hashmap *before* you build the ExtensionValidator?!)

acl-cqc (approved):
Thanks - I think this is an improvement, but we could go further....

PR #481, opened 2023-09-01
perf: use lazy static definittion for prelude registry

ss2165:
and avoid clone

### DIFF
--- a/src/extension/prelude.rs
+++ b/src/extension/prelude.rs
@@ -15,10 +15,11 @@ use crate::{
 
 use super::ExtensionRegistry;
 
+/// Name of prelude extension.
+pub const PRELUDE_ID: &str = "prelude";
 lazy_static! {
-    /// Prelude extension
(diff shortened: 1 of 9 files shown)


src/extension/prelude.rs
```diff
@@ -48,22 +49,23 @@ lazy_static! {
             .unwrap();
         prelude
     };
-}
+    /// An extension registry containing only the prelude
+    pub static ref PRELUDE_REGISTRY: ExtensionRegistry = [PRELUDE_DEF.to_owned()].into();
+
+    /// Prelude extension
+    pub static ref PRELUDE: &'static Extension = PRELUDE_REGISTRY.get(PRELUDE_ID).unwrap();
```
acl-cqc, line 56: Like this! :)

acl-cqc (approved):
Yes, I did wonder about that. Look good, thanks!

PR #487, opened 2023-09-04
feat: Add ConvexChecker

lmondada:
Exposing a HUGR-specific convex checker is required to be able to construct it from outside the hugr crate.


aborgna-q (approved):
LGTM

PR #488, opened 2023-09-04
feat: Add clippy lint for mut calls in a debug_assert

aborgna-q:
https://rust-lang.github.io/rust-clippy/master/index.html#/debug_assert_with_mut_call

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,5 @@
 #![warn(missing_docs)]
+#![warn(clippy::debug_assert_with_mut_call)]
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![warn(missing_docs)]
+#![warn(clippy::debug_assert_with_mut_call)]
```
acl-cqc, line 2: Maybe link to that issue here? I guess `git blame` and looking up the commit/PR description would get there...

acl-cqc (approved):
Looks good to me, at least until it causes trouble ;-)

PR #491, opened 2023-09-04
Check ExtensionId strings well-formed

acl-cqc:
...as a dot-separated list of (dot-free) identifiers, adding a new type `IdentList` of which `ExtensionId` is an alias.
E.g. no `foo..bar` or `.foo` or `5x` but `x5`, `foo.bar` ok.

`pub const fn new_unchecked` method allows bypassing the check. Nasty but those constants are damn useful.

Also a `const_extension_ids` that automatically generates *tests* that the names in the consts are valid.


src/lib.rs
```diff
@@ -9,6 +9,26 @@
 //!   structures.
 //!
 
+// Declares a 'const' member of a test module that's a new ExtensionId.
+// field_name should be an UPPERCASE i.e. name of a const.
+// This is here so that it is usable throughout the crate, but is not
+// visible from outside (as it would be if we used macro_export)
+// - as it won't *work* from outside because of the call to new_unchecked.
+#[allow(unused_macros)] // Because not used *in this file*
+macro_rules! test_const_ext_id {
```
aborgna-q, line 18: - Either put it in `extensions.rs` or in `macros.rs`.

- Use `$crate` to refer to `hugr` in the `use` statements (here it doesn't change much, but if used from another crate `crate::*` refers to the user's crate).

src/hugr/ident.rs
```diff
@@ -0,0 +1,98 @@
...
+
+impl std::ops::Deref for IdentList {
+    type Target = str;
+
+    fn deref(&self) -> &str {
+        self.0.deref()
+    }
+}
+
+impl PartialEq<str> for IdentList {
```
aborgna-q, line 61: I think this should be covered by `Deref`, but ok.

src/lib.rs
aborgna-q, line 18: This is quite confusing at the call-sites since there is no indication whatsoever that we are declaring a variable.

Since this is an internal macro, I'd very much prefer to leave the declaration explicit, and require us to write checks on the side. We can even use a macro to make it a one-liner;
```rust
macro_rules! test_extension_id {
    ($field_name:ident) => {
        paste::paste! {
            #[test]
            fn [<check_ $field_name:lower _wellformed>]() {
                ExtensionId::new(&*$field_name).unwrap();
            }
        }
    };
}
```

And if we were to make this public, I'd be even more against hiding the `const` in the user-facing API

aborgna-q:
Nice! I'd say we should move to an explicit list sooner rather than later.

That would however prevent any const declaration, so perhaps we should look on the other side and see if we shouldn't rather replace the consts with statics... (or if it's even possible)

src/ops/custom.rs
```diff
@@ -222,7 +222,10 @@ pub fn resolve_extension_ops(
     for n in h.nodes() {
         if let OpType::LeafOp(LeafOp::CustomOp(op)) = h.get_optype(n) {
             if let ExternalOp::Opaque(opaque) = op.as_ref() {
-                if let Some(r) = extension_registry.get(&opaque.extension) {
+                if let Some(r) = extension_registry.get({
+                    let k: &str = &opaque.extension;
+                    k
+                }) {
```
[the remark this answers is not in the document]
acl-cqc: Yeah, this is annoying. I tried just `&opaque.extension` and the typechecker complains:
```
error[E0277]: the trait bound `smol_str::SmolStr: Borrow<IdentList>` is not satisfied
   --> src/ops/custom.rs:225:57
    |
225 | ...extension_registry.get(&opaque.extension) {
    |                       --- ^^^^^^^^^^^^^^^^^ the trait `Borrow<IdentList>` is not implemented for `smol_str::SmolStr`
    |                       |
    |                       required by a bound introduced by this call
    |
    = help: the trait `Borrow<str>` is implemented for `smol_str::SmolStr`
```
so that `: &str` forces the `&` to use the `Borrow<str>` of IdentList to get a `&str` and then use the `Borrow<str>` of SmolStr (the keys in the map) that the message suggests. Do you know of a better way?

src/hugr/ident.rs
acl-cqc: Sadly not - removing it gives me about 100 errors like this:
```
error[E0277]: can't compare `IdentList` with `str`
   --> src/std_extensions/logic.rs:118:9
    |
118 |         assert_eq!(r.name(), "logic");
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no implementation for `IdentList == str`
```

I started out without this `impl PartialEq` and got everything OK besides the tests, then played around a bit with various Into/Borrow/etc., but couldn't get anything much else to work. Any ideas?

src/lib.rs
acl-cqc: Can this be referred to from outside the crate? It's not `#[macro_export]` so I didn't think it could be. It'd be useless ATM if it could be referred to, as it generates calls to the `pub(crate) fn new_unchecked`...

acl-cqc:
> That would however prevent any const declaration, so perhaps we should look on the other side and see if we shouldn't rather replace the consts with statics... (or if it's even possible)

I tried a toy example and think this works:
```
type ExtensionId = (Ident, MaybeStaticList<Ident>); // Head+tail to enforce list non-empty
enum MaybeStaticList<T> {
   Owned(Vec<T>),
   StaticData(&'static [T])
}
const FOO_BAR:ExtensionId = ExtensionId(Ident::new_unchecked("foo"), MaybeStaticList::StaticData(&[Ident::new_unchecked("bar")]));
```
However, it's a lot of boilerplate and I have to ask, is it really worth it?

Maybe it is - IMHO using `lazy_static` instead of `const` was a lot more pain at use sites (never mind the defs) because you have to `.to_owned()` all over the place.

src/hugr/ident.rs
acl-cqc: This works (without the PartialEq), but is a pain:
```
let n: &str = r.name();
assert_eq!(n, "logic");
```
so that `&str` is critical. IOW, it's exactly the same issue as [the other comment](https://github.com/CQCL-DEV/hugr/pull/491#discussion_r1315606215)

src/hugr/ident.rs
```diff
@@ -0,0 +1,98 @@
+use std::borrow::Borrow;
+
+use derive_more::Display;
+use lazy_static::lazy_static;
+use regex::Regex;
+use smol_str::SmolStr;
+use thiserror::Error;
+
+lazy_static! {
+    pub static ref PATH_REGEX: Regex = Regex::new(r"^[\w--\d]\w*(\.[\w--\d]\w*)*$").unwrap();
```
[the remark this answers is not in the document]
aborgna-q: nvm, I forgot `\w` includes `_` too. Though I'd say `[_[:alpha]` is a bit more clear.

src/lib.rs
aborgna-q: Unfortunately `#[cfg(test)]` is only internal to the crate, it is never exported in the public API.

Your declaration inside the macro call is very much possible with macros by example (by exactly matching a const declaration), I think that's a good solution.

src/ops/custom.rs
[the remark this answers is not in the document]
aborgna-q: Your `as &str` works 👍 
Another solution would have been `&*opaque.extensions` (borrow and then get the string slice)

src/hugr/ident.rs
aborgna-q: >.<
👍

acl-cqc: Yes, as I think you've already noticed - I've "solved" this (removing the `PartialEq` and the awkward `let`) via a bunch of `as &str`s

src/ops/custom.rs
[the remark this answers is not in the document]
acl-cqc: Yes, I thought there might be something like that. That was the kinda obvious "thing like that" and indeed it works...I've done that here but kept the tests with the explicit `as &str`.

src/hugr/ident.rs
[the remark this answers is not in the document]
acl-cqc: The latter is easier, yes. The "advantage" of `\w` is it includes unicode, which I'm generally in favour of, so I prefer to stick with this. There are concerns that unicode can be expensive tho so I guess I could be persuaded...

FWIW, this is what we went with [in Tierkreis](https://github.com/CQCL-DEV/tierkreis/blob/b0779abaf755d887a8cac219db3ccb08982692f0/tierkreis-core/src/symbol.rs#L12)

src/lib.rs
acl-cqc: Ok, so I now have a `pub fn new_unchecked` and am biting my tongue in protest :-( :-( :-(

However, the rest works well (and making `new_unchecked` pub was necessary for the doctest of the macro itself, which also validates the `$crate` etc. is correct).

For the record, I am totally happy to keep this private to the crate, i.e. just revert 82c9b3dbc42f9fe46a274dbb0feb07f00e93e353

src/macros.rs
```diff
@@ -80,3 +80,33 @@ macro_rules! type_row {
...
+/// # use hugr::macros::test_const_ext_id;
+///   test_const_ext_id! {
+///     pub const EXT_A: ExtensionId = "A";
+///     #[cfg(foobar)] pub (super) const EXT_A_B: ExtensionId = "A.B";
+///     const EXT_BAD: ExtensionId = "..55"; // this will generate a failing #[test] fn ....
+///   }
+/// # }
+/// ```
+#[macro_export]
+macro_rules! test_const_ext_id {
```
aborgna-q, line 98: Since the main effect of the macro is declaring a cost, maybe rename it to `const_extension_id` ?

The docs already explain that it does the validation.

aborgna-q (approved):
Awesome

src/macros.rs
acl-cqc: Yes, fair - it actually works in prod code too, if you fix the `paste` dep (which we won't). I went with `const_extension_ids` as you can declare more than one...

acl-cqc:
Also added a doc-comment to the doctest, and switched `:literal` to `:expr` which we mentioned in person the other day

PR #492, opened 2023-09-04
feat: Default more builder methods to open extension sets

croyzor:
Another step towards #424 (but not all the way there yet)

### DIFF
--- a/src/builder/conditional.rs
+++ b/src/builder/conditional.rs
@@ -178,7 +178,7 @@ impl ConditionalBuilder<Hugr> {
             extension_delta,
         };
         // TODO: Allow input extensions to be specified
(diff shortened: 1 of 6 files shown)


src/builder/conditional.rs
```diff
@@ -197,7 +196,7 @@ impl CaseBuilder<Hugr> {
             signature: signature.clone(),
         };
         // TODO: Allow input extensions to be specified
```
aborgna-q, line 198: Is the todo still valid if the extension set is open?

aborgna-q (approved):
👍

src/builder/conditional.rs
croyzor: Nope, I'll get rid of it!

PR #494, opened 2023-09-05
feat: More general portgraph references

aborgna-q:


aborgna-q:
@acl-cqc This should help with mutable hugr views, so filtered portgraphs can be created on the fly.

src/builder/module.rs
```diff
@@ -4,9 +4,10 @@ use super::{
     BuildError, Container,
 };
 
+use crate::hugr::hugrmut::sealed::HugrMutInternals;
```
acl-cqc, line 7: nit: I'd drop the reformatting/whitespace/rebracing here and in `src/hugr.rs` so the PR doesn't touch them.

acl-cqc (approved):
Looks good to me :). But if portgraph release is rolling around soon, let's wait for that.

PR #498, opened 2023-09-06
feat: Implement petgraph traits on Hugr

aborgna-q:
This lets us treat HUGRs in the same way as views everywhere (e.g. use them as Circuits) requiring `T: PetgraphHugr` where before we had `T: HierarchyView`.


src/hugr/views/petgraph.rs
```diff
@@ -6,12 +6,141 @@ use crate::hugr::HugrView;
...
-use crate::{Node, Port};
+use crate::{Hugr, Node, Port};
 
 use context_iterators::{ContextIterator, IntoContextIterator, MapWithCtx};
 use petgraph::visit as pv;
 use portgraph::NodeIndex;
 
+/// A trait grouping the multiple petgraph traits implemented on Hugrs
+//
+// TODO: We could just require these in `HugrView`.
```
acl-cqc, line 17: Ok, how about
```
struct PetgraphWrapper<T: HugrView>(T)

impl<T:HugrView> pv::XYZ for PetgraphWrapper<T> {...} // for each pv::XYZ
```
i.e. you'd need to build a PetgraphWrapper around each HugrView, but we could move most of those impls out of the macro altogether, and use the same struct for a Hugr or `&Hugr`...

aborgna-q:
@acl-cqc I did a quick impl of the wrapper so I can test in on tket2.
It seems nice, I'll report how it works over there.

src/hugr/views/petgraph.rs
```diff
@@ -1,187 +1,250 @@
...
+    type EdgeWeight = EdgeKind;
+}
 
-        impl<'g, 'a, Root, Base> pv::IntoNeighbors for &'g $hugr<'a, Root, Base>
-        where
-            Root: NodeHandle,
-            Base: HugrInternals + HugrView,
-        {
-            type Neighbors = <$hugr<'a, Root, Base> as HugrView>::Neighbours<'g>;
+impl<'g, 'a, T> pv::IntoNodeReferences for &'g PetgraphWrapper<'a, T>
```
acl-cqc, line 91: This is particularly good - the differences between the two impls (because of which I had to move this outside my macro in #501) seem to have gone

src/hugr/views/petgraph.rs
```diff
@@ -1,187 +1,250 @@
...
+            fn base_hugr(&self) -> &crate::Hugr;
+            fn root_node(&self) -> Node;
         }
-    };
+    }
 }
 
-impl_region_petgraph_traits!(SiblingGraph);
-impl_region_petgraph_traits!(DescendantsGraph);
+impl<'g, T> HugrView for PetgraphWrapper<'g, T>
```
acl-cqc, line 195: Do we really need this? If you've got the PetgraphWrapper then you've got the view, no? (TBH I'd expect to be given the HugrView to start off with, then construct the PetgraphWrapper as/when needed, i.e. only when passing off to petgraph methods that don't care about HugrView, but there)

src/hugr/views/petgraph.rs
```diff
@@ -1,187 +1,250 @@
...
-        {
-            fn node_count(&self) -> usize {
-                HugrView::node_count(self)
-            }
-        }
+impl<'a, T> PetgraphWrapper<'a, T>
+where
+    T: HugrView,
+{
+    /// Wrap a HugrView in a PetgraphWrapper.
```
acl-cqc, line 27: I guess this is good to make sure you don't construct useless PetgraphWrappers (containing other things). But, given PetgraphWrapper's only field is hidden, so they are not publically constructable - how about making the "public constructor" be an `fn as_petgraph(&self)` in <strike>HierarchyView/</strike>HugrView?

acl-cqc (commented):
Approve in principle. If it does what you need in TKET2, I think this is a much better way to go! :-)

src/hugr/views/petgraph.rs
acl-cqc: Note that if PetgraphWrapper implements HugrView then there is the possibility of nested PetgraphWrapper's, which I think we should probably avoid!?

aborgna-q: Yeah, this isn't necessary. I'll drop it.

src/hugr/views/petgraph.rs
aborgna-q: The bound here is just so we get a nice "Missing HugrView" error here instead of "Petgraph trait not implemented, here's some options".

`as_petgraph` would be handy 👍

acl-cqc: Given we have `as_petgraph` now, indeed on HugrView rather than on HierarchyView :-), do we still need this?

acl-cqc (approved):
Looks good to me! :)

PR #500, opened 2023-09-06
refactor: Separate hierarchy views

acl-cqc:
* Move views/hierarchy/petgraph.rs to views/petgraph.rs
* Move views/sibling.rs to views/sibling_subgraph.rs. I think the fact that it's a(n arbitrary) subgraph of a sibling subgraph is important, and more significant than that they are siblings (the USP, if you like - we have sibling views elsewhere), so emphasize this.
* Move HierarchyView from views/hierarchy to views
* Split hierarchy.rs into sibling.rs and descendants.rs


aborgna-q (approved):
LGTM. I'm assuming there are no changes apart from renaming and module docs.

acl-cqc:
> LGTM. I'm assuming there are no changes apart from renaming and module docs.

Indeed, sorry, should have said!

PR #501, opened 2023-09-06
RFC: Into #498: properly parametrize macro

acl-cqc:
I couldn't work out how to do the lifetimes in IntoNodeReferences so had to pull that out.

But, well, I think the objections here are pretty obvious...

The line count is pretty good though!!

### DIFF
--- a/src/hugr/views/petgraph.rs
+++ b/src/hugr/views/petgraph.rs
@@ -30,54 +30,6 @@ where
 {
 }
 
-impl pv::GraphBase for Hugr {
-    type NodeId = Node;
-    type EdgeId = ((Node, Port), (Node, Port));
-}
-
-impl pv::GraphProp for Hugr {
-    type EdgeType = petgraph::Directed;
-}
-
-impl pv::NodeCount for Hugr {
-    fn node_count(&self) -> usize {
-        HugrView::node_count(self)
-    }
-}
-
-impl pv::NodeIndexable for Hugr {
-    fn node_bound(&self) -> usize {
-        HugrView::node_count(self)
-    }
-
-    fn to_index(&self, ix: Self::NodeId) -> usize {
-        ix.index.into()
-    }
-
-    fn from_index(&self, ix: usize) -> Self::NodeId {
-        NodeIndex::new(ix).into()
-    }
-}
-
-impl pv::EdgeCount for Hugr {
-    fn edge_count(&self) -> usize {
-        HugrView::edge_count(self)
-    }
-}
-
-impl pv::Data for Hugr {
-    type NodeWeight = OpType;
-    type EdgeWeight = EdgeKind;
-}
-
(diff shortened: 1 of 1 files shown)


src/hugr/views/petgraph.rs
```diff
@@ -89,91 +41,46 @@ impl<'g> pv::IntoNodeReferences for &'g Hugr {
...
-impl PetgraphHugr for Hugr {}
-
 macro_rules! impl_region_petgraph_traits {
-    ($hugr:ident) => {
-        impl<'a, Root, Base> pv::GraphBase for $hugr<'a, Root, Base>
-        where
-            Root: NodeHandle,
-            Base: HugrInternals + HugrView,
+    ($hugr:ident $(< $($l:lifetime,)? $($v:ident: $b:tt $(+ $b2:tt)*),* >)? ) => {
+        impl $(<$($l,)? $($v: $b $(+ $b2)*),*>)? pv::GraphBase for $hugr$(<$($l,)? $($v),*>)?
```
aborgna-q, line 65: - Some generics may be unbounded. There should be an extra `$(: ...)?`

- `T<>` is valid syntax, so you could do
```rust
    ($hugr:ident) => {
        impl_region_petgraph_traits!{$hugr <>}
    };
    ($hugr:ident < $($l:lifetime,)? $($v:ident: $b:tt $(+ $b2:tt)*),* > ) => {
        ...
```
And then include the `'g` lifetime as needed between the always-present `<...>` for `IntoNodeReferences`.

aborgna-q:
In the balance between line-count vs code complexity, I can't say I object to this.

src/hugr/views/petgraph.rs
acl-cqc: * Done. It was *not* a goal to support all possible type parameterizations (only ones we actually needed), but this isn't too hard.
* Nice! Yes, so `Hugr<>` is valid, and so  is `<'g,>`, that simplifies things a little. I needed a `#[rustfmt::skip]` to prevent `cargo fmt` changing `Hugr<>` to `Hugr`... and it throws away the comment in `Hugr< /* don't throw these away */>`!!

* However, `IntoNodeReferences` is a more complex problem - you need to insert `'g` if `$l` didn't match a lifetime, but use `$l` if it does. I felt there was enough difference between the instances that parameterizing the macro over those differences was not helpful rather than just giving both instances explicitly...

acl-cqc:
> In the balance between line-count and code complexity, I can't say I object to this.

I have to say I am feeling more in favour of `struct PetGraphWrapper<T:HugrView>`....does that work? what do you think of that?

aborgna-q:
I'll merge this into #498 (with a small change to avoid the rustfmt problem).

I think the wrapper you propose may work (unless lifetimes get in the way), and it seems that it'd simplify the trait requirements in tket2... Let's continue the discussion over there.

PR #504, opened 2023-09-07
fix!: Reduce max int width in arithmetic extension to 64

mark-koch:
This is required for serialisation since msgpack doesn't support `i128` and `u128`: https://github.com/msgpack/msgpack/blob/master/spec.md#int-format-family

### DIFF
--- a/src/std_extensions/arithmetic/int_types.rs
+++ b/src/std_extensions/arithmetic/int_types.rs
@@ -44,7 +44,7 @@ const fn is_valid_log_width(n: u8) -> bool {
 }
 
 /// The smallest forbidden log width.
-pub const LOG_WIDTH_BOUND: u8 = 8;
+pub const LOG_WIDTH_BOUND: u8 = 7;
 
 /// Type parameter for the log width of the integer.
 // SAFETY: unsafe block should be ok as the value is definitely not zero.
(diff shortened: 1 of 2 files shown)


src/std_extensions/arithmetic/int_types.rs
```diff
@@ -106,15 +106,14 @@ impl ConstIntU {
...
+    pub fn new(log_width: u8, value: i64) -> Result<Self, ConstTypeError> {
         if !is_valid_log_width(log_width) {
             return Err(ConstTypeError::CustomCheckFail(
                 crate::types::CustomCheckFailure::Message("Invalid integer width.".to_owned()),
             ));
         }
         let width = 1u8 << log_width;
-        if (log_width <= 6) && (value >= (1i128 << (width - 1)) || value < -(1i128 << (width - 1)))
-        {
+        if (log_width <= 6) && (value >= (1i64 << (width - 1)) || value < -(1i64 << (width - 1))) {
```
cqc-alec, line 116:
```suggestion
        if (log_width <= 5) && (value >= (1i64 << (width - 1)) || value < -(1i64 << (width - 1))) {
```

cqc-alec (changes requested):

PR #505, opened 2023-09-07
refactor!: Use named struct for float constants

mark-koch:
This simplifies serialisation with pydantic


src/std_extensions/arithmetic/float_types.rs
```diff
@@ -23,20 +23,23 @@ pub const FLOAT64_TYPE: Type = Type::new_extension(FLOAT64_CUSTOM_TYPE);
 
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 /// A floating-point value.
-pub struct ConstF64(f64);
+pub struct ConstF64 {
+    #[allow(missing_docs)]
```
cqc-alec, line 27:
```suggestion
    /// The value.
```
May as well?

cqc-alec (approved):

PR #507, opened 2023-09-07
fix: HugrView::get_function_type

acl-cqc:
Previously this was implemented separately for Hugr (correctly, according to the doc), and for the other two HugrViews SiblingGraph/DescendantsGraph (by deferring to the get_function_type of the whole Hugr).

This is incorrect - the latter two should (surely?) return the function type of the portion of the Hugr which they view.

However, we can both fix this and delete code, by moving the impl-Hugr version to be the trait default, and removing the other two. See test added (previously both were returning None).

### DIFF
--- a/src/hugr/views.rs
+++ b/src/hugr/views.rs
@@ -179,7 +179,15 @@ pub trait HugrView: sealed::HugrInternals {
 
     /// For function-like HUGRs (DFG, FuncDefn, FuncDecl), report the function
     /// type. Otherwise return None.
-    fn get_function_type(&self) -> Option<&FunctionType>;
+    fn get_function_type(&self) -> Option<&FunctionType> {
+        let op = self.get_nodetype(self.root());
+        match &op.op {
+            OpType::DFG(DFG { signature })
+            | OpType::FuncDecl(FuncDecl { signature, .. })
+            | OpType::FuncDefn(FuncDefn { signature, .. }) => Some(signature),
+            _ => None,
+        }
+    }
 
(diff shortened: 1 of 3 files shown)


acl-cqc:
> This is incorrect - the latter two should (surely?) return the function type of the portion of the Hugr which they view.

I say "surely" here - there are no uses in the Hugr repo, except on whole Hugrs, to confirm one way or the other. The `get_function_type` method was introduced in #431 where it was used for whole Hugrs only, so there are other interpretations. But this is what seems most natural to me....

aborgna-q (approved):
Good catch!

lmondada (approved):
I definitely agree with you. The question of the semantics will come up again when/if we implement `HugrView` for `SiblingSubgraph`. In that case I am not sure what the right answer would be: the signature of the root or the signature of the subgraph input/output... Or maybe None?

acl-cqc:
> I definitely agree with you. The question of the semantics will come up again when/if we implement `HugrView` for `SiblingSubgraph`. In that case I am not sure what the right answer would be: the signature of the root or the signature of the subgraph input/output... Or maybe None?

Interesting question, indeed. (The first question here is, *do* we implement HugrView for SiblingSubgraph?!) A SiblingGraph necessarily includes both Input+Output, and parent - "the whole thing". Whereas a SiblingSubgraph doesn't include the parent, and might not include Input/Output. I think I know what I'd *like* to do, which would be to return the input/output parts of the "signature". But that might also be misleading, and just None might be more appropriate....

Issue #508, opened 2023-09-08
Integrate resolve_extension_ops into validate / infer_and_validate


acl-cqc:
Moreover, `resolve_extension_ops` was intended to use after loading the Hugr, but the checks in here that saved Signatures match the Signatures that would be computed, would make sense to do in validation. `validate` and `infer_and_validate` are also the main entry points for when you do have an ExtensionRegistry. In `validate` the Hugr is immutable but we could do this in `infer_and_validate` just prior to validation....

acl-cqc:
This was done in #603, renaming `infer_and_validate` to `update_validate`

PR #509, opened 2023-09-08
feat: Validate TypeArgs to ExtensionOp

acl-cqc:
The TypeArgs might contain malformed things like "the type List<5>" - we can only check that they make sense (i.e. that `5` is valid as an argument for List) *given the ExtensionRegistry*, and nowhere outside validation has the ExtensionRegistry, so we need to do this here.

Also rename `validate_children` (in ValidateOp) to `validate_op_children)`, and `validate_operation` to `validate_children`.

### DIFF
--- a/src/hugr/validate.rs
+++ b/src/hugr/validate.rs
@@ -169,7 +169,7 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
             }
         }
         // Secondly that the node has correct children
-        self.validate_operation(node, node_type)?;
+        self.validate_children(node, node_type)?;
 
         // If this is a container with I/O nodes, check that the extension they
         // define match the extensions of the container.
@@ -270,7 +270,7 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
     /// Check operation-specific constraints.
     ///
     /// These are flags defined for each operation type as an [`OpValidityFlags`] object.
-    fn validate_operation(&self, node: Node, node_type: &NodeType) -> Result<(), ValidationError> {
+    fn validate_children(&self, node: Node, node_type: &NodeType) -> Result<(), ValidationError> {
         let op_type = &node_type.op;
         let flags = op_type.validity_flags();
 
--- a/src/hugr/validate.rs
+++ b/src/hugr/validate.rs
@@ -17,7 +17,7 @@ use crate::extension::{
     validate::{ExtensionError, ExtensionValidator},
     ExtensionRegistry, ExtensionSolution, InferExtensionError,
 };
-use crate::ops::custom::ExternalOp;
+
 use crate::ops::validate::{ChildrenEdgeData, ChildrenValidationError, EdgeValidationError};
 use crate::ops::{OpTag, OpTrait, OpType, ValidateOp};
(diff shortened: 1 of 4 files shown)


src/hugr/validate.rs
```diff
@@ -158,7 +159,16 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
...
+        // Check operation-specific constraints. Firstly that type args are correct
+        // (note that if the Hugr were mutable here, this could be a good time to do
+        // resolve_extension_ops)
+        if let OpType::LeafOp(crate::ops::LeafOp::CustomOp(b)) = op_type {
+            for arg in b.args() {
+                arg.validate(self.extension_registry)
+                    .map_err(|cause| ValidationError::SignatureError { node, cause })?;
+            }
+        }
+        // Secondly that the node has correct children
```
acl-cqc, line 171: Methinks `validate_operation` is not a good name, given there's already other validation of the operation *here in validate_node*. Rather, validate_operation is all about the op's *children*. However `validate_children` is already taken (as a *subset* of those child-validating checks), so the best renaming I can see would be along the lines of
1. `validate_children` -> `validate_children_op_specific` *or something*
2. `validate_operation` -> `validate_children`

If anyone can think of a good name for the *or something* then I'm happy to proceed with that ;-)

src/hugr/validate.rs
```diff
@@ -158,7 +159,16 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
             }
         }
 
-        // Check operation-specific constraints
+        // Check operation-specific constraints. Firstly that type args are correct
+        // (note that if the Hugr were mutable here, this could be a good time to do
+        // resolve_extension_ops)
```
acl-cqc, line 164: This refers to #508 so maybe isn't a good comment to leave in the code...thoughts welcome here and/or on that issue as appropriate

aborgna-q: I'd say it's useful to keep the read-only validation from any inference/resolution/etc mutation.

In any case, I'd either remove this comment since we have the open Issue, or at least link to the issue from here.

src/hugr/validate.rs
aborgna-q: Uhm. I guess the `OpValidityFlags` are focused on a container's children.
I'd say rename that to `OpContainerFlags` (?) and `ValidateOp::validate_children` to `run_children_validator` ?
(on top of your 2. )

acl-cqc: Ok, gone with `validate_children` -> `validate_op_children` (my #1, your #2 - the only `validate_children` was that defined in ValidateOp), and `validate_operation` -> `validate_children`.

I've left OpValidityFlags because it does have one member that is not related to children - `non_df_port`. I think we were talking about refactoring non-dataflow-ports more widely, so if `non_df_port` goes as part of that refactor.... (as perhaps it should: `non_df_port` doesn't appear to be used *in validation*, but rather in `pub` functions of `OpType`: `other_port_count` and `port_index`....)

Happy to move these two renames into a separate PR if you like, I was going to when I thought there'd be 4 but I haven't bothered at the moment since it looks like there are only 2.

PR #510, opened 2023-09-08
fix: TODO in resolve_extension_ops: copy across input_extensions

acl-cqc:


croyzor (approved):
I don't think there's an easier way to construct nodes like this, but it comes up so often it should definitely be another constructor for `NodeType`

PR #511, opened 2023-09-08
refactor: allow NodeType::new to take any Into<Option<ExtensionSet>>

acl-cqc:
This might be the most extreme refactor/rename/etc. that answers [your comment](https://github.com/CQCL-DEV/hugr/pull/510#pullrequestreview-1617148177) - but, well, is there a reason *not* to do this? :)

### DIFF
--- a/src/builder/dataflow.rs
+++ b/src/builder/dataflow.rs
@@ -49,20 +49,14 @@ impl<T: AsMut<Hugr> + AsRef<Hugr>> DFGBuilder<T> {
         let output = ops::Output {
             types: signature.output().clone(),
         };
+        base.as_mut()
+            .add_node_with_parent(parent, NodeType::new(input, input_extensions.clone()))?;
         base.as_mut().add_node_with_parent(
             parent,
-            match &input_extensions {
-                None => NodeType::open_extensions(input),
-                Some(rs) => NodeType::new(input, rs.clone()),
-            },
-        )?;
-        base.as_mut().add_node_with_parent(
-            parent,
-            match input_extensions.map(|inp| inp.union(&signature.extension_reqs)) {
-                // TODO: Make this NodeType::open_extensions
-                None => NodeType::open_extensions(output),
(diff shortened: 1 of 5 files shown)


src/builder/dataflow.rs
```diff
@@ -49,20 +49,14 @@ impl<T: AsMut<Hugr> + AsRef<Hugr>> DFGBuilder<T> {
...
-        base.as_mut().add_node_with_parent(
-            parent,
-            match input_extensions.map(|inp| inp.union(&signature.extension_reqs)) {
-                // TODO: Make this NodeType::open_extensions
-                None => NodeType::open_extensions(output),
-                Some(rs) => NodeType::new(output, rs),
-            },
+            NodeType::new(
+                output,
+                input_extensions.map(|inp| inp.union(&signature.extension_reqs)),
```
acl-cqc, line 58: Note I've lost your comment here "TODO: make this NodeType::open_extensions". I wasn't sure what to make of that, given the line after it *did* use open_extensions...

src/hugr.rs
```diff
@@ -66,10 +66,10 @@ pub struct NodeType {
 
 impl NodeType {
     /// Create a new optype with some ExtensionSet
-    pub fn new(op: impl Into<OpType>, input_extensions: ExtensionSet) -> Self {
+    pub fn new(op: impl Into<OpType>, input_extensions: impl Into<Option<ExtensionSet>>) -> Self {
```
croyzor, line 69: I don't think the `Into` is necessary - I'm not sure what we would want to put here that implements the interface

croyzor (approved):

src/hugr.rs
acl-cqc: This is because the library provides a blanket implementation of `Into<Option<T>>` for all T (using `Some`). So using that with `T`==`ExtensionSet`, this generalizes the old `new`.

PR #519, opened 2023-09-12
Refactor: move apply_rewrite from Hugr to HugrMut

acl-cqc:
Given `apply_rewrite` just does a bounce dispatch to `Rewrite::apply`, rather than any actual work, this seems the right place to put it! (The alternative might be to drop `apply_rewrite` altogether but I think it's kinda nice and aids discovery)

### DIFF
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -12,7 +12,7 @@ use crate::{Hugr, Port};
 
 use self::sealed::HugrMutInternals;
 
-use super::NodeMetadata;
+use super::{NodeMetadata, Rewrite};
 
 /// Functions for low-level building of a HUGR.
 pub trait HugrMut: HugrView + HugrMutInternals {
@@ -143,6 +143,14 @@ pub trait HugrMut: HugrView + HugrMutInternals {
     fn insert_from_view(&mut self, root: Node, other: &impl HugrView) -> Result<Node, HugrError> {
         self.valid_node(root)?;
         self.hugr_mut().insert_from_view(root, other)
+    }
+
+    /// Applies a rewrite to the graph.
+    fn apply_rewrite<R, E>(&mut self, rw: impl Rewrite<ApplyResult = R, Error = E>) -> Result<R, E>
+    where
+        Self: Sized,
+    {
+        rw.apply(self)
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/hugr/hugrmut.rs
```diff
@@ -144,6 +144,14 @@ pub trait HugrMut: HugrView + HugrMutInternals {
         self.valid_node(root)?;
         self.hugr_mut().insert_from_view(root, other)
     }
+
+    /// Applies a rewrite to the graph.
+    fn apply_rewrite<R, E>(&mut self, rw: impl Rewrite<ApplyResult = R, Error = E>) -> Result<R, E>
+    where
+        Self: Sized,
+    {
+        rw.apply(self)
```
aborgna-q, line 153: In the other hugrmut methods we validate that we're not modifying the graph outside the view.
This isn't possible with the current `Rewrite` interface, so we cannot avoid rewriting arbitrary parts.

Is this something we want to allow?

aborgna-q: Although I guess if we require the user to assert that `rw.validate(self)` passes then this would be valid.

Would it make sense to have an `assert_debug!` with that?

acl-cqc: The idea is that `rw.apply(self)` will have to work through methods of the `impl HugrMut` and so the checking comes there - i.e., when the rewrite tries to modify the view.

(If we relied on checking here, then just calling `Rewrite::apply` would allow to bypass the check!!)

acl-cqc: But are you suggesting (skipping all the typing boilerplate for now):
```
fn apply(rw: impl Rewrite) {
   debug_assert!(rw.verify(self));
   rw.apply(self)
}
```
? I do quite like that idea, but right now a lot of our `Rewrite::verify` impl's are `unimplemented!`....

acl-cqc: I guess we could have an `apply_verified` as a shortcut? (not `debug_assert` just `assert`, then!)

aborgna-q: :+1:

Yeah, I was suggesting the `debug_assert` you mentioned. If it's unimplemented I'd leave it out (maybe a TODO?).

This is definitely not a place where we want `assert`s. So I'd just leave it bare and let the inner calls do the checks.

aborgna-q (approved):

PR #522, opened 2023-09-12
Add SiblingMut allowing modification of a SiblingGraph

acl-cqc:
* Combine `HugrView::get_io` methods - could pull that out into a preliminary PR, it's just tidying a wrapper function
* Also `HugrMut::add_op_with_parent` is just a wrapper around `add_node_with_parent` so make that a default-impl too
* SibingMut implementation parallels SiblingView for many methods, but not all, as does not its own PortGraph
* Reimplement OutlineCfg to use nested SiblingMut instances (escaping to &mut Hugr for `set_parent`)
* Refactor tests so we test calling OutlineCfg on a SiblingMut of only a portion of a Hugr

### DIFF
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -141,19 +141,19 @@ pub trait HugrMut: HugrView + HugrMutInternals {
     }
 
     /// Insert another hugr into this one, under a given root node.
-    ///
-    /// Returns the root node of the inserted hugr.
     #[inline]
-    fn insert_hugr(&mut self, root: Node, other: Hugr) -> Result<Node, HugrError> {
+    fn insert_hugr(&mut self, root: Node, other: Hugr) -> Result<InsertionResult, HugrError> {
         self.valid_node(root)?;
         self.hugr_mut().insert_hugr(root, other)
     }
 
     /// Copy another hugr into this one, under a given root node.
-    ///
-    /// Returns the root node of the inserted hugr.
     #[inline]
-    fn insert_from_view(&mut self, root: Node, other: &impl HugrView) -> Result<Node, HugrError> {
+    fn insert_from_view(
+        &mut self,
+        root: Node,
+        other: &impl HugrView,
+    ) -> Result<InsertionResult, HugrError> {
         self.valid_node(root)?;
         self.hugr_mut().insert_from_view(root, other)
     }
@@ -164,6 +164,26 @@ pub trait HugrMut: HugrView + HugrMutInternals {
         Self: Sized,
     {
         rw.apply(self)
+    }
+}
+
+/// Records the result of inserting a Hugr or view
+/// via [HugrMut::insert_hugr] or [HugrMut::insert_from_view]
+pub struct InsertionResult {
+    /// The node, after insertion, that was the root of the inserted Hugr.
+    /// (That is, the value in [InsertionResult::node_map] under the key that was the [HugrView::root]))
+    pub new_root: Node,
+    /// Map from nodes in the Hugr/view that was inserted, to their new
+    /// positions in the Hugr into which said was inserted.
+    pub node_map: HashMap<Node, Node>,
+}
+
+impl InsertionResult {
+    fn translating_indices(new_root: Node, node_map: HashMap<NodeIndex, NodeIndex>) -> Self {
+        Self {
+            new_root,
+            node_map: HashMap::from_iter(node_map.into_iter().map(|(k, v)| (k.into(), v.into()))),
+        }
     }
 }
 
@@ -249,7 +269,7 @@ where
         Ok((src_port, dst_port))
     }
 
-    fn insert_hugr(&mut self, root: Node, mut other: Hugr) -> Result<Node, HugrError> {
+    fn insert_hugr(&mut self, root: Node, mut other: Hugr) -> Result<InsertionResult, HugrError> {
         let (other_root, node_map) = insert_hugr_internal(self.as_mut(), root, &other)?;
         // Update the optypes and metadata, taking them from the other graph.
         for (&node, &new_node) in node_map.iter() {
@@ -258,10 +278,15 @@ where
             let meta = other.metadata.take(node);
             self.as_mut().set_metadata(node.into(), meta).unwrap();
         }
-        Ok(other_root)
+        debug_assert_eq!(Some(&other_root.index), node_map.get(&other.root().index));
+        Ok(InsertionResult::translating_indices(other_root, node_map))
     }
 
-    fn insert_from_view(&mut self, root: Node, other: &impl HugrView) -> Result<Node, HugrError> {
+    fn insert_from_view(
+        &mut self,
+        root: Node,
+        other: &impl HugrView,
+    ) -> Result<InsertionResult, HugrError> {
         let (other_root, node_map) = insert_hugr_internal(self.as_mut(), root, other)?;
         // Update the optypes and metadata, copying them from the other graph.
         for (&node, &new_node) in node_map.iter() {
@@ -272,7 +297,8 @@ where
                 .set_metadata(node.into(), meta.clone())
                 .unwrap();
         }
-        Ok(other_root)
+        debug_assert_eq!(Some(&other_root.index), node_map.get(&other.root().index));
+        Ok(InsertionResult::translating_indices(other_root, node_map))
     }
 }
 
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -37,8 +37,8 @@ pub trait HugrMut: HugrView + HugrMutInternals {
         parent: Node,
         op: impl Into<OpType>,
     ) -> Result<Node, HugrError> {
-        self.valid_node(parent)?;
-        self.hugr_mut().add_op_with_parent(parent, op)
+        // TODO: Default to `NodeType::open_extensions` once we can infer extensions
+        self.add_node_with_parent(parent, NodeType::pure(op))
     }
 
     /// Add a node to the graph with a parent in the hierarchy.
@@ -163,15 +163,6 @@ impl<T> HugrMut for T
 where
     T: HugrView + AsMut<Hugr>,
 {
-    fn add_op_with_parent(
-        &mut self,
-        parent: Node,
-        op: impl Into<OpType>,
-    ) -> Result<Node, HugrError> {
-        // TODO: Default to `NodeType::open_extensions` once we can infer extensions
-        self.add_node_with_parent(parent, NodeType::pure(op))
-    }
-
     fn add_node_with_parent(&mut self, parent: Node, node: NodeType) -> Result<Node, HugrError> {
         let node = self.as_mut().add_node(node);
         self.as_mut()
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -64,6 +64,15 @@ pub trait HugrMut: HugrView + HugrMutInternals {
         self.hugr_mut().add_op_before(sibling, op)
     }
 
+    /// A generalisation of [`HugrMut::add_op_before`], needed temporarily until
+    /// add_op type methods all default to creating nodes with open extensions.
+    /// See issue #424
+    #[inline]
+    fn add_node_before(&mut self, sibling: Node, nodetype: NodeType) -> Result<Node, HugrError> {
+        self.valid_non_root(sibling)?;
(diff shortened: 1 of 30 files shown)


src/hugr/rewrite/outline_cfg.rs
```diff
@@ -140,13 +143,16 @@ impl Rewrite for OutlineCfg {
         };
 
         // 3. Extract Cfg node created above (it moved when we called insert_hugr)
-        let cfg_node = h
+        // Support filtered Sibling-only views by explicitly descending into new_block
```
acl-cqc, line 146: This is to show the kind of thing I'm thinking of. We can't do everything this way - eventually we have to just break out to the `hugr_mut(): &mut Hugr` - but we can do *most* things remaining within one (nested) SiblingGraph at a time

src/hugr/hugrmut.rs
```diff
@@ -17,7 +17,9 @@ use super::{NodeMetadata, Rewrite};
 /// Functions for low-level building of a HUGR.
 pub trait HugrMut: HugrView + HugrMutInternals {
     /// Returns the metadata associated with a node.
-    fn get_metadata_mut(&mut self, node: Node) -> &mut NodeMetadata;
+    fn get_metadata_mut(&mut self, node: Node) -> &mut NodeMetadata {
+        self.hugr_mut().metadata.get_mut(node.index)
+    }
```
acl-cqc, line 22: Oh, we should do a `contains_node` check here

src/hugr/rewrite/outline_cfg.rs
```diff
@@ -227,19 +239,25 @@ mod test {
...
+    use crate::types::FunctionType;
+    use crate::{type_row, HugrView, Node};
     use cool_asserts::assert_matches;
     use itertools::Itertools;
 
     use super::{OutlineCfg, OutlineCfgError};
 
     fn depth(h: &impl HugrView, n: Node) -> u32 {
-        match h.get_parent(n) {
+        match h.base_hugr().get_parent(n) {
```
acl-cqc, line 260: This is a bit ugly (doing this on every call - alternative would be to do at every callsite, which is probably a bit better), but it kinda makes sense. (Better might be to calculate depth to the root of the *starting* HugrView, or something, but no way to do that)

src/hugr/views/sibling.rs
```diff
@@ -101,7 +103,7 @@ where
 
     #[inline]
     fn get_optype(&self, node: Node) -> &OpType {
-        self.hugr.get_optype(node)
+        self.base_hugr().get_optype(node)
```
acl-cqc, line 106: This is to avoid the double-filtering issue

acl-cqc: Now gone

src/hugr/rewrite/outline_cfg.rs
acl-cqc: However, worth noting that an alternative for these bits is just for `insert_hugr` to return the NodeIndex->NodeIndex map from `insert_hugr_internal`!....should I do that instead?

src/hugr/rewrite/outline_cfg.rs
```diff
@@ -163,31 +171,37 @@ impl Rewrite for OutlineCfg {
...
-        // 5. Children of new CFG.
-        // Entry node must be first
-        h.move_before_sibling(entry, inner_exit).unwrap();
-        // And remaining nodes
-        for n in self.blocks {
-            // Do not move the entry node, as we have already
-            if n != entry {
-                h.set_parent(n, cfg_node).unwrap();
+        {
+            // These operations do not fit into any SiblingView
```
acl-cqc, line 175: I guess I mean SiblingMut. So we'd need a full-on DescendantsMut for that. I think I'm happy to skip for now but I could probably be persuaded....

src/hugr/rewrite/outline_cfg.rs
acl-cqc: Yes, this will get significantly easier after #535. Maybe wait for that to go in first....

acl-cqc:
Ok have merged with #535 which makes things quite a bit simpler. @Agustín Borgna think this is ready now!

src/hugr/hugrmut.rs
```diff
@@ -37,8 +37,8 @@ pub trait HugrMut: HugrView + HugrMutInternals {
         parent: Node,
         op: impl Into<OpType>,
     ) -> Result<Node, HugrError> {
-        self.valid_node(parent)?;
-        self.hugr_mut().add_op_with_parent(parent, op)
+        // TODO: Default to `NodeType::open_extensions` once we can infer extensions
```
acl-cqc, line 40: (This comment just moved from the other `add_op_with_parent`)

src/hugr/rewrite/outline_cfg.rs
```diff
@@ -189,10 +179,36 @@ impl Rewrite for OutlineCfg {
...
+                // Do not move the entry node, as we have already
+                if n != entry {
+                    h.set_parent(n, cfg_node).unwrap();
+                }
+            }
+            inner_exit
+        };
+
+        // 4(b). Reconnect exit edge to the new exit node within the inner CFG
+        let mut in_bb_view: SiblingMut<'_, BasicBlockID> =
```
acl-cqc, line 204: I guess maybe I should comment why we're building these SiblingMut's (i.e. in case the HugrMut we're operating on is a SiblingMut itself), as they would not be necessary if the `h: impl HugrMut` was in fact a `&mut Hugr`

src/hugr/views/sibling.rs
```diff
@@ -215,6 +208,185 @@ where
...
+    fn all_node_ports(&self, node: Node) -> Self::NodePorts<'_> {
+        match self.contains_node(node) {
+            true => self.base_hugr().all_node_ports(node),
+            false => <FlatRegionGraph as PortView>::NodePortOffsets::default().map_into(),
+        }
+    }
+
+    fn linked_ports(&self, node: Node, port: Port) -> Self::PortLinks<'_> {
+        match self.contains_node(node) {
+            true => self.base_hugr().linked_ports(node, port),
```
aborgna-q, line 339: This won't filter external edges.

src/hugr/views/sibling.rs
```diff
@@ -215,6 +208,185 @@ where
...
+            false => EMPTY_GRAPH
+                .port_links(portgraph::PortIndex::new(1))
+                .with_context(self.base_hugr())
+                .map_with_context(|_port, _ctx| panic!("Never called")),
+        }
+    }
+
+    fn node_connections(&self, node: Node, other: Node) -> Self::NodeConnections<'_> {
+        match self.contains_node(node) && self.contains_node(other) {
+            true => self.base_hugr().node_connections(node, other),
```
aborgna-q, line 349: Same here, this isn't filtering the nodes.

src/hugr/views/sibling.rs
```diff
@@ -215,6 +208,185 @@ where
...
+    }
+
+    fn neighbours(&self, node: Node, dir: Direction) -> Self::Neighbours<'_> {
+        self.valid_node(node).unwrap(); // ?? Or return empty iterator?
+        self.base_hugr().neighbours(node, dir) // Or self.hugr ?
+    }
+
+    fn all_neighbours(&self, node: Node) -> Self::Neighbours<'_> {
+        self.valid_node(node).unwrap(); // ?? Or return empty iterator?
+        self.base_hugr().all_neighbours(node)
```
aborgna-q, line 378: I'd say return an empty one.

src/hugr/views/sibling.rs
aborgna-q, line 1: I wonder if `delegate!`ing all the `HugrView` impl to a `SiblingGraph` is feasible. In theory it should be free after all the layers are optimized away...
(there is no caching nor anything on a flat filtered graph, we only do that for descendants).

Have you tried that?

aborgna-q (commented):
Nice!

src/hugr/views/sibling.rs
acl-cqc: Not seriously. The only two strategies I had thought about before, are easily dismissed:
* Keep a SiblingGraph permanently in the SiblingMut. But that means there's an immutable borrow of the underlying Hugr sitting around, which means we can't mutate it, which is pretty useless.
* Construct a SiblingGraph on demand, call the method, then discard it. This is plausible for things like 'get_parent` but falls down for any of the iterators - we need (e.g.) `SiblingMut.nodes` to return the `nodes()` of the SiblingGraph, and the nodes refers to the SiblingGraph, which is a local variable deallocated as soon as `SiblingMut.nodes` returns. (Returning a reference to temporary variable). So that doesn't fly.

However, thinking a bit more, what I'd like to do is declare a `struct SGHolder<T>` which holds (owns) a SiblingGraph, and a T *which can refer to that SiblingGraph*; and then `impl...Iterator for SGHolder<T> where T: Iterator`. However, it doesn't seem there's any way to have a struct like that - e.g. [here](https://users.rust-lang.org/t/how-to-create-a-struct-with-fields-which-reference-each-other/78260/2) or many others.

So...I think the only option remaining is to change the `nodes()` method of `HugrView` itself to return not a `Self::Nodes<'_>` but a `Self::Nodes<'t>` for some longer lifetime `'t` - this might have to be a parameter of the `nodes()` method (`where Self: 't`), and we might need a max constraint from the other direction as part of HugrView (i.e. `trait HugrView<'g> { .... fn<'t> nodes(&self) -> Self::Nodes<'t> where 't: '_, 'g: 't`, I think). Then `nodes()` can return an iterator that references the Hugr inside the view, but not the view itself. However this may lead to other problems... (shout if you want but otherwise I'll see where I get to.)

acl-cqc: No, the latter would work if `SiblingGraph` methods were returning things that referenced the underlying `Hugr`, but not the portgraph, as that's part of SiblingGraph (and gets deallocated when the SiblingMut methods return)...

So we'd have to store a portgraph in the SiblingMut? I.e. keep it up-to-date following each mutation (unless we change HugrView methods to take `&mut self`). I don't think we want to do that, but thoughts?

I also see suggestions that we change other classes (=~= the various iterators) to use `Arc`/`Rc`s rather than `&`s, but given we *want* to prevent mutation of the SiblingMut while the iterators are live, I don't see how that works.

I'll look at the various refs-between-fields-of-a-struct crates linked above, tho I think these are basically wrappers over `unsafe`....

acl-cqc: Ok, `ouroboros` allows the intra-field-refs OK, but still seems to struggle with our fancy lifetime-parameterized trait members (like `type Nodes<'a>`) - see branch `sibling-mut/ouroboros`. Starting to feel like I'm out of ideas here...

acl-cqc: So I tried parameterizing the HugrView *and* using Ouroboros there, to add a `nodes2()` that returns an iterator outliving the view itself...but still got stuck. I think the fundamental issue is that with Ouroboros, you define a struct holding a Foo and a Bar that "borrows" the Foo; but when you create the struct, you provide a function to create the Bar given an `&Foo` but the types we are using require the Bar returned by that function to outlive the `&Foo` reference (passed to the function). Possibly we could get around this by going right the way into portgraph and adding ouroboros-type magic there, but I don't think I'm going there.

Fundamentally, we want some way to convert, say, a `portgraph::Children<'a>` into a `Children<'static>`, i.e. copying the extra data.

But I think there is a much easier way to do that - copy the iterated items, rather than the data over which the iterator runs. I'll give that a go here....

src/hugr/views/sibling.rs
acl-cqc: Ooops, missed those when I fixed a bunch of other `unwrap`s. Done, thanks! :)

src/hugr/views/sibling.rs
```diff
@@ -41,24 +43,63 @@ pub struct SiblingGraph<'g, Root = Node> {
     _phantom: std::marker::PhantomData<Root>,
 }
 
+/// HugrView trait members common to both [SiblingGraph] and [SiblingMut],
+/// i.e. that rely only on [HugrInternals::base_hugr]
+macro_rules! base_members {
```
aborgna-q, line 48: Uhm, this is iffy but I guess it's ok.
I'd just rename it to `impl_base_members` to convey a bit more info at the callsites.

aborgna-q (approved):
Yeah, this is a non-trivial problem :P

This look alright as a first implementation. I say open a `P-low` `perf` issue to investigate the performance here.

PR #527, opened 2023-09-12
feat: Const nodes are built with some extension requirements

### LINKED ISSUE #525: Extension inference for Const nodes

croyzor:
When using builder methods to put `Const` nodes on the graph, users must pass an extra "provenance" argument which specifies which extension the type of the constant value is coming from. The argument is an `Option`, where `None` means that the constant is of a built-in type (like our true and false predicates).

The `Const` nodes will be added to the graph with `input_extensions` which are either the empty set or the singleton set of the const value's provenance. Constant loading methods in the builder will make the `LoadConst` node's `input_extensions` match those of the constant it's connected too.

This also adds a constraint to extension inference that extension requirements are the same on both ends of `Static` and `ControlFlow` edges.

Resolves #525

### DIFF
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -62,10 +62,7 @@ pub trait Container {
         Ok(Wire::new(src, src_port))
     }
 
-    /// Add a constant value to the container and return a handle to it. The
-    /// `provenance` argument says which extension the type and value of the
-    /// constant comes from. If it is `None`, then the constant must be a built
-    /// in type.
+    /// Add a constant value to the container and return a handle to it.
     ///
     /// # Errors
     ///
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -17,7 +17,7 @@ use crate::{
     types::EdgeKind,
 };
 
-use crate::extension::{ExtensionRegistry, ExtensionSet, PRELUDE_REGISTRY};
+use crate::extension::{ExtensionId, ExtensionRegistry, ExtensionSet, PRELUDE_REGISTRY};
 use crate::types::{FunctionType, Signature, Type, TypeRow};
 
 use itertools::Itertools;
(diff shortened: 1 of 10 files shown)


src/extension/infer.rs
```diff
@@ -323,10 +323,11 @@ impl UnificationContext {
...
             let sig: &OpType = hugr.get_nodetype(tgt_node).into();
             // Incoming ports with a dataflow edge
-            for port in hugr
-                .node_inputs(tgt_node)
-                .filter(|src_port| matches!(sig.port_kind(*src_port), Some(EdgeKind::Value(_))))
-            {
+            for port in hugr.node_inputs(tgt_node).filter(|src_port| {
+                matches!(sig.port_kind(*src_port), Some(EdgeKind::Value(_)))
+                    || matches!(sig.port_kind(*src_port), Some(EdgeKind::Static(_)))
+                    || matches!(sig.port_kind(*src_port), Some(EdgeKind::ControlFlow))
```
croyzor, line 329: Adding `ControlFlow` edges as a generator of constraints is kind of irrelevant to this PR, just a drive-by

acl-cqc:
Just a thought @croyzor but couldn't a constant require multiple extensions. For example, a List (that requires the collections extension) of Floats (requiring the arithmetic extension)....you get the idea.

Now, does an empty list of Floats require the arithmetic extension too? Ymmv....

Seems to me it ought to be possible to figure the extensions out from the `Value` rather than passing it separately, too.

src/extension/infer.rs
acl-cqc: I'd strongly suggest moving this into #529. It makes that more self-contained, more obviously complete (!), and it's really not relevant here

acl-cqc: Also, can you do `matches!(sig.port_kind(*src_port), Some(EdgeKind::Value(_)) | Some(EdgeKind::Static(_)) | Some(EdgeKind:ControlFlow))`? That's possible in normal patterns (outside `matches!`), so maybe it works there too...

croyzor: Removed the ControlFlow match

src/hugr/hugrmut.rs
```diff
@@ -60,6 +60,14 @@ pub trait HugrMut: HugrView + HugrMutInternals {
         self.hugr_mut().add_op_before(sibling, op)
     }
 
+    /// A generalisation of [`HugrMut::add_op_before`], needed temporarily until
+    /// add_op type methods all default to creating nodes with open extensions.
```
cqc-alec, line 64: Is there a github issue for this we could reference?

src/builder/build_traits.rs
```diff
@@ -62,14 +62,21 @@ pub trait Container {
         Ok(Wire::new(src, src_port))
     }
 
-    /// Add a constant value to the container and return a handle to it.
+    /// Add a constant value to the container and return a handle to it. The
+    /// `provenance` argument says which extension the type and value of the
```
cqc-alec, line 66: No `provenance` argument...

src/hugr/hugrmut.rs
croyzor: #424 - I'll add it to the comment

src/builder/build_traits.rs
croyzor: Removed

cqc-alec (approved):

PR #528, opened 2023-09-12
fix: Lessen requirements for hugrs in outline_cfg

croyzor:
* Avoid calling validate on the subgraph in outline cfg while it isn't valid (because it has no entry node yet)
* Loosen constraints on tests for rewrites such that we're happy if they pass validation after their extensions are inferred


aborgna-q (approved):
LGTM

acl-cqc:
The alternative here would be to copy the requirements across from the existing nodes? That must be possible, no? We shouldn't *need* to redo inference after the outlining operation, we haven't really changed anything....

PR #529, opened 2023-09-12
feat: Extension inference for CFGs

### LINKED ISSUE #497: Extension inference for control flow graphs
Blocks #473

### LINKED ISSUE #526: Extension inference for CFGs

croyzor:
Based off of #527. Add constraints for control flow graphs and their basic blocks, and a test of a CFG which adds resources in its basic blocks.
Resolves #497


src/extension/infer.rs
```diff
@@ -302,6 +302,18 @@ impl UnificationContext {
...
             }
 
+            if node_type.tag() == OpTag::Cfg {
+                let mut children = hugr.children(node);
+                let entry = children.next().unwrap();
+                let exit = children.next().unwrap();
+                let m_entry = self.make_or_get_meta(entry, Direction::Incoming);
+                let m_exit_in = self.make_or_get_meta(exit, Direction::Incoming);
+                let m_exit_out = self.make_or_get_meta(exit, Direction::Outgoing);
+                self.add_constraint(m_input, Constraint::Equal(m_entry));
```
acl-cqc, line 312: It seems a bit odd that you deal with both Direction::Incoming and Direction::Outgoing for the exit node, but not for the entry node. Do you mean to do similarly for `self.make_or_get_meta(entry, Direction::Outgoing)` too?

If so, how about commoning up with the case for nested dataflow nodes above, i.e. breaking this block out into a helper function:
```
for dir in Direction::BOTH {
                    let m_input_node = self.make_or_get_meta(input, dir);
                    self.add_constraint(m_input_node, Constraint::Equal(m_input));
                    let m_output_node = self.make_or_get_meta(output, dir);
                    self.add_constraint(m_output_node, Constraint::Equal(m_output));
                }
```

src/extension/infer.rs
```diff
@@ -1092,4 +1104,183 @@ mod test {
...
+        let a = ExtensionSet::singleton(&A);
+        let abc = ExtensionSet::from_iter([A, B, C]);
+        let bc = ExtensionSet::from_iter([B, C]);
+        let b = ExtensionSet::singleton(&B);
+        let c = ExtensionSet::singleton(&C);
+
+        let oneway = vec![Type::new_predicate([type_row![NAT]])];
+        let twoway = vec![Type::new_predicate([type_row![NAT], type_row![NAT]])];
+
+        let mut hugr = Hugr::new(NodeType::open_extensions(ops::CFG {
```
acl-cqc, line 1147: driveby, but we should probably use FunctionType here (inside CFG), too

src/extension/infer.rs
croyzor: The exit node of a CFG can never add extension requirements between its inputs and outputs, but an entry node is an arbitrary DFB with an extension_delta, so the output extensions aren't necessarily the same as the CFG's input extensions.

Having said that, we _can_ add a constraint that the entry node's outputs are the CFG input plus that node's delta

src/extension/infer.rs
```diff
@@ -1092,4 +1104,183 @@ mod test {
...
+    ///          |  (Adds [A]) |
+    ///          +-/---------\-+
+    ///           /           \
+    ///  +-------/-----+     +-\-------------+
+    ///  |     BB0     |     |      BB1      |
+    ///  | (Adds [BC]) |     |   (Adds [B])  |
+    ///  +----\--------+     +---/------\----+
+    ///        \                /        \
+    ///         \              /          \
+    ///          \       +----/-------+  +-\---------+
```
acl-cqc, line 1128: I think you could make this a slightly more interesting test by linking BB0 to BB10 (rather than to exit)....but,

I think you could make this a *lot* more interesting by linking, say, `BB10`, *back* to `BB1` or, even better, `Entry` to make a loop. I guess the hope is that unification will deal with loops no problem, without requiring iteration to a fixpoint (say), but it'd be great to have the test of that :)

src/builder/cfg.rs
acl-cqc, line 1: I'd drop changes to this file, you've not got anything else here

src/extension/infer.rs
```diff
@@ -1092,4 +1104,183 @@ mod test {
...
+
+        let mkpred = hugr.add_node_with_parent(
+            entry,
+            NodeType::open_extensions(make_opaque(
+                A,
+                FunctionType::new(vec![NAT], twoway.clone()).with_extension_delta(&a),
+            )),
+        )?;
+
+        let [bb0, bb0_in, bb0_out] = create_with_io(
```
acl-cqc, line 1181: You've got four lots here of `create_with_io` + `add_node_with_parent` + 2*`connect` ("internal wiring"). I'd say that was worth making a helper function or lambda, e.g. inside this test. (It can probably create the BasicBlock::DFB as well.)

src/extension/infer.rs
```diff
@@ -1092,4 +1104,183 @@ mod test {
...
+        hugr.connect(entry, 0, bb0, 0)?;
+        hugr.connect(entry, 1, bb1, 0)?;
+        hugr.connect(bb1, 0, bb10, 0)?;
+        hugr.connect(bb1, 1, bb11, 0)?;
+
+        hugr.connect(bb0, 0, exit, 0)?;
+        hugr.connect(bb10, 0, exit, 0)?;
+        hugr.connect(bb11, 0, exit, 0)?;
+
+        hugr.infer_extensions()?;
```
acl-cqc, line 1282: The idea is that you've fixed them, so this also checks consistency with the extensions specified already?

acl-cqc (commented):
Looks good, thanks - I presume you've already handled the case of ControlFlow edges?

croyzor:
> I presume you've already handled the case of ControlFlow edges?

Yes, this is done as a drive-by in #527

src/extension/infer.rs
acl-cqc: Ah yes sorry, I forget that Entry could have anything in it. Hmmm. So that case is handled elsewhere. Does the op_signature for an Output node not specify that it adds no delta, then? (i.e. do we need to make a constraint for Exit node output extensions here at all?)

croyzor: Yes you're right the output constraint is probably redundant

acl-cqc: Ok in that case I suggest you do either *one* of the following two options:
* Don't add output constraint for either node. *Ideally* I think you'd make the same change to the `if let Some([input, output]) = hugr.get_io(node)` block above too, and perhaps elsewhere as a general principle, tho, and that might be better done in a later PR. In which case:
* Just add a comment here as to why entry and exit are different. (I should know, I wrote the spec for them, but having not worked on CFGs for awhile I forgot, and so others will too - it is an oddity of the spec that they are not symmetric, so I think that justifies a comment.)

src/algorithm/nest_cfgs.rs
```diff
@@ -428,8 +428,8 @@ pub(crate) mod test {
         //               \-> right -/             \-<--<-/
         let mut cfg_builder = CFGBuilder::new(FunctionType::new(type_row![NAT], type_row![NAT]))?;
 
-        let pred_const = cfg_builder.add_constant(Const::simple_predicate(0, 2))?; // Nothing here cares which
-        let const_unit = cfg_builder.add_constant(Const::simple_unary_predicate())?;
+        let pred_const = cfg_builder.add_constant(Const::simple_predicate(0, 2), None)?; // Nothing here cares which
```
acl-cqc, line 431: Good to update the comment here - "Nothing here cares which *branch*" to be clear that the comment isn't about the extensions. (However, I think this should not actually be showing up in this PR, so this might disappear post-merge)

src/algorithm/nest_cfgs.rs
```diff
@@ -675,8 +675,8 @@ pub(crate) mod test {
         separate: bool,
     ) -> Result<(Hugr, BasicBlockID, BasicBlockID), BuildError> {
         let mut cfg_builder = CFGBuilder::new(FunctionType::new(type_row![NAT], type_row![NAT]))?;
-        let pred_const = cfg_builder.add_constant(Const::simple_predicate(0, 2))?; // Nothing here cares which
-        let const_unit = cfg_builder.add_constant(Const::simple_unary_predicate())?;
+        let pred_const = cfg_builder.add_constant(Const::simple_predicate(0, 2), None)?; // Nothing here cares which
```
acl-cqc, line 678: ditto "Nothing here cares which branch"

src/builder/build_traits.rs
```diff
@@ -62,14 +62,25 @@ pub trait Container {
...
     /// # Errors
     ///
     /// This function will return an error if there is an error in adding the
     /// [`OpType::Const`] node.
-    fn add_constant(&mut self, constant: ops::Const) -> Result<ConstID, BuildError> {
-        let const_n = self.add_child_op(constant)?;
+    fn add_constant(
+        &mut self,
+        constant: ops::Const,
+        provenance: Option<ExtensionId>,
```
acl-cqc, line 77: Didn't we say 'provenance' should not be an Id but a Set? I'm not sure whether an Option<ExtensionSet> or just an ExtensionSet, depending on whether None means "no extensions" (=empty ExtensionSet) or "unknown extensions" (different from an empty ExtensionSet)...

acl-cqc: Oh, again this might be the wrong PR

src/builder/build_traits.rs
```diff
@@ -190,7 +201,7 @@ pub trait Dataflow: Container {
         op: impl Into<OpType>,
         input_wires: impl IntoIterator<Item = Wire>,
     ) -> Result<BuildHandle<DataflowOpID>, BuildError> {
-        self.add_dataflow_node(NodeType::pure(op), input_wires)
+        self.add_dataflow_node(NodeType::open_extensions(op), input_wires)
```
acl-cqc, line 204: I'm gonna stop commenting about things like this that look a bit out of date until you've done some merging/rebasing/something....

src/extension/infer.rs
```diff
@@ -323,10 +335,11 @@ impl UnificationContext {
         for tgt_node in hugr.nodes() {
             let sig: &OpType = hugr.get_nodetype(tgt_node).into();
             // Incoming ports with a dataflow edge
-            for port in hugr
-                .node_inputs(tgt_node)
-                .filter(|src_port| matches!(sig.port_kind(*src_port), Some(EdgeKind::Value(_))))
-            {
+            for port in hugr.node_inputs(tgt_node).filter(|src_port| {
+                matches!(sig.port_kind(*src_port), Some(EdgeKind::Value(_)))
```
acl-cqc, line 339: Yes, you *can* combine these into a single `matches!(sig.port_kind(*src_port), Some(EdgeKind::Value(_)) | Some(EdgeKind::Static(_)) | Some(EdgeKind::ControlFlow))`, I checked...

src/extension/infer.rs
acl-cqc: Really I think that would be better as a separate PR but I might be ok...

src/extension/infer.rs
```diff
@@ -324,8 +334,12 @@ impl UnificationContext {
             let sig: &OpType = hugr.get_nodetype(tgt_node).into();
             // Incoming ports with a dataflow edge
             for port in hugr.node_inputs(tgt_node).filter(|src_port| {
-                matches!(sig.port_kind(*src_port), Some(EdgeKind::Value(_)))
-                    || matches!(sig.port_kind(*src_port), Some(EdgeKind::Static(_)))
+                matches!(
+                    sig.port_kind(*src_port),
+                    Some(EdgeKind::Value(_))
+                        | Some(EdgeKind::Static(_))
+                        | Some(EdgeKind::ControlFlow)
```
acl-cqc, line 341: I think this neat little one-line change is doing most of the work!?

Which is cool. But can you
1. update the comment two lines above: "Incoming ports with a dataflow edge" doesn't seem right anymore.
2. (super-nit) 10 lines above, "Seperate loop so that...." -> "Separate loop so that". The former is an adjective, the latter a verb...

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+        let opaque =
+            ops::custom::OpaqueOp::new(extension.into(), "", "".into(), vec![], Some(signature));
+        ops::custom::ExternalOp::from(opaque).into()
+    }
+
+    fn make_block(
+        hugr: &mut Hugr,
+        bb_parent: Node,
+        dfb: impl Into<OpType>,
+        dfb_sig: FunctionType,
```
acl-cqc, line 1177: I note the following in `impl Case`: https://github.com/CQCL-DEV/hugr/blob/554d658f7afd38f0eb567ceff0599e65e74541c9/src/ops/controlflow.rs#L237-L238

You could add something similar in `controlflow.rs` for `impl BasicBlock` (I guess it would have to return an Option to allow for BasicBlock::Exit), and then drop the `dfg_sig` here, making `dfb` be an `impl Into<BasicBlock>`. (More thorough still would be to turn the contents of BasicBlock::DFB into a separate struct and impl for *that*, but I'd be happy not to go as far as that in this PR)

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+                extension_delta: c.clone(),
+            },
+            FunctionType::new(vec![NAT], oneway(NAT)),
+            make_opaque(
+                C,
+                FunctionType::new(vec![NAT], oneway(NAT)).with_extension_delta(&c),
+            ),
+        )?;
+
+        let bb11 = make_block(
```
acl-cqc, line 1309:
```suggestion
        let bb11 = bb10.clone();
```

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+                extension_delta: c.clone(),
+            },
+            FunctionType::new(vec![NAT], oneway(NAT)),
+            make_opaque(
+                C,
+                FunctionType::new(vec![NAT], oneway(NAT)).with_extension_delta(&c),
+            ),
+        )?;
+
+        // Internal wiring for DFGs
```
acl-cqc, line 1325: Move this up to where `entry`(_in,_out) and `mkpred` are declared

acl-cqc: (Like you are doing with `let entry =` in the `multi_entry` test below, tho I won't insist on hiding all these in a nested scope)

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+        )?;
+        let entry = {
+            let dfb = ops::BasicBlock::DFB {
+                inputs: type_row![NAT],
+                other_outputs: type_row![],
+                predicate_variants: vec![type_row![NAT], type_row![NAT]],
+                extension_delta: ExtensionSet::new(),
+            };
+
+            let entry = hugr.add_node_before(exit, NodeType::open_extensions(dfb))?;
```
acl-cqc, line 1364: Is this not `let (entry, entry_in, entry_out) = create_with_io(.....)` ? (If you do `entry` before `exit` maybe)

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+                inputs: type_row![NAT],
+                other_outputs: type_row![],
+                predicate_variants: vec![type_row![NAT]],
+                extension_delta: ExtensionSet::new(),
+            },
+            FunctionType::new(vec![NAT], oneway(NAT)),
+            make_opaque(PRELUDE_ID, FunctionType::new(vec![NAT], oneway(NAT))),
+        )?;
+
+        let bb1 = make_block(
```
acl-cqc, line 1407:
```suggestion
        let bb1 = bb0.clone()
```

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+                other_outputs: type_row![],
+                predicate_variants: vec![type_row![NAT]],
+                extension_delta: ExtensionSet::new(),
+            },
+            FunctionType::new(vec![NAT], oneway(NAT)),
+            make_opaque(PRELUDE_ID, FunctionType::new(vec![NAT], oneway(NAT))),
+        )?;
+
+        // Mult
+        let bb2 = make_block(
```
acl-cqc, line 1421:
```suggestion
        let bb2 = bb0.clone();
```

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+        hugr.connect(bb10, 0, exit, 0)?;
+        hugr.connect(bb11, 0, exit, 0)?;
+
+        hugr.infer_extensions()?;
+
+        Ok(())
+    }
+
+    #[test]
+    fn multi_entry() -> Result<(), Box<dyn Error>> {
```
acl-cqc, line 1345: No ascii art here, but this looks like entry -> (if(P) bb0; else bb1;) -> bb2 -> exit, which seems like a strictly simpler/easier case than the above. I guess that's OK, ascii art is nice tho (even if just a three-line comment), and why "multi_entry" ??

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+            let dfb: OpType = ops::BasicBlock::DFB {
+                inputs: type_row![NAT],
+                other_outputs: type_row![],
+                predicate_variants: vec![type_row![NAT]],
+                extension_delta: entry_ext.clone(),
+            }
+            .into();
+
+            let entry = hugr.add_node_before(exit, NodeType::open_extensions(dfb))?;
+            let entry_in = hugr.add_node_with_parent(
```
acl-cqc, line 1493: again this looks like `let (entry, entry_in, entry_out) = create_with_io(.....)` ?

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+        hugr.connect(bb2, 0, entry, 0)?;
+
+        Ok(hugr)
+    }
+
+    #[test]
+    fn test_cfg_loops() -> Result<(), Box<dyn Error>> {
+        let just_a = ExtensionSet::singleton(&A);
+        let variants = vec![
+            (
```
acl-cqc, line 1563: You could add a couple more here - with `just_a` repeated, and one with two blocks that do something. I mean, since you've set up this nifty framework already, I mean :-)

src/extension/infer.rs
acl-cqc: Or just make `make_block` here take a TypeRow for inputs and a Vec<TypeRow> for predicate, and do the rest here

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+                inputs: type_row![NAT],
+                other_outputs: type_row![],
+                predicate_variants: vec![type_row![NAT]],
+                extension_delta: ExtensionSet::new(),
+            },
+            FunctionType::new(vec![NAT], oneway(NAT)),
+            make_opaque(PRELUDE_ID, FunctionType::new(vec![NAT], oneway(NAT))),
+        )?;
+
+        // Mult
```
acl-cqc, line 1420: ??

acl-cqc: Multiple predecessors?

src/extension/infer.rs
```diff
@@ -1133,6 +1160,502 @@ mod test {
...
+        let root = hugr.root();
+
+        let exit = hugr.add_node_with_parent(
+            root,
+            NodeType::open_extensions(ops::BasicBlock::Exit {
+                cfg_outputs: type_row![NAT],
+            }),
+        )?;
+
+        let entry = {
```
acl-cqc, line 1604: One more create_with_io ? Or, better, why not make `make_block` return all three blocks

acl-cqc (approved):
A few ideas here and there, but this is great, thank you - there's nothing unexpected, mostly things just work and look as it feels they should, and most of the PR is a great set of tests :-). (So most comments are on the tests!)

src/extension/infer.rs
croyzor: Not quite, because `make_block` is adding a bunch of nodes to the graph, but I can certainly abstract this a bit!

src/extension/infer.rs
croyzor: Hm, I thought about this but opted against it because defining entry before exit like that is taking advantage of implementation details and using `add_node_before` is more "the right thing"

src/extension/infer.rs
croyzor: Here (and below) `create_with_io` is avoided because the dataflow parent needs to be added with `add_node_before` rather than just `add_node_with_parent`

src/extension/infer.rs
acl-cqc: ah yes, good point, `clone`ing an op doesn't clone its children, sorry!

src/extension/infer.rs
acl-cqc: Sorry, I am barking up the wrong tree here. This is nothing like `create_with_io` - that makes an Input and Output (dataflow) nodes, this makes an entry (BasicBlock::DFB) and exit (BasicBlock::Exit) node.

That said, given you do this a lot across three or four tests, maybe you could just have a helper `create_with_entry_exit`....

PR #530, opened 2023-09-12
feat: Derive Debug & Clone for `ExtensionRegistry`.

aborgna-q:


acl-cqc (approved):
Yeah, OK. Normal use should not involve cloning (they're big, that'd be very expensive), but I can see we might want to in e.g. tests and there's no real reason not to allow it.

PR #531, opened 2023-09-12
HugrMut: validate nodes for set_metadata/get_metadata_mut, too

acl-cqc:
...bringing these in line with the other methods. Yes, this involves a `Result<&mut, ...>` but that works ok!

Also inline `HugrMutInternals::add_node` and `add_op`, these are only used once and it's fairly trivial (but against, restores uniformity)

### DIFF
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -115,12 +115,16 @@ pub trait Container {
     /// Add metadata to the container node.
     fn set_metadata(&mut self, meta: NodeMetadata) {
         let parent = self.container_node();
+        // Implementor's container_node() should be a valid node
         self.hugr_mut().set_metadata(parent, meta).unwrap();
     }
 
     /// Add metadata to a child node.
+    ///
+    /// Returns an error if the specified `child` is not a child of this container
     fn set_child_metadata(&mut self, child: Node, meta: NodeMetadata) -> Result<(), BuildError> {
-        Ok(self.hugr_mut().set_metadata(child, meta)?)
+        self.hugr_mut().set_metadata(child, meta)?;
+        Ok(())
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/builder/build_traits.rs
```diff
@@ -115,12 +115,12 @@ pub trait Container {
...
         let parent = self.container_node();
-        self.hugr_mut().set_metadata(parent, meta);
+        self.hugr_mut().set_metadata(parent, meta).unwrap();
     }
 
     /// Add metadata to a child node.
-    fn set_child_metadata(&mut self, child: Node, meta: NodeMetadata) {
-        self.hugr_mut().set_metadata(child, meta);
+    fn set_child_metadata(&mut self, child: Node, meta: NodeMetadata) -> Result<(), BuildError> {
+        Ok(self.hugr_mut().set_metadata(child, meta)?)
```
aborgna-q, line 123:
```suggestion
    fn set_child_metadata(&mut self, child: Node, meta: NodeMetadata) -> Result<(), BuildError> {
        self.hugr_mut().set_metadata(child, meta)?;
        Ok()
```

src/builder/build_traits.rs
```diff
@@ -115,12 +115,12 @@ pub trait Container {
     /// Add metadata to the container node.
     fn set_metadata(&mut self, meta: NodeMetadata) {
         let parent = self.container_node();
-        self.hugr_mut().set_metadata(parent, meta);
+        self.hugr_mut().set_metadata(parent, meta).unwrap();
     }
 
     /// Add metadata to a child node.
```
aborgna-q, line 121: Mention why this is failable while `set_metadata` isn't ?

aborgna-q (approved):

PR #533, opened 2023-09-13
feat: Some python errors and bindings

aborgna-q:
Adds multiple custom python errors, and implements `Into<PyErr>` using them.
Unfortunately, we have to name the `pyrs` python module name here due to how pyo3 works.

Also, adds a python wrapper for `EdgeKind`. We don't export any python methods yet, just the class definitions.


src/types/type_row.rs
```diff
@@ -11,9 +11,12 @@ use super::Type;
 use crate::utils::display_list;
 use delegate::delegate;
 
+#[cfg(feature = "pyo3")]
+use pyo3::prelude::*;
```
croyzor, line 15: Why are we always importing all of the pyo3 prelude when we seem to just be using `pyclass`?

src/types.rs
```diff
@@ -46,6 +49,12 @@ impl EdgeKind {
     }
 }
 
+/// Python representation for [`EdgeKind`], the kinds of edges in a HUGR.
+#[cfg_attr(feature = "pyo3", pyclass)]
+#[repr(transparent)]
+#[derive(Clone, PartialEq, Eq, Debug, From, Into)]
+pub struct PyEdgeKind(EdgeKind);
```
croyzor, line 56: Why can't we just use `EdgeKind` for this?

aborgna-q: Because pyo3 doesn't like enums with non-unit variants. Wrapping it in a struct we hide the variants so it works alright (we'll need some python-level methods to access the internals later).

croyzor (approved):

PR #535, opened 2023-09-14
feat: insert_hugr/insert_view return node map

acl-cqc:
Also corresponding builder methods. `insert_hugr_internal` already returns this, so just wrap in a new `struct InsertionResult` translating the NodeIndex's into Nodes.

### DIFF
--- a/src/hugr/rewrite/outline_cfg.rs
+++ b/src/hugr/rewrite/outline_cfg.rs
@@ -139,10 +139,13 @@ impl Rewrite for OutlineCfg {
             new_block_bldr
                 .set_outputs(pred_wire, cfg.outputs())
                 .unwrap();
-            let (new_block, node_map) = h
+            let ins_res = h
                 .insert_hugr(outer_cfg, new_block_bldr.hugr().clone())
                 .unwrap();
-            (new_block, *node_map.get(&cfg.node()).unwrap())
+            (
+                ins_res.new_root,
+                *ins_res.node_map.get(&cfg.node()).unwrap(),
+            )
         };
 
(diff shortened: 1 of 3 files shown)


croyzor:
I do think having an `InsertionResult` type struct would be a bit clearer

src/hugr/rewrite/outline_cfg.rs
```diff
@@ -128,26 +129,21 @@ impl Rewrite for OutlineCfg {
...
-            new_block_bldr.set_outputs(pred_wire, cfg_outputs).unwrap();
-            h.insert_hugr(outer_cfg, new_block_bldr.hugr().clone())
-                .unwrap()
+            new_block_bldr
+                .set_outputs(pred_wire, cfg.outputs())
+                .unwrap();
+            let (new_block, node_map) = h
+                .insert_hugr(outer_cfg, new_block_bldr.hugr().clone())
+                .unwrap();
+            (new_block, *node_map.get(&cfg.node()).unwrap())
```
croyzor, line 143: The second element here should just be `new_block`, no?

croyzor: In which case all of the changes to return nodemaps are redundant and all we needed was to bind the result of `insert_hugr`

croyzor: Ah I see that's what was happening before and the Cfg is meant to be its child

croyzor (approved):

acl-cqc:
> I do think having an `InsertionResult` type struct would be a bit clearer

Fair enough! Done :)

PR #537, opened 2023-09-15
refactor!: More exception handling for SimpleReplacement

lmondada:
I needed this for debbuging in tket2, so I thought we might as well merge it in.

BREAKING CHANGE: `SimpleReplacement::new` is now `SimpleReplacement::try_new`.

### DIFF
--- a/src/hugr/rewrite/simple_replace.rs
+++ b/src/hugr/rewrite/simple_replace.rs
@@ -6,6 +6,7 @@ use itertools::Itertools;
 
 use crate::hugr::views::SiblingSubgraph;
 use crate::hugr::{HugrMut, HugrView, NodeMetadata};
+use crate::Direction;
 use crate::{
     hugr::{Node, Rewrite},
     ops::{OpTag, OpTrait, OpType},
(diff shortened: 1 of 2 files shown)


src/hugr/rewrite/simple_replace.rs
```diff
@@ -121,8 +124,7 @@ impl Rewrite for SimpleReplacement {
                 let (rem_inp_pred_node, rem_inp_pred_port) = h
                     .linked_ports(*rem_inp_node, *rem_inp_port)
                     .exactly_one()
-                    .ok() // PortLinks does not implement Debug
-                    .unwrap();
+                    .map_err(|_| SimpleReplacementError::InvalidBoundary())?;
```
cqc-alec, line 127: Since the maps were checked on construction, and an in-port must have exactly one linked port in any valid hugr, isn't it OK to panic here?

src/hugr/rewrite/simple_replace.rs
```diff
@@ -30,18 +31,20 @@ pub struct SimpleReplacement {
 
 impl SimpleReplacement {
     /// Create a new [`SimpleReplacement`] specification.
-    pub fn new(
+    pub fn try_new(
```
cqc-alec, line 34: Could we add a test where this fails?

PR #538, opened 2023-09-18
fix: Make unification logic less strict

croyzor:
The `Plus` constraint has up till now been too strict in extension unification. So far we have been deriving things like the following:
```
a := Plus("C", b)
a := {"A", "B", "C"}
implies
b := {"A", "B"}
```
(as in `minus_test`, which this PR removes)
However, plus should be meant as the union of the the singleton set that it specifies with the extension set of another metavariable. Hence in the above example the solution for `b` could be either `{"A", "B"}` or `{"A", "B", "C"}`.

This means that if we have chains of circuit operations which all add the "quantum" resource, they can all be represented as `Plus` constraints

### DIFF
--- a/src/extension/infer.rs
+++ b/src/extension/infer.rs
@@ -516,10 +516,12 @@ impl UnificationContext {
                                 solved = true;
                             }
                         };
-                    } else if let Some(superset) = self.get_solution(&meta) {
-                        let subset = ExtensionSet::singleton(r).missing_from(superset);
-                        self.add_solution(self.resolve(*other_meta), subset);
-                        solved = true;
+                    } else if let Some(_superset) = self.get_solution(&meta) {
+                        // Here, we're stuck because the Plus constraint only
+                        // signifies a preorder. I.e. if m0 = m1 + 'R', it's
+                        // still possible that the solution is m0 = m1 because
+                        // it's possible that we're adding 'R' to a set which
+                        // already contained it.
                     };
                 }
             }
@@ -936,58 +938,6 @@ mod test {
         assert_eq!(ctx.resolve(m0), ctx.resolve(m2));
         assert_eq!(ctx.resolve(m1), ctx.resolve(m2));
         assert!(ctx.resolve(m0) != mid0);
-        Ok(())
-    }
-
-    #[test]
-    fn minus_test() -> Result<(), Box<dyn Error>> {
-        let const_true = ops::Const::true_val();
-        const BOOLEAN: Type = Type::new_simple_predicate(2);
-        let just_bool = type_row![BOOLEAN];
-
-        let abc = ExtensionSet::from_iter([A, B, C]);
-
-        // Parent graph is closed
-        let mut hugr = closed_dfg_root_hugr(
-            FunctionType::new(type_row![], just_bool.clone()).with_extension_delta(&abc),
-        );
-
-        let [_, output] = hugr.get_io(hugr.root()).unwrap();
-
-        let root = hugr.root();
-        let [child, _, ochild] = create_with_io(
-            &mut hugr,
-            root,
-            ops::DFG {
-                signature: FunctionType::new(type_row![], just_bool.clone())
-                    .with_extension_delta(&abc),
-            },
-        )?;
-
-        let const_node = hugr.add_node_with_parent(child, NodeType::open_extensions(const_true))?;
-        let lift_node = hugr.add_node_with_parent(
-            child,
-            NodeType::open_extensions(ops::LeafOp::Lift {
-                type_row: just_bool,
-                new_extension: C,
-            }),
-        )?;
-
-        hugr.connect(const_node, 0, lift_node, 0)?;
-        hugr.connect(lift_node, 0, ochild, 0)?;
-        hugr.connect(child, 0, output, 0)?;
-
-        hugr.infer_extensions()?;
-
-        // The solution for the const node should be {A, B}!
-        assert_eq!(
-            hugr.get_nodetype(const_node)
-                .signature()
-                .unwrap()
-                .output_extensions(),
-            ExtensionSet::from_iter([A, B])
-        );
-
         Ok(())
     }
 
--- a/src/extension/infer.rs
+++ b/src/extension/infer.rs
@@ -672,7 +672,7 @@ mod test {
     use super::*;
     use crate::builder::test::closed_dfg_root_hugr;
     use crate::builder::{BuildError, DFGBuilder, Dataflow, DataflowHugr};
-    use crate::extension::{ExtensionSet, EMPTY_REG};
+    use crate::extension::{ExtensionSet, EMPTY_REG, PRELUDE_REGISTRY};
     use crate::hugr::{validate::ValidationError, Hugr, HugrMut, HugrView, NodeType};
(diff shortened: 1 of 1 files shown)


acl-cqc:
As it stands, this PR isn't doing anything - you are removing a test OK, but what about adding one of something that you couldn't verify before?? I see #539 but is that test passing already??

> Potential improvement: we could assert that if the node is a `Lift` node, then it generates a "greater than" Plus constraint and otherwise generates a "greater than or equal" Plus constraint. What do you think, @acl-cqc ?

Good question. ISTR this relates to the issues Alex Rice had with resource inference in Tierkreis too. I can see a few possibilities.
* As you suggest, keep lift nodes generating an input-does-not-contain constraint. I think this might be a useful thing to do - if nothing need be added, we don't need a lift node. Where *else* do we generate "greater than" Plus constraints atm, and should those be "strictly equal"? 
    * This may lead to problems with substitution of variables - i.e. lift nodes disappear when graphs are moved into new contexts (perhaps this will only happen during Tierkreis execution and we won't see it yet?). If so, it might help to specify that *every* Input node has empty extensions; the containing DFG node then takes a (non-disjoint, greater-than-or-equal) union of its inputs with the delta of the containing graph. (This might be a good idea anyway).

* Drop Lift nodes. What if we just say, the input extensions of every node must be exactly equal to the union of the output extensions of every predecessor.

croyzor:
> As it stands, this PR isn't doing anything - you are removing a test OK, but what about adding one of something that you couldn't verify before?? I see #539 but is that test passing already??

#539 was passing on main by mistake - the PR now shows that the test is failing on main, and that same commit has been added to this branch (and the test now passes.)

> As you suggest, keep lift nodes generating an input-does-not-contain constraint. I think this might be a useful thing to do - if nothing need be added, we don't need a lift node. Where _else_ do we generate "greater than" Plus constraints atm, and should those be "strictly equal"?

Before this PR we *only* generated "strictly greater than" constraints, and following this PR we *never* generate such constraints.

> This may lead to problems with substitution of variables - i.e. lift nodes disappear when graphs are moved into new contexts (perhaps this will only happen during Tierkreis execution and we won't see it yet?). If so, it might help to specify that _every_ Input node has empty extensions; the containing DFG node then takes a (non-disjoint, greater-than-or-equal) union of its inputs with the delta of the containing graph. (This might be a good idea anyway).

Hmm, good point about recontextualising graphs, but i *think* we can just ignore lift nodes in that context and run inference to re-add them where necessary (once we're adding lift nodes automatically)

> Drop Lift nodes. What if we just say, the input extensions of every node must be exactly equal to the union of the output extensions of every predecessor.

This is an option, but definitely one for another PR...

src/extension/infer.rs
```diff
@@ -516,10 +516,12 @@ impl UnificationContext {
...
-                    } else if let Some(superset) = self.get_solution(&meta) {
-                        let subset = ExtensionSet::singleton(r).missing_from(superset);
-                        self.add_solution(self.resolve(*other_meta), subset);
-                        solved = true;
+                    } else if let Some(_superset) = self.get_solution(&meta) {
+                        // Here, we're stuck because the Plus constraint only
+                        // signifies a preorder. I.e. if m0 = m1 + 'R', it's
+                        // still possible that the solution is m0 = m1 because
+                        // it's possible that we're adding 'R' to a set which
+                        // already contained it.
```
acl-cqc, line 524: Either just add `So, do nothing.` to make it clear this `if` here is just to scope the comment; Or, comment out the `else if` as well. Of course that'll then become noncompiling and soon break, so if you go that way, might be better to translate it from rust to english ;)

src/extension/infer.rs
```diff
@@ -516,10 +516,12 @@ impl UnificationContext {
                                 solved = true;
                             }
                         };
-                    } else if let Some(superset) = self.get_solution(&meta) {
-                        let subset = ExtensionSet::singleton(r).missing_from(superset);
-                        self.add_solution(self.resolve(*other_meta), subset);
-                        solved = true;
+                    } else if let Some(_superset) = self.get_solution(&meta) {
+                        // Here, we're stuck because the Plus constraint only
+                        // signifies a preorder. I.e. if m0 = m1 + 'R', it's
```
acl-cqc, line 521: It might be nicer still to use `meta` and `other_meta` instead of `m0` and `m1` (whichever way around it is)

src/extension/infer.rs
```diff
@@ -1091,4 +1042,76 @@ mod test {
...
+        let node4 = mknode(&mut hugr, root, df_sig.clone(), A)?;
+        let node5 = mknode(&mut hugr, root, df_sig.clone(), B)?;
+
+        // Connect nodes in order (0 -> 1 -> 2 ...)
+        let nodes = [input, node0, node1, node2, node3, node4, node5, output];
+        for (src, tgt) in nodes.into_iter().tuple_windows() {
+            hugr.connect(src, 0, tgt, 0)?;
+        }
+
+        hugr.infer_and_validate(&PRELUDE_REGISTRY)?;
```
acl-cqc, line 1113: Right - so without the fix, this line would break, yes?

acl-cqc (approved):
Ok, makes sense now, ta. I'm leaning towards not keeping a "disjoint union" constraint for lift because if we use that, then substitution of context could mean *requiring* removal of lift nodes, whereas with only the one kind of union, removal of lift nodes is always an optimization, never required. (And removal of nodes during `infer_and_validate` feels like a rather big thing to be sneaking in behind the user's back. Inserting them might be even bigger, I suppose....)

That said, I'm still holding out hope that this moves us towards eventual removal of the lift node type altogether, but indeed, not for this PR!

src/extension/infer.rs
```diff
@@ -1091,4 +1042,76 @@ mod test {
...
+
+        let node0 = mknode(&mut hugr, root, df_sig.clone(), A)?;
+        let node1 = mknode(&mut hugr, root, df_sig.clone(), A)?;
+        let node2 = mknode(&mut hugr, root, df_sig.clone(), B)?;
+        let node3 = mknode(&mut hugr, root, df_sig.clone(), B)?;
+        let node4 = mknode(&mut hugr, root, df_sig.clone(), A)?;
+        let node5 = mknode(&mut hugr, root, df_sig.clone(), B)?;
+
+        // Connect nodes in order (0 -> 1 -> 2 ...)
+        let nodes = [input, node0, node1, node2, node3, node4, node5, output];
```
acl-cqc, line 1108: How about something like
```
let mid_nodes = [A,A,B,B,A,B].map(|ext| mknode(&mut hugr, root, df_sig.clone(), ext)).collect::<Result<Vec<Node>,_>>()?;
let nodes = [[input], mid_nodes, [output]].concat()
```
(there is a collect that turns iterator of Result<T,...> into Result<Vec<T>,...>)

acl-cqc: You could even define `mknode` as a closure `|ext| {...}` after you've done `let hugr =` and then it'd be just `.map(mknode).collect....`

src/extension/infer.rs
acl-cqc: In which case LGTM ;-)

croyzor: Exactly!

PR #539, opened 2023-09-18
tests: Add test of repeated adding of the same extension

croyzor:


src/extension/infer.rs
```diff
@@ -1091,4 +1091,63 @@ mod test {
...
+        let node3 = mknode(&mut hugr, root, df_sig.clone(), B)?;
+        let node4 = mknode(&mut hugr, root, df_sig.clone(), A)?;
+        let node5 = mknode(&mut hugr, root, df_sig.clone(), B)?;
+
+        // Connect nodes in order (0 -> 1 -> 2 ...)
+        let nodes = [node0, node1, node2, node3, node4, node5];
+        let mut windows = nodes.windows(2);
+        while let Some(&[src, tgt]) = windows.next() {
+            hugr.connect(src, 0, tgt, 0)?;
+        }
```
aborgna-q, line 1149: Validate it afterwards? Where would a broken inference fail?

aborgna-q (approved):

src/extension/infer.rs
croyzor: 🤦

src/extension/infer.rs
```diff
@@ -1091,4 +1091,63 @@ mod test {
...
+        let node1 = mknode(&mut hugr, root, df_sig.clone(), A)?;
+        let node2 = mknode(&mut hugr, root, df_sig.clone(), B)?;
+        let node3 = mknode(&mut hugr, root, df_sig.clone(), B)?;
+        let node4 = mknode(&mut hugr, root, df_sig.clone(), A)?;
+        let node5 = mknode(&mut hugr, root, df_sig.clone(), B)?;
+
+        // Connect nodes in order (0 -> 1 -> 2 ...)
+        let nodes = [node0, node1, node2, node3, node4, node5];
+        let mut windows = nodes.windows(2);
+        while let Some(&[src, tgt]) = windows.next() {
```
acl-cqc, line 1147: Can this be `for &[src, tgt] in nodes.windows(2) {...}` or something like that?

croyzor: Unfortunately that doesn't work quite as well because rust doesn't know at compile time the size of the window and complains that `&[src, tgt]` is a refutable pattern.

aborgna-q: You can use [`tuple_windows](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.tuple_windows) instead.

croyzor: 🙌

croyzor:
This test should've been added to #538 instead of its own PR! But it acts as proof that this test doesn't pass on main, so #538 is doing something!

PR #541, opened 2023-09-18
Pass &ExtensionRegistry to compute_signature (etc.)

acl-cqc:
Without this, impl's can't necessarily construct instances of custom types like List - we rely on implementations to have the definitions themselves already, by some backdoor. (And for implementations to store a reference the ExtensionRegistry, leads to cyclic data, as they are going to be part of it!)

### DIFF
--- a/src/extension.rs
+++ b/src/extension.rs
@@ -300,9 +300,10 @@ impl Extension {
         &self,
         op_name: &str,
         args: impl Into<Vec<TypeArg>>,
+        ext_reg: &ExtensionRegistry,
     ) -> Result<ExtensionOp, SignatureError> {
         let op_def = self.get_op(op_name).expect("Op not found.");
-        ExtensionOp::new(op_def.clone(), args)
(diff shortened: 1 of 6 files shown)


aborgna-q (approved):
LGTM

Would it make sense to start using this mechanism in some of the existing extensions?

acl-cqc:
> LGTM
> 
> Would it make sense to start using this mechanism in some of the existing extensions?

There are two aspects to this.
1. Yes, we could start changing some of e.g. the logic extension, to use this to look up types like BOOL_T defined in the prelude (rather than using static variables). I think that's reasonable, and I am happy to do that, but I think it can be another PR
2. At some point (probably after we have type schemes) we are going to realize that building extensions like this is a bit of a pain. I think we may end up adding all the *types* to the extensions first, then adding the ops (as the ops can refer to types in other extensions, but types don't refer to types and ops don't refer to ops). That'll probably require some extra fun with mutable ExtensionRegistries. So that'll be coming up but I don't see any great urgency ;-) ;-)

PR #542, opened 2023-09-19
Polymorphic Function Types

acl-cqc:
* Based off #484
* Rename+move extension::OpDefTypeScheme (a Vec<TypeParam> and a FunctionType) to types::PolyFuncType
* PrimType now contains a PolyFuncType, not a FunctionType
   -> as hence we must, implement substitution and validation for that!
* The main change was to make `fn substitute` take a new `enum Substitution`. We need to deal with (a) substituting into things that have more binders deeper within them, and (b) substituting in values (on the RHS) that themselves have free variables (and binders) in them. Yes, this is the main bit of extra complexity beyond #484, and so I've put that in a new file `subst.rs`.
* Secondly there is the addition of a new `TypeApply` op, whose signature (required TypeParams) cannot be expressed as an extension OpDef, but (of course) this doesn't get used until we have some `indirect_call` or similar operation.
* CustomSignatureFunc now returns a PolyFuncType; you have to declare both sets of params (those passed to the binary, and those substituted into the polymorphic type scheme) in advance. We could change this that you only have to declare the static/binary params, if we wanted; this would change the requirements on supporting (builder/frontend) tools but again I don't think there are major problems. (You couldn't check the TypeArgs for an op "fitted" without calling the binary `compute_signature` but then, if you have a binary `compute_signature` then the `check_type_args` check against the TypeParams isn't complete anyway so no big difference)
* Note, this PR does *not* cover
   * Instances (values) of these polymorphic types. Hugrs themselves are NOT polymorphic yet. That can follow in another PR (should be fairly straightforward but we'll need to add a `Vec<TypeParam>` to either Hugr or Value::Graph)
   * Substitution on said values (i.e. the runtime implementation of the new `TypeApply` op) - in that we don't have runtime implementations of any other ops yet either, those belong in tools that use Hugr.
   * Updating all those extension operations, that currently use custom binary code, to use type schemes. This would be a really good thing to do, but that doesn't belong here (it doesn't even depend on this PR but would just follow #484 largely in parallel with this).

Note I believe all individual commits are passing except for e51b984b617e6b82db57ec90d54d7505d0761241 ("CustomSignatureFunc returns PolyFuncType"), immediately followed by "Fix tests....". However I have jumped around a lot with the representation of Substitution.


src/types/subst.rs
```diff
@@ -0,0 +1,82 @@
...
+        match &self.mapping {
+            Mapping::Values(args) => args
+                .get(idx - self.leave_lowest)
+                .expect("Unexpected free type var")
+                .clone(),
+            Mapping::AddToIndex(diff) => TypeArg::new_var_use(idx + diff, decl.clone()),
+        }
+    }
+
+    pub(super) fn get_type(&self, idx: usize, bound: TypeBound) -> Type {
```
acl-cqc, line 50: Hmmm, maybe this should be `apply_type_var` to be more consistent with `apply_var`

src/types/subst.rs
```diff
@@ -0,0 +1,82 @@
...
+        let TypeArg::Type {ty} = self.apply_var(idx, &TypeParam::Type(bound))
+           else {panic!("Var of kind 'type' did not produce a Type")};
+        ty
+    }
+
+    // A bit unfortunate to need the Extension registry here. We could move `exts` into Substitution?
```
acl-cqc, line 56: Or indeed, move all the `substitute` methods into `Substitution`....(they need some `types`-level access but that's it really)

src/extension/op_def.rs
```diff
@@ -96,10 +95,14 @@ pub(super) enum SignatureFunc {
...
     // However instead we treat all CustomFunc's as non-serializable.
     #[serde(rename = "signature")]
-    TypeScheme(OpDefTypeScheme),
+    TypeScheme(PolyFuncType),
     #[serde(skip)]
     CustomFunc {
-        params: Vec<TypeParam>,
+        /// All type parameters (both those passed to [func] and any passed
+        /// to the PolyFuncType returned by that)
+        all_params: Vec<TypeParam>,
```
acl-cqc, line 103: I considered having two separate vectors here, but we are constrained by having to implement `fn params(&self) -> &[TypeParam]` so we need this vec to return a view/slice thereof

src/extension/op_def.rs
```diff
@@ -326,36 +343,40 @@ impl Extension {
...
+                num_static_params,
                 func: Box::new(signature_func),
             },
         )
     }
 
-    /// Create an OpDef with custom binary code to compute the signature, and no "misc" or "lowering
-    /// functions" defined.
+    /// Create an OpDef with custom binary code to compute the signature,
+    /// which is not polymorphic; and no "misc" or "lowering functions" defined.
```
acl-cqc, line 368: we might well change this when we've converted the existing uses in the extensions to use type schemes rather than their own custom binary functions

src/hugr/validate.rs
```diff
@@ -162,13 +162,19 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
         // Check operation-specific constraints. Firstly that type args are correct
         // (Good to call `resolve_extension_ops` immediately before this
         //   - see https://github.com/CQCL-DEV/hugr/issues/508 )
-        if let OpType::LeafOp(crate::ops::LeafOp::CustomOp(b)) = op_type {
-            for arg in b.args() {
+        match op_type {
```
acl-cqc, line 165: Beginnings of a new trait here (currently we have a trait+method for `validate_op_children` but this is not about children)

src/types/poly_func.rs
```diff
@@ -0,0 +1,526 @@
...
+/// A polymorphic function type, e.g. of a [Graph], or perhaps an [OpDef].
+/// (Nodes/operations in the Hugr are not polymorphic.)
+///
+/// [Graph]: crate::values::PrimValue::Function
+/// [OpDef]: crate::extension::OpDef
+#[derive(
+    Clone, PartialEq, Debug, Eq, derive_more::Display, serde::Serialize, serde::Deserialize,
+)]
+#[display(
+    fmt = "forall {}. {}",
```
acl-cqc, line 20: Perhaps we could omit the `forall .` if there are no bound variables; it does look a little odd. But in that case we'd want some outer label (e.g. "PolyFuncType") to differentiate from `FunctionType`, a distinct type in Rust

src/types/serialize.rs
```diff
@@ -36,7 +34,7 @@ impl From<Type> for SerSimpleType {
             TypeEnum::Prim(t) => match t {
                 PrimType::Extension(c) => SerSimpleType::Opaque(c),
                 PrimType::Alias(a) => SerSimpleType::Alias(a),
-                PrimType::Function(sig) => SerSimpleType::G(Box::new(*sig)),
+                PrimType::Function(sig) => SerSimpleType::G(sig),
```
acl-cqc, line 37: Hmmmm, reckon we could have done this one-line change without this PR, but I think it's a good drive-by

src/types/poly_func.rs
```diff
@@ -0,0 +1,524 @@
...
+            &PRELUDE_REGISTRY,
+        )?;
+        let res = array_max.instantiate(&[USIZE_TA], &PRELUDE_REGISTRY)?;
+        assert_eq!(res, partial);
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_type_apply_nested() -> Result<(), SignatureError> {
```
mark-koch, line 488: What's the difference between this test and the first case in `substitute_under_binder`?

src/types/check.rs
```diff
@@ -60,7 +60,8 @@ impl PrimType {
                 Ok(())
             }
             (PrimType::Function(t), PrimValue::Function { hugr: v })
-                if Some(t.as_ref()) == v.get_function_type() =>
+                if t.params.is_empty() // Hugr's still monomorphic
+                && Some(&t.body) == v.get_function_type() =>
```
mark-koch, line 64: Is this `is_empty` check necessary? I would presume `v.get_function_type()` will have empty params, so the equality check should take care of it

src/types/subst.rs
mark-koch: Or rename `apply_var` to `get_arg`?

src/types/subst.rs
mark-koch: I like the idea of moving all the substitution logic into here 👍

src/ops/leaf.rs
```diff
@@ -49,6 +52,59 @@ pub enum LeafOp {
...
+    ///
+    /// [TypeParam]: crate::types::type_param::TypeParam
+    pub fn try_new(
+        input: PolyFuncType,
+        args: impl Into<Vec<TypeArg>>,
+        extension_registry: &ExtensionRegistry,
+    ) -> Result<Self, SignatureError> {
+        let args = args.into();
+        // Should we require >=1 `arg`s here? Or that input declares >=1 params?
+        // At the moment we allow an identity TypeApply on a monomorphic function type.
```
mark-koch, line 85: Mhm I don't have a strong preference here. Allowing identity TypeApplies wouldn't hurt but I also don't see a reason why you would want to use them

src/extension/op_def.rs
mark-koch, line 103: Maybe clarify that the first `num_static` params go to func and the remaining ones to the resulting `PolyFuncType`

src/extension/op_def.rs
```diff
@@ -326,36 +343,40 @@ impl Extension {
         &mut self,
         name: SmolStr,
         description: String,
-        params: Vec<TypeParam>,
+        params: (Vec<TypeParam>, Vec<TypeParam>),
```
mark-koch, line 346: Why a tuple of param vectors instead of two arguments?

src/extension/op_def.rs
mark-koch: Maybe add a TODO?

src/std_extensions/collections.rs
```diff
@@ -174,7 +170,12 @@ mod test {
 
     #[test]
     fn test_list_ops() {
-        let reg = &[EXTENSION.to_owned()].into();
+        let reg = &[
+            EXTENSION.to_owned(),
+            PRELUDE.to_owned(),
+            float_types::extension(),
```
mark-koch, line 176: Why are those two extra extensions required now?

src/types/poly_func.rs
```diff
@@ -0,0 +1,524 @@
...
+        body.validate(extension_registry, &params)?;
+        Ok(Self { params, body })
+    }
+
+    pub(super) fn validate(
+        &self,
+        reg: &ExtensionRegistry,
+        external_type_vars: &[TypeParam],
+    ) -> Result<(), SignatureError> {
+        // TODO should we add a mechanism to validate a TypeParam?
```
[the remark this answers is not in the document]
acl-cqc: oh, a value of `List<5>`, let's say, or `Array<usize>`

src/types/poly_func.rs
```diff
@@ -0,0 +1,524 @@
...
+        let mut v;
```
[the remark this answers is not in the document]
acl-cqc: Yes!!! I followed `rustc` and `clippy`s direction. The point of declaring here is so that the value in it lives until the end of this scope, rather than the much-shorter else scope, so we can borrow said value until end of this scope

src/types/subst.rs
acl-cqc: I've gone with `apply_to_var` and `apply_to_type_var`. If we later move all the `substitute` methods into the Substitution class, we can call those `apply_to_arg`, `apply_to_custom`, etc...

src/types/subst.rs
acl-cqc: Yeah, I think that'd be good to do in a followup. For now I'm trying not to move stuff around unnecessarily just to make the actual change more obvious.

src/types/poly_func.rs
```diff
@@ -0,0 +1,524 @@
...
+    ) -> Result<Self, SignatureError> {
+        let params = params.into();
+        body.validate(extension_registry, &params)?;
+        Ok(Self { params, body })
+    }
+
+    pub(super) fn validate(
+        &self,
+        reg: &ExtensionRegistry,
+        external_type_vars: &[TypeParam],
```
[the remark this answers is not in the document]
acl-cqc: Good spot - following discussion in #484, renamed to `external_var_decls` (and `all_type_vars` within to `all_var_decls` similarly)

src/types/poly_func.rs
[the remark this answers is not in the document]
acl-cqc: Added a comment

src/types/poly_func.rs
```diff
@@ -0,0 +1,524 @@
...
+            params: self.params.clone(),
+        }
+    }
+
+    pub(crate) fn instantiate(
+        &self,
+        args: &[TypeArg],
+        exts: &ExtensionRegistry,
+    ) -> Result<Self, SignatureError> {
+        let remaining = self.params.get(args.len()..).unwrap_or_default();
```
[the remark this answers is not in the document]
acl-cqc: Yes, fair! IIRC I had to think quite a bit about this myself to persuade myself that it worked.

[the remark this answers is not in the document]
acl-cqc: (Comment added.)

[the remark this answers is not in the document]
acl-cqc: I tried to get rid of the `unwrap_or_default` but you really need it for return value

src/extension/op_def.rs
acl-cqc: Ok. A bit repetitive given the docstring on `num_static_params` but harmless enough!

src/extension/op_def.rs
acl-cqc: `clippy: too many arguments`....

But yeah, better to disable that for this function, and add a TODO to sort this out when we've converted existing binary cases to type schemes.

@ss2165 if we went with "declare TypeParams to binary function only" (rather than declaring in advance the params that will be taken by the type scheme that the binary function gives you back) then that'd solve this. Of course this is not a reason to do that....

src/std_extensions/collections.rs
acl-cqc: Haha. Subtle. Because with the new "binary compute_signature can return a type-scheme", after executing the binary as we did before, we then do call `substitute` (with no args) on the type scheme (which has no type params). This traverses the type applying the identity substitution, and uses the extension registry to recompute TypeBounds of everything - even e.g. `USIZE`.

So, we could skip all of that - if the type scheme we get back from compute_signature has no type-params (and there are no typeargs), we could call `clone()` rather than `substitute`. This might be a performance advantage, but asymptotically it'd still be O(size of Type tree)...

Moreover, I kind of feel this *should* error if those extensions aren't present; when we convert the binary function to a type scheme, then it'll have to do a non-trivial substitution, and we'll almost certainly need these extensions then.

I mean, I guess we could "optimize" more substitutions into clones, i.e. for arbitrary bits of the type tree rather than just the whole tree, e.g. for `CustomType`s with no args - but those are close enough to leaves anyway, so I think the saving could well be even smaller.

Thoughts welcome....

src/ops/leaf.rs
ss2165: I suspect allowing identity type applies will make certain rewrites and simplifications easier to express, can always be removed in cleanup phases

src/types/subst.rs
ss2165: I think this would be a good and high priority follow up

src/extension/op_def.rs
ss2165: I find this specification unintuitive. If we went with "declare params for binary only" those would be the only ones here right?

ss2165 (commented):
I've trusted the detailed code review to @mark-koch but I can also do one of those later.

Overall, I like it! In particular I am a fan of binary-defined ops returning a polyfunc to be substituted, and that many/most ops can be declared as polyfunctypes without binaries.

My only big comment is that since we are nicely reducing the number of cases in binary signature computation is necessary anyway, we can choose to make that a little more back-doory - by letting the PolyFuncType returned by such a computation have unknown (not pre-declared) TypeParams. In short: I support going with "only declare binary type params" in this PR, as I think it makes this PR nicer.

src/std_extensions/collections.rs
acl-cqc: That seems a bit of a nonsequitur but that doesn't matter - however, I'm wondering what you mean. Do you have any (more concrete) examples? I prefer not to add a Type variant that's "a monomorphic FunctionType" because, sitting alongside the variant that's a PolyFuncType, it'd be strictly less general.

src/ops/leaf.rs
acl-cqc: Makes sense yes

src/extension/op_def.rs
acl-cqc: Yes. I've done this now. Everything is much neater - but the downside is, supposing you are writing a Hugr using an LSP in your editor or similar - there is now no method to say, "here are the static/binary params, what more params do I need?)". This is clearly something we will want at some stage - at the moment I'm leaving that until someone embarks upon such an LSP, but I could add something now if we like?

src/types/subst.rs
acl-cqc: I've done the small part that's moving the `&ExtensionRegistry` into the Substitution class - that just tidies things up. Moving the rest of the substitute methods can follow later.

src/types/poly_func.rs
```diff
@@ -0,0 +1,524 @@
...
+                    Type::new_var_use(0, TypeBound::Copyable),
+                    list_of_tup(
+                        Type::new_var_use(0, TypeBound::Copyable),
+                        Type::new_var_use(1, TypeBound::Any), // The outer variable (renumbered)
+                    ),
+                ))],
+            ),
+            &reg,
+        )
+        .unwrap();
```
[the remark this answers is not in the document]
acl-cqc: Good idea - I've added here and a few others. Wasn't sure what would be less confusing, between putting the comment just before the call to make a FunctionType/PolyFuncType or just inside, so went with the former - take a look and see if that's clear??

src/types/poly_func.rs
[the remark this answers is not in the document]
mark-koch: Ah I see. It might be useful to validate them at some point then. But I'm not sure if we should do it here. Would it make sense to assume that the params that are passed to `validate` are themselves already validated? (similar to how we assume that the extension registry that is passed here contains valid extensions, right?)

src/types/subst.rs
```diff
@@ -13,6 +13,7 @@ pub(crate) struct Substitution<'a> {
     leave_lowest: usize,
     /// What to do to variables that are affected
     mapping: Mapping<'a>,
+    reg: &'a ExtensionRegistry,
```
mark-koch, line 16: Could add a docstring explaining why the registry is needed for a substitution

src/types/poly_func.rs
mark-koch: So it's `forall A. A -> (forall C. C -> List(Tuple(C, A))` vs `forall N. -> (forall A. Array(A, N) -> A)`. Isn't this mostly the same?

src/types/poly_func.rs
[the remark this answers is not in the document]
mark-koch: I think the way you did it is clear 👍

src/types/subst.rs
acl-cqc: Yes, good point, this is better than the other places we have that comment! Can I find them....(I haven't tried to remove old comments saying the same thing)

src/types/poly_func.rs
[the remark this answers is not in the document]
acl-cqc: Yes, that's true! Ok - I've raised #624, and dropped the TODO

src/types/check.rs
acl-cqc: 1. At the moment `v.get_function_type()` returns a `&FunctionType`, not a `PolyFuncType`. I can't map that into a no-binders PolyFuncType without `clone`ing it, which seems a bit wasteful (but you if say "premature optimization" I'm willing to change that I guess).
2. I could add an `PolyFuncType::eq_monomorphic(&FunctionType)`, which is a little nicer in some ways, but...
3. In fact `v.get_function_type()` refers an *Option* of `&FunctionType` and by the time I've dealt with that, basically *anything else* is 3x longer for unclear benefit

Other suggestions welcome!

acl-cqc: Ah, I missed that there is Option::is_some_and. So we could add `PolyFuncType::eq_monomorphic(&self, &FunctionType)` and then make this here `v.get_function_type().is_some_and(|f|t.eq_monomorphic(f))`, that might be a bit better?

acl-cqc: Ok, and rather than `eq_monomorphic`, the trait "PartialEq" is parametrised, so that does look like the right way to do this! Done :)

acl-cqc: Done!

mark-koch: Very nice!

src/types/poly_func.rs
acl-cqc: Ooof! Yes, you are right. (The later tests were originally in `type_apply.rs` and then as I shifted more logic from there to here I realized that the tests should move too....well spotted, sir!)

Ok, so I have commoned up into a helper function `nested_func()` used in both. Saves quite a few lines, but the two tests remain (one is testing more complex RHSes than the other). Thanks! :)

acl-cqc:
Going for #630...

PR #543, opened 2023-09-19
docs: add operation constraint to quantum extension

### LINKED ISSUE #170: Expand on bijection of linear inputs to outputs in quantum resource
To finish the tket2 tangent above, concretely I propose TKET2 should have a constraint that it will only accept _operations_ on qubits on linear types:
1. That take in as many of linear type _L_ as they return
2. Will have the bijection semantics that when grouping by concrete type _L_, the nth input of type _L_ maps to the nth output of type _L_

_Functions_ like `Sum(Qubit) -&gt;i1` will need to use `qfree` in their definition. I think this gives us another distinction between `Operation` and `Function`?

[Slack Message](https://cambridgequantum.slack.com/archives/C04SHCL4FKP/p1686654856303309?thread_ts=1686651648.102459&cid=C04SHCL4FKP)

ss2165:
Closes #170

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1733,13 +1733,14 @@ compatibility:
 1. The `main` function takes one `Array<N, Qubit>`
 input and has one output of the same type (the same statically known
 size).
-2. All Operations that have a `FunctionType` involving `Qubit`:
-    1. Have as many `Qubit` input wires as output.
-    2. Guarantee semantics that when filtered to `Qubit` only, there is a
-       bijection such that the _nth_ input maps to the _nth_ output.
+2. All Operations that have a `FunctionType` involving `Qubit` have as
+ many `Qubit` input wires as output.
 
-This means we can treat all `Qubit` operations as returning all qubits they take
-in. If further the program does not contain any `qalloc` or `qfree`
+
+With these constraints, we can treat all `Qubit` operations as returning all qubits they take
+in. The implicit bijection from input `Qubit` to output allows register
+allocation for all `Qubit` wires. 
+If further the program does not contain any `qalloc` or `qfree`
 operations we can state the program only uses `N` qubits.
 
 ### Higher-order (Tierkreis) Extension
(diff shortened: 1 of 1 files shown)


specification/hugr.md
```diff
@@ -1727,10 +1727,19 @@ allocation, and the compiler can add/remove/move these operations to use
...
+compatibility:
+
+1. The `main` function takes one `Array<N, Qubit>`
 input and has one output of the same type (the same statically known
-size). If further the program does not contain any `qalloc` or `qfree`
+size).
+2. All Operations that have a `FunctionType` involving `Qubit`:
+    1. Have as many `Qubit` input wires as output.
+    2. Guarantee semantics that when filtered to `Qubit` only, there is a
+       bijection such that the _nth_ input maps to the _nth_ output.
```
cqc-alec, line 1739: It's a little unclear what this means.

"There is a bijection such that the _nth_ input maps to the _nth_ output" is trivially true if there are the same number of inputs and outputs.

Actually I had a look at the original slack thread and I am still not quite sure what it is trying to say. We are saying that the nth input "is" the nth output in some sense, but how to make this precise without introducing explicit qubit references (and saying that the nth input has the same qubit reference as the nth output)?

cqc-alec: To put it another way: what would failure of this condition look like?

ss2165: Yeah I am not sure how to express it either. I'm trying to say that the operation makes some guarantee that the qubit returned on the nth wire is in fact the same qubit that was taken in on the nth wire. To make this more precise we can say that when executed they will be the same logical qubit?

specification/hugr.md
```diff
@@ -1797,7 +1806,7 @@ an edge weight.
     Dataflow container node, with all edges between them. Includes
     exactly one input node (unique node having no input edges) and one
     output node (unique node having no output edges). Nodes are
-    processes that operate on input data and produce output data. Edges
+    proceses that operate on input data and produce output data. Edges
```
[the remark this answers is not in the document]
ss2165: no idea how this happened, didn't intend to edit anything over there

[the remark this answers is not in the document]
ss2165: i suspect a rogue cursor + backspace

specification/hugr.md
cqc-alec: I'm still not sure what we lose by omitting point 2. What do we need this guarantee for?

cqc-alec (approved):

PR #544, opened 2023-09-19
feat: add `new_array` operation to prelude

ss2165:


src/extension/prelude.rs
```diff
@@ -82,6 +100,21 @@ pub fn new_array(typ: Type, size: u64) -> Type {
     Type::new_extension(custom_t)
 }
 
+/// Name of the operation in the prelude for creating new arrays.
+pub const NEW_ARRAY_OP_ID: &str = "new_array";
+
+/// Initialize a new array op of type `typ` of length `size`
```
acl-cqc, line 106: super-nit: we talk about the "type" of an array, but the array is itself an array; the thing we are referring to is the *element type*.

(You could rename `typ` to `elem_ty` or similar if you wanted, but if not, then super-super-nit `typ` -> `ty` would make the subsequent construction easier)

src/extension/prelude.rs
```diff
@@ -71,7 +89,7 @@ pub const USIZE_T: Type = Type::new_extension(USIZE_CUSTOM_T);
 pub const BOOL_T: Type = Type::new_simple_predicate(2);
 
 /// Initialize a new array of type `typ` of length `size`
-pub fn new_array(typ: Type, size: u64) -> Type {
+pub fn array_type(typ: Type, size: u64) -> Type {
```
acl-cqc, line 92: driveby repeat of other nit (inc the comment on the line above), it's the *element* type

acl-cqc (approved):
Looks good to me, I particularly approve of renaming `new_array` to `array_type`.

That said, I don't think that's what the issue relates to (it was pre-type-redux when we actually had a Type::Array and there was the business of SimpleType::Classic(ClassicType::Container(Container::Array(...))) vs SimpleType::Container(Container::Array(SimpleType::Clasic(...))) which the factory methods were designed to enforce). So we haven't really thought about what array operations we'd want - e.g. given a *Copyable* type one could have `n_times`. But this seems fine to me.

PR #548, opened 2023-09-21
Add RootChecked to guarantee root node type

acl-cqc:
* Factor out RootTagged as a subtrait of HugrView to allow separate implementation. Most things can still just use HugrView (as it happens everything that implements HugrView also implements RootTagged, but this is not required).
    * HugrMut a subtrait of RootTagged not just HugrView
* Then add a new RootChecked struct storing any `AsRef<Hugr>` with a fixed RootHandle.
     * It allows `.as_ref()`, throwing away the extra information in the RootHandle
     * If the underlying view is a HugrMut then it is too
     * Use trait-default implementations of Hugr(View,(Mut)Internals), as the latter check that `replace_op` does not break the bound on root-type
     * But do not provide `.as_mut()` as that would allow bypassing and invalidating the extra info in the RootHandle
     * Like SiblingMut, check that nested instances only narrow the bound.
* Change the overridden-for-&(mut) Hugr impls of Hugr(Mut)Internals to only work for `RootHandle=Node` to ensure the lack of checking in replace_op there is safe.

### DIFF
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -436,12 +436,13 @@ fn insert_subgraph_internal(
 
 pub(crate) mod sealed {
     use super::*;
+    use crate::ops::handle::NodeHandle;
 
     /// Trait for accessing the mutable internals of a Hugr(Mut).
     ///
     /// Specifically, this trait lets you apply arbitrary modifications that may
     /// invalidate the HUGR.
-    pub trait HugrMutInternals: HugrView {
+    pub trait HugrMutInternals: RootTagged {
         /// Returns the Hugr at the base of a chain of views.
         fn hugr_mut(&mut self) -> &mut Hugr;
 
@@ -501,11 +502,20 @@ pub(crate) mod sealed {
         /// # Errors
         /// Returns a [`HugrError::InvalidTag`] if this would break the bound
         /// ([`Self::RootHandle`]) on the root node's [OpTag]
-        fn replace_op(&mut self, node: Node, op: NodeType) -> Result<NodeType, HugrError>;
+        fn replace_op(&mut self, node: Node, op: NodeType) -> Result<NodeType, HugrError> {
+            self.valid_node(node)?;
+            if node == self.root() && !Self::RootHandle::TAG.is_superset(op.tag()) {
+                return Err(HugrError::InvalidTag {
+                    required: Self::RootHandle::TAG,
+                    actual: op.tag(),
+                });
+            }
+            self.hugr_mut().replace_op(node, op)
+        }
     }
 
     /// Impl for non-wrapped Hugrs. Overwrites the recursive default-impls to directly use the hugr.
-    impl<T: HugrView + AsMut<Hugr>> HugrMutInternals for T {
+    impl<T: RootTagged + AsMut<Hugr>> HugrMutInternals for T {
         fn hugr_mut(&mut self) -> &mut Hugr {
             self.as_mut()
         }
@@ -561,7 +571,7 @@ pub(crate) mod sealed {
         }
 
         fn replace_op(&mut self, node: Node, op: NodeType) -> Result<NodeType, HugrError> {
-            // No possibility of failure here since there is no Self::RootHandle
+            // No possibility of failure here since Self::RootHandle == Any
             let cur = self.hugr_mut().op_types.get_mut(node.index);
             Ok(std::mem::replace(cur, op))
         }
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -216,8 +216,7 @@ impl InsertionResult {
 }
 
 /// Impl for non-wrapped Hugrs. Overwrites the recursive default-impls to directly use the hugr.
-impl<T: HugrView + AsMut<Hugr>> HugrMut for T
-{
+impl<T: HugrView + AsMut<Hugr>> HugrMut for T {
     fn add_node_with_parent(&mut self, parent: Node, node: NodeType) -> Result<Node, HugrError> {
         let node = self.as_mut().add_node(node);
         self.as_mut()
@@ -516,8 +515,7 @@ pub(crate) mod sealed {
     }
 
     /// Impl for non-wrapped Hugrs. Overwrites the recursive default-impls to directly use the hugr.
-    impl<T: HugrView + AsMut<Hugr>> HugrMutInternals for T
-    {
+    impl<T: HugrView + AsMut<Hugr>> HugrMutInternals for T {
         fn hugr_mut(&mut self) -> &mut Hugr {
             self.as_mut()
         }
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -571,7 +571,12 @@ pub(crate) mod sealed {
         }
 
         fn replace_op(&mut self, node: Node, op: NodeType) -> Result<NodeType, HugrError> {
-            // No possibility of failure here since Self::RootHandle == Any
+            if node == self.root() && !Self::RootHandle::TAG.is_superset(op.tag()) {
+                return Err(HugrError::InvalidTag {
+                    required: Self::RootHandle::TAG,
+                    actual: op.tag(),
+                });
+            }
             let cur = self.hugr_mut().op_types.get_mut(node.index);
             Ok(std::mem::replace(cur, op))
         }
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -216,7 +216,7 @@ impl InsertionResult {
 }
 
 /// Impl for non-wrapped Hugrs. Overwrites the recursive default-impls to directly use the hugr.
-impl<T: RootTagged<RootHandle = Node> + AsMut<Hugr>> HugrMut for T {
+impl<T: RootTagged + AsMut<Hugr>> HugrMut for T {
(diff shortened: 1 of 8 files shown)


ss2165 (commented):
Seems fine to me, but I would suggest waiting for Agustin's review too

src/hugr/views.rs
```diff
@@ -303,6 +305,29 @@ pub trait HugrView: sealed::HugrInternals {
...
 
+/// A view of the whole Hugr.
+/// (Just provides static checking of the type of the root node)
+pub struct WholeHugrView<'a, Root = Node>(&'a Hugr, PhantomData<Root>);
+
+impl<'a, Root: NodeHandle> WholeHugrView<'a, Root> {
+    /// Create a hierarchical view of a whole HUGR
+    ///
+    /// # Errors
+    /// Returns [`HugrError::InvalidNode`] if the root isn't a node of the required [OpTag]
```
aborgna-q, line 316: I feel we could use a `HugrError::InvalidTag` variant for these cases. This could be another PR, since we do the same on all the views.

aborgna-q, line 316:
```suggestion
    /// Returns [`HugrError::InvalidNode`] if the root isn't a node of the required [`OpTag`]
    ///
    /// [`OpTag`]: crate::ops::tag::OpTag
```

src/hugr/views.rs
```diff
@@ -303,6 +305,29 @@ pub trait HugrView: sealed::HugrInternals {
...
+    pub fn try_new(hugr: &'a Hugr) -> Result<Self, HugrError> {
```
aborgna-q, line 317: impl `TryInto<WholeHugrView<'a, R>> for &'a Hugr` ?

aborgna-q (commented):
I agree with doing root-checked references, but from knowing how the rest of the codebase went we will also need a mutable versions at some point.

What do you think about having a
```rust
pub struct RootTagged<H, Root>(H, PhantomData<Root>);

impl<H: AsRef<Hugr>, Root: NodeHandle> RootTagged<H, Root> { ... }
```
instead? (and derive `AsMut` too).

src/hugr/views.rs
acl-cqc: Done (the extra link definition didn't seem to be required)

src/hugr/views.rs
acl-cqc: Nice!! :-)

acl-cqc:
> I agree with doing root-checked references, but from knowing how the rest of the codebase went we will also need a mutable versions at some point.
> 
> What do you think about having....
> 
> ```rust
> impl<H: AsRef<Hugr>, Root: NodeHandle> RootTagged<H, Root> { ... }
> ```
> 
> instead? (and derive `AsMut` too).

Not sure I understand the lifetime stuff well enough, but don't we need the explicit `'a` lifetime to make this useful?

Hmmm, I'll have a play and see where I get. `AsMut` (or `HugrMut` for the `WholeHugrView` as it stands) both seem reasonable, yes can do (but will try to figure out the AsRef-vs-`'a` before I do that)

acl-cqc:
> don't we need the explicit `'a` lifetime to make this useful?

Hmmm, probably not. The likely use case here is when you have an (&)Hugr and want to pass it to something that takes an `&impl HugrView<NodeHandle=...>`, see for example #247. Returning a WholeHugrView with a particular lifetime specified in the type is...possible, but likely not the main use case.

So yeah, maybe we don't need the lifetime after all....but, that said, what's the advantage of using `AsRef<Hugr>`? You gain the ability to store a whole Hugr in a WholeHugrView, if you want to. In general I doubt that's the main use case *either*, but again, probably has uses in other/different corner cases...

src/hugr/views.rs
acl-cqc: Yes, #581, I think it'd be good to do that first, and then I'll make this PR update those tests

acl-cqc:
> I agree with doing root-checked references, but from knowing how the rest of the codebase went we will also need a mutable versions at some point.

Yup, now done :).

> What do you think about having a
> 
> ```rust
> pub struct RootTagged<H, Root>(H, PhantomData<Root>);
> ```

Ok, I persuade myself that the use case of returning a `RootTagged<Hugr, DfgId>` (owning the Hugr) is useful - otherwise you would need to return a Hugr and a WholeHugrView of that same thing, which is generally impossible. So, good call.

That said, it's a shame to see `try_from` has to go, and it's a bit of a pain having to do `RootTagged<Hugr, XXXId>` when you want to specify just the XXX. But the only way around that that I can see is something like:
```rust
struct RootTagFactory<H>(PhantomData<H>);
impl<H: AsRef<Hugr>> RootTagFactory<H> {
  fn try_new<Root: NodeHandle>(h: H) -> Result<RootTagged<H, Root>> {....}
}
```
so you can then do `RootTagFactory::try_new::<CfgID>(h)`. (I think? Right?)....I am not sure that's worth it.

src/hugr/hugrmut.rs
```diff
@@ -216,10 +216,7 @@ impl InsertionResult {
 }
 
 /// Impl for non-wrapped Hugrs. Overwrites the recursive default-impls to directly use the hugr.
-impl<T> HugrMut for T
-where
-    T: HugrView + AsMut<Hugr>,
-{
+impl<T: HugrView + AsMut<Hugr>> HugrMut for T {
```
acl-cqc, line 219: Hmmm, these were a bit drive-by, but they make searching for these quasi-blanket impls much easier. (A one-line search hit `impl<T> HugrMut for T` is really not very informative!)

acl-cqc:
CI failing due to updates in clippy. Let's fix those in another PR.

acl-cqc:
> CI failing on existing code due to updates in clippy. Let's fix those in another PR.

Done (#592)

src/hugr/views.rs
```diff
@@ -303,8 +300,59 @@ pub trait HugrView: sealed::HugrInternals {
...
+    pub fn into_hugr(self) -> Hugr {
+        self.0
+    }
+}
+
+impl<H: AsRef<Hugr>, Root: NodeHandle> RootTagged for RootChecked<H, Root> {
+    type RootHandle = Root;
+}
+
+// Note do not implement AsMut<Hugr> - that would get us the `impl HugrMutInternals`
```
acl-cqc, line 338: Note this contrary to your initial idea @aborgna-q. If you try to impl AsMut<Hugr>, you'll have to remove the impl of HugrMutInternals (and HugrMut) here, and then the test will fail. The test may be enough to justify removing this comment, possibly.

src/hugr/views.rs
```diff
@@ -451,10 +508,7 @@ pub(crate) mod sealed {
         fn root_node(&self) -> Node;
     }
 
-    impl<T> HugrInternals for T
-    where
-        T: AsRef<super::Hugr>,
-    {
+    impl<T: AsRef<super::Hugr>> HugrInternals for T {
```
aborgna-q, line 511:
```suggestion
    impl<T: AsRef<Hugr>> HugrInternals for T {
```

src/hugr/views.rs
```diff
@@ -303,8 +300,61 @@ pub trait HugrView: sealed::HugrInternals {
...
+    /// The kind of handle that can be used to refer to the root node.
+    ///
+    /// The handle is guaranteed to be able to contain the operation returned by
+    /// [`HugrView::root_type`].
+    type RootHandle: NodeHandle;
+}
+
+/// A view of the whole Hugr.
+/// (Just provides static checking of the type of the root node)
+pub struct RootChecked<H, Root = Node>(H, PhantomData<Root>);
```
aborgna-q, line 314: This should go in a submodule, as all the other views.

src/hugr/views.rs
aborgna-q: Uhm, this is an easy pit to fall in.

Since the auto-impl for `HugrMutInternals` requires `RootTagged`, why not check the root type before dereferencing with `as_mut` there?
The generic constraints in `impl<T: RootTagged + AsMut<Hugr>> HugrMutInts for T` kind of tell you that you're adding the extra check on top of the reference.

acl-cqc: Hmmm, the answer might be - premature optimization, i.e. that the `impl HugrMutInternals` for `T: AsMut<Hugr>` then gets to skip the check....so ok, let's not prematurely optimize, SGTM ;)

Note there was an unexpected subtle consequence that I had to rule out in 07e4f1097cf936d8327881bdc0e2b118ba627613

acl-cqc: (and this does let anyone call `.as_mut().replace_op(...)` and bypass the check, which I'm not very keen on....)

acl-cqc: So overall, I have done as you say, but I'm not sure why this is better, TBH

acl-cqc: See 2be330dc7a16dccc7dacbf04bbe84c63fa5e283c, which HEAD reverts - I actually think I prefer the former; it needs an extra `impl HugrMut` (but, empty!) so that seems minimally harmful to me

aborgna-q (approved):
👍

acl-cqc:
I'm now allowing making one RootChecked on top of another, or indeed on top of a SiblingMut, so long as the root-type bound is only narrowed, rather than only on top of something whose RootHandle was Any, as before. And hiding `.as_mut()` (which would allow breaking the bound) by using the trait-default HugrMut methods rather than the versions overridden for operating directly on a Hugr.

PR #550, opened 2023-09-22
tests: Update inference test to not use DFG builder

croyzor:
Nowadays, DFGBuilder creates parent and child I/O nodes all with open extensions. Since this development, this test has been too strict since, in the absence of any other constraints, it can resolve all extension sets on the graph to `{A}`. Improvements to inference in the near future to address #497 will cause it to fail


aborgna-q (approved):
LGTM

PR #552, opened 2023-09-22
feat: Insert/extract subgraphs from a HugrView

aborgna-q:
With this we can efficiently extract subgraphs from a HUGR, and insert other HUGR's subgraphs.

### DIFF
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -161,6 +161,12 @@ pub trait HugrMut: HugrView + HugrMutInternals {
     }
 
     /// Copy a subgraph from another hugr into this one, under a given root node.
+    ///
+    /// Sibling order is not preserved.
+    //
+    // TODO: Try to preserve the order when possible? We cannot always ensure
+    // it, since the subgraph may have arbitrary nodes without including their
+    // parent.
     fn insert_subgraph(
         &mut self,
         root: Node,
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -3,6 +3,7 @@
 use std::collections::HashMap;
 use std::ops::Range;
 
+use portgraph::view::{NodeFilter, NodeFiltered};
 use portgraph::{LinkMut, NodeIndex, PortMut, PortView, SecondaryMap};
 
 use crate::hugr::{Direction, HugrError, HugrView, Node, NodeType};
@@ -12,6 +13,7 @@ use crate::{Hugr, Port};
 
 use self::sealed::HugrMutInternals;
 
+use super::views::SiblingSubgraph;
 use super::{NodeMetadata, Rewrite};
 
 /// Functions for low-level building of a HUGR.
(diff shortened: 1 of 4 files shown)


src/hugr/hugrmut.rs
```diff
@@ -294,11 +322,40 @@ where
             self.as_mut().op_types.set(new_node, nodetype.clone());
             let meta = other.get_metadata(node.into());
             self.as_mut()
-                .set_metadata(node.into(), meta.clone())
+                .set_metadata(new_node.into(), meta.clone())
```
croyzor, line 325: This was done in an earlier PR, no?

croyzor: Oh no, that was outline_cfg. but same error 😲

src/hugr/hugrmut.rs
```diff
@@ -341,6 +398,37 @@ fn insert_hugr_internal(
...
+///
+/// Returns a mapping from the nodes in the inserted graph to their new indices
+/// in `hugr`.
+///
+/// This function does not update the optypes of the inserted nodes, so the
+/// caller must do that.
+///
+/// In contrast to `insert_hugr_internal`, this function does not preserve
+/// sibling order in the hierarchy.
+fn insert_hugr_internal_with_portgraph(
```
croyzor, line 411: I don't understand why this function is being defined instead of using the existing one? I assume in the existing function the graph insertion is done as a side effect, plus returning the inserted root node in the graph seems preferable to putting an `Option` there in the `InsertionResult`

src/hugr/hugrmut.rs
aborgna-q: Yep, seems like an easy to miss error :/
And we never do anything with the metadata, so the error is never triggered.

src/hugr/hugrmut.rs
aborgna-q: `insert_hugr_internal` assumes that you are inserting a complete view. That is, a complete region with a single root node (and maybe more descendants). So we can re-connect the hierarchy of the inserted nodes simply by doing a BFS traversal from the old root.

When inserting a subgraph, we may potentially include nodes from any part of the hugr. We then preserve the hierarchy when both parent and children are being inserted, but if a node's parent is missing then we connect it to the indicated `root`.

Because of that, we cannot do the efficient traversal of `insert_hugr_internal` and instead have to iterate the inserted nodes and ask if their parent is also in the chosen set. That causes the sibling order to not be preserved (unless we add some costly pre-processing).

As for the other point, note the definition of the `InsertionResult`'s `root`:
>     /// The node, after insertion, that was the root of the inserted Hugr.

For a subgraph there is no previous root.

croyzor: Ah I see! Can you add some of this to docstring please? And maybe something like `insert_subgraph` would be a better name since I don't think we can really say "insert hugr" if the hugr we're inserting doesn't have a root

src/hugr/views/sibling_subgraph.rs
```diff
@@ -386,6 +400,51 @@ impl SiblingSubgraph {
...
+        input_extensions: ExtensionSet,
+    ) -> Result<Hugr, HugrError> {
+        let signature = Signature {
+            signature: self.signature(hugr),
+            input_extensions,
+        };
+        let builder = FunctionBuilder::new(name, signature).unwrap();
+        let inputs = builder.input_wires();
+        let mut extracted = builder
+            .finish_hugr_with_outputs(inputs, &PRELUDE_REGISTRY)
```
croyzor, line 421: What's the reason for wiring the input and output nodes here? Seems unnecessary

croyzor (approved):

PR #553, opened 2023-09-22
feat: PortIndex trait for undirected port parameters

aborgna-q:
Add a `PortIndex` trait implemented by both `Port` and `usize`, to use in all the methods that expect an undirected port offset (instead of asking for an `usize`).

### DIFF
--- a/src/hugr.rs
+++ b/src/hugr.rs
@@ -209,7 +209,8 @@ pub struct Port {
 
 /// A trait for getting the undirected index of a port.
 ///
-/// Implemented for both [`Port`] and `usize`.
+/// This allows functions to admit both [`Port`]s and explicit `usize`s for
+/// identifying port offsets.
 pub trait PortIndex {
     /// Returns the offset of the port.
     fn index(self) -> usize;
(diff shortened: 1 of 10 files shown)


src/hugr.rs
```diff
@@ -207,6 +207,14 @@ pub struct Port {
     offset: portgraph::PortOffset,
 }
 
+/// A trait for getting the undirected index of a port.
+///
+/// Implemented for both [`Port`] and `usize`.
```
croyzor, line 212: * Can you specify that `usize` is included because it's the port type in `portgraph`.
* I quite like that the comment specifies that this is to unite Port and usize, but writing something like this seems doomed to fall out of date. Maybe we can be a bit more abstract?
* Also, maybe in the future we can add something like `portgraph::PortType` to implement this on rather than usize so that the doclinks can restrict the ways this comment can go out of date.

croyzor (approved):
Nice!

src/hugr.rs
aborgna-q: - I reworded it a bit so it's not necessarily an exhaustive list.
- Now that it mentions this is for offsets, I don't think `usize` needs explanation since that's the default index type everywhere (`portgraph` uses `u16` internally for performance, but we don't care at this abstraction level).

PR #554, opened 2023-09-22
OutlineCfg rewrite returns new BB and CFG nodes

acl-cqc:
This seems like a good use of the facility for a rewrite to return a rewrite-specific result, and preliminary to other upcoming PRs (specifically #247)

The new CFG node is non-trivial to find given its parent (basic block) as that has several children: not only Input and Output but also constant nodes to build the predicate.

### DIFF
--- a/src/hugr/rewrite/outline_cfg.rs
+++ b/src/hugr/rewrite/outline_cfg.rs
@@ -93,14 +93,17 @@ impl OutlineCfg {
 
 impl Rewrite for OutlineCfg {
     type Error = OutlineCfgError;
-    type ApplyResult = ();
+    /// The newly-created basic block, and the [CFG] node inside it
+    ///
+    /// [CFG]: OpType::CFG
+    type ApplyResult = (Node, Node);
 
     const UNCHANGED_ON_FAILURE: bool = true;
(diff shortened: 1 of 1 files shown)


src/hugr/rewrite/outline_cfg.rs
```diff
@@ -389,15 +402,22 @@ mod test {
         for &n in blocks_to_move.iter().chain(other_blocks.iter()) {
             assert_eq!(depth(&h, n), 1);
         }
-        h.apply_rewrite(OutlineCfg::new(blocks_to_move.iter().copied()))
+        let (new_block, new_cfg) = h
+            .apply_rewrite(OutlineCfg::new(blocks_to_move.iter().copied()))
             .unwrap();
         h.validate(&PRELUDE_REGISTRY).unwrap();
-        let new_entry = h.children(h.root()).next().unwrap();
+        assert_eq!(new_block, h.children(h.root()).next().unwrap());
```
acl-cqc, line 409: `new_entry`, although correct, could also suggest the entry node inside the new sub-cfg, hence "renaming" to `new_block`

cqc-alec (approved):

PR #555, opened 2023-09-26
fix(Inference): Consider shunted metas when comparing equality

croyzor:
When comparing for equality of metavars, we need to consider that they might have been merged to form a new meta, so `UnificationContext::resolve` must be used. In this case, the variable (LHS) is updated to the merged value in `UnificationContext::merge_equal_metas` so we don't need to resolve that one.

### DIFF
--- a/src/extension/infer.rs
+++ b/src/extension/infer.rs
@@ -664,7 +664,9 @@ impl UnificationContext {
                 let solution =
                     ExtensionSet::from_iter(self.get_constraints(&m).unwrap().iter().filter_map(
                         |c| match c {
-                            Constraint::Plus(x, other_m) if &m == other_m => Some(x.clone()),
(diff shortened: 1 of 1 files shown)


src/extension/infer.rs
```diff
@@ -664,7 +664,9 @@ impl UnificationContext {
                 let solution =
                     ExtensionSet::from_iter(self.get_constraints(&m).unwrap().iter().filter_map(
                         |c| match c {
-                            Constraint::Plus(x, other_m) if &m == other_m => Some(x.clone()),
+                            Constraint::Plus(x, other_m) if m == self.resolve(*other_m) => {
```
ss2165, line 667: I think the following from the PR description is worth adding as a comment here: " In this case, the variable (LHS) is updated to the merged value in UnificationContext::merge_equal_metas so we don't need to resolve that one."

ss2165 (approved):

PR #557, opened 2023-09-26
refactor!: rename predicate to TupleSum/UnitSum

### LINKED ISSUE #448: Rename Predicate/SimplePredicate (to Choice + UnitChoice?)
I was thinking the same (BoundedUSize) but I take your argument. Really I think "predicate" is the wrong word to start. We're calling it that because we use it to control branches, but IMHO predicate is inherently binary - true or false - Google gives me this:

> 2.
> LOGIC
> something which is affirmed or denied concerning an argument of a proposition

So perhaps go with predicate for now but a better word might be "selector" or "choice"? or wilder ideas like "Nway"

_Originally posted by @acl-cqc in https://github.com/CQCL-DEV/hugr/pull/445#discussion_r1302914040_

ss2165:
Closes #448

BREAKING CHANGE: previous uses of type/value constructor methods for "predicates" will need updating

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -345,8 +345,8 @@ not. That is, Conditional-nodes act as "if-then-else" followed by a
 control-flow merge.
 
 A **TupleSum(T0, T1…TN)** type is an algebraic “sum of products” type,
-defined as `Sum(Tuple(#t0), Tuple(#t1), ...Tuple(#tn))` (see [type
-system](#type-system)), where `#ti` is the *i*th Row defining it.
+defined as `Sum(Tuple(#T0), Tuple(#T1), ...Tuple(#Tn))` (see [type
+system](#type-system)), where `#Ti` is the *i*th Row defining it.
 
 ```mermaid
 flowchart
@@ -424,7 +424,7 @@ output of each of these is a sum type, whose arity is the number of outgoing
 control edges; the remaining outputs are those that are passed to all
 succeeding nodes.
 
-The three nodes labelled "Const" are simply generating a choice with one empty
+The three nodes labelled "Const" are simply generating a TupleSum with one empty
 value to pass to the Output node.
 
 ```mermaid
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -335,16 +335,16 @@ express control flow, i.e. conditional or repeated evaluation.
 ##### `Conditional` nodes
 
 These are parents to multiple `Case` nodes; the children have no edges.
-The first input to the Conditional-node is of Predicate type (see below), whose
+The first input to the Conditional-node is of Choice type (see below), whose
 arity matches the number of children of the Conditional-node. At runtime
 the constructor (tag) selects which child to execute; the unpacked
-contents of the Predicate with all remaining inputs to Conditional
+contents of the Choice with all remaining inputs to Conditional
 appended are sent to this child, and all outputs of the child are the
 outputs of the Conditional; that child is evaluated, but the others are
 not. That is, Conditional-nodes act as "if-then-else" followed by a
 control-flow merge.
 
-A **Predicate(T0, T1…TN)** type is an algebraic “sum of products” type,
+A **Choice(T0, T1…TN)** type is an algebraic “sum of products” type,
 defined as `Sum(Tuple(#t0), Tuple(#t1), ...Tuple(#tn))` (see [type
 system](#type-system)), where `#ti` is the *i*th Row defining it.
 
@@ -362,7 +362,7 @@ flowchart
         end
         Case0 ~~~ Case1
     end
-    Pred["case 0 inputs | case 1 inputs"] --> Conditional
+    Choice["case 0 inputs | case 1 inputs"] --> Conditional
     OI["other inputs"] --> Conditional
     Conditional --> outputs
 ```
@@ -370,7 +370,7 @@ flowchart
 ##### `TailLoop` nodes
 
 These provide tail-controlled loops: the data sibling graph within the
-TailLoop-node computes a value of 2-ary `Predicate(#i, #o)`; the first
+TailLoop-node computes a value of 2-ary `Choice(#i, #o)`; the first
 variant means to repeat the loop with the values of the tuple unpacked
 and “fed” in at at the top; the second variant means to exit the loop
 with those values unpacked. The graph may additionally take in a row
@@ -398,7 +398,7 @@ The first child is the entry block and must be a `DFB`, with inputs the same as 
 The remaining children are either `DFB`s or [scoped definitions](#scoped-definitions).
 
 The first output of the DSG contained in a `BasicBlock` has type
-`Predicate(#t0,...#t(n-1))`, where the node has `n` successors, and the
+`Choice(#t0,...#t(n-1))`, where the node has `n` successors, and the
 remaining outputs are a row `#x`. `#ti` with `#x` appended matches the
 inputs of successor `i`.
 
@@ -424,7 +424,7 @@ output of each of these is a sum type, whose arity is the number of outgoing
 control edges; the remaining outputs are those that are passed to all
 succeeding nodes.
 
-The three nodes labelled "Const" are simply generating a predicate with one empty
+The three nodes labelled "Const" are simply generating a choice with one empty
 value to pass to the Output node.
 
 ```mermaid
@@ -1376,8 +1376,8 @@ use an empty node in the replacement and have B map this node to the old
 one.
 
 We can, for example, implement “turning a Conditional-node with known
-predicate into a DFG-node” by a `Replace` where the Conditional (and its
-preceding predicate) is replaced by an empty DFG and the map B specifies
+Choice into a DFG-node” by a `Replace` where the Conditional (and its
+preceding Choice) is replaced by an empty DFG and the map B specifies
 the “good” child of the Conditional as the surrogate parent of the new
 DFG’s children. (If the good child was just an Op, we could either
 remove it and include it in the replacement, or – to avoid this overhead
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1,4 +1,4 @@
-# HUGR design document, Draft 3
+# HUGR design document
 
 The Hierarchical Unified Graph Representation (HUGR, pronounced *hugger* 
 🫂) is a proposed new
@@ -335,16 +335,16 @@ express control flow, i.e. conditional or repeated evaluation.
 ##### `Conditional` nodes
 
 These are parents to multiple `Case` nodes; the children have no edges.
-The first input to the Conditional-node is of Choice type (see below), whose
+The first input to the Conditional-node is of TupleSum type (see below), whose
 arity matches the number of children of the Conditional-node. At runtime
 the constructor (tag) selects which child to execute; the unpacked
-contents of the Choice with all remaining inputs to Conditional
+contents of the TupleSum with all remaining inputs to Conditional
 appended are sent to this child, and all outputs of the child are the
 outputs of the Conditional; that child is evaluated, but the others are
 not. That is, Conditional-nodes act as "if-then-else" followed by a
 control-flow merge.
 
-A **Choice(T0, T1…TN)** type is an algebraic “sum of products” type,
+A **TupleSum(T0, T1…TN)** type is an algebraic “sum of products” type,
 defined as `Sum(Tuple(#t0), Tuple(#t1), ...Tuple(#tn))` (see [type
 system](#type-system)), where `#ti` is the *i*th Row defining it.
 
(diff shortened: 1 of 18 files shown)


specification/hugr.md
```diff
@@ -424,7 +424,7 @@ output of each of these is a sum type, whose arity is the number of outgoing
 control edges; the remaining outputs are those that are passed to all
 succeeding nodes.
 
-The three nodes labelled "Const" are simply generating a predicate with one empty
+The three nodes labelled "Const" are simply generating a choice with one empty
```
acl-cqc, line 427: If a Choice (capitalized) is a type, is the corresponding value a choice (lowercase)? Or should "choice" here be "value of Choice type"? (the "one empty value to pass" might need changing too, not quite sure I understand that

specification/hugr.md
```diff
@@ -335,16 +335,16 @@ express control flow, i.e. conditional or repeated evaluation.
...
 the constructor (tag) selects which child to execute; the unpacked
-contents of the Predicate with all remaining inputs to Conditional
+contents of the Choice with all remaining inputs to Conditional
 appended are sent to this child, and all outputs of the child are the
 outputs of the Conditional; that child is evaluated, but the others are
 not. That is, Conditional-nodes act as "if-then-else" followed by a
 control-flow merge.
 
-A **Predicate(T0, T1…TN)** type is an algebraic “sum of products” type,
+A **Choice(T0, T1…TN)** type is an algebraic “sum of products” type,
```
acl-cqc, line 347: Drive-by, but I think we should be consistent between `T<i>` on this line and `#t<i>` on the next

src/builder/cfg.rs
```diff
@@ -152,14 +152,14 @@ impl<B: AsMut<Hugr> + AsRef<Hugr>> CFGBuilder<B> {
...
-            predicate_variants,
+            choice_variants,
             other_outputs,
             inputs,
         )
     }
 
     /// Return a builder for a non-entry [`BasicBlock::DFB`] child graph with `inputs`
-    /// and `outputs` and a simple predicate type: a Sum of `n_cases` unit types.
+    /// and `outputs` and a unit choice type: a Sum of `n_cases` unit types.
```
acl-cqc, line 162:
```suggestion
    /// and `outputs` and a unit Choice type: a Sum of `n_cases` unit types.
```

src/builder/cfg.rs
```diff
@@ -196,15 +196,15 @@ impl<B: AsMut<Hugr> + AsRef<Hugr>> CFGBuilder<B> {
...
+            choice_variants,
             other_outputs,
             extension_delta,
             true,
         )
     }
 
     /// Return a builder for the entry [`BasicBlock::DFB`] child graph with `inputs`
-    /// and `outputs` and a simple predicate type: a Sum of `n_cases` unit types.
+    /// and `outputs` and a unit choice type: a Sum of `n_cases` unit types.
```
acl-cqc, line 207:
```suggestion
    /// and `outputs` and a unit Choice type: a Sum of `n_cases` unit types.
```

src/builder/cfg.rs
```diff
@@ -245,7 +245,7 @@ pub type BlockBuilder<B> = DFGWrapper<B, BasicBlockID>;
 
 impl<B: AsMut<Hugr> + AsRef<Hugr>> BlockBuilder<B> {
     /// Set the outputs of the block, with `branch_wire` being the value of the
-    /// predicate.  `outputs` are the remaining outputs.
+    /// Choice.  `outputs` are the remaining outputs.
```
acl-cqc, line 248: Why is it `branch_wire` not `choice_wire` (why wasn't it `predicate_wire` before)?

src/ops/constant.rs
```diff
@@ -34,41 +34,41 @@ impl Const {
...
         }
     }
 
     /// Constant Sum over units, with only one variant.
-    pub fn simple_unary_predicate() -> Self {
+    pub fn unary_unit_choice() -> Self {
         Self {
-            value: Value::simple_unary_predicate(),
-            typ: Type::new_simple_predicate(1),
+            value: Value::unary_unit_choice(),
```
acl-cqc, line 59: or just call `Self::unit_choice(0,1)`

src/ops/validate.rs
```diff
@@ -146,18 +146,18 @@ impl ValidateOp for super::Conditional {
...
-        if self.predicate_inputs.len() != children.len() {
-            return Err(ChildrenValidationError::InvalidConditionalPredicate {
+        if self.choice_inputs.len() != children.len() {
+            return Err(ChildrenValidationError::InvalidConditionalChoice {
                 child: children[0].0, // Pass an arbitrary child
                 expected_count: children.len(),
-                actual_count: self.predicate_inputs.len(),
-                actual_predicate_rows: self.predicate_inputs.clone(),
+                actual_count: self.choice_inputs.len(),
+                actual_choice_rows: self.choice_inputs.clone(),
```
acl-cqc, line 156: just a side note, elsewhere this would be "choice_variants". I wonder whether "choice_rows" would be better everywhere

src/types.rs
```diff
@@ -241,17 +241,17 @@ impl Type {
...
     where
         V: Into<TypeRow>,
     {
-        Self::new_sum(predicate_variants_row(variant_rows))
+        Self::new_sum(choice_variant_row(variant_rows))
     }
 
-    /// New simple predicate with empty Tuple variants
-    pub const fn new_simple_predicate(size: u8) -> Self {
+    /// New unit choice with empty Tuple variants
```
acl-cqc, line 253:
```suggestion
    /// New unit Choice with empty Tuple variants
```

src/types.rs
```diff
@@ -142,7 +142,7 @@ impl SumType {
 impl From<SumType> for Type {
     fn from(sum: SumType) -> Type {
         match sum {
-            SumType::Simple { size } => Type::new_simple_predicate(size),
+            SumType::Simple { size } => Type::new_unit_choice(size),
```
acl-cqc, line 145: Reckon you should rename SumType::Simple. Maybe SumType::OfUnits ? (SumType::Unit perhaps but seems more likely to lead to confusion)

src/types/serialize.rs
```diff
@@ -80,8 +80,8 @@ mod test {
         let t = Type::new_sum(vec![USIZE_T, FLOAT64_TYPE]);
         assert_eq!(ser_roundtrip(&t), t);
 
-        // A simple predicate
-        let t = Type::new_simple_predicate(4);
+        // A unit choice
```
acl-cqc, line 83:
```suggestion
        // A unit Choice
```
But I don't think this comment adds much tbh....(!)

src/values.rs
```diff
@@ -96,14 +96,14 @@ impl Value {
         Self::Tuple { vs: vec![] }
     }
 
-    /// Constant Sum over units, used as predicates.
-    pub fn simple_predicate(tag: usize) -> Self {
+    /// Constant Sum over units, used as branching values.
```
acl-cqc, line 99:
```suggestion
    /// Constant Sum of a unit value, used to control branches.
```
(Other variants could be non-unit, since we are checkable not synthesizable here)

src/values.rs
```diff
@@ -96,14 +96,14 @@ impl Value {
...
+    pub fn unit_choice(tag: usize) -> Self {
         Self::sum(tag, Self::unit())
     }
 
     /// Constant Sum over Tuples with just one variant of unit type
```
acl-cqc, line 104:
```suggestion
    /// Constant Sum with just one variant of unit type
```

src/values.rs
```diff
@@ -113,7 +113,7 @@ impl Value {
         }
     }
 
-    /// Sum value (could be of any compatible type, e.g. a predicate)
+    /// Sum value (could be of any compatible type, e.g. a Choice)
```
acl-cqc, line 116:
```suggestion
    /// Sum value (could be of any compatible type - e.g., if `value` was a Tuple, a Choice type)
```

acl-cqc (commented):
Yes-ish. I think I'd like @cqc-alec's looking over here, not for nitpicking details, but whether this name-change "works" overall.
* "Unit Choice" is clearly an improvement over "Simple Predicate", but then "Unit Predicate" would have been too
* My general feeling is I'm not as persuaded as I thought I'd be. I guess I just feel that one is far more likely to use the word choice by accident, when one didn't mean a Choice, and in spoken English capitalization is hard to determine.
* Perhaps this is OK if we are 100% consistent on always capitalizing here, and it may also help to tighten up the distinction between Choice types and Choice values (if that's what they are called) - by which perhaps I mean never to use "Choice" on its own without being immediately followed by either "type" or "value"...
* I do like "choice_variants" or "choice_rows", too. Perhaps a general theme here that maybe Choice works better if we always use it as an adjective not as a noun?

I'm not sure whether I see a good alternative that is less likely to be used by accident. "Selector" perhaps, but is that better than Predicate? "Chooser"? Partway through I was actually wondering if we could just say "sum-of-tuples" explicitly, but then I gave up when I looked at the builder code....

cqc-alec:
> Yes-ish. I think I'd like @cqc-alec's looking over here, not for nitpicking details, but whether this name-change "works" overall.
> 
>     * "Unit Choice" is clearly an improvement over "Simple Predicate", but then "Unit Predicate" would have been too

Or "Simple Choice"?

>     * My general feeling is I'm not as persuaded as I thought I'd be. I guess I just feel that one is far more likely to use the word choice by accident, when one didn't mean a Choice, and in spoken English capitalization is hard to determine.
> 
>     * Perhaps this is OK if we are 100% consistent on always capitalizing here, and it may also help to tighten up the distinction between Choice types and Choice values (if that's what they are called) - by which perhaps I mean never to use "Choice" on its own without being immediately followed by either "type" or "value"...
> 
>     * I do like "choice_variants" or "choice_rows", too. Perhaps a general theme here that maybe Choice works better if we always use it as an adjective not as a noun?
> 
> 
> I'm not sure whether I see a good alternative that is less likely to be used by accident. "Selector" perhaps, but is that better than Predicate? "Chooser"? Partway through I was actually wondering if we could just say "sum-of-tuples" explicitly, but then I gave up when I looked at the builder code....

"Choice" seems better than "predicate" but perhaps it isn't ideal. We could use "variant" (variant type or variant value), as in C++17. That can be heard as a noun or adjective, but I'm not sure if that ambiguity matters. I'm less keen on "chooser" or "selector" because they imply an _action_ that is not really part of the type.

ss2165:
what about borrowing SOP (sum of products) from boolean algebra?

cqc-alec:
> what about borrowing SOP (sum of products) from boolean algebra?

Has the advantage of being precise, but I feel like a more understandable name (for exactly that concept) would be preferable. If "Choice" and "Variant" are a bit too vague, could we use "DataChoice" or "DataVariant"? Not really convinced, just trying out ideas...

ss2165:
Variant is too close to Sum for me (indeed we use it in Tierkreis instead of Sum). How about just simple `SumTuple` and `SumUnit` then?
`new_sum_tuple`, `new_sum_unit` etc.
We could keep using Choice just as a shorthand in the spec, and use precise naming in the code.

cqc-alec:
> Variant is too close to Sum for me (indeed we use it in Tierkreis instead of Sum). How about just simple `SumTuple` and `SumUnit` then? `new_sum_tuple`, `new_sum_unit` etc. We could keep using Choice just as a shorthand in the spec, and use precise naming in the code.

I'd personally find it confusing to have different terminology used in the spec and implementation. I'd be OK with something like `SumTuple` and `SumUnit`, though I would turn these around to `TupleSum` and `UnitSum`!

acl-cqc:
If `TupleSum` seems ok (I guess it's no longer than predicate!) then that could work. I'm wondering if "choice value" and "choice type" might be alright tho - specifically, if we always use choice as an adjective, then that's different from its normal/accidental use in English, i.e. as a noun. Perhaps `Type::new_choice` is obviously-enough a new choice *type* to avoid having to write `Type::new_choice_type`...

ss2165:
Yeah I see what you mean Alan, `Choice` reads nicer in the spec, but I'm happy with `TupleSum` and `UnitSum` everywhere

cqc-alec:
> Yeah I see what you mean Alan, `Choice` reads nicer in the spec, but I'm happy with `TupleSum` and `UnitSum` everywhere

OK for me

cqc-alec:
Sorry closed by mistake!

cqc-alec (commented):
Do we go with `TupleSum` and `UnitSum` then?

src/ops/validate.rs
ss2165: yes I've tried to go with "tuple_sum_rows" everywhere

src/extension/infer.rs
```diff
@@ -1209,11 +1210,11 @@ mod test {
         entry_extensions: ExtensionSet,
         exit_types: impl Into<TypeRow>,
     ) -> Result<([Node; 3], Node), Box<dyn Error>> {
-        let entry_predicate_type = Type::new_predicate(entry_predicates.clone());
+        let entry_predicate_type = Type::new_tuple_sum(entry_predicates.clone());
```
cqc-alec, line 1213: Should we rename `entry_predicates`?

cqc-alec (approved):

PR #558, opened 2023-09-27
feat: Make TypeEnum and it's contents public

doug-q:
A missing doc is added for `PrimType::check_type` while I was there.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -11,6 +11,7 @@ pub mod type_row;
 pub use check::{ConstTypeError, CustomCheckFailure};
 pub use custom::CustomType;
 pub use signature::{FunctionType, Signature, SignatureDescription};
+pub use type_param::TypeArg;
 pub use type_row::TypeRow;
 
 use derive_more::{From, Into};
@@ -23,7 +24,7 @@ use crate::ops::AliasDecl;
 use crate::type_row;
 use std::fmt::Debug;
 
-use self::primitive::PrimType;
+pub use self::primitive::PrimType;
 
 #[cfg(feature = "pyo3")]
 use pyo3::pyclass;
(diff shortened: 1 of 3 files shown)


src/types.rs
```diff
@@ -103,12 +104,13 @@ pub(crate) fn least_upper_bound(mut tags: impl Iterator<Item = TypeBound>) -> Ty
     .into_inner()
 }
 
+#[allow(missing_docs)]
 #[derive(Clone, PartialEq, Debug, Eq, derive_more::Display, Serialize, Deserialize)]
 #[serde(tag = "s")]
 /// Representation of a Sum type.
 /// Either store the types of the variants, or in the special (but common) case
 /// of a "simple predicate" (sum over empty tuples), store only the size of the predicate.
```
ss2165, line 112: i think this would allow no docs on the enum itself, maybe instead remove the #[allow] and split the docstring up between the variants (it currently describes both of them at the top)

src/types/primitive.rs
```diff
@@ -4,10 +4,11 @@ use crate::ops::AliasDecl;
 
 use super::{CustomType, FunctionType, TypeBound};
 
+#[allow(missing_docs)]
```
ss2165, line 7: i think this should have a docstring, at least for the enum itself, I think you can place the `allow` within the enum definition to allow missing docs on variants

ss2165 (changes requested):
looks good, just a couple of docs changes

src/types.rs
doug-q: I like the single comment approach and so have moved the #[allow ] to the variants. Will change if you prefer.
I applied the same change to TypeEnum.

src/types/primitive.rs
doug-q: Done

ss2165 (approved):
👍

PR #562, opened 2023-09-27
fix: Comment out broken priority code

croyzor:
Setting the priority in Jira didn't work because it's illegal to assign a list to a variable (apparently).
Until such a time that it's fixed, comment out the broken line so that we can still add issues

### DIFF
--- a/.github/workflows/issue.yml
+++ b/.github/workflows/issue.yml
@@ -50,7 +50,9 @@ jobs:
         issuetype: Bug
         summary: « [hugr] ${{ github.event.issue.title }}»
         description: ${{ github.event.issue.html_url }}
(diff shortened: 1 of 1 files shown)


aborgna-q (approved):
Failing test for reference: https://github.com/CQCL-DEV/hugr/actions/runs/6326625908/job/17180598059

You may need to do some json encoding...
https://stackoverflow.com/questions/75391000/define-list-of-elements-in-a-github-action-variable

croyzor:
New failure: https://github.com/CQCL-DEV/hugr/actions/runs/6327027264
I've updated this PR to comment out a different part that hopefully gets back to a working state while I think about this error

PR #563, opened 2023-09-27
fix: Join labels in issue workflow

croyzor:
Instead of merging #562, I think this should work to fix the problem


aborgna-q (approved):
LGTM

PR #566, opened 2023-09-27
feat: Optional direction check when querying a port index

### LINKED ISSUE #565: `HugrMut::connect` does not handle backwards edges
A function call of the kind

```rust
hugr.connect(n1, Port::new_incoming(0), n2, Port::new_outgoing(1))
```

will fail with an `UnknownOffset` error, because it will ignore the port directions and assume the first port is outgoing, the second is incoming. I feel that we should either
- keep the current behaviour but change the API to take port offsets instead of Port,
- accept arguments that are in the reversed order, as above, but handle them accordingly.

I prefer the latter. At a minimum, calling this with arguments in the wrong order should return an error.

aborgna-q:
So `connect` throws an error when passed invalid ports, but we still support `usize`s.

Closes #565. I choose not to allow backwards edges as that introduces a new failure mode when the directions are the same (and weird behaviour when passed one `Port` and one `usize`).

### DIFF
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -28,7 +28,7 @@ use super::{
     tail_loop::TailLoopBuilder, BuildError, Wire,
 };
 
-use crate::Hugr;
+use crate::{Direction, Hugr};
 
 use crate::hugr::HugrMut;
 
@@ -662,8 +662,8 @@ fn wire_up<T: Dataflow + ?Sized>(
     dst: Node,
     dst_port: impl PortIndex,
 ) -> Result<bool, BuildError> {
-    let src_port = src_port.index();
-    let dst_port = dst_port.index();
+    let src_port = src_port.try_index(Direction::Outgoing)?;
+    let dst_port = dst_port.try_index(Direction::Incoming)?;
     let base = data_builder.hugr_mut();
     let src_offset = Port::new_outgoing(src_port);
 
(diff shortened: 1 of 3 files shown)


aborgna-q:
Accepting ports in reverse order in `connect` is possible, but it'd mean that in
```rust
hugr.connect(n1, Port::new_outgoing(4), n2, 3);
hugr.connect(n1, Port::new_incoming(4), n2, 3);
```
the `3` offset would refer to an input or output depending on the other port.

I feel that may lead to buggy code, but I'm happy to implement it anyways.

lmondada (approved):
Excellent, thank you! 

I'm finding the `PortIndex` trait confusing for two reasons
- it's a very different thing from the `PortIndex` in `portgraph`,
- it seems a slightly hacky/untransparent way of accepting both integer and ports as function arguments.

I have a suggestion you can consider: replace `trait PortIndex` with two traits: `trait IncomingPort` and `trait OutgoingPort`. `usize` would implement both, and for `Port` you could implement `TryFrom<Port>` for both traits.

Then in functions you take `impl TryInto<IncomingPort>` and `impl TryInto<IncomingPort>` arguments. At least for me, this would make the API clearer.

lmondada:
> Then in functions you take `impl TryInto<IncomingPort>` and `impl TryInto<OutgoingPort>` arguments. At least for me, this would make the API clearer.

Or leave it to the client to do the `.try_into()` conversion.

lmondada (approved):
Let's go 🚀 

Is `PortIndex` still needed?

aborgna-q:
Not strictly needed, but it unifies all the thinks you can call `index()` on.

PR #569, opened 2023-09-27
RFC: {infer=>intern}_and_validate calls resolve_extension_ops

### LINKED ISSUE #508: Integrate resolve_extension_ops into validate / infer_and_validate

acl-cqc:
There are comments in validate that it'd be good to call `resolve_extension_ops` immediately before validation, and those comments are correct ;-). (Thus, fixes #508.)

Moreover, we actually have a place that mutates the Hugr immediately before validation. It's currently called `infer_and_validate`, but that can be changed, so let's use that.

To my surprise, this didn't fail any tests. I guess we don't have any that validate Hugrs containing `ExternalOp::Opaque` instances! I shall probably add some shortly, but in the meantime....

The big question here is what to call `infer_and_validate` now it also includes `resolve_extension_ops`. `resolve_infer_validate` is too long. A look in the thesaurus suggested things like `normalize`, `internalize`, `regulize` - hence what I have here. I also wonder about using the word `update` - `update_and_validate`, `update_validate`, `valid_update`, `valiupdate` (!), `upvalidate` etc....? Other ideas?

There is another alternative, which is to make validation perform the same checks as `resolve_extension_ops` (i.e. that the data in the serialized OpaqueOp is consistent with the ExtensionRegistry) but without mutating the Hugr. Given that we'd also want to keep the mutating version, this is more work, but possibly it is more consistent with principle. <strike>Please shout if you think this is worth doing?</strike>EDIT: I'll give this a try; it's an extra step to the above, that I suppose could follow in another PR, but this may not be as hard as I thought, so I may be able to just do it here.

### DIFF
--- a/src/builder/cfg.rs
+++ b/src/builder/cfg.rs
@@ -73,7 +73,7 @@ impl HugrBuilder for CFGBuilder<Hugr> {
         mut self,
         extension_registry: &ExtensionRegistry,
     ) -> Result<Hugr, crate::hugr::ValidationError> {
-        self.base.infer_and_validate(extension_registry)?;
+        self.base.intern_and_validate(extension_registry)?;
         Ok(self.base)
     }
 }
--- a/src/builder/conditional.rs
+++ b/src/builder/conditional.rs
@@ -150,7 +150,7 @@ impl HugrBuilder for ConditionalBuilder<Hugr> {
         mut self,
         extension_registry: &ExtensionRegistry,
     ) -> Result<Hugr, crate::hugr::ValidationError> {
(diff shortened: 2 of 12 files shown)


acl-cqc:
Closing in favour of doing #583 first

PR #570, opened 2023-09-28
chore: Add priority to Jira issues

croyzor:
In Jira, "priority" is not a mere string field, but an object, whose "name" field must be specified. I think this is what was going wrong with the last attempt


aborgna-q (approved):
👍
Hopefully this works

PR #573, opened 2023-09-28
fix(Jira): Handling of issues with no priority label

croyzor:
The code seems to work! ...In the case where there is a priority label on the issue. Jira complains if the priority field is just "", so we need to get rid of it


aborgna-q (approved):
nth time's a charm

PR #581, opened 2023-09-29
[new + bugfix] replace_op checks bound, adding HugrError::InvalidTag

acl-cqc:
* Add HugrError::InvalidTag, use in HierarchyView::try_new and SiblingMut::try_new
* In HugrMut::replace_op, check that changing the root-node's op does not invalidate the Root: NodeHandle tag.
* Hence, make HugrMut::replace_op return a Result


aborgna-q (approved):
👍

PR #582, opened 2023-09-29
fix: Don't insert temporary wires when extracting a subgraph

aborgna-q:
We connected the temporary built graph inputs and outputs, but the logic didn't take into account reordering in the signature, or non-linear parameters.

We can avoid the headache by not validating the intermediate construct.

### DIFF
--- a/src/hugr/views/sibling_subgraph.rs
+++ b/src/hugr/views/sibling_subgraph.rs
@@ -10,13 +10,14 @@
 //! hierarchy.
 
 use std::collections::HashSet;
+use std::mem;
 
 use itertools::Itertools;
 use portgraph::{view::Subgraph, Direction, PortView};
 use thiserror::Error;
 
-use crate::builder::{Dataflow, DataflowHugr, FunctionBuilder};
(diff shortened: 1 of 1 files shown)


src/hugr/views/sibling_subgraph.rs
```diff
@@ -415,23 +416,20 @@ impl SiblingSubgraph {
...
             input_extensions,
         };
-        let builder = FunctionBuilder::new(name, signature).unwrap();
-        let inputs = builder.input_wires();
-        let mut extracted = builder
-            .finish_hugr_with_outputs(inputs, &PRELUDE_REGISTRY)
-            .unwrap();
+        let mut builder = FunctionBuilder::new(name, signature).unwrap();
+        // Take the unfinished Hugr from the builder, to avoid unnecessary
+        // validation checks that require connecting the inputs an outputs.
```
ss2165, line 421:
```suggestion
        // validation checks that require connecting the inputs and outputs.
```

ss2165 (approved):

PR #583, opened 2023-09-29
Resolve opaque ops during validation (no mutation)

acl-cqc:
* Refactor `resolve_extension_ops` to allow resolving, one node at a time, without mutation
* Use this in validation, to check OpaqueOp's are correct
* Improve the CustomOpError structure a bit
* Remove `check_opaque` - we've just done it properly, and it has no callers
* Update `infer.rs` tests - opaque operations must be from an unknown extension

### DIFF
--- a/src/hugr/validate.rs
+++ b/src/hugr/validate.rs
@@ -180,7 +180,6 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
             };
             // If successful, check TypeArgs are valid for the declared TypeParams
             if let Some(ext_op) = ext_op {
-                // WTF looks like check_opaque might be duplicating some of this?!
                 ext_op
                     .def()
                     .check_args(ext_op.args())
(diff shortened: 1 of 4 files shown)


src/hugr/validate.rs
```diff
@@ -159,14 +161,31 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
...
+                ExternalOp::Opaque(op) => {
+                    // If resolve_extension_ops has been called first, this would always return Ok(None)
+                    e = resolve_opaque_op(node, op, self.extension_registry)?;
+                    e.as_ref()
+                }
+                ExternalOp::Extension(ext) => Some(ext),
+            };
+            // If successful, check TypeArgs are valid for the declared TypeParams
+            if let Some(ext_op) = ext_op {
+                // WTF looks like check_opaque might be duplicating some of this?!
```
acl-cqc, line 183: Actually I think I can just drop this comment and maybe sort that out in another PR.

ss2165 (approved):
much nicer!

PR #584, opened 2023-09-29
[refactor] Remove never-called OpDef::instantiate_opaque

acl-cqc:
`Extension::instantiate_extension_op` is the main user-facing entry point, and already calls ExtensionOp::new(), as it should.


ss2165 (approved):
Happy to remove this, we can consider broader interface simplifications once we know whether OpaqueOp and ExtensionOp are both necessary.

PR #585, opened 2023-10-02
fix: Improve convexity checking and fix test

### LINKED ISSUE #518: SiblingSubgraph does not handle Copies at output boundary
The following test

```rust
    #[test]
    fn edge_both_output_and_copy() {
        let one_bit = type_row![BOOL_T];
        let two_bit = type_row![BOOL_T, BOOL_T];

        let mut builder =
            DFGBuilder::new(FunctionType::new(one_bit.clone(), two_bit.clone())).unwrap();
        let inw = builder.input_wires().exactly_one().unwrap();
        let outw1 = builder.add_dataflow_op(not_op(), [inw]).unwrap().out_wire(0);
        let outw2 = builder
            .add_dataflow_op(and_op(), [inw, outw1])
            .unwrap()
            .outputs();
        let outw = [outw1]
            .into_iter()
            .chain(outw2);
        let h = builder.finish_hugr_with_outputs(outw, &EMPTY_REG).unwrap();
        let view = SiblingGraph::<DfgID>::new(&h, h.root());
        let subg = SiblingSubgraph::try_new_dataflow_subgraph(&view).unwrap();
        assert_eq!(subg.nodes().len(), 1);
    }
```

encodes the computation

```mermaid
graph TD
IN --> NOT
IN --> AND
NOT --> AND
NOT --> OUT
AND --> OUT
```

This currently fails with the error `NotConvex` as the output of the `NOT` is taken to be an output of the subgraph, even though one copy is actually still within the subgraph and used for the `AND` gate.

cqc-alec:
Fixes #518 .

I changed the `non_convex_subgraph` test because the graph it was constructing seemed to meet the criteria for convexity, and replaced it with another one which doesn't. But please check -- I was a little confused by the definitions and may be missing something.

### DIFF
--- a/src/hugr/views/sibling_subgraph.rs
+++ b/src/hugr/views/sibling_subgraph.rs
@@ -517,16 +517,13 @@ fn validate_subgraph<H: HugrView>(
     }
 
     let mut ports_inside = inputs.iter().flatten().chain(outputs).copied();
-    let mut ports_outside = ports_inside
-        .clone()
-        .flat_map(|(n, p)| hugr.linked_ports(n, p));
     // Check incoming & outgoing ports have target resp. source inside
     let nodes = nodes.iter().copied().collect::<HashSet<_>>();
     if ports_inside.any(|(n, _)| !nodes.contains(&n)) {
         return Err(InvalidSubgraph::InvalidBoundary);
     }
-    // Check incoming & outgoing ports have source resp. target outside
-    if ports_outside.any(|(n, _)| nodes.contains(&n)) {
+    // Check that every inside port has at least one linked port outside.
+    if ports_inside.any(|(n, p)| hugr.linked_ports(n, p).all(|(n1, _)| nodes.contains(&n1))) {
         return Err(InvalidSubgraph::NotConvex);
     }
 
--- a/src/hugr/views/sibling_subgraph.rs
+++ b/src/hugr/views/sibling_subgraph.rs
@@ -662,10 +662,13 @@ mod tests {
         hugr::views::{HierarchyView, SiblingGraph},
         hugr::HugrMut,
         ops::{
-            handle::{FuncID, NodeHandle},
+            handle::{DfgID, FuncID, NodeHandle},
             OpType,
         },
-        std_extensions::{logic::test::and_op, quantum::test::cx_gate},
+        std_extensions::{
+            logic::test::{and_op, not_op},
+            quantum::test::cx_gate,
+        },
         type_row,
     };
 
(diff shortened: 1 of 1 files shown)


src/hugr/views/sibling_subgraph.rs
```diff
@@ -474,8 +474,6 @@ fn get_edge_type<H: HugrView>(hugr: &H, ports: &[(Node, Port)]) -> Option<Type>
 }
 
 /// Whether a subgraph is valid.
```
lmondada, line 476:
```suggestion
/// Whether a subgraph is valid.
///
/// Verifies that input and output ports are valid subgraph boundaries, i.e. they belong
/// to nodes within the subgraph and are linked to at least one node outside of the subgraph.
/// This does NOT check convexity proper, i.e. whether the set of nodes form a convex
/// induced graph.
```

src/hugr/views/sibling_subgraph.rs
```diff
@@ -517,16 +515,34 @@ fn validate_subgraph<H: HugrView>(
...
     // Check incoming & outgoing ports have target resp. source inside
     let nodes = nodes.iter().copied().collect::<HashSet<_>>();
     if ports_inside.any(|(n, _)| !nodes.contains(&n)) {
         return Err(InvalidSubgraph::InvalidBoundary);
     }
-    // Check incoming & outgoing ports have source resp. target outside
-    if ports_outside.any(|(n, _)| nodes.contains(&n)) {
+    // Check that every inside port has at least one linked port outside.
+    if ports_inside.any(|(n, p)| hugr.linked_ports(n, p).all(|(n1, _)| nodes.contains(&n1))) {
+        return Err(InvalidSubgraph::NotConvex);
```
lmondada, line 525: I notice that my use of the notion of convexity is a bit misleading here. Maybe replacing the `NotConvex` error with `InvalidBoundary` would be more appropriate. The function could also be renamed to e.g. `validate_boundary_ports`.

Feel free to choose whatever you find clearer.

src/hugr/views/sibling_subgraph.rs
```diff
@@ -854,17 +890,21 @@ mod tests {
...
+        let not1_out = hugr.node_outputs(not1).next().unwrap();
+        let not3_inp = hugr.node_inputs(not3).next().unwrap();
+        let not3_out = hugr.node_outputs(not3).next().unwrap();
         assert!(matches!(
             SiblingSubgraph::try_new(
-                vec![vec![(out, snd_cx_edge)]],
-                vec![(inp, first_cx_edge)],
-                &func,
+                vec![vec![(not1, not1_inp)], vec![(not3, not3_inp)]],
+                vec![(not1, not1_out), (not3, not3_out)],
```
lmondada, line 906: The updated test is definitely a good example of non-convex subgraph. The old test was more of the pathological kind. But it should still fail the (updated) convexity check, doesn't it?

If possible it would be good to keep the old test too. Another, slightly less pathological but also non-convex subgraph should be
```rust
        let (hugr, func_root) = build_hugr().unwrap();
        let func: SiblingGraph<'_> = SiblingGraph::try_new(&hugr, func_root).unwrap();
        let (inp, out) = hugr.children(func_root).take(2).collect_tuple().unwrap();
        let cx_edges_in = hugr.node_outputs(inp);
        let cx_edges_out = hugr.node_inputs(out);
        // All graph but the CX
        assert!(matches!(
            SiblingSubgraph::try_new(
                cx_edges_out.map(|p| vec![(out, p)]).collect(),
                cx_edges_in.map(|p| (in, p)).collect(),
                &func,
            ),
            Err(InvalidSubgraph::NotConvex)
        ));
```

Does this make sense (I didnt run it)?

lmondada (commented):
Hi Alec,

Thanks a lot for going through this again. I agree it's confusing.

I hope my comments make sense. I wanted to attempt to clarify things further but I will have to come back to this in ~an hour, I need to finish some other benchmarks asap.

Let me know what you think.

src/hugr/views/sibling_subgraph.rs
cqc-alec: > Does this make sense (I didnt run it)?

Yes I'll add this test, but it produces `InvalidBoundary` rather than `NotConvex`. Is this what you would expect?

PR #587, opened 2023-10-04
feat: Better subgraph verification errors

aborgna-q:
Added a `InvalidSubgraphBoundary` error variant so we get some info when the subgraph verification fails.

Fixes:
- It seems that the check for boundary nodes having outside neighbours (now `DisconnectedBoundaryPort`) was using an exhausted iterator, so it was not running.

  With that fixed, now I get errors in the `portmatching` crate.
  
- `try_new_dataflow_subgraph` would return invalid subgraphs when the input is directly connected to an output.
  It now ignores the wires. I'll add an issue to discuss what should be best behaviour in those cases.

### DIFF
--- a/src/hugr/views/sibling_subgraph.rs
+++ b/src/hugr/views/sibling_subgraph.rs
@@ -508,19 +508,19 @@ fn validate_subgraph<H: HugrView>(
     }
 
     // Check inputs are incoming ports and outputs are outgoing ports
-    inputs
+    if let Some(&(n, p)) = inputs
         .iter()
         .flatten()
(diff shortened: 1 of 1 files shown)


lmondada (approved):
Thanks for the much improved code.

Any reason you prefer `.find().map_or(Ok(()), ...)` over `if ... .any() { return ...}`? I think I find the latter easier to read

aborgna-q:
I need to bound the failing element to report it in the error, so I went with
```rust
if let Some(...) = ... .find(...) { Err(...)?; };
```

PR #591, opened 2023-10-05
fix(Inference): Work harder in variable instantiation

### LINKED ISSUE #598: Fix variable instantiation in extension inference
Code has been implemented to find a minimum extension solution for metavariables which loop into themselves, but we can also arrive at the state during this phase of unification where we end up with *multiple* variables which form a loop, which should be treated similarly. Currently, when we reach this condition an error is thrown.

croyzor:
Extend the `instantiate_variables` method in extension inference to handle the case where we end up with variables which form loops of `Plus` constraints with other metavariables. This is required to get the rest of the test variants in `test_cfg_loops` to work

Resolves #598


src/extension/infer.rs
```diff
@@ -109,48 +110,71 @@ pub enum InferExtensionError {
...
 
-impl EqGraph {
-    /// Create a new `EqGraph`
-    fn new() -> Self {
-        EqGraph {
-            equalities: pg::Graph::new_undirected(),
-            node_map: HashMap::new(),
+macro_rules! impl_graph_container {
+    ($dir:ty) => {
+        impl GraphContainer<$dir> {
```
acl-cqc, line 120: I presume you can't do sthg like
```
impl<T: petgraph::EdgeType> GraphContainer<T> { ... }
```

croyzor: Hmm, I wrote this macro before I found out about the `EdgeType` trait, so I'll see if I can redo it!

src/extension/infer.rs
```diff
@@ -109,51 +110,67 @@ pub enum InferExtensionError {
 
 /// A graph of metavariables which we've found equality constraints for. Edges
```
acl-cqc, line 111: Remove `equality` here. If this was only for equality constraints, there'd be no need for a graph - equality being reflexive and transitive, you'd just be dividing things into sets, right. It's only for Plus constraints that we need a graph?

src/extension/infer.rs
```diff
@@ -109,51 +110,67 @@ pub enum InferExtensionError {
...
+    fn new_directed() -> Self {
+        GraphContainer {
+            graph: pg::Graph::new(),
+            node_map: HashMap::new(),
+        }
+    }
+}
+
+type EqGraph = GraphContainer<Undirected>;
+impl EqGraph {
```
acl-cqc, line 168: Note that as far as Rust is concerned, this is just `impl GraphContainer<Undirected>`, which you already have one of above. Perhaps just rename both `new_directed` and `new_undirected` to `new`? Rust will probably make you put in type annotations (such as `EqGraph::`!) if it's ambiguous

src/extension/infer.rs
croyzor: For the equality case we are dividing things into sets, but the mechanism for doing so is by finding the connected components of the undirected version of this graph

src/extension/infer.rs
```diff
@@ -663,31 +680,85 @@ impl UnificationContext {
...
+    /// Gather all the transitive dependencies (induced by constraints) of the
+    /// variables in the context.
+    fn search_variable_deps(&self) -> HashSet<Meta> {
+        let mut variable_scope: HashSet<Meta> = self.variables.clone();
+        let mut scope_size = 0;
+        let mut new_scope_size = variable_scope.len();
+        while new_scope_size > scope_size {
+            scope_size = new_scope_size;
+            for m in variable_scope.clone().iter() {
+                for c in self.get_constraints(m).unwrap().iter() {
```
acl-cqc, line 692: So it's not clear to me from this what happens when we call `self.get_constraints(m)` for an `m` *that we've seen before* (on a previous iteration of the outermost loop). Will we get anything back? If so, is there any chance that we'll actually get a new thing to add to `variable_scope`? Or is it only the new/never-seen-before values of `m` that lead to finding new things? The latter would allow us to rewrite this loop much more simply but I don't know whether it's true....

src/extension/infer.rs
```diff
@@ -663,31 +680,85 @@ impl UnificationContext {
         self.results()
     }
 
-    /// Instantiate all variables in the graph with the empty extension set.
+    /// Gather all the transitive dependencies (induced by constraints) of the
+    /// variables in the context.
+    fn search_variable_deps(&self) -> HashSet<Meta> {
+        let mut variable_scope: HashSet<Meta> = self.variables.clone();
+        let mut scope_size = 0;
+        let mut new_scope_size = variable_scope.len();
```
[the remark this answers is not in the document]
croyzor: I _think_ that doesn't work because `variable_scope` is borrowed as immutable for the loop condition and borrowed as mutable for the loop body

src/extension/infer.rs
croyzor: Done

src/extension/infer.rs
```diff
@@ -663,31 +680,85 @@ impl UnificationContext {
...
-                        },
-                    ));
-                self.add_solution(m, solution);
+                let mut solution = ExtensionSet::new();
+                // Is the solution complete - does this variable *only* refer to itself?
+                let mut solution_complete = true;
+
+                for c in self.get_constraints(&m).unwrap().iter() {
+                    if let Constraint::Plus(r, other_m) = c {
+                        // If `m` has been merged, [`self.variables`] entry
```
acl-cqc, line 728: Would this comment still be true if it were just at the moment `m` were brought into scope (in `for m in variable_scope.into_iter()`) ? I kinda feel it'l just giving me more things to worry about here (in the most complicated bit where I have enough to worry about as it is ;-) )

src/extension/infer.rs
```diff
@@ -663,31 +680,85 @@ impl UnificationContext {
...
+                        // will have already been updated to the merged
+                        // value by [`self.merge_equal_metas`] so we don't
+                        // need to worry about resolving it.
+                        let other_m = self.resolve(*other_m);
+                        solution.insert(r);
+                        if m != other_m {
+                            // This `m` depends on the solution of `other_m` so
+                            // `solution` is not enough info on its own.
+                            solution_complete = false;
+                            relations.add_edge(m, self.resolve(other_m));
```
acl-cqc, line 738: Do you need to `self.resolve` here? You just did `self.resolve(*other_m)` a moment ago

croyzor: Good point, I was being overly cautious 🙂

src/extension/infer.rs
```diff
@@ -663,31 +680,85 @@ impl UnificationContext {
...
-                            }
-                            _ => None,
-                        },
-                    ));
-                self.add_solution(m, solution);
+                let mut solution = ExtensionSet::new();
+                // Is the solution complete - does this variable *only* refer to itself?
+                let mut solution_complete = true;
+
+                for c in self.get_constraints(&m).unwrap().iter() {
```
acl-cqc, line 726: Trying a more-functional approach. See what you think.
```suggestion
let plus_constraints = self.get_constraints(&m).unwrap().iter().filter_map(match |c| {
   Constraint::Plus(r, other_m) => Some((r, self.resolve(*other_m)))
   _ => None
})
let (rs, other_ms) : (Vec<_>, Vec<_>) = plus_constraints.unzip();
let solution = ExtensionSet::from(rs); // We must have something like that right?
let unresolved_ms = other_ms.filter(|other_m| m != other_m); // probably other ways to do that!
if unresolved_ms.empty() {
    self.add_solution(m, solution);
} else {
    unresolved_ms.for_each(|other_m| relations.add_edge(m,other_m));
}
solutions.insert(m, solution);
```

src/extension/infer.rs
```diff
@@ -1509,17 +1580,14 @@ mod test {
...
-                ExtensionSet::new(),
-                ExtensionSet::new(),
-                ExtensionSet::new(),
-            ),
-            (just_a.clone(), ExtensionSet::new(), ExtensionSet::new()),
-            (ExtensionSet::new(), just_a.clone(), ExtensionSet::new()),
-            (ExtensionSet::new(), ExtensionSet::new(), just_a.clone()),
-        ];
-
+        let mut variants = Vec::new();
```
acl-cqc, line 1579: Hahaha! Neat :-) You might consider `flat_map` (*3) to avoid `mut` but like it either way :)

src/extension/infer.rs
```diff
@@ -663,31 +680,85 @@ impl UnificationContext {
...
+                solutions.insert(m, solution.clone());
+                if solution_complete {
+                    self.add_solution(m, solution);
+                }
+            }
+        }
+
+        // Strongly connected components are looping constraint dependencies.
+        // This means that each metavariable in the CC has the same solution.
+        for cc in relations.ccs() {
```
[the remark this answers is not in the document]
acl-cqc: (There are other ways, you could record the `m`s for which the first loop failed to find a solution, then tick them off here. But that'd be more expensive when debug_assert is disabled)

src/extension/infer.rs
croyzor: Yeah it should still be true

src/extension/infer.rs
croyzor: Nice!

src/extension/infer.rs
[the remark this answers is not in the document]
acl-cqc: It totally does and I've just pushed it!

src/extension/infer.rs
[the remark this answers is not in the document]
croyzor: We could add an assert that the number of nodes in `relations.css()` (flattened) is the same as the length of `variable_scope`?

src/extension/infer.rs
acl-cqc: Ok, so if each iteration made a set of new stuff to process in the next iteration, you could do away with the sizes and just have two sets - something like
```
let mut seen_variable_scope = HashSet::new();
let mut new_variables = self.variables.clone();
while !new_variables.empty() {
   let mut newer_variables = HashSet::new();
   for m in new_variables.iter() {
      ....add to newer_variables....
   }
   seen_variable_scope.extend(new_variables);
   new_variables = newer_variables
}
```
Moreover, this'll change your algorithm from O(n^2) to something closer to linear, so....

src/extension/infer.rs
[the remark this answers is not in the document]
acl-cqc: You'd need to filter out the number of those that were already in `self.solved` (there's an `if !self.solved.contains`)...

But the flattening idea is good - how about just that the number of nodes in flattened-`relations.css()` is equal to the number of *edges* in relations?

src/extension/infer.rs
croyzor: I've gone for the former

src/extension/infer.rs
```diff
@@ -107,53 +108,65 @@ pub enum InferExtensionError {
     EdgeMismatch(#[from] ExtensionError),
 }
 
-/// A graph of metavariables which we've found equality constraints for. Edges
-/// between nodes represent equality constraints.
-struct EqGraph {
-    equalities: pg::Graph<Meta, (), petgraph::Undirected>,
+/// A graph of metavariables connected by constraints.
+/// The edges represent `Equal` constraints in the undirected graph and `Plus`
```
acl-cqc, line 112: 👍

src/extension/infer.rs
```diff
@@ -107,53 +108,65 @@ pub enum InferExtensionError {
...
                     .collect()
             })
             .collect()
     }
 }
 
+impl GraphContainer<Undirected> {
+    fn new() -> Self {
+        GraphContainer {
+            graph: pg::Graph::new_undirected(),
```
acl-cqc, line 153: Is this actually different to `pg::Graph::new`? You could just
```
impl<T:EdgeType> GraphContainer<T> {
   fn new() -> Self { ... }
}
```
Or, have different names for each - I know `EqGraph::new_undirected` is a little bit redundant/unnecessary, but one might say harmless

src/extension/infer.rs
```diff
@@ -663,31 +676,98 @@ impl UnificationContext {
...
-    /// Instantiate all variables in the graph with the empty extension set.
+    /// Gather all the transitive dependencies (induced by constraints) of the
+    /// variables in the context.
+    fn search_variable_deps(&self) -> HashSet<Meta> {
+        let mut seen = HashSet::new();
+        let mut new_variables: HashSet<Meta> = self.variables.clone();
+        while !new_variables.is_empty() {
+            let mut newer_variables = HashSet::new();
+            for m in new_variables.into_iter() {
+                if !seen.contains(&m) {
```
acl-cqc, line 687: You could do `if !seen.insert(m)` here (it returns a bool saying whether anything was added or not), rather than having a separate `insert` afterwards

src/extension/infer.rs
```diff
@@ -663,31 +676,98 @@ impl UnificationContext {
...
+    /// then 1 and 2 both definitely contain X, even if we don't know what else.
+    /// So instead of instantiating to the empty set, we'll instantiate to `{X}`
     pub fn instantiate_variables(&mut self) {
-        for m in self.variables.clone().into_iter() {
+        // A directed graph to keep track of `Plus` constraint relationships
+        let mut relations = GraphContainer::<Directed>::new();
+        let mut solutions: HashMap<Meta, ExtensionSet> = HashMap::new();
+
+        let variable_scope = self.search_variable_deps();
+        // If `m` has been merged, [`self.variables`] entry
```
acl-cqc, line 723: Move this comment just inside the loop (the moment when you have just defined `m`)

src/extension/infer.rs
```diff
@@ -663,31 +676,98 @@ impl UnificationContext {
...
-                            // value by [`self.merge_equal_metas`] so we don't
-                            // need to worry about resolving it.
-                            Constraint::Plus(x, other_m) if m == self.resolve(*other_m) => {
-                                Some(x.clone())
-                            }
+                let plus_constraints =
+                    self.get_constraints(&m)
+                        .unwrap()
+                        .iter()
+                        .cloned()
```
acl-cqc, line 736: You probably don't need `.cloned()`? I'd expect you can do something on the next line like `|c| match *c {` or maybe just a `*other_m` on the line after

src/extension/infer.rs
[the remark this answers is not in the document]
acl-cqc: Ah ok, some edges share endpoints. This works:
```suggestion
   assert_eq!(relations.node_map.len(), relations.ccs().iter().map(Vec::len).sum::<usize>());
```
Common up the two `relations.ccs` tho!

acl-cqc (approved):
Thanks Craig - looks good to me, some minor suggestions

src/extension/infer.rs
```diff
@@ -663,31 +676,98 @@ impl UnificationContext {
...
+    /// variables in the context.
+    fn search_variable_deps(&self) -> HashSet<Meta> {
+        let mut seen = HashSet::new();
+        let mut new_variables: HashSet<Meta> = self.variables.clone();
+        while !new_variables.is_empty() {
+            let mut newer_variables = HashSet::new();
+            for m in new_variables.into_iter() {
+                if !seen.contains(&m) {
+                    for c in self.get_constraints(&m).unwrap().iter() {
+                        match c {
```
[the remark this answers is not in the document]
acl-cqc: (if that works, you could just assign to `new_variables`, no need for a separate `newer_variables`!)

src/extension/infer.rs
croyzor: We need to copy the meta and clone the `r`, so it amounts to the same thing?

src/extension/infer.rs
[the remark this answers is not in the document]
croyzor: I'm not sure that check really makes sense, since it's apparent that the size of `node_map` and the number of nodes in the graph will be the same from looking at `GraphContainer::add_or_retrieve`, and `scc`s will return a clustered version of every node in the graph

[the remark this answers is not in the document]
acl-cqc: Agreed size of node map equals number of nodes in the graph, yes the latter is what I meant.

But SCC = Strongly Connected Component, i.e. every node reaches every other? In a directed graph? Ah, are `ccs` undirectly-connected?

So A -> B -> C -> A is one SCC, with all three nodes. But A -> B -> C is NOT an SCC by that definition, but are you saying you'd get A as one `cc`, B as one `cc` and C as one `cc` ?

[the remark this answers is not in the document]
acl-cqc: Yes, indeed it does. In which case yeah, no assert/check really seems necessary, we're only really protecting against somehow getting in a muddle ourselves between the two loops.

Does make me wonder whether we could just put *everything* through the `relations` map, then. A node with no connections will trivially (and cheaply) be its own SCC. So this above:
```
if unresolved_metas,is_empty() {
   self.add_solution(m, solution.clone())
} else {
  unresolved_metas.iter().for_each(.....)
}
```
could just be
```
relations.add_or_retrieve(m);
unresolved_metas.iter().for_each(....)
```
and then all the `add_solution` calls would be in the second loop.

src/extension/infer.rs
acl-cqc: Ah yes, didn't see the `r`, fair point

src/extension/infer.rs
```diff
@@ -663,31 +676,90 @@ impl UnificationContext {
...
+
+        // Strongly connected components are looping constraint dependencies.
+        // This means that each metavariable in the CC has the same solution.
+        for cc in relations.ccs() {
+            let mut combined_solution = ExtensionSet::new();
+            for sol in cc.iter().filter_map(|m| solutions.get(m)) {
+                combined_solution = combined_solution.union(sol);
+            }
+            for m in cc.iter() {
+                self.add_solution(*m, combined_solution.clone());
```
acl-cqc, line 762: We don't need to add to `solutions` here do we? (If so, things'll get a lot more complicated - you'd need to process SCCs in dependency order....)

src/extension/infer.rs
```diff
@@ -107,53 +108,65 @@ pub enum InferExtensionError {
...
         let src_ix = self.add_or_retrieve(src);
         let tgt_ix = self.add_or_retrieve(tgt);
-        self.equalities.add_edge(src_ix, tgt_ix, ());
+        self.graph.add_edge(src_ix, tgt_ix, ());
     }
 
     /// Return the connected components of the graph in terms of metavariables
     fn ccs(&self) -> Vec<Vec<Meta>> {
-        petgraph::algo::tarjan_scc(&self.equalities)
+        petgraph::algo::tarjan_scc(&self.graph)
```
acl-cqc, line 139: Ok, I think the naming and commenting of this function has been confusing me. Graph theory says that an undirected graph has CCs but a directed graph has SCCs. (I sort of assumed that the non-strongly CC of a directed graph was just what you got by forgetting the edge directions, but that may not be common usage.)

So, petgraph's `tarjan_scc` is misnamed for undirected graphs - it returns CCs (not SCCs, there's no such thing). But naming this function CC is misnamed in the opposite way...

Ideal might be to define `sccs` for `GraphContainer<Directed>` and `ccs` for `GraphContainer<Undirected>` - that would need a macro. Alternatively if having one name for each, I'd favour misnaming it in the same way as petgraph :-( rather than another way.... and most importantly, adjust the comment.

src/extension/infer.rs
acl-cqc: Try branch `inference-variable/fix-dependent-sccs` (the last commit) - a test is failing because we're inferring something different so would be good to look into that....

src/extension/infer.rs
croyzor: Renamed

acl-cqc:
Also, I have been wondering. Why is it sufficient to do this in `instantiate_variables`?

Firstly, shouldn't we include Metas 'solved' by finding they are in cycles, in the solution rather than the closure? But secondly, is it actually sufficient. Or, should we detect cycles of plus constraints earlier - perhaps before the *first* `main_loop` - and then convert them to equality constraints?

PR #594, opened 2023-10-09
fix: enforce covariance of SiblingMut::RootHandle

acl-cqc:
When creating a SiblingMut with a root whose type is constrained by being the root of the parent, ensure the "child" SiblingMut only narrows the constraint


aborgna-q (approved):
LGTM

PR #595, opened 2023-10-09
perf: O(1) optag superset checks

aborgna-q:
Pre-computes a superset matrix, so the checks can be done in a single operation.

Adds a benchmark to see the change.
Before the changes:
```
optag/tag_superset/(Const, Any)
                        time:   [1.2409 ns 1.2595 ns 1.2802 ns]
```
After the changes:
```
optag/tag_superset/(Const, Any)
                        time:   [129.44 ps 130.92 ps 132.46 ps]
                        change: [-89.435% -89.267% -89.098%] (p = 0.00 < 0.05)
```

### DIFF
--- a/benches/bench_main.rs
+++ b/benches/bench_main.rs
@@ -5,4 +5,5 @@ use criterion::criterion_main;
 
 criterion_main! {
     benchmarks::hugr::benches,
+    benchmarks::tag::benches,
 }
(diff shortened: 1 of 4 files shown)


acl-cqc:
Love it, but there is NO WAY we should micro-optimize this until we know that the optag-superset check (in `replace_op` or elsewhere) is actually a bottleneck in any real-world usecase.

aborgna-q:
:(

PR #596, opened 2023-10-09
Update definition of Replace (and clarify SimpleReplace)

acl-cqc:
* Clarify various characteristics of mu_in and mu_out maps in SimpleReplace
* Remove definitions (inc duplicate) of Partial Hugr and Separated (inlining into single use)
* Replace requires unique parent node P, removes set of children of P, adds new subtrees only as children of P (so drop `T` map)
* nu_in/nu_out use $\Gamma \setminus R$ rather than $\Gamma \setminus S^*$
* New children of $P$ added at positions vacated by removed children where possible
* Extend/clarify definition of what can be missing in the RHS, and what edges can be in the nu_in/nu_out

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1143,12 +1143,6 @@ is itself in S.
 The meaning of “convex” is: if A and B are nodes in the convex set S,
 then any sibling node on a path from A to B is also in S.
 
-Given a set S of nodes in a hugr, let S\* be the set of all nodes
-descended from nodes in S, including S itself.
-
-Call two nodes a, b in Γ *separated* if a is not in {b}\* and b is not
-in {a}\* (i.e. there is no hierarchy relation between them).
-
 #### API methods
 
 There are the following primitive operations.
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1171,8 +1171,8 @@ The method takes as input:
     Ext edges;
   - a hugr $H$ whose root is a DFG node $R$ with only leaf nodes as children --
     let $T$ be the set of children of $R$;
-  - a map $\nu\_\textrm{inp}: \textrm{inp}\_H(T \setminus \\{\texttt{Input}\\}) \to \textrm{inp}\_{\Gamma}(S)$;
-  - a map $\nu_\textrm{out}: \textrm{out}_{\Gamma}(S) \to \textrm{out}_H(T \setminus \\{\texttt{Output}\\})$.
+  - a map $\nu\_\textrm{inp}: \textrm{inp}\_H(T \setminus \\{\texttt{Input}\\}) \to \textrm{inp}\_{\Gamma}(S)$ (*note* in order to produce a valid Hugr, all keys must be present; and all possible values must be present exactly once unless Copyable);
+  - a map $\nu_\textrm{out}: \textrm{out}_{\Gamma}(S) \to \textrm{out}_H(T \setminus \\{\texttt{Output}\\})$ (*note* (similarly) in order to produce a valid hugr, all keys $\textrm{out}_{\Gamma}(S)$ must be present; and each possible value must be either Copyable and/or present exactly once. Any possible value not present could just be omitted from $H$...).
   
 The new hugr is then derived as follows:
   
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1198,22 +1198,10 @@ The new hugr is then derived as follows:
 
 This is the general subgraph-replacement method.
 
-<!-- A _partial hugr_ is a graph formed by a subset of nodes of a valid hugr together
-# with a subset of their adjoining edges.
-
-# Given a partial hugr $G$, let
-
-#  - $\bot(G)$ be the set of container nodes in $G$ without an outgoing hierarchy edge.
--->
 Given a set $S$ of nodes in a hugr, let $S^\*$ be the set of all nodes
 descended from nodes in $S$ (i.e. reachable from $S$ by following hierarchy edges),
 including $S$ itself.
 
-<!-- Call two nodes $a, b \in \Gamma$ _separated_ if $a \notin \\{b\\}^\*$ and
-$b \notin \\{a\\}^\*$ (i.e. there is no hierarchy relation between them). Note
-that this is much the same requirement as convexity, but following the hierarchy
-edges rather than dataflow edges.
--->
 A `NewEdgeSpec` specifies an edge inserted between an existing node and a new node.
 It contains the following fields:
 
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1178,8 +1178,8 @@ The new hugr is then derived as follows:
   
   1. Make a copy in $\Gamma$ of all children of $R$, excluding Input and Output,
      and all edges between them. Make all the newly added nodes children of $P$.
-     Notation: if $p$ is a port of a node in $R$, write $p^*$ for the copy of
-     the port in $\Gamma$.
+     Notation: for each port $p$ of a node in $R$ of which a copy is made, write
+     $p^*$ for the copy of the port in $\Gamma$.
   2. For each $(q, p = \nu_\textrm{inp}(q))$ such that $q \notin \texttt{Output}$,
      add an edge from $p^-$ to $q^*$.
   3. For each $(p, q = \nu_\textrm{out}(p))$ such that $q^- \notin \texttt{Input}$,
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1231,13 +1231,15 @@ in the hugr and the other is a new node.
 The `Replace` method takes as input:
   - the ID of a container node $P$ in $\Gamma$;
   - a set $S$ of IDs of nodes that are children of $P$
-  - a Hugr $G$ whose root is a node of the same type as $P$. Note this Hugr need not be valid, in that it may be missing:
+  - a Hugr $G$ whose root is a node of the same type as $P$.
+    Note this Hugr need not be valid, in that it may be missing:
       * edges to/from some ports (i.e. it may have unconnected ports)---not just Copyable dataflow outputs, which may occur even in valid Hugrs, but also incoming and/or non-Copyable dataflow ports, and ControlFlow ports,
       * all children for some container nodes strictly beneath the root (i.e. it may have container nodes with no outgoing hierarchy edges)
       * some children of the root, for container nodes that require particular children (e.g.
         $\mathtt{Input}$ and/or $\mathtt{Output}$ if $P$ is a dataflow container, the exit node
         of a CFG, the required number of children of a conditional)
-  - a map $B$ *from* container nodes in $G$ that have no children *to* container nodes in $S^\*$ none of which is an ancestor of another.
+  - a map $B$ *from* container nodes in $G$ that have no children *to* container nodes in $S^\*$
+    none of which is an ancestor of another.
     Let $X$ be the set of children of nodes in the image of $B$, and $R = S^\* \setminus X^\*$.
   - a list $\mu\_\textrm{inp}$ of `NewEdgeSpec` which all have their `TgtNode`in
     $G$ and `SrcNode` in $\Gamma \setminus R$;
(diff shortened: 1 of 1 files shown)


specification/hugr.md
```diff
@@ -1351,13 +1346,12 @@ it (and its incoming value and Order edges) from the hugr.
 
 ###### `InsertConst`
 
-Given a `Const<T>` node `c` and a DSG `P`, add `c` as a child of `P`,
-inserting an Order edge from the Input under `P` to `c`.
+Given a `Const<T>` node `c` and a container node `P` (either a `Module`,
+ a `CFG` node or a dataflow container), add `c` as a child of `P`.
```
acl-cqc, line 1350: This is not as closely related, but Const nodes can go anywhere according to the Scoped Definitions section, and the lack of order edge follows #447 and #468

specification/hugr.md
```diff
@@ -1212,7 +1206,9 @@ descended from nodes in $S$ (i.e. reachable from $S$ by following hierarchy edge
 including $S$ itself.
 
 Call two nodes $a, b \in \Gamma$ _separated_ if $a \notin \\{b\\}^\*$ and
-$b \notin \\{a\\}^\*$ (i.e. there is no hierarchy relation between them).
+$b \notin \\{a\\}^\*$ (i.e. there is no hierarchy relation between them). Note
+that this is much the same requirement as convexity, but following the hierarchy
```
cqc-alec, line 1210: I don't really see why separatedness is the same (or similar) to convexity. Certainly convex subgraphs of the hierarchy tree need not be separated.

specification/hugr.md
```diff
@@ -1177,15 +1171,15 @@ The method takes as input:
     Ext edges;
   - a hugr $H$ whose root is a DFG node $R$ with only leaf nodes as children --
     let $T$ be the set of children of $R$;
-  - a map $\nu\_\textrm{inp}: \textrm{inp}\_H(T \setminus \\{\texttt{Input}\\}) \to \textrm{inp}\_{\Gamma}(S)$;
-  - a map $\nu_\textrm{out}: \textrm{out}_{\Gamma}(S) \to \textrm{out}_H(T \setminus \\{\texttt{Output}\\})$.
+  - a map $\nu\_\textrm{inp}: \textrm{inp}\_H(T \setminus \\{\texttt{Input}\\}) \to \textrm{inp}\_{\Gamma}(S)$ (*note* in order to produce a valid Hugr, all keys must be present; and all possible values must be present exactly once unless Copyable);
+  - a map $\nu_\textrm{out}: \textrm{out}_{\Gamma}(S) \to \textrm{out}_H(T \setminus \\{\texttt{Output}\\})$ (*note* (similarly) in order to produce a valid hugr, all keys $\textrm{out}_{\Gamma}(S)$ must be present; and each possible value must be either Copyable and/or present exactly once. Any possible value not present could just be omitted from $H$...).
```
cqc-alec, line 1175: We are quite inconsistent with capitalization of hugr/Hugr/HUGR, but this can be fixed later.

cqc-alec:
This seems like a useful set of clarifications, though more changes may be needed...

acl-cqc:
Ok, I've had a pass at updating Replace.
* Inlined definition of partial Hugr, and tried to make it a lot more precise
* Also inlined definition of separated, as now we have single-parent restriction, separated is only used in one place
* I'm not sure about treatment of Input/Output or Exit BasicBlock, also what if the parent $P$ is a Module node....next

specification/hugr.md
```diff
@@ -1225,11 +1229,14 @@ Note that in a `NewEdgeSpec` one of `SrcNode` and `TgtNode` is an existing node
 in the hugr and the other is a new node.
 
 The `Replace` method takes as input:
-
-  - a set $S$ of mutually-separated nodes in $\Gamma$;
-  - a partial hugr $G$;
-  - a map $T : \top(G) \to \Gamma \setminus S^*$ whose image consists of container nodes;
-  - a map $B : \bot(G) \to S^\*$ whose image consists of mutually-separated container nodes.
+  - the ID of a DFG node $P$ in $\Gamma$;
```
cqc-alec, line 1232: Not necessarily DFG...

specification/hugr.md
```diff
@@ -1239,17 +1246,16 @@ The `Replace` method takes as input:
 
 The new hugr is then derived as follows:
 
-1.  Make a copy in $\Gamma$ of all the nodes in $G$, and all edges between them.
-2.  For each $\sigma\_\mathrm{inp} \in \mu\_\textrm{inp}$, insert a new edge going into the new
+1.  Make a copy in $\Gamma$ of all the nodes in $H$ *except the root*, and all edges where both source and target are copied.
```
cqc-alec, line 1249: Isn't "where both source and target are copied" redundant? The only node that isn't copied is the root...

acl-cqc: Yes, good point...:-)

acl-cqc: Clarified

specification/hugr.md
acl-cqc: (Separatedness now removed/inlined)

specification/hugr.md
```diff
@@ -1229,22 +1229,26 @@ Note that in a `NewEdgeSpec` one of `SrcNode` and `TgtNode` is an existing node
...
+        $\mathtt{Input}$ and/or $\mathtt{Output}$ if $P$ is a dataflow container, the exit node
+        of a CFG, the required number of children of a conditional)
+  - a map $B$ *from* container nodes in $H$ that have no children *to* container nodes in $S^\*$ none of which is an ancestor of another.
+    Let $X$ be the set of children of nodes in the image of $B$, and $R = S^\* \setminus X^\*$.
   - a list $\mu\_\textrm{inp}$ of `NewEdgeSpec` which all have their `TgtNode`in
-    $G$ and `SrcNode` in $\Gamma \setminus S^*$;
+    $G$ and `SrcNode` in $\Gamma \setminus R$;
   - a list $\mu\_\textrm{out}$ of `NewEdgeSpec` which all have their `SrcNode`in
-    $G$ and `TgtNode` in $\Gamma \setminus S^*$ (and `TgtNode` has an existing
+    $G$ and `TgtNode` in $\Gamma \setminus R$ (and `TgtNode` has an existing
```
acl-cqc, line 1245: I'd like to be more precise about the target port/whatever having an existing incoming edge, but that's a little awkward. I'm still wondering about a bigger restructure of edges from being two "lists", but I think the "well-formedness" requirements are so complex anyway that one might as well give up....

One thing I think one might want is to add new edges between nodes in `$\Gamma \setminus R$ i.e. where the source has been removed, the "new" source might also be in the remainder rather than new in the replacement. Forcing all such edges to be "passed through" Input -> Output doesn't work if $P$ isn't a DFG, and doesn't work well for static/order edges even for DFGs.

One possibility might be to allow mu_out to have any SrcNode, although that's tricky because until we've copied $G$ into $\Gamma$ we don't have a consistent notion of ID, hmmm.

A third set of new edges?? There may be call for being able to add arbitrary *new* Order edges, too....

specification/hugr.md
```diff
@@ -1253,8 +1257,10 @@ The new hugr is then derived as follows:
     copy of the `SrcNode` of $\sigma\_\mathrm{out}$ according to the specification $\sigma\_\mathrm{out}$.
     The target port must have an existing edge whose source is in $R$; this edge
     is removed.
-4.  For each $(n, t = T(n))$, append the copy of $n$ to the list
-    of children of $t$ (adding a hierachy edge from $t$ to $n$).
+4.  Let $N$ be the ordered list of the copies made in $\Gamma$ of the children of the root node of $H$.
```
acl-cqc: This is a bit more complex than append-to-end, but I think it may be necessary for replacing particular children of Conditionals (even without changing their number); CFG entry nodes; and Input/Output nodes. Generally you'll need to provide exactly one Input if-and-only-if you remove an Input, sort of thing, but it seems reasonable that you might???

specification/hugr.md
```diff
@@ -1244,8 +1244,10 @@ The `Replace` method takes as input:
   - a list $\mu\_\textrm{inp}$ of `NewEdgeSpec` which all have their `TgtNode`in
     $G$ and `SrcNode` in $\Gamma \setminus R$;
   - a list $\mu\_\textrm{out}$ of `NewEdgeSpec` which all have their `SrcNode`in
-    $G$ and `TgtNode` in $\Gamma \setminus R$ (and `TgtNode` has an existing
-    incoming edge from a node in $R$).
+    $G$ and `TgtNode` in $\Gamma \setminus R$ (where `TgtNode` and `TgtPos` describe
+    an existing incoming edge of that kind from a node in $R$).
+
+Note that the well-formedness requirements of Hugr imply that $\mu\_\textrm{inp}$ and $\mu\_\textrm{out}$ only contain Order edges if $P$ is a dataflow container, and only contain ControlFlow edges if $P$ is a CFG-node; and that any such Order or ControlFlow edges will (more precisely) be between nodes $S$ and children of the root of $G$.
```
cqc-alec, line 1250: Do you mean "Value/Order/Static" when you say "Order" here?

specification/hugr.md
```diff
@@ -1229,22 +1229,30 @@ Note that in a `NewEdgeSpec` one of `SrcNode` and `TgtNode` is an existing node
...
+      * all children for some container nodes strictly beneath the root (i.e. it may have container nodes with no outgoing hierarchy edges)
+      * some children of the root, for container nodes that require particular children (e.g.
+        $\mathtt{Input}$ and/or $\mathtt{Output}$ if $P$ is a dataflow container, the exit node
+        of a CFG, the required number of children of a conditional)
+  - a map $B$ *from* container nodes in $G$ that have no children *to* container nodes in $S^\*$
+    none of which is an ancestor of another.
+    Let $X$ be the set of children of nodes in the image of $B$, and $R = S^\* \setminus X^\*$.
   - a list $\mu\_\textrm{inp}$ of `NewEdgeSpec` which all have their `TgtNode`in
-    $G$ and `SrcNode` in $\Gamma \setminus S^*$;
+    $G$ and `SrcNode` in $\Gamma \setminus R$;
```
cqc-alec, line 1245: So this change is allowing `SrcNode` to be in S* ∩ X*. OK I think I see why this is allowed!

specification/hugr.md
acl-cqc: I definitely do *not* mean "contain only Order edges if ....". What I mean is that if $P$ is not a dataflow container, then there will be no Order edges. And if $P$ is not a CFG node, there will be no ControlFlow edges. Ok, will rephrase.

specification/hugr.md
acl-cqc: If you move a subtree from the old Hugr to a new location....
* There won't be any Ext edges *out* of it (because they'd only go to siblings, i.e. new bits added INTO the subtree you've moved, and Replace doesn't let you add any such)
* But you might need to add Ext edges *into* it (if the bit you kept, had an Ext edge from a bit you removed)
* And, you could have Dom edges *out* of the bit you're moving into new nodes, or Dom edges *in* as previous point.

Should I expand, do you think? I admit that change looks quite small but has bigger ramifications!

specification/hugr.md
```diff
@@ -1253,8 +1261,10 @@ The new hugr is then derived as follows:
     copy of the `SrcNode` of $\sigma\_\mathrm{out}$ according to the specification $\sigma\_\mathrm{out}$.
     The target port must have an existing edge whose source is in $R$; this edge
     is removed.
-4.  For each $(n, t = T(n))$, append the copy of $n$ to the list
-    of children of $t$ (adding a hierachy edge from $t$ to $n$).
+4.  Let $N$ be the ordered list of the copies made in $\Gamma$ of the children of the root node of $G$.
```
acl-cqc, line 1264: This is a bit more complex than append-to-end, but I think it may be necessary for replacing particular children of Conditionals (even without changing their number); CFG entry nodes; and Input/Output nodes. Generally you'll need to provide exactly one Input if-and-only-if you remove an Input, sort of thing, but it seems reasonable that you might???

cqc-alec (approved):

specification/hugr.md
acl-cqc: Of course one can always use `HugrMut::connect`, I guess we can add this later if we really find a need

PR #597, opened 2023-10-10
[Refactor] Tidy some CFG structuring code, require RootHandle=CfgID

acl-cqc:
* IdentityCfgMap and HalfNodeView require RootTagged w/ CfgID rather than just HugrView
* Derive `Clone` for RootHandle
* implement a `fn borrow` (returning `RootHandle<&Hugr>`) for `RootHandle<&mut Hugr`
* Avoid some redundant checks and unwraps, remove a misleading `pub` (the structure it's on isn't pub)

### DIFF
--- a/src/algorithm/half_node.rs
+++ b/src/algorithm/half_node.rs
@@ -2,8 +2,9 @@ use std::hash::Hash;
 
 use super::nest_cfgs::CfgNodeMap;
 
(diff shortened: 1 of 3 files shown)


aborgna-q (approved):
Nice! It looks like we won't have much problem in expanding the `RootHandle` constraints around.

PR #600, opened 2023-10-11
feat: Compute affected nodes for `SimpleReplacement`

aborgna-q:


src/hugr/rewrite/simple_replace.rs
```diff
@@ -45,14 +46,30 @@ impl SimpleReplacement {
...
+    ///
+    /// This includes the nodes in the subgraph and the boundary neighbours that
+    /// are referenced by the replacement.
+    ///
+    /// Two `SimpleReplacement`s can be composed if their affected nodes are
+    /// disjoint.
+    #[inline]
+    pub fn affected_nodes(&self) -> impl Iterator<Item = Node> + '_ {
+        let subcirc = self.subgraph.nodes().iter().copied();
+        let out_neighs = self.nu_out.keys().map(|&(n, _)| n);
```
cqc-alec, line 70: This only picks up the output neighbours, I think? Do we want to include the input neighbours?

aborgna-q: The input nodes referenced in the rewrite are part of the removed set.

aborgna-q: (so modifying the input neighbours doesn't invalidate the rewrite)

cqc-alec: OK I see what you mean, though I think maybe the documentation needs clarifying, or the function renaming? E.g. in the test case, intuitively first two H gates in the circuit (preceding the replaced subcircuit) would count as "affected by" the replacement just as much as the succeeding output node.

Hmm, "referenced nodes"? Not convinced. Anyway no quibbles with the implementation.

cqc-alec (approved):

src/hugr/rewrite/simple_replace.rs
aborgna-q: Maybe `invalidation_set`?

aborgna-q:
This works for our immediate needs in tket2. I opened issue #601 to generalise it to the `Rewrite` trait.

PR #608, opened 2023-10-16
feat: Add hashable `Angle` type to Quantum extension

cqc-alec:
After coding it up I have a couple of reservations about this:

 * Calling the parameter "precision" suggests that it is an _approximate_ representation of a value, rather than an _exact_ representation of a value that is restricted in its possibilities. I think that the latter makes more sense, since "approximate values" are (I think) outside the scope of HUGR, so another name might be better.
 * Compilation passes that squash rotation gates using Euler-angle decomposition will in general _not_ transform `Angle<N>` values into `Angle<N>` values exactly. Therefore care will be needed in applying such passes. Of course, even when applied to `float64` they are not exact and can accumulate rounding error, so this isn't a new problem, just one that's made more obvious.

The big advantage is that it makes angle values (and therefore circuits containing them) equatable and hashable.


src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,168 @@
...
+    /// Returns the precision of the constant
+    pub fn precision(&self) -> u8 {
+        self.precision
+    }
+}
+
+#[typetag::serde]
+impl CustomConst for ConstAngle {
+    fn name(&self) -> SmolStr {
+        format!("a(2π*{}/{})", self.value, 1u64 << self.precision).into()
```
ss2165, line 108:
```suggestion
        format!("a(2π*{}/2^{})", self.value, self.precision).into()
```
I prefer more explicit

src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,168 @@
...
+fn atrunc_sig(arg_values: &[TypeArg]) -> Result<FunctionType, SignatureError> {
+    let [arg0, arg1] = collect_array(arg_values);
+    let m: u8 = get_precision(arg0)?;
+    let n: u8 = get_precision(arg1)?;
+    if m < n {
+        return Err(SignatureError::InvalidTypeArgs);
+    }
+    Ok(FunctionType::new(
+        vec![angle_type(arg0.clone())],
+        vec![Type::new_sum(vec![angle_type(arg1.clone()), ERROR_TYPE])],
```
ss2165, line 133: what causes the error state? It's not in the spec

src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,168 @@
...
+/// An angle
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct ConstAngle {
+    precision: u8,
+    value: u64,
+}
+
+impl ConstAngle {
+    /// Create a new [`ConstAngle`]
+    pub fn new(precision: u8, value: u64) -> Result<Self, ConstTypeError> {
```
ss2165, line 78: would it be useful to add a `from_f64(value: f64, precision: u8)`, with rounding down?

ss2165 (changes requested):
There is an angle type still in the rotation extension, I think it is worth removing that one while adding this.

I agree about the name `precision`. Not sure of a better alternative, only one I've got is the descriptive route `denom_power`

specification/hugr.md
```diff
@@ -1751,6 +1751,26 @@ allocation for all `Qubit` wires.
...
+ hashable.
+
+The following operations are defined:
+
+| Name           | Inputs     | Outputs    | Meaning |
+| -------------- | ---------- | ---------- | ------- |
+| `aconst<N, x>` | none       | `angle<N>` | const node producing angle $2 \pi x / 2^N$ (where $0 \leq x \lt 2^N$) |
+| `atrunc<M,N>`  | `angle<M>` | `angle<N>` | round `angle<M>` to `angle<N>`, where $M \geq N$, rounding down in $[0, 2\pi)$ if necessary |
+| `awiden<M,N>`  | `angle<M>` | `angle<N>` | widen `angle<M>` to `angle<N>`, where $M \leq N$ |
+| `aadd<N>`      | `angle<N>`, `angle<N>` | `angle<N>` | add two angles |
```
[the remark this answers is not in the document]
cqc-alec: Yes, or even `aadd<N,M> : (Angle<N>, Angle<M>) --> Angle<max(N,M)>` assuming that's something we can express (I think so).

src/std_extensions/quantum.rs
cqc-alec: It could be. Will add.

cqc-alec:
> There is an angle type still in the rotation extension, I think it is worth removing that one while adding this.

I wondered about this But wasn't sure what the consequences would be (does anything else depend on the rotation extension?).

cqc-alec:
> I agree about the name `precision`. Not sure of a better alternative, only one I've got is the descriptive route `denom_power`

I may go with "log-denominator", bit of a mouthful though it is.

src/std_extensions/quantum.rs
cqc-alec: Right, this can't error -- I think I was confusing this with an operation that would fail (instead of rounding down) if the angles didn't match exactly. That could actually be a useful operation, and I should maybe add it -- or replace `awiden<M,N>` with a more general one that doesn't require M <= N but can fail if M > N and the angle doesn't "fit" exactly.

specification/hugr.md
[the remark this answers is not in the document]
cqc-alec: Done.

src/std_extensions/quantum.rs
cqc-alec: Done.

src/std_extensions/quantum.rs
cqc-alec: Made a `from_radians()` method that takes an `f64`.

cqc-alec:
> > There is an angle type still in the rotation extension, I think it is worth removing that one while adding this.
> 
> I wondered about this But wasn't sure what the consequences would be (does anything else depend on the rotation extension?).

Have left this for now, as this would probably involve rewriting the whole of `rotations`, which we may want to do but I think in another PR.

acl-cqc:
> There is an angle type still in the rotation extension, I think it is worth removing that one while adding this.
> 
> I agree about the name `precision`. Not sure of a better alternative, only one I've got is the descriptive route `denom_power`

To me precision means the number of digits. Whether those digits are correct or not would be *accuracy*. However, a low precision number (say 1.0) vs a high precision number (1.0153245) - the former identifies a range that would include the latter (indeed, all numbers between 0.95 and 1.05). Whereas for Angle I think you are saying that the angle is *exactly* what you get by dividing the numerator by the denominator, in which case yes, some variant of denominator seems appropriate (or "unit"? As in unit of measurement, not the 0-element tuple!)...

src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,190 @@
...
+pub const LOG_DENOM_BOUND: u8 = 54;
+
+const fn is_valid_log_denom(n: u8) -> bool {
+    n < LOG_DENOM_BOUND
+}
+
+/// Type parameter for the log-denominator of an angle.
+// SAFETY: unsafe block should be ok as the value is definitely not zero.
+#[allow(clippy::assertions_on_constants)]
+pub const LOG_DENOM_TYPE_PARAM: TypeParam = TypeParam::bounded_nat(unsafe {
```
acl-cqc, line 49: Whoa! `new_unchecked` needs `unsafe`, I see. Hmmmm. Ick. (I think? Or should we make our `new_unchecked`s also need unsafe?)

The following works without an `unsafe`, but you can keep the assert:
`NonZeroU64::MIN.saturating_add((LOG_WIDTH_BOUND-1) as u64)`

or even
`NonZeroU64::MIN.saturating_add((LOG_WIDTH_BOUND as u64)-NonZeroU64::MIN.get())`

Ymmv :)

src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,190 @@
...
+/// An angle
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct ConstAngle {
+    log_denom: u8,
+    value: u64,
+}
+
+impl ConstAngle {
+    /// Create a new [`ConstAngle`] from a log-denominator and a numerator
+    pub fn new(log_denom: u8, value: u64) -> Result<Self, ConstTypeError> {
```
acl-cqc, line 80: Is it possible to make this `const` ?

src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,190 @@
...
+                    "Invalid unsigned integer value.".to_owned(),
+                ),
+            ));
+        }
+        Ok(Self { log_denom, value })
+    }
+
+    /// Create a new [`ConstAngle`] from a log-denominator and a floating-point value in radians,
+    /// rounding to the nearest corresponding value. (Ties round away from zero.)
+    pub fn from_radians(log_denom: u8, theta: f64) -> Result<Self, ConstTypeError> {
```
acl-cqc, line 100: nit: maybe `rounding_radians` ?

src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,190 @@
...
+
+fn atrunc_sig(arg_values: &[TypeArg]) -> Result<FunctionType, SignatureError> {
+    let [arg0, arg1] = collect_array(arg_values);
+    let m: u8 = get_log_denom(arg0)?;
+    let n: u8 = get_log_denom(arg1)?;
+    if m < n {
+        return Err(SignatureError::InvalidTypeArgs);
+    }
+    Ok(FunctionType::new(
+        vec![angle_type(arg0.clone())],
```
acl-cqc, line 153: Nit - do you need to clone these? (not sure why `clippy` doesn't complain?!)

acl-cqc (commented):
Looks good to me! Of course there is no hash defined here, or any implementation (e.g. Rust definition) of the type.

src/std_extensions/quantum.rs
cqc-alec: Without `clone()` I get an error:

```
error[E0308]: mismatched types
   --> src/std_extensions/quantum.rs:153:25
    |
153 |         vec![angle_type(arg0)],
    |              ---------- ^^^^ expected `TypeArg`, found `&TypeArg`
    |              |
    |              arguments to this function are incorrect
    |
note: function defined here
   --> src/std_extensions/quantum.rs:35:15
    |
35  | pub(super) fn angle_type(log_denom_arg: TypeArg) -> Type {
    |               ^^^^^^^^^^ ----------------------
help: consider using clone here
    |
153 |         vec![angle_type(arg0.clone())],
```

and the suggested fix is to use `clone()`.

specification/hugr.md
```diff
@@ -1757,6 +1757,27 @@ allocation for all `Qubit` wires.
...
 
+#### Angles
+
+The Quantum extension also defines a specialized `angle<N>` type which is used
+to express parameters of rotation gates. The type is parametrized by the
+_log-denominator_, which is an integer $N \in [0, 53]$; angles with
+log-denominator $N$ are multiples of $2 \pi / 2^N$, where the multiplier is an
+unsigned `int<N>`. The maximum log-denominator $53$ effectively gives the
+resolution of a `float64` value; but note that unlike `float64` all angle values
+are equatable and hashable.
```
acl-cqc, line 1768: I'm wondering whether `x / (2pi/2^N)` has the same hash as `(2x)/(2pi/2^(N+1))`, but I don't think you give the hash "algorithm" so fine

specification/hugr.md
```diff
@@ -1757,6 +1757,27 @@ allocation for all `Qubit` wires.
 If further the program does not contain any `qalloc` or `qfree`
 operations we can state the program only uses `N` qubits.
 
+#### Angles
+
+The Quantum extension also defines a specialized `angle<N>` type which is used
+to express parameters of rotation gates. The type is parametrized by the
+_log-denominator_, which is an integer $N \in [0, 53]$; angles with
```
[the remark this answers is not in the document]
cqc-alec: They don't do much, but an Rz(2*pi) operation is a global phase flip so it's conceivable. I don't think there's any harm in allowing 0?

src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,190 @@
...
+/// Angle type with a given log-denominator (specified by the TypeArg).
+///
+/// This type is capable of representing angles that are multiples of 2π / 2^N where N is the
+/// log-denominator.
+pub(super) fn angle_type(log_denom_arg: TypeArg) -> Type {
+    Type::new_extension(angle_custom_type(log_denom_arg))
+}
+
+/// The smallest forbidden log-denominator.
+pub const LOG_DENOM_BOUND: u8 = 54;
```
acl-cqc, line 40: super-nit: I find `LOG_DENOM_MAX` clearer (without having to read the doc) as to whether it's inclusive/exclusive; then you'd change `<` to `<=`

src/std_extensions/quantum.rs
acl-cqc: Ok - think I was forgetting how slices work, sorry!

specification/hugr.md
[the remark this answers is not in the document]
cqc-alec: Actually because we restrict the numerator to [0, 2^n) we couldn't express that. Hmm, this is a bit annoying, because in tket we can.

src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,190 @@
...
+            angle_type(TypeArg::BoundedNat { n: m as u64 }),
+            angle_type(TypeArg::BoundedNat { n: n as u64 }),
+        ],
+        vec![angle_type(TypeArg::BoundedNat { n: l as u64 })],
+    ))
+}
+
+fn aunop_sig(arg_values: &[TypeArg]) -> Result<FunctionType, SignatureError> {
+    let [arg] = collect_array(arg_values);
+    Ok(FunctionType::new(
```
acl-cqc, line 182: nit: `FunctionType::new_linear`

src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,190 @@
...
+        vec![
+            angle_type(TypeArg::BoundedNat { n: m as u64 }),
+            angle_type(TypeArg::BoundedNat { n: n as u64 }),
+        ],
+        vec![angle_type(TypeArg::BoundedNat { n: l as u64 })],
+    ))
+}
+
+fn aunop_sig(arg_values: &[TypeArg]) -> Result<FunctionType, SignatureError> {
```
acl-cqc, line 180: Naming perhaps a little terse here...took me a while to figure out that`aun` is short for "angle unary"

specification/hugr.md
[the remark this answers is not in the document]
cqc-alec: Morally an angle type _should_ be modulo 2*pi (not modulo 4*pi) as it sometimes is in tket), so the parameter to Rz is not _really_ an angle (but a half-angle).

src/std_extensions/quantum.rs
```diff
@@ -1,19 +1,190 @@
...
+}
+
+pub(super) const fn type_arg(log_denom: u8) -> TypeArg {
+    TypeArg::BoundedNat {
+        n: log_denom as u64,
+    }
+}
+
+/// An angle
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
```
acl-cqc, line 72: I note we aren't actually deriving (or implementing) Hash here...that may be right though...

acl-cqc (approved):
Generally LGTM, thanks Alec - and sorry for slow review. There are a couple things that make me slightly nervous - that `unsafe` rings alarm bells (for now, so I guess gradually I'll reduce my sensitivity), and I wonder about specifying some properties of the hash (which we aren't implementing anyway....)

src/std_extensions/quantum.rs
cqc-alec: Nice, we can probably do this in the integer extension code too where there is a similar unsafe block.

src/std_extensions/quantum.rs
cqc-alec: Seems not, because of the `to_owned()` inside it.

specification/hugr.md
cqc-alec: If we restrict the numerator to the range $[0, 2^n]$, which I think we should, then all values with a given denominator $2^n$ have different hashes. Interesting question is whether values with different denominators should have the same hash if they represent the same angle: e.g. $1 \times 2\pi/2^2$ and $2 \times 2\pi/2^3$. I think probably not, since they are different types and you can do different things with them.

src/std_extensions/quantum.rs
cqc-alec: Changed.

src/std_extensions/quantum.rs
cqc-alec: Yup, changed this and a couple of others.

ss2165 (changes requested):

src/std_extensions/quantum.rs
```diff
@@ -100,4 +336,34 @@ pub(crate) mod test {
...
+        assert_ne!(const_a32_7, const_a33_7);
+        assert_ne!(const_a32_7, const_a32_8);
+        assert_eq!(const_a32_7, ConstAngle::new(5, 7));
+        assert_matches!(
+            ConstAngle::new(3, 256),
+            Err(ConstTypeError::CustomCheckFail(_))
+        );
+        assert_matches!(
+            ConstAngle::new(54, 256),
+            Err(ConstTypeError::CustomCheckFail(_))
```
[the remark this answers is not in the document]
cqc-alec: Now added.

ss2165 (approved):
👍

PR #609, opened 2023-10-16
refactor: insert_subgraph just return HashMap, make InsertionResult new_root compulsory

acl-cqc:
insert_subgraph reused InsertionResult by making new_root into an Option. The "reuse" is trivial - what's left is a single field - so this avoids confusion as to when `new_root` is or is not None and thus a bunch of unwraps.


aborgna-q (approved):
👍

PR #613, opened 2023-10-20
feat: Provide+implement Replace API

acl-cqc:
* Add struct Replace, implement both apply() and verify()
* Generalize some SimpleReplace tests to convert requests into Replaces too
* Tiny change to spec to allow Replaces to patch up broken inter-graph edges
* Generalize builder::Container::add_constant to allow specifying "None" for open Extensions


specification/hugr.md
```diff
@@ -1236,10 +1236,10 @@ The `Replace` method takes as input:
     $G$ and `SrcNode` in $\Gamma \setminus R$;
   - a list $\mu\_\textrm{out}$ of `NewEdgeSpec` which all have their `SrcNode`in
     $G$ and `TgtNode` in $\Gamma \setminus R$, where `TgtNode` and `TgtPos` describe
-    an existing incoming edge of that kind from a node in $R$.
+    an existing incoming edge of that kind from a node in $S\*$.
```
acl-cqc, line 1239: These tweaks allow one to replace an edge into/out-of a part of the Hugr that you are keeping but moving to a new location (the old edge might be invalid if it were nonlocal, say)

src/hugr/rewrite/replace.rs
```diff
@@ -0,0 +1,591 @@
...
+    },
+    /// A [EdgeKind::ControlFlow] edge (between CFG nodes only)
+    ControlFlow {
+        /// Identifies a control-flow output (successor) of the source node.
+        src_pos: OutgoingPort,
+    },
+}
+
+/// Specification of a `Replace` operation
+#[derive(Debug, Clone)] // PartialEq? Eq probably doesn't make sense because of ordering.
```
acl-cqc, line 59: Thoughts here?

src/hugr/rewrite/replace.rs
```diff
@@ -0,0 +1,591 @@
...
+        let exset = ExtensionSet::singleton(&collections::EXTENSION_NAME);
+        let intermed = TypeRow::from(vec![listy.clone(), USIZE_T]);
+
+        let mut cfg = CFGBuilder::new(
+            FunctionType::new_linear(just_list.clone()).with_extension_delta(&exset),
+        )?;
+
+        let pred_const = cfg.add_constant(
+            ops::Const::unary_unit_sum(),
+            // Slightly awkward we have to specify this, and the choice is non-obvious:
```
acl-cqc, line 478: I felt I had to say something here, as this relates to #633....
* at the moment Builder `add_constant` forces you to specify the Extensions. The obvious choice would be `ExtensionSet::new()` but then you would need lift nodes, ugh
* If we generalized `add_constant` to allow specifying "open" (None), then in theory inference should figure out the choice here, but in practice #633 kicks in and it only does so *sometimes* - infinite-looping on other runs :(

So without generalizing add_constant (to no immediate win here), I don't really see what I can say, so I will try to come back to this after fixing #633....

src/hugr/rewrite/replace.rs
acl-cqc: It is (I had that nondeterministic bug!!) - the new nodes (in the fixed ordering in which `replacement` keeps them) are put into the places of the nodes in `removal` (in that order) - so you can put a new node in as the entry node of a CFG, or an Input/Output, etc.

Also, Hugr itself is only `PartialEq` not `Eq`...but, I guess there is no reason not to make this PartialEq; false negatives is exactly what PartialEq was intended for, even if I suspect we have greater scope for them than, say, `f64` !

src/hugr/rewrite/replace.rs
```diff
@@ -0,0 +1,591 @@
...
+        let parent = self
+            .removal
+            .iter()
+            .map(|n| h.get_parent(*n))
+            .unique()
+            .exactly_one()
+            .map_err(|ex_one| ReplaceError::MultipleParents(ex_one.flatten().collect()))?
+            .ok_or(ReplaceError::CantReplaceRoot)?; // If no parent
+
+        // Check replacement parent is of same tag. Should we require exactly OpType equality?
```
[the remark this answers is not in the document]
acl-cqc: Yes. I mean, given the replacement Hugr might be invalid, the signature is in fact meaningless! But for cases where you want to change the row of inputs/outputs to a nested DFG, say (which will temporarily make the Hugr invalid - presumably you'll then follow up with some more replaces or add/remove wires), checking signature equality here is only going to make things weirder and the replacement harder to construct...

src/hugr/rewrite/replace.rs
```diff
@@ -0,0 +1,591 @@
...
+    /// The node(s) to replace had no parent i.e. were root(s).
+    // (Perhaps if there is only one node to replace we should be able to?)
+    #[error("Cannot replace the root node of the Hugr")]
+    CantReplaceRoot,
+    /// The nodes to replace did not have a unique common parent
+    #[error("Removed nodes had different parents {0:?}")]
+    MultipleParents(Vec<Node>),
+    /// Replacement root node had different tag from parent of removed nodes
+    #[error("Expected replacement root with tag {expected} but found {actual}")]
+    #[allow(missing_docs)]
```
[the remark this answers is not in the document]
acl-cqc: For the fields of WrongRootNodeTag...well, so I've renamed them, which I think is the better solution :), but ok, docs added

src/hugr/rewrite/replace.rs
```diff
@@ -0,0 +1,591 @@
...
+        // Check the keys of the transfer map too, the values we'll use imminently
+        self.transfers.keys().try_for_each(|&n| {
+            (self.replacement.contains_node(n)
+                && self.replacement.get_optype(n).is_container()
+                && self.replacement.children(n).next().is_none())
+            .then_some(())
+            .ok_or(ReplaceError::InvalidTransferTarget(n))
+        })?;
+        let mut transferred: HashSet<Node> = self.transfers.values().copied().collect();
+        if transferred.len() != self.transfers.values().len() {
```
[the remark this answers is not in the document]
acl-cqc: This is tempting, but really "all values distinct" is only a subpart (and implied by) "no value may be a descendant or ancestor of another", which we have to check anyway, so I think the gain is quite small

src/hugr/rewrite/replace.rs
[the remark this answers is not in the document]
acl-cqc: Noted we do not check exact signature.

src/hugr/rewrite/replace.rs
```diff
@@ -0,0 +1,591 @@
...
+        src_pos: OutgoingPort,
+    },
+}
+
+/// Specification of a `Replace` operation
+#[derive(Debug, Clone)] // PartialEq? Eq probably doesn't make sense because of ordering.
+pub struct Replacement {
+    /// The nodes to remove from the existing Hugr (known as Gamma).
+    /// These must all have a common parent (i.e. be siblings).  Called "S" in the spec.
+    /// Must be non-empty - otherwise there is no parent under which to place [replacement],
```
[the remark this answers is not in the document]
acl-cqc: Then it would be just "add nodes" rather than "Replacement" - and, we'd be unable to connect the newly-added nodes to the rest (as there'd be no removed edges...) so I think I'm gonna stick with this.

src/builder/build_traits.rs
```diff
@@ -73,9 +73,9 @@ pub trait Container {
     fn add_constant(
         &mut self,
         constant: ops::Const,
-        extensions: ExtensionSet,
+        extensions: impl Into<Option<ExtensionSet>>,
```
acl-cqc, line 76: Slight generalization here @ss2165 if you are happy with that? All existing calls remain legal.

acl-cqc: I can do something slightly odd in this PR (i.e. explicitly specify an ExtensionSet that is...not what you might expect) if we'd rather keep like this, 'coz either way we should probably at least try to make another PR where we remove this (and use None/open-extensions) everywhere

acl-cqc:
* Renamed "transfers" to "adoptions" which I think is the appropriate English word for finding someone a new parent....I wondered about something involving "orphans" but didn't quite figure that out.
* Combined ConflictingTransfers and TransfersNotSeparateDescendants into AdopteesNotSeparateDescendants (could be ConflictingAdoptees) - either means the same thing, that we are trying to "move" some old nodes into multiple places
* Refactoring of `check_existing_edge` to allow reporting the correct NewEdgeSpec if there's an error (aaiieeee)
* Assert more about the Hugr in the "main" `cfg` test
* Add `test_invalid` which both verifies and applies at least one case for every variant of ReplaceError, although I do not claim to have got one error out of each possible error path!

cqc-alec (approved):

PR #614, opened 2023-10-20
Static checking of Port

acl-cqc:
A bunch of code used to go via `Port::try_new_incoming` (and similarly `try_new_outgoing`), allowing to use both existing Ports (failing at runtime if direction incorrect) as well as ints etc.

Instead, separate out the only fallible case (Port -> IncomingPort/OutgoingPort) - add fallible methods `Port::as_incoming` and `Port::as_outgoing` to perform those conversions. APIs that want to enforce a direction now take `Into<IncomingPort>` (rather than TryInto).

This is generally not too painful, indeed in some cases quite the opposite, the key part was to extend directionality further - e.g. `Wire` stores an `OutgoingPort`, `HugrView::node_inputs`/`node_outputs` return iterators over that element type, etc. - and new methods `linked_inputs` and `linked_outputs` propagate the knowledge that IncomingPorts are only connected to OutgoingPorts, etc.

### DIFF
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -60,7 +60,7 @@ pub trait Container {
     /// [`other_outputs`]: crate::ops::OpTrait::other_output
     fn add_other_wire(&mut self, src: Node, dst: Node) -> Result<Wire, BuildError> {
         let (src_port, _) = self.hugr_mut().add_other_edge(src, dst)?;
-        Ok(Wire::new(src, src_port.as_outgoing().unwrap()))
+        Ok(Wire::new(src, src_port))
     }
 
     /// Add a constant value to the container and return a handle to it.
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -1,7 +1,7 @@
 use crate::hugr::hugrmut::InsertionResult;
 use crate::hugr::validate::InterGraphEdgeError;
 use crate::hugr::views::HugrView;
-use crate::hugr::{IncomingPort, Node, NodeMetadata, OutgoingPort, Port, ValidationError};
+use crate::hugr::{IncomingPort, Node, NodeMetadata, OutgoingPort, ValidationError};
 use crate::ops::{self, LeafOp, OpTrait, OpType};
 
 use std::iter;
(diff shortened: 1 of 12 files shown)


src/builder/build_traits.rs
```diff
@@ -675,9 +675,9 @@ fn wire_up<T: Dataflow + ?Sized>(
             if !typ.copyable() {
                 let val_err: ValidationError = InterGraphEdgeError::NonCopyableData {
                     from: src,
-                    from_offset: src_port,
+                    from_offset: src_port.into(),
```
acl-cqc, line 678: Possibly we could extend `NonCopyableData` to take an OutgoingPort and an IncomingPort, etc., etc. - by all means shout out cases like that, I am not claiming to have found them all

src/hugr/rewrite/outline_cfg.rs
```diff
@@ -162,7 +162,8 @@ impl Rewrite for OutlineCfg {
         for (pred, br) in preds {
             if !self.blocks.contains(&pred) {
                 h.disconnect(pred, br).unwrap();
-                h.connect(pred, br, new_block, 0).unwrap();
+                h.connect(pred, br.as_outgoing().unwrap(), new_block, 0)
```
aborgna-q, line 165: You could avoid the conversion by using `preds = h.linked_outputs(...)` above.

src/hugr/views.rs
```diff
@@ -300,6 +320,48 @@ pub trait HugrView: sealed::HugrInternals {
     }
 }
 
+/// Wraps an iterator over [Port]s that are known to be [OutgoingPort]s
+#[derive(Debug)]
+pub struct OutgoingPorts<T>(T);
```
aborgna-q, line 325: Alternatively,
```rust
pub type OutgoingPorts<I> = std::iter::Map<I, fn(Port) -> OutgoingPort>
```
and then create it with
```rust
self.node_ports(node, Direction::Outgoing).map(|p| p.as_outgoing().unwrap());
```

Otherwise, you need to derive the other iterator traits: `FusedIterator`, `DoubleEndedIterator`, `ExactSizeIterator`.

aborgna-q (commented):
Nice! This feels much better.

Only nitpick is the iterator wrappers.

src/hugr/rewrite/outline_cfg.rs
acl-cqc: Good spot! :)

src/hugr/views.rs
acl-cqc: Nice!! I was hoping to do something like this but somehow got stuck looking at much more complicated `type` definitions than your simple one - moreover, more complicated ones that didn't work!

The other possibility would be to redefine `NodeInputs` as parameterized (over `T: Into<Port>`); and then leave `node_inputs` and `node_outputs` both to subclasses, with `node_ports` picking one or the other and then `.map_into`....no, that doesn't work, because the type of the MapInto depends on the source type, bah. Ok - thanks! :)

acl-cqc: Also, we could inline the definition of `IncomingNodes` (etc.) - each is only used once now...

aborgna-q (approved):
Nice!

PR #616, opened 2023-10-23
refactor: Default to open extensions in branch of any_block_builder

croyzor:


acl-cqc (commented):
If we made `add_op_with_parent` automatically use `add_node_with_parent` + `open_extensions`, then this change would have no effect, right.

#618 appears to do that? So if #618 works...???

croyzor:
> If we made add_op_with_parent automatically use add_node_with_parent + open_extensions, then this change would have no effect, right.

Yes, but...

There are other callers of `add_op_with_parent`, so that doesn't quite work to fix this problem without breaking other things. Also `add_node_with_parent`, calls `add_node` which means that we can't delete it in #618, only generalise the `add_op` version

acl-cqc:
Superceded by #622

PR #617, opened 2023-10-23
refactor: Remove `add_node_before`

croyzor:


src/hugr/hugrmut.rs
```diff
@@ -216,11 +207,7 @@ impl<T: RootTagged<RootHandle = Node> + AsMut<Hugr>> HugrMut for T {
     }
 
     fn add_op_before(&mut self, sibling: Node, op: impl Into<OpType>) -> Result<Node, HugrError> {
-        self.add_node_before(sibling, NodeType::pure(op))
-    }
-
-    fn add_node_before(&mut self, sibling: Node, nodetype: NodeType) -> Result<Node, HugrError> {
-        let node = self.as_mut().add_node(nodetype);
+        let node = self.as_mut().add_node(NodeType::open_extensions(op));
```
croyzor, line 210: I think we can remove `add_node` too, but I'll open another PR for that

acl-cqc (commented):
I'm very nervous of this on its own because it leaves `add_op_before` using `open_extensions` whereas `add_op_with_parent` uses `pure`....

However, perhaps we can just fold #618 into this. Does *that* pass tests?

acl-cqc:
I think we can drop this following #622 ??

PR #618, opened 2023-10-23
refactor: Generalise extensions for `add_op`

croyzor:


acl-cqc (commented):
Arrgh all these different add_{node,op}{,_with_parent,_after,_before} methods and the different implementations of each are looking increasingly mazelike. Does or doesn't this affect `add_op_with_parent`?

acl-cqc:
Superceded by #622

PR #619, opened 2023-10-23
refactor: Default more functions to using open_extensions

croyzor:
A combined version of #616 #617 and #618

### DIFF
--- a/src/builder/cfg.rs
+++ b/src/builder/cfg.rs
@@ -146,8 +146,8 @@ impl<B: AsMut<Hugr> + AsRef<Hugr>> CFGBuilder<B> {
             // TODO: Make extensions a parameter
             self.hugr_mut().add_op_before(exit, op)
         } else {
-            // TODO: Make extensions a parameter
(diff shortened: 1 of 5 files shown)


acl-cqc (commented):
This looks pretty good, but I've been prodding around here myself too and discovering how much of a maze these various methods are :-(. Does this make `add_op_with_parent` also use `open_extensions`?

acl-cqc:
> Does this make `add_op_with_parent` also use `open_extensions`?

No, it doesn't :(. So I think it's non-obvious, indeed counterintuitive, that which `add_op`/`add_node` function you use, changes what `Option<ExtensionSet>` you get. I think we should probably focus on figuring out why using `open_extensions` everywhere doesn't work.

croyzor:
`add_op_with_parent` still uses `pure`, or there'd be no reason to change the implementation of `any_block_builder`.
Still need to investigate what needs done to update `add_op_with_parent`, unfortunately

acl-cqc:
Superceded by #622

PR #620, opened 2023-10-23
refactor: Remove add_op/node_after and move_after

acl-cqc:
We only use these in a couple of places, and we never need to - we're always adding after the last sibling of a parent that we have handy anyway.

The `_before` versions are much more widely used, but maybe we only need one rather than both?

Note [this run](https://github.com/CQCL/hugr/actions/runs/6614791644/job/17965516726) without the last commit: `clippy --all-targets` complains that `add_op` is unused, even though it *is* used. But it's not that hard to remove...

### DIFF
--- a/src/hugr.rs
+++ b/src/hugr.rs
@@ -303,12 +303,6 @@ impl Hugr {
         }
     }
 
-    /// Add a node to the graph, with the default conversion from OpType to NodeType
-    pub(crate) fn add_op(&mut self, op: impl Into<OpType>) -> Node {
-        // TODO: Default to `NodeType::open_extensions` once we can infer extensions
-        self.add_node(NodeType::pure(op))
-    }
-
     /// Add a node to the graph.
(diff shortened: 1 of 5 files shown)


aborgna-q (commented):
I don't really see a motivation to drop these. Yes, we mostly only add `_before` in hugr, but the footprint of this is minimal.

Outside of tests, these are used for manipulating basic blocks where the `exit` has to be last, or ordering conditional branches. Any of these could have a use for `_after` at some point.

acl-cqc:
> I don't really see a motivation to drop these. Yes, we mostly only add `_before` in hugr, but the footprint of this is minimal.
> 
> Outside of tests, these are used for manipulating basic blocks where the `exit` has to be last, or ordering conditional branches. Any of these could have a use for `_after` at some point.

Yeah, ok, was trying to drain the swamp ;-), but as you say this is only really chipping around the edges, and #622 is much more the right direction; can probably do more after that. Ta :)

Issue #457, opened 2023-08-25
Validation for graphs with resource variables


croyzor:
Currently, extension validation assumes that hugrs should be extensions-closed. Which is to say that every node in the hugr should have concrete extensions for its incoming and outgoing edges. This makes sense for a static function definition, but may be needlessly restrictive for other types of hugrs.

Concretely, in the below diagram, input0 could have an open extension variable as its input extensions. Extension inference could conclude that the rest of the graph makes sense given any value for this variable, but validation would complain that `DFG0`'s input doesn't have a concrete extension set!
```mermaid
flowchart
    direction TB
    subgraph DFG0
        direction TB
        subgraph DFG1
            direction TB
            Input1 --> op0
            op0 --> op1
            op1 --> Output1
        end
        Input0 --> DFG1
        Input1 --> DFG1
        DFG1 --> Output0
    end
```
The workaround for this lack of more permissive extension validation as introduced by #456 is for extension inference to return a _closure_ -- a mapping of solutions for extension sets in the graph which depend on open variables which it then passes to validation, allowing validation to operate on an extension-closed hugr.

I.e. extension inference checks that there is a solution given any value for the remaining extension variables, and for validation it sets their values to the empty extension set.

acl-cqc:
Update: following #713 the main thing missing - given that we already [validate type-variables](https://github.com/CQCL/hugr/blob/d98fb791a5ca5651fba659e0e55da59db451a975/hugr/src/hugr/validate.rs#L525-L565) (including ExtensionSet ones) appearing in TypeArgs are declared - and that [all port kinds are valid similarly](https://github.com/CQCL/hugr/blob/d98fb791a5ca5651fba659e0e55da59db451a975/hugr/src/hugr/validate.rs#L265-L278) - we think the only things remaining are:
* the first `match` might want to include other LeafOps that can include Types (e.g. Noop). However, given #816, this may be unnecessary
* as well as validating the port-kinds of each node's signature, we should also validate the extension-delta

acl-cqc:
New approach using extension typevars should let us do something similar to this following #709

PR #622, opened 2023-10-24
new: Default to open extensions

### LINKED ISSUE #424: Default graph builder nodes to open ext requirements, rather than empty

acl-cqc:
This continues in the theme of #618, #619, etc., but I think is now beyond "refactor" :).

Algorithmically, there was only one change - in `infer.rs` - where Meta's on which no constraints would be placed are now considered variables. I think that's reasonable, but the definition is a bit loose. Maybe there is a better way (e.g. in `instantiate_variables`)?

Secondly some gymnastics in one of the tests where we want the correct error out of validation but can't do inference because the Hugr is, well, invalid....

closes #424


src/extension/infer.rs
```diff
@@ -338,16 +341,16 @@ impl UnificationContext {
                         | Some(EdgeKind::ControlFlow)
                 )
             }) {
+                let m_tgt = *self
```
acl-cqc, line 344: This is a driveby lifting of a loop invariant, nothing more

src/hugr/validate.rs
```diff
@@ -997,16 +997,20 @@ mod test {
...
+        let mut closure = b.infer_extensions().unwrap();
+        assert!(closure.remove(&copy).is_some());
         b.replace_op(
             copy,
             NodeType::pure(ops::CFG {
                 signature: FunctionType::new(type_row![BOOL_T], type_row![BOOL_T]),
             }),
         )
         .unwrap();
+        // We feed in the closure from above because we can't compute the closure
```
acl-cqc, line 1009: This suggests we might want a `&mut Hugr` method that writes the values from the closure into the Hugr (Changing `input_extensions` from `None` to `Some`)

src/hugr/validate.rs
```diff
@@ -997,7 +997,11 @@ mod test {
             .map_into()
             .collect_tuple()
             .unwrap();
-
+        // Write Extension annotations into the Hugr while it's still well-formed
+        // enough for us to compute them
```
acl-cqc, line 1001: This was the tricky one ;-). I found another way that also works, using `validate_with_extension_closure`; commit 923415479183d54d94d753dc55059aac136ac965 swapped that out for this version, I'd be happy to go with either.

src/extension/infer.rs
```diff
@@ -312,6 +307,14 @@ impl UnificationContext {
             match node_type.signature() {
                 // Input extensions are open
                 None => {
+                    if node == hugr.root()
```
acl-cqc, line 310: The check on this line is equivalent to the one removed above (lines 271-275); the `||` on next line extends it and is a first stab at what nodes we'd expect to be unconnected. (An alternative might be to skip creating `m_input` and `m_output` earlier in the loop...that might be better, happy to give that a go). Note #457 .

This still won't address random unconnected bits of graph, which is to say, malformed graphs, tho. (Only really an issue for tests of validation - some of these have to jump through hoops to get the error they expect rather than "CantInfer")

src/hugr/validate.rs
acl-cqc: (Using that way in the other test!)

src/hugr/validate.rs
acl-cqc: Oh, hang on, that's `instantiate_extensions`

src/extension/infer.rs
```diff
@@ -317,6 +317,12 @@ impl UnificationContext {
                         m_output,
                         node_type.op_signature().extension_reqs,
                     );
+                    if matches!(
```
acl-cqc, line 320: I've tried a few things here - see e.g. commit 59eb260 and [this comment](https://github.com/CQCL/hugr/pull/622/files#r1372772845). Think this is better than that...

but a better-still way (???) might be to define `OpType::default_extensions() -> Option<ExtensionSet>` and then make `add_op_xxxx` use not `open_extensions` but `default`. (Where `default_extensions` is `None` i.e. open for most OpTypes, but `ExtensionSet::new()` i.e. pure for Alias/Function/FuncDefn.) Thoughts?

acl-cqc: (Could be good for a later refactor - at the same time as sorting out `NodeType::pure` etc. as that should be `new_pure` or `OpType::pure(self)`, like `FunctionType::pure(self) -> Signature`)

croyzor: Good ideas, both imo! The first one, making "open_extensions" a more well supported "default" option was hampered previously by the lack of extension inference so now we can reconsider

croyzor (approved):

PR #623, opened 2023-10-24
refactor: change &NodeType->&OpType conversion into op() accessor

acl-cqc:
* Remove `impl From<&NodeType> for &OpType`
* Add new NodeType `fn op(&self) -> &OpType`


aborgna-q (approved):
:+1:

PR #625, opened 2023-10-25
feat!: Add panicking integer division ops

doug-q:
TODO

will need to find better names before this is merged.
Marking as a breaking change as I anticipate using the old names for the new ops


doug-q:
I think the wasm names should go to these new ops, since they match wasm semantics. What then should the error value returning ops be names?

cqc-alec:
Agree with keeping the old names for the panicking operations. For the ones that don't panic, how about adding `_checked` to the names, e.g. `idivmod_checked_u`?

cqc-alec (approved):

PR #627, opened 2023-10-26
refactor: move the core types to their own module

aborgna-q:
Moves `CircuitUnit, Direction, IncomingPort, Node, NodeIndex, OutgoingPort, Port, PortIndex, Wire`
to a private `::core` module, and re-exports this from the root.

Most of the noise comes from modules accessing the internals of `Port` and `Node`, that now have to go through an accessor.


mark-koch (approved):
Looks good to me 👍

PR #630, opened 2023-10-30
feat: polymorphic function types (inc OpDefs) using dyn trait

acl-cqc:
* Add type variable as a variant of PrimType and TypeArg
* Add types/poly_func.rs containing PolyFuncType - a list of binders ('TypeParam's) and a FunctionType
   * can be instantiated to a FunctionType, or (perhaps-partially) a PolyFuncType
* PrimType contains PolyFuncType, not FunctionType
* Add `substitute` method to Type, TypeArg, et al., taking a Substitution trait which provides values for type variables
   * Implementations SubstValues (when instantiating) as well as InsideBinders and Renumber for dealing with PolyFuncTypes 
* OpDefs may have either binary compute_signature functions or PolyFuncType "type schemes"; the former (i.e. binary functions) return the latter.
* Add LeafOp::TypeApply whose signature represents applying a PolyFuncType to a type-argument - the argument of kind (i.e. TypeParam) depending on the former

BREAKING CHANGE: * Rename OpDef::instantiate_concrete to just instantiate; * PrimType::Function and Type::new_function take PolyFuncType not FunctionType


mark-koch (commented):
I like this version! I'm not a Rust expert so can't really comment on the performance implications, but I find the substitution logic in this PR easier to understand compared to #542

src/types.rs
```diff
@@ -320,36 +318,41 @@ impl Type {
...
+            TypeEnum::Prim(PrimType::Extension(cty)) => Type::new_extension(cty.transform(t)),
+            TypeEnum::Prim(PrimType::Function(bf)) => Type::new_function(bf.transform(t)),
+            TypeEnum::Tuple(elems) => Type::new_tuple(transform_row(elems, t)),
+            TypeEnum::Sum(SumType::General { row }) => Type::new_sum(transform_row(row, t)),
         }
     }
 }
 
+pub(crate) trait TypeTransformer {
+    fn apply_typevar(&self, idx: usize, bound: TypeBound) -> Type {
```
ss2165, line 336: I think a dedicated type/newtype for type var index would help disambiguate a bit

ss2165 (approved):
Yes, agree with Mark this is much nicer

src/types.rs
acl-cqc: Great idea, thanks, will do

src/types/poly_func.rs
```diff
@@ -115,16 +120,66 @@ impl PolyFuncType {
...
+impl<'a> TypeTransformer for InsideBinders<'a> {
+    fn apply_var(&self, idx: usize, decl: &TypeParam) -> TypeArg {
+        // Don't touch the first <self.0> variables
+        if idx < self.0 {
+            return TypeArg::new_var_use(idx, decl.clone());
+        }
+        let under = self.1.apply_var(idx - self.0, decl);
+        // Make returned value (from underlying substitution, outside the
+        // new binders) avoid the variables newly bound
+        under.transform(&Renumber(self.0, self.extension_registry()))
```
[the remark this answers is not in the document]
acl-cqc: Yes! Also much easier to write, the old `Substitution` was really hard!!

acl-cqc:
@ss2165 I've done most of what I wanted, but I haven't converted usize to a VariableIndex struct yet. Initial impression was that this was a lot more boilerplate and less gain than I hoped - the big win I was hoping for was cleaning up `InsideBinders::apply_var` and I found `usize::checked_sub` actually does what I wanted there anyway. So I'm still investigating that concurrently, but in the meantime I think this is ready to take a look as it stand.

src/types/poly_func.rs
```diff
@@ -0,0 +1,613 @@
...
+use crate::{
+    extension::{ExtensionRegistry, SignatureError},
+    types::type_param::check_type_arg,
+};
+use itertools::Itertools;
+
+use super::type_param::{check_type_args, TypeArg, TypeParam};
+use super::{FunctionType, Substitution};
+
+/// A polymorphic function type, e.g. of a [Graph], or perhaps an [OpDef].
```
acl-cqc, line 12: strictly, not "of a Graph" yet/in this PR

src/types/poly_func.rs
```diff
@@ -0,0 +1,613 @@
...
+)]
+#[display(
+    fmt = "forall {}. {}",
+    "params.iter().map(ToString::to_string).join(\" \")",
+    "body"
+)]
+pub struct PolyFuncType {
+    /// The declared type parameters, i.e., these must be instantiated with
+    /// the same number of [TypeArg]s before the function can be called. Note that within
+    /// the [Self::body], variable (DeBruijn) index 0 is element 0 of this array, i.e. the
```
acl-cqc, line 28: This will be very familiar to BRAT developers as the Bwd list...

src/types/poly_func.rs
```diff
@@ -0,0 +1,613 @@
...
+            params: self.params.clone(),
+            body: self.body.substitute(&InsideBinders {
+                skip_lowest: self.params.len(),
+                underlying: t,
+            }),
+        }
+    }
+
+    /// (Perhaps-partially) instantiates this [PolyFuncType] into another with fewer binders.
+    /// Note that indices into `args` correspond to the same index within [Self::params],
```
acl-cqc, line 97: The alternative - better in theory but I feel more surprising to users - would be to make the end of the `args` list line up with the end of `params`, consistent with them both being Bwd lists...

src/types/poly_func.rs
acl-cqc: I guess the other way would be to treat all variable indices as negative offsets from the end, but that seems pretty nasty

mark-koch: I agree, using the variable as an index into the list is probably more natural. I feel the reverse indexing would be more surprising

acl-cqc: Yes, so branch `new/poly_types_varidx` manages to hide all that away behind the VarIdx struct, and it is largely restricted to the `types` module - HOWEVER anyone building PolyFuncTypes will have to work with indices directly, so unless we offer Yaml parsing as the only interface, the L-R / R-L indexing is exposed to endusers, and I am leaning towards the current scheme.

I'll put the VarIdx changes *without changing indexing* up as a separate "refactor" PR so we can consider them later if we want, but it seems ok to do that as a second step.

src/types/poly_func.rs
```diff
@@ -147,36 +152,45 @@ impl<'a> Substitution for SubstValues<'a> {
...
-        self.1
+        self.exts
     }
 }
 
 /// Given a [Substitution] defined outside a binder (i.e. [PolyFuncType]),
 /// applies that transformer to types inside the binder (i.e. arguments/results of said function)
-struct InsideBinders<'a>(usize, &'a dyn Substitution);
+struct InsideBinders<'a> {
+    skip_lowest: usize,
```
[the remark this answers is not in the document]
acl-cqc: Yes, done

acl-cqc:
Also renamed the old `instantiate` to `instantiate_poly` and `instantiate_all` to `instantiate` - the latter is then consistent with OpDef/TypeDef, the former is the odd-one-out hence the extra suffix to indicate it's unusual

mark-koch:
> Also renamed what was `instantiate` to `instantiate_poly`

Maybe `instantiate_partial`?

acl-cqc:
> > Also renamed what was `instantiate` to `instantiate_poly`
> 
> Maybe `instantiate_partial`?

It doesn't have to be partial - the thing is that if total, you'd get back a PolyFuncType (with no binders) rather than a FunctionType

ss2165 (approved):

ss2165 (approved):

ss2165 (approved):
Very nice! Looking forward to using this all over extensions.

src/extension/op_def.rs
```diff
@@ -274,58 +277,111 @@ impl Extension {
...
+        let reg1 = [PRELUDE.to_owned(), EXTENSION.to_owned()].into();
+        let list_def = EXTENSION.get_type(&LIST_TYPENAME).unwrap();
+        let mut e = Extension::new(EXT_ID);
+        const TP: TypeParam = TypeParam::Type(TypeBound::Any);
+        let list_of_var =
+            Type::new_extension(list_def.instantiate(vec![TypeArg::new_var_use(0, TP)])?);
+        const OP_NAME: SmolStr = SmolStr::new_inline("Reverse");
+        let type_scheme = PolyFuncType::new_validated(
+            vec![TP],
+            FunctionType::new_linear(vec![list_of_var]),
```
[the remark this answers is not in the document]
acl-cqc: Agreed I don't really like `new_linear` either. Raised #664.

PR #631, opened 2023-10-30
refactor: infer.rs 'fn results'

acl-cqc:
* No need for `&self` to be mutable
* Make control-flow more obvious
* Also, in merge_equal_metas(), do if contains ... remove.unwrap in one step

No change to semantics.


acl-cqc:
@cqc-alec reassigning this as Seyon sounds busy and I'm not looking for any subject area expertise - this is semantics-preserving!

cqc-alec (approved):

PR #632, opened 2023-10-30
ci: Use codecov for coverage reports

aborgna-q:
The slack reporting can be added later with codecov's slack app


.github/workflows/coverage.yml
```diff
@@ -1,68 +1,37 @@
...
+      - name: Run tests with coverage instrumentation
         run: |
-          change="`cat coverage-summary-prev.txt`% --> `cat coverage-summary.txt`%"
-          if (( $(echo "`cat coverage-summary-prev.txt` < `cat coverage-summary.txt` + 0.04" | bc -l) ))
-          then
-            echo "msg=Coverage check for hugr shows no regression (${change}). ✅ ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" >> "$GITHUB_OUTPUT"
-          else
-            echo "msg=Coverage check for hugr shows regression (${change}). ❌ ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" >> "$GITHUB_OUTPUT"
-          fi
-  notify-slack:
```
cqc-alec, line 57: Is there a way we can still send a notification to slack with the headline number?

cqc-alec (approved):

.github/workflows/coverage.yml
aborgna-q: I hoped the slack app would have this option, but I didn't find it.

I brought back the cron job (simplified a bit with cargo-llvm-cov). Instead of storing the full coverage result it now just links to the codecov page.
Unfortunately there is no web view for comparing two commits (that's only on the API), so I just link to the [landing page](https://app.codecov.io/gh/CQCL/hugr?search=&trend=7%20days) with a trend graph.

PR #635, opened 2023-10-31
refactor: NodeType constructors, adding new_auto

acl-cqc:
* Rename NodeType::open_extensions to NodeType::new_open
* Rename NodeType::pure to NodeType::new_pure
* Add NodeType::new_auto, which uses Pure for module-ops and Open for others
* Remove special-case in infer.rs solving some module-ops to empty set
* Switch builder/HugrMut methods from new_open to new_auto

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -146,7 +146,7 @@ pub(crate) mod test {
     /// inference. Using DFGBuilder will default to a root node with an open
     /// extension variable
     pub(crate) fn closed_dfg_root_hugr(signature: FunctionType) -> Hugr {
-        let mut hugr = Hugr::new(NodeType::pure(ops::DFG {
+        let mut hugr = Hugr::new(NodeType::new_pure(ops::DFG {
             signature: signature.clone(),
         }));
         hugr.add_op_with_parent(
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -200,7 +200,7 @@ pub trait Dataflow: Container {
         op: impl Into<OpType>,
         input_wires: impl IntoIterator<Item = Wire>,
     ) -> Result<BuildHandle<DataflowOpID>, BuildError> {
-        self.add_dataflow_node(NodeType::new_open(op), input_wires)
+        self.add_dataflow_node(NodeType::new_default(op), input_wires)
     }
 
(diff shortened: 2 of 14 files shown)


croyzor:
I think while we're doing this, we should probably go for `new_open` and `new_closed` or just `open` and `closed` (I prefer the latter). The name for the `new_default` could be clearer about what it does. If it were actually the default option, it should just be called `new` and used as the de facto default, but I think it's a bit to unintuative for that

acl-cqc:
> I think while we're doing this, we should probably go for `new_open` and `new_closed` or just `open` and `closed` (I prefer the latter).

Happy to change pure to closed. (I never really liked "pure", but I admit changing it to "closed" does make me think maybe "pure" isn't so bad! Closed is opposite to open so should be *any fixed ExtensionSet* ??)

I'm pretty clear these should be `new_` methods, the alternative would be to make them methods on Op (like FunctionType::pure, which I've renamed to FunctionType::closed). The unfortunate thing there is that that's a trait, so we could define a Trait-default `fn open(self) -> NodeType` but we'd have to be saying, Trait implementors should never override this...

> The name for the `new_default` could be clearer about what it does. If it were actually the default option, it should just be called `new` and used as the de facto default, but I think it's a bit to unintuitive for that

I take `new` as being the maximally-explicit version that allows anything; the alternative would be to rename new to `new_with_extensions` or something (except that it can also accept None = extensions unspecified!). Then we *could* rename new_default to new but as you say I think that'd be rather surprising.

So instead I've renamed new_default to new_auto, how's that?

croyzor:
You make a good point, `closed` could apply to any fixed extension set, so I'm happy to revert to `pure`.
I feel the symmetry of open/closed constructors is somehow less important when they're prefixed with new, so if we're sticking with that then it's all good.

> So instead I've renamed new_default to new_auto, how's that?

 :+1:

acl-cqc:
> You make a good point, `closed` could apply to any fixed extension set, so I'm happy to revert to `pure`. I feel the symmetry of open/closed constructors is somehow less important when they're prefixed with new, so if we're sticking with that then it's all good.
> 
> > So instead I've renamed new_default to new_auto, how's that?
> 
> 👍

Ok, so new_pure, new_open, and new_auto it is...

croyzor (approved):

PR #636, opened 2023-11-01
refactor: Constraint::Plus stores an ExtensionSet, which is a BTreeSet

acl-cqc:
* Convert ExtensionSet from HashSet to BTreeSe => trivially derive Hash
* Constraint::Plus stores an ExtensionSet, rather than a single Extension => gen_union_constraint avoids long chains of metas
* Inline gen_union_constraint (no longer recursive)

### DIFF
--- a/src/extension/infer.rs
+++ b/src/extension/infer.rs
@@ -230,17 +230,6 @@ impl UnificationContext {
         self.solved.get(&self.resolve(*m))
     }
 
-    fn gen_union_constraint(&mut self, input: Meta, output: Meta, delta: ExtensionSet) {
-        self.add_constraint(
-            output,
-            if delta.is_empty() {
-                Constraint::Equal(input)
-            } else {
-                Constraint::Plus(delta, input)
-            },
-        );
-    }
-
     /// Return the metavariable corresponding to the given location on the
     /// graph, either by making a new meta, or looking it up
     fn make_or_get_meta(&mut self, node: Node, dir: Direction) -> Meta {
@@ -302,11 +291,13 @@ impl UnificationContext {
             match node_type.signature() {
                 // Input extensions are open
                 None => {
-                    self.gen_union_constraint(
-                        m_input,
-                        m_output,
-                        node_type.op_signature().extension_reqs,
-                    );
+                    let delta = node_type.op_signature().extension_reqs;
+                    let c = if delta.is_empty() {
+                        Constraint::Equal(m_input)
+                    } else {
+                        Constraint::Plus(delta, m_input)
+                    };
+                    self.add_constraint(m_output, c);
                     if matches!(
                         node_type.tag(),
                         OpTag::Alias | OpTag::Function | OpTag::FuncDefn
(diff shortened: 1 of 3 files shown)


aborgna-q:
> the Hash implementation here deliberately goes against the instruction of how to use Hash

It doesn't. The [requirement](https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq) of `Hash` is that $k_1 = k_2 \implies \texttt{hash}(k_1) = \texttt{hash}(k_2)$.
The `HashSet` equality is order-invariant, so hashing should be too.

aborgna-q:
Ah, I see what you mean. You are forced to introduce a hand-picked hasher.

You could sort the `ExtensionId`s instead, or just use a `BTreeSet`.

acl-cqc:
[Found it again](https://doc.rust-lang.org/std/hash/trait.BuildHasher.html):
```This must not be used in any code which creates hashes, such as in an implementation of [Hash](https://doc.rust-lang.org/std/hash/trait.Hash.html). The way to create a combined hash of multiple values is to call [Hash::hash](https://doc.rust-lang.org/std/hash/trait.Hash.html#tymethod.hash) multiple times using the same [Hasher](https://doc.rust-lang.org/std/hash/trait.Hasher.html), not to call this method repeatedly and combine the results.```

**Comment by acl-cqc** [comment]
Actually a BTreeSet might not be a bad way to go. IIUC this might even enable making some ExtensionSet methods `const`....I'll have a go

**Comment by acl-cqc** [comment]
> Actually a BTreeSet might not be a bad way to go. IIUC this might even enable making some ExtensionSet methods `const`....I'll have a go

Ok, I thought I remembered a `BTreeSet::singleton` const method, but it is not so. However, converted to BTreeSet, that was painless, and Hash is ready-provided :-) :-)

**Comment by acl-cqc** [comment]
AFAICS this is reducing code coverage by shortening and simplifying the tested part of the codebase, therefore the untested part of the codebase (not changed by this PR) is now a greater proportion....?

**Comment by aborgna-q** [comment]
Yep, that's why the `patch` test passes, but `project` fails. Just ignore it.

File: src/extension/infer.rs [hunk 1]
```diff
@@ -311,11 +291,13 @@ impl UnificationContext {
...
                 // Input extensions are open
                 None => {
-                    self.gen_union_constraint(
-                        m_input,
-                        m_output,
-                        node_type.op_signature().extension_reqs,
-                    );
+                    let delta = node_type.op_signature().extension_reqs;
+                    let c = if delta.is_empty() {
+                        Constraint::Equal(m_input)
```
acl-cqc, line 296: The alternative would be to *store* this with the empty ExtensionSet, and then we would only be storing one kind of Constraint (so, a struct, not an enum - might be syntactically easier to work with!). Then we'd be testing `ExtensionSet::is_empty` instead of matching on `Constraint::Plus` vs `::Equals`...

croyzor (approved):

PR #637, opened 2023-11-01
refactor: Be safe

cqc-alec:
Use a trick suggested by @acl-cqc to remove the `unsafe`.

Drive-by: Remove unneeded clippy annotation.


aborgna-q:
Nice! And it all gets compiled to the same code
https://godbolt.org/z/M4EcWaWMT

aborgna-q (approved):

PR #638, opened 2023-11-02
fix: erratic stack overflow in infer.rs (live_var)

### LINKED ISSUE #633: Erratic stack overflow in infer.rs 'live_vars'
* See branch `tests/infer_stack_overflow` for the test (not very minimal, sorry - I haven't tried to reduce)
* Commit 6d4b446b3d46e2afbbec2c80dc105a2a675c2643 precisely identifies the place where the HashSet's iteration order determines the outcome, and makes the test panic every time (without affecting any other test - obviously this hasn't come up before)
* The problem seems to be that `merge_equal_metas` has found a set of equal variables where one also has a Plus constraint on another. This results in the merged variable effectively having a plus constraint on itself!
* I also tried the fix of #591 (see branch `tests/infer_stack_overflow2`) , but although that PR adds code to deal with cycles, it doesn't fix the bug. I think the problem is that `live_var` (where the infinite-recursion happens) is called from `results` *before* we get to instantiating variables. This suggests that we may need to deal with cycles of Plus nodes (as #591 does) earlier, i.e. in `main_loop` rather than `instantiate_variables` ??

acl-cqc:
closes #633

* Add test case. Reduced this to a 2-node DFG (plus Input and Output). That this hasn't been showing up repeatedly in other tests - I put down to #388: most custom ops have empty `extension_reqs` so do not generate Plus constraints.
* Run (50%-failing) test 10 times -> failure rate ~~ 1 in 2^10
* Fix calculation of `live_vars` and `live_metas` by one-off traversal of constraints+solutions in `fn results()`.


src/extension/infer.rs
```diff
@@ -546,57 +546,51 @@ impl UnificationContext {
     pub fn results(&self) -> Result<ExtensionSolution, InferExtensionError> {
         // Check that all of the metavariables associated with nodes of the
         // graph are solved
+        let dependees = {
+            let mut h: HashMap<Meta, Vec<Meta>> = HashMap::new();
+            for (m, m2) in self.constraints.iter().flat_map(|(m, cs)| {
+                cs.iter().flat_map(|c| match c {
```
acl-cqc, line 552: Hmmm, this one could be a filter_map; or we could do `constraints.iter().flat_map(|(m,cs)| cs.map(|c|(m,c))).filter_map(|(m,c) match c.....)`. "Golf", to some extent

src/extension/infer.rs
```diff
@@ -546,57 +546,50 @@ impl UnificationContext {
...
                 if loc.1 == Direction::Incoming {
                     results.insert(loc.0, rs.clone());
                 }
-            } else if self.live_var(meta).is_some() {
+            } else if !depends_on_var.contains(&self.resolve(*meta)) {
                 // If it depends on some other live meta, that's bad news.
                 return Err(InferExtensionError::Unsolved { location: *loc });
             }
-            // If it only depends on graph variables, then we don't have
+            // If it ("only"??) depends on graph variables, then we don't have
```
acl-cqc, line 585: This only is what bothers me; the above calculation includes things that depend on both graph variables and other things.

See branch `fix/infer_stack_overflow2` for a simpler check that things depend upon solved values are solved; but (step through the commits) this is basically pointless.

See branch `fix/infer_stack_overflow3` for a more-complex check. Upon reflection I think maybe that is what's wanted?

croyzor: Yes good point, the logic wasn't checking the "only" part before

acl-cqc: Any thoughts as to which of these is "right" ?

croyzor: I think the "only" can be removed. All the comment is saying is that we can't find a solution as long as this meta depends on a variable.

acl-cqc: So if we have a Meta with two constraints: Plus({Ext}, Meta1) and Plus({Ext2}, V) where Meta1 has a concrete solution and V is the graph variable...we should avoid solving it? Is that because solving it now mean we lose the ability to check the graph variable provides the same solution?

acl-cqc: If so, then yes, I'll remove the "only", and to be honest, the assert below seems pretty pointless now too (it exactly repeats the two if-conditions inside this loop over the same `self.extensions`).

I was thinking maybe the "only" was genuine, and if we can compute a solution without the variables, then we should - in which case branch `tests/infer_stack_overflow3`

croyzor: > So if we have a Meta with two constraints: Plus({Ext}, Meta1) and Plus({Ext2}, V) where Meta1 has a concrete solution and V is the graph variable...we should avoid solving it?

I think in that case we would've found the solution (Ext + get_solution(Meta1)) in `main_loop`, and would skip this step on that meta because a solution exists for it

src/extension/infer.rs
```diff
@@ -546,57 +546,50 @@ impl UnificationContext {
...
+
         let mut results: ExtensionSolution = HashMap::new();
         for (loc, meta) in self.extensions.iter() {
             if let Some(rs) = self.get_solution(meta) {
                 if loc.1 == Direction::Incoming {
                     results.insert(loc.0, rs.clone());
                 }
-            } else if self.live_var(meta).is_some() {
+            } else if !depends_on_var.contains(&self.resolve(*meta)) {
                 // If it depends on some other live meta, that's bad news.
```
croyzor, line 582: This comment needs updating, since the one explaining what "live var" means is gone!

croyzor (approved):

src/extension/infer.rs
acl-cqc: Ah, good point, yes thanks. Ok, will update comment + assert

PR #642, opened 2023-11-06
refactor: remove add_op_<posn> by generalizing add_node_<posn> with "impl Into"

acl-cqc:
* Add a mildly-funky `Into` conversion from OpType to NodeType using the `new_auto` added in #635
* Generalize add_node_(with_parent,etc.) methods to take `impl Into<NodeType>`, and then remove corresponding `add_op_....`
* `add_op_after` had no corresponding `add_node_after`, so just turn it into that
* leave Hugr::add_node in simple form, without the `impl Into"

BREAKING CHANGE: add_op_{with_parent,before,after} removed => use add_node_....


src/hugr.rs
```diff
@@ -150,13 +150,9 @@ impl NodeType {
...
 }
 
-impl OpType {
-    /// Convert an OpType to a NodeType by giving it some input extensions
-    pub fn with_extensions(self, rs: ExtensionSet) -> NodeType {
-        NodeType {
-            op: self,
-            input_extensions: Some(rs),
-        }
+impl<T: Into<OpType>> From<T> for NodeType {
```
aborgna-q, line 153: Funky.

This may restrict some other impls for things that don't want the transitiveness. But it should be alright as long as `new_auto` is the sane default.

aborgna-q (approved):
LGTM

PR #643, opened 2023-11-06
test!: Improve coverage in signature and validate

ss2165:
Builder couldn't handle non-local  edges between cfg blocks so fix that and add tests.

BREAKING CHANGES: FunctionType::linear no longer exists


ss2165:
Highly recommend reviewing commit by commit!

PR #647, opened 2023-11-07
feat: port.as_directed to match on either direction

aborgna-q:
This replaces boilerplate like
```rust
match port.direction() {
    Direction::Incoming => {
        let incoming = port.as_incoming().unwrap();
        ...
    }
    Direction::Outgoing => {
        let outgoing = port.as_outgoing().unwrap();
        ...
    }
}
```
with
```rust
match port.as_directed() {
    Left(incoming) => ...
    Right(outgoing) => ...
}
```

drive-by: Drop leftover `Port::try_new_outgoing`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,6 @@ petgraph = { version = "0.6.3", default-features = false }
 context-iterators = "0.2.0"
 serde_json = "1.0.97"
 delegate = "0.10.0"
-either = "1.9.0"
 
 [features]
 pyo3 = ["dep:pyo3"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ petgraph = { version = "0.6.3", default-features = false }
 context-iterators = "0.2.0"
 serde_json = "1.0.97"
 delegate = "0.10.0"
+either = "1.9.0"
 
 [features]
 pyo3 = ["dep:pyo3"]
--- a/src/core.rs
+++ b/src/core.rs
@@ -4,6 +4,7 @@
 
 use derive_more::From;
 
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -50,6 +50,7 @@ petgraph = { version = "0.6.3", default-features = false }
 context-iterators = "0.2.0"
 serde_json = "1.0.97"
 delegate = "0.10.0"
+either = "1.9.0"
```
acl-cqc, line 53: There is an `Either` in `itertools` already, can you use that?

src/core.rs
```diff
@@ -92,35 +93,32 @@ impl Port {
...
     #[inline]
     pub fn as_outgoing(&self) -> Result<OutgoingPort, HugrError> {
+        self.as_directed()
+            .right()
+            .ok_or(HugrError::InvalidPortDirection(self.direction()))
+    }
+
+    /// Converts to either an [IncomingPort] or an [OutgoingPort], as appropriate.
+    #[inline]
+    pub fn as_directed(&self) -> Either<IncomingPort, OutgoingPort> {
```
acl-cqc, line 112: Possibly `in_or_out`? But happy with `as_directed` too

acl-cqc (approved):
Neat! :-). Like it. Think we don't need the import tho

src/core.rs
aborgna-q: I'd say `in_or_out` would be more similar to `direction`, where `as_*` is a cast.

Cargo.toml
aborgna-q: Done, and re-exported it so we don't require `itertools` on the dependants.

PR #651, opened 2023-11-07
feat: impl GraphRef for PetgraphWrapper

doug-q:
Since PetgraphWrapper is a reference to a HugrView, we implement GraphRef directly. This simplifies the implementation of several traits.


doug-q:
I don't quite understand why I need to impl Clone + Copy manually. Something complication with generics I suppose.

aborgna-q (approved):
Looks good.

Deriving `Copy`/`Clone` requires the traits from all the type parameters transitively, so it was restricting `T: Copy + Clone` where `GraphRef: Copy` is unconditional.

The reason behind that limitation is due to API stability; you don't want changes to the internal definition to implicitly affect the public API.

PR #660, opened 2023-11-08
tests: strength "failing_sccs_test", rename to "sccs" as it's not failing!

acl-cqc:
* Include situation with two loops
* Using four extensions, finally obtained a version where the iteration order of the `sccs()` mattered. Turns out its correct now (but reversing it breaks the test - many simpler tests worked in both orders).
* Driveby switch a lambda `|a,b| a.union(b)` to just `ExtensionSet::union`

### DIFF
--- a/src/extension/infer.rs
+++ b/src/extension/infer.rs
@@ -687,7 +687,7 @@ impl UnificationContext {
                         });
 
                 let (rs, other_ms): (Vec<_>, Vec<_>) = plus_constraints.unzip();
-                let solution = rs.iter().fold(ExtensionSet::new(), |e1, e2| e1.union(e2));
+                let solution = rs.iter().fold(ExtensionSet::new(), ExtensionSet::union);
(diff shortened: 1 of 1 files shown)


src/extension/infer.rs
```diff
@@ -705,9 +705,8 @@ impl UnificationContext {
         println!("{:?}", relations.node_map);
         println!("{:?}", relations.graph);
 
-        // Process the strongly-connected components. We need to deal with these
-        // depended-upon before depender. ccs() gives them back in some order
-        // - this might need to be reversed????
+        // Process the strongly-connected components. petgraph/sccs() returns these
+        // depended-upon before depender, as we need.
```
croyzor, line 709: depender -> dependant

croyzor (approved):

src/extension/infer.rs
acl-cqc: Ah of course yes, thank you!

PR #661, opened 2023-11-08
feat: Require the node's metadata to always be a Map

### LINKED ISSUE #652: feat: Require the node metadata to always be a map
It is used as such everywhere (specially on tket2).
Restricting each node's metadata to `Option<serde_json::Map<String, serde_json::Value>>` would allow us to rework the metadata API to directly access the values in the map.

The HUGR spec doesn't currently specify the metadata structure.

aborgna-q:
Closes #652

### DIFF
--- a/src/hugr/hugrmut.rs
+++ b/src/hugr/hugrmut.rs
@@ -14,17 +14,52 @@ use crate::{Hugr, IncomingPort, OutgoingPort, Port, PortIndex};
 
 use self::sealed::HugrMutInternals;
 
+use super::NodeMetadataMap;
+
 /// Functions for low-level building of a HUGR.
 pub trait HugrMut: HugrMutInternals {
-    /// Returns the metadata associated with a node.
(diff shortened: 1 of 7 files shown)


src/hugr/hugrmut.rs
```diff
@@ -14,17 +14,52 @@ use crate::{Hugr, IncomingPort, OutgoingPort, Port, PortIndex};
...
+    }
+
+    /// Sets a metadata value associated with a node.
+    fn set_metadata(
+        &mut self,
+        node: Node,
+        key: impl AsRef<str>,
+        metadata: impl Into<NodeMetadata>,
+    ) -> Result<(), HugrError> {
+        *self.get_metadata_mut(node, key)? = metadata.into();
```
croyzor, line 45: There's a lot going on on the LHS of an assignment here 👀

croyzor (approved):

PR #669, opened 2023-11-09
feat: make prelude error type public

ss2165:
to allow angle extension to be moved out


cqc-alec (approved):
Should this be `feat` rather than `refactor`?

Issue #673, opened 2023-11-10
FuncDefn uses wrong extension_reqs delta in inference


acl-cqc:
1. FuncDefn stores a FunctionType (inputs, outputs, extension_reqs) as the type of the function declared. This FunctionType is what you get from the static_output of the FuncDefn node.

2. The FuncDefn node also has its own `signature: FunctionType` for its dataflow ports. Since it's not a dataflow node, this is:
https://github.com/CQCL/hugr/blob/b256c2bc42e8499dac89c6f8bff7a0d478a1ff59/src/ops.rs#L189-L191

i.e. no ports, and empty-set extension reqs.

3. In extension inference, we make the input/output extensions of the FuncDefn node equal to the extensions of the Input/Output nodes beneath it (FuncDefn is a DataflowParent just like BasicBlock, Dfg, etc.):
https://github.com/CQCL/hugr/blob/b256c2bc42e8499dac89c6f8bff7a0d478a1ff59/src/extension/infer.rs#L277-L284

4. Also in inference, we use the FunctionType from step 2 (not step 1) to constrain the input/output extensions of the FuncDefn node:
https://github.com/CQCL/hugr/blob/b256c2bc42e8499dac89c6f8bff7a0d478a1ff59/src/extension/infer.rs#L305-L315

Putting 2,3,4 together we have that the body of the function must have a zero extension delta (Input and Output nodes of a FuncDefn are constrained to have equal extension requirements), ooops

acl-cqc:
To be honest I find it quite confusing that you can even get a value for `input_extensions` or delta/`extension_reqs` for a FuncDefn *node*....

PR #674, opened 2023-11-10
Alternative start to automatically including extension-reqs

### LINKED ISSUE #388: Custom operations implicitly require their extension

acl-cqc:
Intended to fix #388, alternative approach to #389.

However, the tiny fix for #388 breaks many tests. Some (4) are updated+fixed here, but many more still failing. At least some of these are due to #673 which will need fixing first


acl-cqc:
Actually, I think even the fixes I've applied should be changing delta, not input_extensions, and then would also fall foul of #673. So closing.

PR #677, opened 2023-11-10
chore: devenv support for local coverage

ss2165:


devenv.lock
```diff
@@ -110,6 +110,22 @@
...
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-23.05",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "nixpkgs-stable_2": {
```
aborgna-q, line 128: Why are there multiple stables?

aborgna-q (approved):
LGTM

devenv.lock
ss2165: i have no idea

Issue #676, opened 2023-11-10
Constructing Extensions with ExtensionRegistries is difficult


acl-cqc:
Extension::add_op_.... takes an ExtensionRegistry which it uses to validate the type params, type scheme etc. of the new op. But if the op refers to types defined in the same extension, this is very tricky!!

One *possible* but far-from-ideal situation is to break every extension into two - the types and the ops - then you can build an ExtensionRegistry from the types before starting on the ops.

Another possibility would be for `add_op_xxx` to include the extension (to which the op is added) together in the ExtensionRegistry when validating the type scheme. That allows an Extension's ops to refer to types in the same Extension, but is still difficult if you are making two Extensions each of which has ops referring to the others types, say.

Instead we propose a solution where adding an op, etc., does not perform validation or require an ExtensionRegistry (?). Instead validation of the whole ExtensionRegistry is performed in a single step (when building the ExtensionRegistry out of a Vec<Extension>, say). Feedback is a little slower but this recovers most flexibility.

PR #678, opened 2023-11-10
refactor: use type schemes in extension definitions wherever possible

### LINKED ISSUE #658: Convert binary compute_signatures to type schemes
After #630 is merged, many existing Rust/binary implementations of `compute_signature` can be converted to a type scheme using a PolyFuncType

ss2165:
Closes #658 

As a drive by improve extensions test coverage as I went.

Definition can be a bit unintuitive, should improve after #676 is done.

### DIFF
--- a/src/extension/op_def.rs
+++ b/src/extension/op_def.rs
@@ -331,6 +331,23 @@ impl Extension {
             SignatureFunc::TypeScheme(type_scheme),
         )
     }
+
+    /// Create an OpDef with a signature (inputs+outputs) read from e.g.
+    /// declarative YAML; and no "misc" or "lowering functions" defined.
+    pub fn add_op_type_scheme_simple(
+        &mut self,
+        name: SmolStr,
+        description: String,
+        type_scheme: PolyFuncType,
+    ) -> Result<&OpDef, ExtensionBuildError> {
+        self.add_op(
+            name,
+            description,
+            Default::default(),
+            vec![],
+            SignatureFunc::TypeScheme(type_scheme),
+        )
+    }
 }
 
 #[cfg(test)]
--- a/src/std_extensions/arithmetic/conversions.rs
+++ b/src/std_extensions/arithmetic/conversions.rs
@@ -6,30 +6,27 @@ use crate::{
         PRELUDE,
     },
     type_row,
-    types::{FunctionType, PolyFuncType, Type},
+    types::{FunctionType, PolyFuncType},
     Extension,
 };
 
-use super::int_types::{int_type_var, INT_TYPE_ID};
+use super::int_types::int_type_var;
 use super::{float_types::FLOAT64_TYPE, int_types::LOG_WIDTH_TYPE_PARAM};
 
 /// The extension identifier.
 pub const EXTENSION_ID: ExtensionId = ExtensionId::new_unchecked("arithmetic.conversions");
 
-fn ftoi_sig(
-    int_type_var: Type,
-    temp_reg: &ExtensionRegistry,
-) -> Result<PolyFuncType, SignatureError> {
-    let body = FunctionType::new(type_row![FLOAT64_TYPE], vec![sum_with_error(int_type_var)]);
+fn ftoi_sig(temp_reg: &ExtensionRegistry) -> Result<PolyFuncType, SignatureError> {
+    let body = FunctionType::new(
+        type_row![FLOAT64_TYPE],
+        vec![sum_with_error(int_type_var(0))],
+    );
 
     PolyFuncType::new_validated(vec![LOG_WIDTH_TYPE_PARAM], body, temp_reg)
 }
 
-fn itof_sig(
-    int_type_var: Type,
-    temp_reg: &ExtensionRegistry,
-) -> Result<PolyFuncType, SignatureError> {
-    let body = FunctionType::new(vec![int_type_var], type_row![FLOAT64_TYPE]);
+fn itof_sig(temp_reg: &ExtensionRegistry) -> Result<PolyFuncType, SignatureError> {
+    let body = FunctionType::new(vec![int_type_var(0)], type_row![FLOAT64_TYPE]);
 
     PolyFuncType::new_validated(vec![LOG_WIDTH_TYPE_PARAM], body, temp_reg)
 }
@@ -43,12 +40,8 @@ pub fn extension() -> Extension {
             super::float_types::EXTENSION_ID,
         ]),
     );
-    let int_types_extension = super::int_types::extension();
-    let int_type_def = int_types_extension.get_type(&INT_TYPE_ID).unwrap();
-    let int_type_var = int_type_var(0, int_type_def).unwrap();
     let temp_reg: ExtensionRegistry = [
-        extension.clone(),
-        int_types_extension,
+        super::int_types::EXTENSION.to_owned(),
         super::float_types::extension(),
         PRELUDE.to_owned(),
     ]
@@ -57,28 +50,28 @@ pub fn extension() -> Extension {
         .add_op_type_scheme_simple(
             "trunc_u".into(),
             "float to unsigned int".to_owned(),
-            ftoi_sig(int_type_var.clone(), &temp_reg).unwrap(),
+            ftoi_sig(&temp_reg).unwrap(),
         )
         .unwrap();
     extension
         .add_op_type_scheme_simple(
             "trunc_s".into(),
             "float to signed int".to_owned(),
-            ftoi_sig(int_type_var.clone(), &temp_reg).unwrap(),
+            ftoi_sig(&temp_reg).unwrap(),
         )
         .unwrap();
     extension
         .add_op_type_scheme_simple(
             "convert_u".into(),
             "unsigned int to float".to_owned(),
-            itof_sig(int_type_var.clone(), &temp_reg).unwrap(),
+            itof_sig(&temp_reg).unwrap(),
         )
         .unwrap();
     extension
         .add_op_type_scheme_simple(
             "convert_s".into(),
             "signed int to float".to_owned(),
-            itof_sig(int_type_var, &temp_reg).unwrap(),
+            itof_sig(&temp_reg).unwrap(),
         )
         .unwrap();
 
--- a/src/std_extensions/arithmetic/conversions.rs
+++ b/src/std_extensions/arithmetic/conversions.rs
@@ -1,7 +1,10 @@
 //! Conversions between integer and floating-point values.
 
 use crate::{
-    extension::{ExtensionId, ExtensionRegistry, ExtensionSet, SignatureError, PRELUDE},
+    extension::{
+        prelude::sum_with_error, ExtensionId, ExtensionRegistry, ExtensionSet, SignatureError,
+        PRELUDE,
+    },
     type_row,
     types::{FunctionType, PolyFuncType, Type},
     Extension,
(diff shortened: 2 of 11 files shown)


src/std_extensions/arithmetic/int_ops.rs
```diff
@@ -40,100 +41,145 @@ fn inarrow_sig(arg_values: &[TypeArg]) -> Result<FunctionType, SignatureError> {
...
+        vec![int_type_var.clone()],
+        temp_reg,
+    )
 }
 
-fn idivmod_checked_sig(arg_values: &[TypeArg]) -> Result<FunctionType, SignatureError> {
-    let [arg0, arg1] = collect_array(arg_values);
-    let intpair: TypeRow = vec![int_type(arg0.clone()), int_type(arg1.clone())].into();
-    Ok(FunctionType::new(
+fn iunop_sig(
```
acl-cqc, line 90: Hmmm, I thought we had renamed these to something more intelligible, like `int_unary_op`. I mean, it makes sense now I've seen it a few times before...

src/std_extensions/arithmetic/int_ops.rs
```diff
@@ -40,100 +41,145 @@ fn inarrow_sig(arg_values: &[TypeArg]) -> Result<FunctionType, SignatureError> {
...
 }
 
-fn idiv_checked_sig(arg_values: &[TypeArg]) -> Result<FunctionType, SignatureError> {
-    let [arg0, arg1] = collect_array(arg_values);
-    Ok(FunctionType::new(
-        vec![int_type(arg0.clone()), int_type(arg1.clone())],
-        vec![Type::new_sum(vec![int_type(arg0.clone()), ERROR_TYPE])],
-    ))
+fn idiv_checked_sig(
+    int_type_var_0: Type,
```
acl-cqc, line 121: Yeah, presumably the only legal values for these two are `int_type_var(0, ...)` and `int_type_var(1, ...)`

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -198,6 +198,12 @@ pub fn extension() -> Extension {
     extension
 }
 
+/// get an integer type variable, given the integer type definition
+pub(super) fn int_type_var(var_id: usize, int_type_def: &TypeDef) -> Result<Type, SignatureError> {
```
acl-cqc, line 202: You could make this find the `int_type_def`, right? (So we don't have to pass the values returned by the function around everywhere? I've commented elsewhere but it does seem a bit error-prone)

src/std_extensions/arithmetic/conversions.rs
```diff
@@ -1,36 +1,40 @@
...
-    Ok(FunctionType::new(
+fn ftoi_sig(
+    int_type_var: Type,
+    temp_reg: &ExtensionRegistry,
+) -> Result<PolyFuncType, SignatureError> {
+    let body = FunctionType::new(
         type_row![FLOAT64_TYPE],
         vec![Type::new_sum(vec![
-            int_type(arg.clone()),
+            int_type_var,
```
acl-cqc, line 23: Consider a helper function `sum_with_error(Type) -> Type` (in some ancestor module I suspect)

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -244,6 +251,40 @@ mod test {
             ConstIntS::new(3, 128),
             Err(ConstTypeError::CustomCheckFail(_))
         );
-        assert_matches!(ConstIntS::new(3, -128), Ok(_));
+        assert!(ConstIntS::new(3, -128).is_ok());
```
acl-cqc, line 254: If the old version doesn't compile, fair, but if the assert fails, this new form will lose the reason why

src/std_extensions/arithmetic/int_types.rs
```diff
@@ -244,6 +251,40 @@ mod test {
...
+            .is_ok());
+        assert!(const_i32_2
+            .check_custom_type(&int_custom_type(TypeArg::BoundedNat { n: 6 }))
+            .is_err());
+        assert!(const_i32_2
+            .check_custom_type(&int_custom_type(TypeArg::BoundedNat { n: 19 }))
+            .is_err());
+        assert_eq!(const_i32_2.name(), "i5(-2)");
+
+        ConstIntS::new(50, -2).unwrap_err();
```
acl-cqc, line 287: You could combine these two into a single test `new(50, 0)` - 0 is legal for *any* integer type...

src/std_extensions/collections.rs
```diff
@@ -59,50 +59,60 @@ impl CustomConst for ListValue {
...
+                FunctionType::new(vec![l.clone(), e], vec![l]),
+                &temp_reg,
+            )
+            .unwrap(),
         )
         .unwrap();
     extension
 }
+
+fn list_types(list_type_def: &TypeDef) -> (Type, Type) {
```
acl-cqc, line 106: Did you mean to move this?

`list_and_elem_type` (or `list_type_and_elem`) might be a better name

src/std_extensions/logic.rs
```diff
@@ -48,19 +47,12 @@ fn extension() -> Extension {
...
-                let n: u64 = match a {
-                    TypeArg::BoundedNat { n } => *n,
-                    _ => {
-                        return Err(TypeArgError::TypeMismatch {
-                            arg: a.clone(),
-                            param: H_INT,
-                        }
-                        .into());
-                    }
+                let TypeArg::BoundedNat { n } = arg_values.iter().exactly_one().unwrap() else {
```
acl-cqc, line 50:
```suggestion
                let Ok(TypeArg::BoundedNat { n }) = arg_values.iter().exactly_one() else {
```

acl-cqc: Good use of `let...else` tho!

src/std_extensions/logic.rs
```diff
@@ -73,19 +65,12 @@ fn extension() -> Extension {
...
```
acl-cqc, line 68: ditto

acl-cqc (commented):
Looks great! Thanks @ss2165. I can't quite bring myself to hit "approve" with 21 comments but certainly in principle ;)

src/std_extensions/arithmetic/float_ops.rs
```diff
@@ -41,77 +36,60 @@ pub fn extension() -> Extension {
...
-        .add_op_custom_sig_simple("feq".into(), "equality test".to_owned(), vec![], fcmp_sig)
+        .add_op_type_scheme_simple("feq".into(), "equality test".to_owned(), fcmp_sig())
         .unwrap();
     extension
-        .add_op_custom_sig_simple("fne".into(), "inequality test".to_owned(), vec![], fcmp_sig)
+        .add_op_type_scheme_simple("fne".into(), "inequality test".to_owned(), fcmp_sig())
         .unwrap();
     extension
-        .add_op_custom_sig_simple("flt".into(), "\"less than\"".to_owned(), vec![], fcmp_sig)
+        .add_op_type_scheme_simple("flt".into(), "\"less than\"".to_owned(), fcmp_sig())
```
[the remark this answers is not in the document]
ss2165: @cqc-alec ?

src/std_extensions/arithmetic/float_types.rs
```diff
@@ -100,6 +100,11 @@ mod test {
     fn test_float_consts() {
         let const_f64_1 = ConstF64::new(1.0);
         let const_f64_2 = ConstF64::new(2.0);
+
+        assert_eq!(const_f64_1.value(), 1.0);
```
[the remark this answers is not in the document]
ss2165: no, the commits are separate so i could cherry pick them, but its "related" in the sense that the changes in this PR highlight the remaining missing coverage better

src/std_extensions/arithmetic/int_types.rs
ss2165: it does compile, it just adds an uncovered line 😅

src/extension/op_def.rs
```diff
@@ -331,6 +331,23 @@ impl Extension {
             SignatureFunc::TypeScheme(type_scheme),
         )
     }
+
+    /// Create an OpDef with a signature (inputs+outputs) read from e.g.
+    /// declarative YAML; and no "misc" or "lowering functions" defined.
+    pub fn add_op_type_scheme_simple(
+        &mut self,
+        name: SmolStr,
+        description: String,
```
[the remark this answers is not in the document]
ss2165: All the other methods take String so I think I'd rather leave that for a
separate issue

src/std_extensions/arithmetic/float_ops.rs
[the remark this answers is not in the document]
cqc-alec: The name of the operation is "less than" in quotes because the convention for operation names is that they should be nouns (addition, maximum etc), and "less than" without quotes would be inconsistent with that -- it would have to be "less-than comparator" or somthing, which is more verbose. But it is rather awkward I agree; we could remove them if you like.

src/std_extensions/arithmetic/conversions.rs
```diff
@@ -42,37 +46,42 @@ pub fn extension() -> Extension {
             super::float_types::EXTENSION_ID,
         ]),
     );
-
+    let int_types_extension = super::int_types::extension();
+    let int_type_def = int_types_extension.get_type(&INT_TYPE_ID).unwrap();
+    let int_type_var = int_type_var(0, int_type_def).unwrap();
+    let temp_reg: ExtensionRegistry = [
+        extension.clone(),
+        int_types_extension,
```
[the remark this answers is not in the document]
ss2165: because int_types_extension is used multiple times

src/std_extensions/arithmetic/int_types.rs
ss2165: not sure what you mean here - I'd still need two lines for the unsigned and signed

src/std_extensions/arithmetic/conversions.rs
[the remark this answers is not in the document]
ss2165: (or at least, it was)

src/std_extensions/arithmetic/int_types.rs
acl-cqc: Oh of course, right. Remembered there was some issue with `assert_matches` but not what. Consider `unwrap()`...

src/std_extensions/arithmetic/int_types.rs
acl-cqc: I had to dive into `fn new` to understand what this was about but after that I thought the point was that it was gonna ignore the second argument

ss2165: it's testing that both constants error on invalid widths (coverage)

src/std_extensions/arithmetic/conversions.rs
```diff
@@ -1,36 +1,40 @@
...
         type_row![FLOAT64_TYPE],
         vec![Type::new_sum(vec![
-            int_type(arg.clone()),
+            int_type_var,
             crate::extension::prelude::ERROR_TYPE,
         ])],
-    ))
+    );
+
+    PolyFuncType::new_validated(vec![LOG_WIDTH_TYPE_PARAM], body, temp_reg)
```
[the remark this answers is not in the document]
ss2165: yes ver very good catch. I think I was getting tangled being worried about circular
definition dependance but fortunately the int_types extension is standalone so
we can call it freely in conversions and int_ops

src/std_extensions/arithmetic/int_types.rs
acl-cqc: Yes, so they both cover exactly the same code, as the "is width valid" check is done before even looking at the second argument??

src/std_extensions/arithmetic/float_ops.rs
```diff
@@ -40,78 +35,76 @@ pub fn extension() -> Extension {
         ExtensionSet::singleton(&super::float_types::EXTENSION_ID),
     );
 
+    let fcmp_sig = fcmp_sig();
```
acl-cqc, line 38: You could just inline the functions here, these are the only callsites

src/std_extensions/arithmetic/int_types.rs
ss2165: One covers 90-94, the other covers 120-124 ?

acl-cqc: Ok, my bad. Somehow I'd seen +2/-2 but not S/U, sorry!

acl-cqc (approved):
Looks great, thanks!

PR #680, opened 2023-11-11
feat!: HugrView API improvements

### LINKED ISSUE #495: Clean up OpType methods
- [ ] Add a method returning the `Option<Port>` for the static input (`static_input_port`?)
- [ ] `OpType::other_port` returns the EdgeKind, and `other_port_index` returns the `Port`.
    Rename them to `other_port_kind` and `other_port` to avoid confusion.
- [ ] Add more `*_input`/`*_output` helpers to the port getters to avoid writing the `Direction`.

### LINKED ISSUE #499: Shorthand HugrView method for "connected port"
Lots of places where we write things like:

```rust
  let (node, port) = h
      .linked_ports(node, port)
      .exactly_one()
      .ok()
      .unwrap();

```

### LINKED ISSUE #506: Edge kind specific HugrView methods
Building on from #499, given we have a finite number of edge kinds and restrictions on their use, we should have specific methods for querying those connections. I propose:

```rust
    // just value ports
    fn value_ports(&self, node: Node, direction: Direction) -> Self::PortLinks<'_>;
    fn value_neighbours(&self, node: Node, direction: Direction) -> Self::Neighbours<'_>;
    // companion to linked_ports, a linear value port must have one target
    fn linked_liner_port(&self, node: Node, port: Port) -> Option<(Node, Port)>;

    // all Value + Order neighbours
    fn dataflow_neighbours(&self, node: Node, direction: Direction) -> Self::Neighbours<'_>;
    
    fn order_neighbours(&self, node: Node, direction: Direction) -> impl Iterator<Item=Node>;

    // can only be one static input
    fn static_source(&self, node: Node) -> Option<Node>;

    // can be many static targets from one port
    fn static_targets(&self, node: Node) -> impl Iterator<Node>;

    // same as static but for cfg nodes
    fn previous_block(&self, node: Node) -> Option<Node>;
    fn next_blocks(&self, node: Node) -> impl Iterator<Node>;
```

with these we should be able to minimise/nullify the use of `OpType` methods like `other_edge_kind` `static_port` `other_edge_port` (so this is also an alternative to #495)

### LINKED ISSUE #521: Refactor: move `non_df_count` out from OpValidityFlags
....as it's not used in validation, and *is* used in public methods of OpType: `non_df_count` and `get_port_index`

Could then follow up by renaming `OpValidityFlags` to something like `OpChildFlags` as all remaining flags would relate to children.

### LINKED ISSUE #653: feat: HugrView Value port enhancements
Methods for:
* iterate over only Value edges (from @doug-q): 
> I believe that it's not guaranteed that ports are iterated over in order, and so to do this in a bulletproof way one should get all ports, sort by port index, and fail if there are not enough ports to match the signature.
* get the nth Value ports
* get type of value port `HugrView::get_value_type(Node, Port) -> Option<Type>`

### LINKED ISSUE #655: feat: HugrView method for iterating over neighbouring ports
```rust 
HugrView::neighbour_ports(&self, node: Node, direction: Direction) -> impl Iterator<Item = (Node, Port)> + '_;
```

ss2165:
- [x] Closes #495
- [x] Closes #521
- [x] Closes #655
- [x] Closes #499
- [x] Closes #506 
- [x] Closes #653
- [x] `OpType::signature` returns option (non-dataflow ops don't return signature)
- [x] Implement `try_into` from OpType references in to inner references.

Doesn't necessarily do exactly as those issues specify - instead considers them holistically for a more unified interface. Easiest to review commit by commit.

Uses rust_version crate to use return position impl for the new `HugrView` methods that return iterators. This will be stable with 1.75 (which enters beta in a few days).

BREAKING_CHANGES: `OpType` and `FunctionType` methods renamed for clarity; `OpType::signature` returns `Option<FuncType>`.


src/hugr/views.rs
```diff
@@ -505,6 +590,19 @@ impl<T: AsRef<Hugr>> HugrView for T {
     }
 }
 
+/// Filter an iterator of node-ports to only dataflow dependency specifying
+/// ports (Value and StateOrder)
+pub fn dataflow_ports_only<'i, 'a: 'i, P: Into<Port> + Copy>(
```
aborgna-q, line 595: It would be a bit more practical to allow chained applications;

```rust
pub trait PortIterator {
    fn dataflow_ports_only(self, hugr) -> impl Iterator<...> { ... };
}
impl<I, P> PortIterator for I where I: Iterator<Item = (Node, P)> {}
```

But either way works alright.

src/hugr.rs
```diff
@@ -111,13 +111,15 @@ impl NodeType {
 
     /// Use the input extensions to calculate the concrete signature of the node
     pub fn signature(&self) -> Option<Signature> {
-        self.input_extensions
-            .as_ref()
-            .map(|rs| self.op.signature().with_input_extensions(rs.clone()))
+        self.input_extensions.as_ref().and_then(|rs| {
```
acl-cqc, line 114: I kept pondering whether there's a "right" order to check these in but failed to come up with one. How about
`if let (Some(rs), Some(sig)) = (self.input_extensions, self.op.signature()) {sig.with_input_extensions(rs.clone())} else {None}`

ss2165: I'm going to change this to keep this as previous behaviour (with an unwrap_or_default), I think changes to the inference logic should be a follow on

src/hugr/rewrite/replace.rs
```diff
@@ -616,7 +616,7 @@ mod test {
         entry: bool,
     ) -> Result<BasicBlockID, BuildError> {
         let op: OpType = op.into();
-        let op_sig = op.signature();
+        let op_sig = op.signature().expect("dataflow op needs signature");
```
acl-cqc, line 619: Hmmmm, we can't do something like `impl Into<DataflowOpTrait>` can we? Then we'd know we have a signature. I guess nothing to worry about in a test but generally it feels like it'd be good to try to know statically when we have an Op(Type) with a signature

src/hugr/rewrite/simple_replace.rs
```diff
@@ -505,15 +496,27 @@ pub(in crate::hugr::rewrite) mod test {
             .collect_vec();
         let inputs = h
             .node_outputs(input)
-            .filter(|&p| h.get_optype(input).signature().get(p).is_some())
+            .filter(|&p| {
+                h.get_optype(input)
+                    .signature()
+                    .unwrap()
+                    .port_type(p)
+                    .is_some()
```
aborgna-q, line 504: Perhaps an `OpType::df_port_type` helper would be useful?

Alternatively,
```suggestion
                h.signature(input).is_some_and(|s| s.port_type(p).is_some())
```
(there's many other cases of `hugr.get_optype(n).signature()`).

aborgna-q (commented):
Awesome!
A much needed API update.

src/ops.rs
```diff
@@ -214,6 +260,14 @@ pub trait OpTrait {
     fn other_output(&self) -> Option<EdgeKind> {
         None
     }
+
+    /// Get the number of non-dataflow multiports.
+    ///
+    /// If None, the operation must have exactly one non-dataflow port
```
acl-cqc, line 266: Erm.....aaiieeee....

If the operation has other_edges, "None" means "exactly one non-dataflow port". Is that different from `None` meaning `Some(1)` ?

And if the operation does not have other edges, "None" means "zero". Is that different from `None` meaning `Some(0)`?

i.e. can we turn this Option into a usize?

src/ops/controlflow.rs
```diff
@@ -169,12 +169,21 @@ impl OpTrait for BasicBlock {
         Some(EdgeKind::ControlFlow)
     }
 
-    fn signature(&self) -> FunctionType {
-        match self {
+    fn signature(&self) -> Option<FunctionType> {
```
acl-cqc, line 172: Hmmmm, so this is because we need to have a FunctionType in order to have a `Signature` carrying the input-extensions, right?

Is it time to decouple input-extensions from signature? (Another issue, not here.)

src/ops/dataflow.rs
```diff
@@ -115,8 +111,8 @@ impl<T: DataflowOpTrait> OpTrait for T {
     fn tag(&self) -> OpTag {
         T::TAG
     }
-    fn signature(&self) -> FunctionType {
-        DataflowOpTrait::signature(self)
+    fn signature(&self) -> Option<FunctionType> {
+        Some(DataflowOpTrait::dataflow_signature(self))
```
acl-cqc, line 115: This kinda makes me think the names should be the other way around. Anything defining `dataflow_signature` definitely has one, whereas the broader set of nodes defining `signature` might or might not...(but if they do, it's a dataflow signature, no?)

src/ops.rs
```diff
@@ -77,54 +77,100 @@ impl OpType {
...
         if port.index() < port_count {
-            signature.get(port).cloned().map(EdgeKind::Value)
-        } else if port.index() == port_count
-            && dir == Direction::Incoming
-            && self.static_input().is_some()
-        {
-            self.static_input().map(EdgeKind::Static)
+            signature.port_type(port).cloned().map(EdgeKind::Value)
+        } else if port_as_in.is_some() && port_as_in == self.static_input_port() {
+            Some(EdgeKind::Static(static_in_type(self)))
```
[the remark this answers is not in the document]
acl-cqc: (that said, I'm still a bit confused by the static_input/other_input stuff. I don't mean to say that this PR isn't making things better, though)

src/hugr/rewrite/replace.rs
ss2165: done, I've used this logic in ~3 places, there are probably more. (Could be a follow up task for builder to require `DataflowOpTrait` in more places rather than `Into<OpType>`?)

src/ops.rs
```diff
@@ -77,54 +77,100 @@ impl OpType {
...
-            ))
+            Some(Port::new(dir, self.value_port_count(dir) + static_input))
         } else {
             None
         }
     }
 
+    /// The number of Value ports in given direction.
+    pub fn value_port_count(&self, dir: portgraph::Direction) -> usize {
+        self.signature().map(|sig| sig.port_count(dir)).unwrap_or(0)
```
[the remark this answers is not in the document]
acl-cqc: true, but that's two empty TypeRow's and one empty ExtensionSet, all of which are `const` so allocation-free... (the closure involves an allocation to capture `dir`)

src/hugr/rewrite/simple_replace.rs
ss2165: I've taken a different approach - casting to known types and using
`dataflow_signature` rather than continuing to wrap `FunctionType` methods on to `OpType`

Cargo.toml
```diff
@@ -50,6 +50,8 @@ petgraph = { version = "0.6.3", default-features = false }
 context-iterators = "0.2.0"
 serde_json = "1.0.97"
 delegate = "0.10.0"
+rustversion = "1.0.14"
+paste = "1.0"
```
ss2165, line 54: I continue my brazen injecting of dtolnay crates

acl-cqc: Note this one is already in `dev-dependencies`

src/ops/controlflow.rs
ss2165: yes exactly, I think `Signature` should be a) renamed, and b) hold an `Option<FunctionType>` ?

src/hugr/views.rs
```diff
@@ -179,6 +181,45 @@ pub trait HugrView: sealed::HugrInternals {
     /// Iterator over the nodes and ports connected to a port.
     fn linked_ports(&self, node: Node, port: impl Into<Port>) -> Self::PortLinks<'_>;
 
+    #[rustversion::since(1.75)] // uses impl in return position
+    /// Iterator over all the nodes and ports connected to a node's inputs..
+    fn all_linked_outputs(&self, node: Node) -> impl Iterator<Item = (Node, OutgoingPort)> {
```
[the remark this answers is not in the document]
aborgna-q: Ah, I meant a parametric method.
```rust
fn all_linked_ports(&self, node: Node, dir: Direction) -> impl Iterator<Item = (Node, Port)> {
    match dir {
        Direction::Incoming => Either::Left(self.all_linked_inputs(node).map(|(n,p)| (n, p.into())),
        Direction::Outgoing => Either::Left(self.all_linked_outputs(node).map(|(n,p)| (n, p.into())),
    }
}
```

src/ops/controlflow.rs
acl-cqc: I actually think two separate members on NodeType (`input_extensions: Option<ExtensionSet>, signature: Option<FunctionType>`) might be better, but if keeping them together then yes

src/ops/dataflow.rs
acl-cqc: As in, I suggest `DataflowOpTrait::signature(&self) -> FunctionType` and `OpTrait::dataflow_signature(&self) -> Option<FunctionType>`

aborgna-q (approved):
I'm happy with merging this, and iterating on other open issues as new PRs.

PR #682, opened 2023-11-13
test: use insta snapshots to add dot_string coverage

ss2165:


src/hugr/views/tests.rs
```diff
@@ -38,7 +51,19 @@ fn node_connections() -> Result<(), BuildError> {
...
         ]
     );
     Ok(())
 }
+
+#[rstest]
+fn dot_string(sample_hugr: (Hugr, BuildHandle<DataflowOpID>, BuildHandle<DataflowOpID>)) {
+    let (h, _, _) = sample_hugr;
+
+    insta::assert_yaml_snapshot!(h.dot_string());
```
aborgna-q, line 68: This is not yaml, so probably should be `assert_snapshot`? Or is it needed for whitespace stuff?

aborgna-q (commented):
I'm not sure if we want to test exact matching of the dot format. It was always meant as debug information, so the output is not stable.

src/hugr/views/tests.rs
ss2165: it's used for easier debug, its the snap format

ss2165:
Is unstable output so bad? insta makes life easier there - you just review the change is one you expected and accept it.

PR #687, opened 2023-11-14
chore: Move extension inference tests to their own file

croyzor:
These files are still too big, but it's a start!


aborgna-q (approved):
LGTM

PR #688, opened 2023-11-14
fix: FuncDefns don't require that their extensions match their children

### LINKED ISSUE #673: FuncDefn uses wrong extension_reqs delta in inference

croyzor:
Resolves #673


src/extension/infer.rs
```diff
@@ -279,7 +279,13 @@ impl UnificationContext {
                     let m_input_node = self.make_or_get_meta(input, dir);
                     self.add_constraint(m_input_node, Constraint::Equal(m_input));
                     let m_output_node = self.make_or_get_meta(output, dir);
-                    self.add_constraint(m_output_node, Constraint::Equal(m_output));
+                    // If the parent node is a FuncDefn, it's input_resources
```
acl-cqc, line 282: input_resources -> input_extensions

src/extension/infer.rs
```diff
@@ -279,7 +279,13 @@ impl UnificationContext {
...
+                    // should always be empty, meaning if the function adds
+                    // resources, they shouldn't be equal to the parent's output
+                    // (which will also be empty.)
```
acl-cqc, line 285: Is that true? parent's output is equal to (its input, probably empty, maybe not necessarily) plus delta

src/extension/infer.rs
```diff
@@ -279,7 +279,13 @@ impl UnificationContext {
...
+                    if node_type.tag() != OpTag::FuncDefn {
```
acl-cqc, line 286: ok, that is neater than I realized, nice

src/extension/infer/test.rs
```diff
@@ -962,3 +967,73 @@ fn sccs() {
...
+    let [w] = lift.outputs_arr();
+    func_builder.finish_with_outputs([w])?;
+    builder.finish_hugr(&PRELUDE_REGISTRY)?;
+    Ok(())
+}
+
+#[test]
+// Define a function "g" inside "f", which defines a function which adds to
+// the graph. Tests that the resources of the graph of "g" aren't being
+// constrained to match those of the FuncDefn node.
```
acl-cqc, line 998: struggling to see how the test does what you say here - sorry! - can you put a comment or two inside the method?

src/hugr/validate.rs
```diff
@@ -200,11 +200,15 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
...
         self.validate_children(node, node_type)?;
 
-        // If this is a container with I/O nodes, check that the extension they
-        // define match the extensions of the container.
-        if let Some([input, output]) = self.hugr.get_io(node) {
-            self.extension_validator
-                .validate_io_extensions(node, input, output)?;
+        // FuncDefns have no resources since they're static nodes, but the
+        // functions they define can have any extension delta.
+        if node_type.tag() != OpTag::FuncDefn {
```
acl-cqc, line 205: 👍

src/hugr/validate/test.rs
```diff
@@ -750,3 +750,51 @@ fn invalid_types() {
         SignatureError::TypeArgMismatch(TypeArgError::WrongNumberArgs(2, 1))
     );
 }
+
+#[test]
+fn parent_io_mismatch() {
+    let mut hugr = Hugr::new(NodeType::new_pure(ops::DFG {
+        signature: FunctionType::new(type_row![USIZE_T], type_row![USIZE_T]),
```
acl-cqc, line 757: The problem here is that the FunctionType has zero delta? Whereas if you did `.with_extension_delta(...XB_T...)` it'd pass? (Might be worth doing both ways via a loop / helper fn just to show the difference between working and not working)

acl-cqc (commented):
Thanks, Craig - I see this is, as ever, more subtle than I realized, well done for figuring it out. I'm not sure I understand all the tests tho - any chance of some comments (as per comments)?

src/extension/infer.rs
```diff
@@ -279,7 +279,13 @@ impl UnificationContext {
                     let m_input_node = self.make_or_get_meta(input, dir);
                     self.add_constraint(m_input_node, Constraint::Equal(m_input));
                     let m_output_node = self.make_or_get_meta(output, dir);
-                    self.add_constraint(m_output_node, Constraint::Equal(m_output));
+                    // If the parent node is a FuncDefn, it's input_resources
+                    // should always be empty, meaning if the function adds
```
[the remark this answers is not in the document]
acl-cqc: I think it's the default, but we don't insist. OTOH I think it might be sane if we *did* insist, and I'd be happy to make that change in this PR

src/extension/infer.rs
acl-cqc: Sorry, yes, well, the "its input_extensions should always be empty" bit is as above, but the point to make is that the parent's output is equal to its input, because these do not include the delta of the function itself.

src/extension/infer.rs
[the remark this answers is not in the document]
acl-cqc: We might have to change EdgeKind::Static not to require equality of extensions at both ends. But we should probably do that anyway - ATM all calls will have to have the same input-extensions, as you can't `lift` the static input

src/extension/infer.rs
acl-cqc: Hang on. (Hang on!) Don't we need an additional constraint, then, that the (input-extensions to the) Output node are Plus of (the delta *of the FuncDefn's declared FunctionType* and the Input node)?

Could add a test of a FuncDefn that declares delta is {A} and then has a body that Lift's only by {B}, say - that should be rejected, but is it?

src/extension/infer.rs
acl-cqc: in fact, "it's input resources" -> "its input_extensions" (it's = it is, its = belonging to it)

src/extension/infer.rs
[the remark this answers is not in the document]
acl-cqc: Ok, if #693 doesn't fly, we need an alternative approach for dealing with functions. Maybe splitting EdgeKind::Static into EdgeKind::Function and EdgeKind::Const (as suggested [here](https://github.com/CQCL/hugr/discussions/285#discussioncomment-7551985)) would be a good way to go *anyway*....

What about resurrecting PR #693, and giving the LoadConstant node a nonzero delta built from the constant being loaded? No, that would leave the input-extensions for the LoadConstant undefined :-(.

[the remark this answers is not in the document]
acl-cqc: Maybe raise an issue "All calls to a function must have the same input_extensions" - I think that's a fair description of the problem (and it sounds like a problem to me)?

src/extension/infer/test.rs
croyzor: I fixed it but it doesn't pass! It's really requires us to be ignoring "Static" edges for functions, so I'll re-add it when I open the PR fixing #695

src/extension/infer.rs
croyzor: I've added a test for this "funcdefn_signature_mismatch"

src/extension/infer.rs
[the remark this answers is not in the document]
croyzor: Raised #695

src/extension/infer.rs
croyzor: Removed the stuff about empty input_extensions, since it isn't true

src/hugr/validate/test.rs
croyzor: Added a comment explaining this

PR #692, opened 2023-11-14
feat: make FuncDecl/FuncDefn polymorphic

acl-cqc:
* Change FuncDecl and FuncDefn from storing FunctionType to PolyFuncType
* Change `declare_function` and other builder methods to take PolyFuncType rather than Signature. The input-extensions of the Signature were only used for the FuncDefn node itself and it seems saner, and tests pass ok, for the builder to set the input-extensions of the FuncDefn node itself to pure==empty in all cases. Note this makes #695 slightly even more severe, but I feel we should just take this into account when fixing that bug.
    * Much of the bulk of the PR is thus changing (someFunctionType)`.pure()` to `.into()`...
* Restructure validation. Break out the op-type specific checks that type args were declared, into a hierarchical traversal (rather than iterating over the Hugr's nodes), so they can gather binders from the nearest enclosing FuncDefn (potentially several levels above in the hierarchy). Move `validate_port` (and hence `validate_edge`) here because ports/edges inside a FuncDefn may have types using its parameters.
   * Also do not allow Static edges to refer to TVs declared outside. (The Static edge from a polymorphic FuncDefn contains a PolyFuncType i.e. declaring its own type vars, so they are not free.)
* Update OpDef to properly check_args etc. reflecting that we may need to pass some args to the binary function and then more args to the *returned* PolyFuncType.
   * Also drop `impl TypeParametrised for OpDef`
* Add TypeArgs to Call, and builder `fn call()`; also give the latter an ExtensionRegistry argument
* Finally add `impl From<TypeBound> for TypeParam`, which we probably should have done much earlier.


src/extension/op_def.rs
```diff
@@ -143,26 +141,36 @@ pub struct OpDef {
...
-        self.extension()
-    }
-}
-
 impl OpDef {
-    /// Check provided type arguments are valid against parameters.
-    pub fn check_args(&self, args: &[TypeArg]) -> Result<(), SignatureError> {
-        self.check_args_impl(args)
+    /// Check provided type arguments are valid against [ExtensionRegistry],
+    /// against parameters, and that no type variables are used as static arguments
```
acl-cqc, line 146: Hmmm, I could link to `[Self::params()]`, but I also wonder if we should keep that, or switch to `static_params()` and `substitutable_params(&[TypeArg])` (where the latter has to call the custom binary function)

src/hugr/validate.rs
```diff
@@ -558,6 +528,62 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
...
     }
+
+    fn validate_subtree(&self, node: Node, var_decls: &[TypeParam]) -> Result<(), ValidationError> {
+        let op_type = self.hugr.get_optype(node);
+        // TODO consider turning this match into a trait method?
+        match op_type {
+            OpType::LeafOp(crate::ops::LeafOp::CustomOp(b)) => {
+                // Try to resolve serialized names to actual OpDefs in Extensions.
+                let e: Option<ExtensionOp>;
+                let ext_or_opaq = match &**b {
```
acl-cqc, line 539: This was quite hard to get right, but the Either is necessary because we want only a *reference* - an `&ExtensionOp` or `&OpaqueOp`, whereas an `ExternalOp` stores an actual `ExtensionOp` or `OpaqueOp`

src/hugr/validate.rs
```diff
@@ -558,6 +521,77 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
...
+
+        // Check port connections.
+        for dir in Direction::BOTH {
+            for (i, port_index) in self.hugr.graph.ports(node.pg_index(), dir).enumerate() {
+                let port = Port::new(dir, i);
+                self.validate_port(node, port, port_index, op_type, var_decls)?;
+            }
+        }
+
+        // Compute cumulative TypeParams for nested FuncDefns, i.e. allow inner FuncDefns
```
acl-cqc, line 576: This is a policy that we could change, but it'd not simplify things much - we'd lose this appending logic, but we'd still need to do the hierarchical traversal, because nodes/ports/edges could still refer to type variables bound by the closest-ancestral FuncDefn node, which could be many levels above in a deep nest of Dfg/Cfg nodes.

src/builder/build_traits.rs
```diff
@@ -612,6 +612,14 @@ pub trait Dataflow: Container {
                 })
             }
         };
+        let signature = type_scheme.instantiate(type_args, exts).map_err(|e| {
+            BuildError::InvalidHUGR(ValidationError::SignatureError {
+                // TODO this is rather a horrendous hack. Do we need some way of returning a SignatureError without a node
```
croyzor, line 617: I guess we should make a new error constructor for this

src/hugr/validate.rs
acl-cqc: No, managed to do this with ExternalOp after all

src/builder/build_traits.rs
acl-cqc: Yes, done, hope the comment clarifies

src/hugr/validate.rs
acl-cqc: It'd not simplify things much *unless* we are about to simplify the type hierarchy (and remove polymorphic lambdas), whereupon we might then also be able to remove InsideBinders, Renumber, Substitution, etc....hmmm

src/hugr/views.rs
```diff
@@ -336,12 +336,12 @@ pub trait HugrView: sealed::HugrInternals {
...
     /// For function-like HUGRs (DFG, FuncDefn, FuncDecl), report the function
     /// type. Otherwise return None.
-    fn get_function_type(&self) -> Option<&FunctionType> {
+    fn get_function_type(&self) -> Option<PolyFuncType> {
         let op = self.get_nodetype(self.root());
         match &op.op {
-            OpType::DFG(DFG { signature })
-            | OpType::FuncDecl(FuncDecl { signature, .. })
-            | OpType::FuncDefn(FuncDefn { signature, .. }) => Some(signature),
+            OpType::DFG(DFG { signature }) => Some(signature.clone().into()),
```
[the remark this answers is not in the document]
acl-cqc: In the signature for a DFG, there is no requirement that the body is closed. The body could well have type variables bound in the nearest enclosing FuncDefn ?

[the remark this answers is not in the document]
croyzor: Ah, I see. in my head there was the distinction between `body`'s of `PolyFuncType` which could have free variables, and I was assuming that `FunctionType`s created every other way would be closed

src/ops/validate.rs
```diff
@@ -77,12 +77,8 @@ impl ValidateOp for super::FuncDefn {
         &self,
         children: impl DoubleEndedIterator<Item = (NodeIndex, &'a OpType)>,
     ) -> Result<(), ChildrenValidationError> {
-        validate_io_nodes(
-            &self.signature.input,
-            &self.signature.output,
-            "function definition",
-            children,
-        )
+        let FunctionType { input, output, .. } = &self.signature.body;
```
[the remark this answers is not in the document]
acl-cqc: FuncDefn signatures aren't closed ATM - they could include type variables bound by an *outer* FuncDefn. Maybe I should change that...but I realize this is perhaps a little sneaky, or at least non-obvious, so added a comment on the previous line.

src/hugr/views.rs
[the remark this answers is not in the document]
acl-cqc: We have a FunctionType for every operation in the implementation/definition of the function, so if the latter has a polymorphic type, operations will too... (and similarly we have a Type for most edges too)

src/hugr/validate.rs
```diff
@@ -558,6 +521,82 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
         }
         .into())
     }
+
+    fn validate_subtree(
```
croyzor, line 525: Add a comment explaining what this does, like the previous "validate op-specific constraints"

src/hugr/validate.rs
```diff
@@ -558,6 +521,82 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
...
+                    .map_err(|cause| ValidationError::SignatureError { node, cause })?;
+            }
+            _ => (),
+        }
+
+        // Check port connections.
+        for dir in Direction::BOTH {
+            for (i, port_index) in self.hugr.graph.ports(node.pg_index(), dir).enumerate() {
+                let port = Port::new(dir, i);
+                self.validate_port(node, port, port_index, op_type, var_decls)?;
```
croyzor, line 577: Say we have a ModuleOp "outer" containing an op "x" and a nested module "inner" which contains a node "y". Each module is free to bind some type variables. Can we have an inter graph edge between "x" and "y"? If so, the current de Bruijn indices aren't sufficient.

If this is true and you agree, I'll make an issue for it

acl-cqc: Ouch! You evil.... ;-) ;-)

I think though that you couldn't have an inter-graph/non-local Value edge. The only thing you could have would be a Static edge. But yeah, if such can refer to enclosing type variables, that would lead to exactly the kind of problem you describe (?)

This is sounding like the death knell of "nested Function nodes can refer to type variables declared by an enclosing FuncDefn"....

acl-cqc: Good spot!!

src/hugr/validate.rs
acl-cqc: Simplified.

src/hugr/validate.rs
acl-cqc: Been a bit more specific!

src/hugr/validate.rs
acl-cqc: Ok, I've added extra restriction on Static Edges, to rule out the case of polymorphic consts which would otherwise break this. (Also ruled out non-closed FuncDefn's by replacing rather than concatenating lists of binders, which dealt with the function case, but needed to go further for Const.)

croyzor: I was going to suggest instantiating variables with a different naming scheme during validation rather than scrapping the concept, but this might be a more appropriate level of abstraction for an IR (and we can always change our minds later)

croyzor (approved):

src/ops/validate.rs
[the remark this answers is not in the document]
acl-cqc: No, think this is still correct, it delegates all responsibility for the TVs elsewhere so no updates re. closest/any-enclosing required here

PR #693, opened 2023-11-15
fix(Inference): Don't add constraints between static edges

croyzor:
As mentioned in #688, ([here](https://github.com/CQCL/hugr/pull/688#discussion_r1393404921)) don't require that both sides of a static edge have the same extension requirements, since the static edge often has the empty set.

### DIFF
--- a/src/hugr/rewrite/replace.rs
+++ b/src/hugr/rewrite/replace.rs
@@ -482,7 +482,7 @@ mod test {
             FunctionType::new_linear(just_list.clone()).with_extension_delta(&exset),
         )?;
 
-        let pred_const = cfg.add_constant(ops::Const::unary_unit_sum(), None)?;
+        let pred_const = cfg.add_constant(ops::Const::unary_unit_sum(), exset)?;
 
         let entry = single_node_block(&mut cfg, pop, &pred_const, true)?;
         let bb2 = single_node_block(&mut cfg, push, &pred_const, false)?;
--- a/src/extension/infer.rs
+++ b/src/extension/infer.rs
@@ -328,9 +328,7 @@ impl UnificationContext {
             for port in hugr.node_inputs(tgt_node).filter(|src_port| {
                 matches!(
                     sig.port_kind(*src_port),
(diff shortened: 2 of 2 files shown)


src/hugr/rewrite/replace.rs
```diff
@@ -482,7 +482,7 @@ mod test {
             FunctionType::new_linear(just_list.clone()).with_extension_delta(&exset),
         )?;
 
-        let pred_const = cfg.add_constant(ops::Const::unary_unit_sum(), None)?;
+        let pred_const = cfg.add_constant(ops::Const::unary_unit_sum(), exset)?;
```
acl-cqc, line 485: This is because we can no longer infer an extension set for the Const itself, right? Presumably we could still use None for the LoadConstant? So what about changing the builder to automatically use `pure` for the Const node, but still allow specifying something for the LoadConstant?

acl-cqc (approved):
Seems sane, thanks - if tests are passing then looks good to me :-)

src/hugr/rewrite/replace.rs
acl-cqc: Ah, no, hang on - this is just the Const, there is no LoadConstant here. So IOW, the `exset` could be anything, that will play no further part in inference?

Hmmm, we should drop the param. That could be another PR/issue, tho.

acl-cqc:
There is also the question of, if there's a constant value from a particular extension, do we want that to be reflected in the extension-delta of the containing Dfg/Function. This PR means it won't be. That might be fine (if values are passed around as generic YAML, say) or not fine (if we need to understand the extension to load CustomConst implementations)... this is probably worth making an issue?

src/hugr/rewrite/replace.rs
croyzor: I think we should drop the parameter, default to `pure` for const and move this param to `load_constant`. I'll open another PR for this.
At the moment though, `load_constant` uses the extension set that is provided with the ConstId, which is why `exset

PR #694, opened 2023-11-15
docs: Remove "quantum extension" from HUGR spec.

cqc-alec:


ss2165 (approved):
I think it's worth keeping this text around though - maybe move to the tket2 repository?

PR #697, opened 2023-11-15
refactor!: Rename `new_linear()` to `new_endo()`.

### LINKED ISSUE #664: Rename FunctionType::new_linear to new_endo
Linearity is a property of a value, not a function on that value. So, possibilities?
* `new_symmetric`? (thanks to @seyon)
* `self_composable`? But would have to be `new_self_composable`, too long.
* Mathematically speaking, such functions are iterable - one can compute f^2, f^3 ... f^n. So `new_iterable`? But in a Rusty world, that sounds confusing
* `new_elidable`? as in, we could just wire the inputs through to the outputs
* `new_identity_type`? (It has the same type as the identity function, and/or, the signature is what you get by applying the identity function to the input *types*)

cqc-alec:
Closes #664 .

BREAKING_CHANGES: `FunctionType::new_linear()` renamed to `FunctionType::new_endo()`.


acl-cqc (approved):
Looks good to me, thanks Alec :)

PR #699, opened 2023-11-15
ci: Compare coverage using specific commit shas

aborgna-q:
We store the last commit we notified about, so the notifications actually cover the period we care about.


cqc-alec (approved):
Looks good, thanks!

PR #700, opened 2023-11-15
refactor: Replace NodeType::signature() with io_extensions()

acl-cqc:
The signature() method on NodeType basically only existed so the caller could get both the input and output extensions (Signature providing the `.output_extensions()` method). Instead, provide `nodetype.io_extensions()` returning the same information.

This does leave Signature itself, which is a bit of a carbuncle.

### DIFF
--- a/src/hugr.rs
+++ b/src/hugr.rs
@@ -31,7 +31,7 @@ use crate::extension::{
 };
 use crate::ops::custom::resolve_extension_ops;
 use crate::ops::{OpTag, OpTrait, OpType, DEFAULT_OPTYPE};
-use crate::types::{FunctionType, Signature};
+use crate::types::FunctionType;
 use crate::{Direction, Node};
 
 use delegate::delegate;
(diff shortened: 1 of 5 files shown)


src/hugr.rs
```diff
@@ -134,6 +124,23 @@ impl NodeType {
...
+    /// `None`` if the [Self::input_extensions] is `None`.
+    /// Otherwise, will return Some, with the output extensions computed from the node's delta
+    pub fn io_extensions(&self) -> Option<(&ExtensionSet, ExtensionSet)> {
+        self.input_extensions.as_ref().map(|e| {
+            (
+                e,
+                self.op
+                    .dataflow_signature()
+                    .map(|ft| ft.extension_reqs)
+                    .unwrap_or_default()
```
acl-cqc, line 138: This is a change to the semantics introduced in #680 but I believe closer to what we had before - where a non-dataflow node would return `FunctionType::default()` (rather than `None` in #680) which meant the empty set of `extension_reqs`

ss2165 (approved):
I'm ok with making this change for now and removing `Signature` after #692

PR #704, opened 2023-11-20
refactor!: one way to add_op to extension

ss2165:
Methods to allow appending extra data to opdef after adding to extension.

BREAKING_CHANGES: `add_op` behaves like the now-removed `add_op_simple`

### DIFF
--- a/src/extension/op_def.rs
+++ b/src/extension/op_def.rs
@@ -423,8 +423,8 @@ impl Extension {
 
         match self.operations.entry(op.name.clone()) {
             Entry::Occupied(_) => Err(ExtensionBuildError::OpDefExists(op.name)),
-            Entry::Vacant(ve) => Ok(Arc::get_mut(ve.insert(Arc::new(op)))
-                .expect("Just made the arc so should only be one reference to it, can get_mut")),
+            // Just made the arc so should only be one reference to it, can get_mut,
+            Entry::Vacant(ve) => Ok(Arc::get_mut(ve.insert(Arc::new(op))).unwrap()),
         }
     }
 }
--- a/src/extension/op_def.rs
+++ b/src/extension/op_def.rs
@@ -339,6 +339,7 @@ impl OpDef {
     /// Fallibly returns a Hugr that may replace an instance of this OpDef
     /// given a set of available extensions that may be used in the Hugr.
     pub fn try_lower(&self, args: &[TypeArg], available_extensions: &ExtensionSet) -> Option<Hugr> {
+        // TODO test this
         self.lower_funcs
             .iter()
             .flat_map(|f| match f {
@@ -384,6 +385,20 @@ impl OpDef {
         }
         Ok(())
     }
+
+    /// Add a lowering function to the [OpDef]
+    pub fn add_lower_func(&mut self, lower: LowerFunc) {
+        self.lower_funcs.push(lower);
+    }
+
+    /// Insert miscellaneous data `v` to the [OpDef], keyed by `k`.
+    pub fn add_misc(
+        &mut self,
+        k: impl ToString,
+        v: serde_yaml::Value,
+    ) -> Option<serde_yaml::Value> {
+        self.misc.insert(k.to_string(), v)
+    }
 }
 
 impl Extension {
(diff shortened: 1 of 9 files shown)


src/extension/op_def.rs
```diff
@@ -70,28 +70,47 @@ pub trait CustomLowerFunc: Send + Sync {
     ) -> Option<Hugr>;
 }
 
+/// Type parameters passed to [func]. (The returned [PolyFuncType]
+/// may require further type parameters, not declared here.)
+pub struct CustomFunc {
+    static_params: Vec<TypeParam>,
```
acl-cqc, line 76: Now you've done your #678, I'd say moving this static-params getter into `trait CustomSignatureFunc` might be a good next step (perhaps before the rest of this?)

src/extension/op_def.rs
```diff
@@ -256,106 +276,78 @@ impl OpDef {
     }
 }
 
-impl Extension {
-    /// Add an operation definition to the extension.
-    fn add_op(
+/// Wrapper to allow editing an OpDef before it is added to an extension.
```
acl-cqc, line 279: Do you really mean "before"? Or, you could make `add_op` return the OpDefMut (and be the only way to do so); you'll have to give up the OpDefMut when you move the Extension into an ExtensionRegistry....

src/extension/op_def.rs
ss2165: Done in #706

src/extension/op_def.rs
ss2165: OpDefMut removed using `Arc::get_mut`

src/extension/op_def.rs
```diff
@@ -339,6 +339,7 @@ impl OpDef {
     /// Fallibly returns a Hugr that may replace an instance of this OpDef
     /// given a set of available extensions that may be used in the Hugr.
     pub fn try_lower(&self, args: &[TypeArg], available_extensions: &ExtensionSet) -> Option<Hugr> {
+        // TODO test this
```
ss2165, line 342: drive-by: this function isn't covered

src/extension/op_def.rs
```diff
@@ -384,6 +385,20 @@ impl OpDef {
         }
         Ok(())
     }
+
+    /// Add a lowering function to the [OpDef]
+    pub fn add_lower_func(&mut self, lower: LowerFunc) {
+        self.lower_funcs.push(lower);
```
croyzor, line 391: Why is this a vector?

src/extension/op_def.rs
```diff
@@ -395,41 +410,23 @@ impl Extension {
...
-            lower_funcs,
+            misc: Default::default(),
+            lower_funcs: Default::default(),
         };
 
         match self.operations.entry(op.name.clone()) {
             Entry::Occupied(_) => Err(ExtensionBuildError::OpDefExists(op.name)),
-            Entry::Vacant(ve) => Ok(ve.insert(Arc::new(op))),
+            Entry::Vacant(ve) => Ok(Arc::get_mut(ve.insert(Arc::new(op)))
+                .expect("Just made the arc so should only be one reference to it, can get_mut")),
```
croyzor, line 427: Maybe this should just be an `unwrap`?

src/extension/op_def.rs
ss2165: very historic at this point but the idea is that you might have different lowerings for the same op, something to revisit when we consider loweriing in anger I think

croyzor (approved):
This is almost a builder interface and would maybe be better as one, to keep the mutation out of the hands of the user

ss2165:
> to keep the mutable references out of the hands of the user

what is the danger?

croyzor:
> > to keep the mutable references out of the hands of the user
> 
> what is the danger?

It's not dangerous, per se, it just feels a bit gross to pass around mutable references for an interface that could be pure and functional

acl-cqc:
> agreed, ultimately it comes from the constraint that we don't want `OpDef`s to exist by themselves, only inside extensions. I'm not sure whether this is actually a useful constraint.

The issue here is the cyclic pseudo-reference: from the extension to the OpDef but also from the OpDef to the name of the extension that owns it

PR #715, opened 2023-11-23
ci: Fix incorrect coverage reported

aborgna-q:
We computed everything correctly, but printed the old value twice >.<


cqc-alec (approved):
oops

croyzor (approved):
🙌

PR #716, opened 2023-11-23
ci: Use sccache instead of rust-cache

aborgna-q:
Performance is about the same, but this is consistent with tket2's CI worflows.


doug-q:
I'm not sure that this is working. 
The documentation says that you need `export RUSTC_WRAPPER=/path/to/sccache` (or set rustc-wrapper in config.toml). 
I don't see any cache info on the action page, which I suspect means nothing is cached?

aborgna-q:
Good catch, seems I missed copying that line from
https://github.com/CQCL/tket2/blob/85ce5f9aa1233d0837d93f7666e58a318aba01ae/.github/workflows/ci.yml#L19-L20
:facepalm:

aborgna-q:
You are right on that, but even then the performance seem to be the same to rust-cache.
I'll keep an eye on it.

PR #720, opened 2023-11-23
feat: constant folding for arithmetic conversion operations

ss2165:


cqc-alec:
Could you resolve the conflicts? (Seems to be quite a lot of overlap with the already-merged PR.)

ss2165:
> Could you resolve the conflicts?

Yes, sorry, done

src/std_extensions/arithmetic/conversions/const_fold.rs
```diff
@@ -0,0 +1,134 @@
...
+struct TruncU;
+
+impl ConstFold for TruncU {
+    fn fold(
+        &self,
+        type_args: &[crate::types::TypeArg],
+        consts: &[(IncomingPort, ops::Const)],
+    ) -> ConstFoldResult {
+        fold_trunc(type_args, consts, |f, log_width| {
+            ConstIntU::new(log_width, f.trunc() as u64).map(Into::into)
```
cqc-alec, line 89: What if `f` is negative? Does `as u64` cause a panic?

cqc-alec (approved):

src/std_extensions/arithmetic/conversions/const_fold.rs
ss2165: `as u64` maps all negative floats to 0, should I add a panic?

cqc-alec: Hmm, this is a can of worms. The spec for `trunc_u` currently says "Returns an error when the float is non-finite or cannot be exactly stored in N bits". This should probably say "cannot be exactly represented as a `u<N>`". But what we're doing here is rounding, which is much more forgiving. WASM [seems to say](https://webassembly.github.io/spec/core/bikeshed/#-hrefop-trunc-umathrmtruncmathsfu_m-n-z) that the result is undefined if the number is negative, but rounded otherwise, which makes me think we should change the definition of `trunc_{u,s}` in the spec.

ss2165: sounds like this could be a follow up issue?

cqc-alec: Yes, I'll make one.

PR #722, opened 2023-11-27
fix!: Add serde tag to TypeParam enum

mark-koch:
Needed for Pydantic serialisation.

BREAKING_CHANGES: Turn the `TypeParam` enum variants into struct variants


ss2165 (changes requested):
this is a lot of noise, at least for `TypeParam::Type` could you use `bound.into()` 

https://github.com/CQCL/hugr/blob/7cb6751bad2e9687141f28055d5494f70adb76ff/src/types/type_param.rs#L94-L95

Issue #702, opened 2023-11-16
Extension Inference for functions relies on builder


acl-cqc:
When defining a function, the builder fixes solutions for the extensions of the Input and Output child of the FuncDefn:
https://github.com/CQCL/hugr/blob/762839da26fef556f1dbb0365e49f71aa2c63fdd/src/builder/build_traits.rs#L103-L107 (note last line)
https://github.com/CQCL/hugr/blob/41e15da4360eeaaba48e6bda74bffcf3d24bc067/src/builder/dataflow.rs#L53-L58 (note the two `input_extensions`)

The relationship between Input and Output is not modelled by inference - they should have a Plus constraint, but Plus constraints are only generated here:
https://github.com/CQCL/hugr/blob/41e15da4360eeaaba48e6bda74bffcf3d24bc067/src/extension/infer.rs#L317-L330 (note this only applies if the node *has* an op_signature, which FuncDefn doesn't, and only if extension-annotations need to be inferred; if they are pre-specified, inference just accepts them)....and neither is it checked by validation, as the only checks on Input/Output are done here:
https://github.com/CQCL/hugr/blob/41e15da4360eeaaba48e6bda74bffcf3d24bc067/src/hugr/validate.rs#L205-L212
(note FuncDefns are explicitly excluded, because `validate_io_children` would use the wrong delta for them).

Thus, it should be possible to construct a Hugr (*not* using the builder) where the body of the function adds extension-requirements not specified in the function delta, and to get this past validation (possibly past inference too, but otherwise by hand-specifying extensions). Step 1, obviously, produce that test case ;-)

acl-cqc:
May be best to do this after #692 where the builder changes to pass `None` (no extensions specified) to `create_with_io`

acl-cqc:
I think we can say this was done in #1142 / #1183

PR #733, opened 2023-12-01
feat: Values (and hence Consts) know their extensions

acl-cqc:
* Make each Value able to report its ExtensionSet needed at runtime (and hence, also each CustomConst)
* Also give OpTrait a method for the extension-delta. This is the delta of the FunctionType, *for dataflow ops*, but can be defined for other optypes too - specifically (here), Const ops and also Case. Use this in both inference and for `NodeType::io_extensions()`
* Input extensions of every Const node can now be the empty set (i.e. the default-for-ModuleOp `pure`), as the relevant extension will be added in the delta - thus, drop separate extension parameter in builder (add_constant, add_load_const, etc.). LoadConstant ops can now also have an empty delta and open input-extensions as these can be figured out from the Const. This is thus a step towards fixing #702...
* Also note slight issue in replace::test::cfg, pending #388
* Add `ExtensionSet::union_over(impl IntoIterator<Item=Self>) -> Self` utility method

This should ease the way towards solving the rest of #702 and moreover to removing `new_auto` - we should be able to constrain the input-extensions for any ModuleOp to `pure` in inference and thus *every* node can be created open rather than a mix, but those are all for follow-up PRs.

### DIFF
--- a/src/values.rs
+++ b/src/values.rs
@@ -122,10 +122,7 @@ impl Value {
         match self {
             Value::Extension { c } => c.0.extension_reqs().clone(),
             Value::Function { .. } => ExtensionSet::new(), // no extensions reqd to load Hugr (only to run)
-            Value::Tuple { vs } => vs
-                .iter()
-                .map(Value::extension_reqs)
-                .fold(ExtensionSet::new(), |a, b| a.union(&b)),
+            Value::Tuple { vs } => ExtensionSet::union_over(vs.iter().map(Value::extension_reqs)),
             Value::Sum { value, .. } => value.extension_reqs(),
         }
     }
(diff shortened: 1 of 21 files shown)


src/values.rs
```diff
@@ -115,6 +116,19 @@ impl Value {
...
+
+    /// The Extensions that must be supported to handle the value at runtime
+    pub fn extension_reqs(&self) -> ExtensionSet {
+        match self {
+            Value::Extension { c } => c.0.extension_reqs().clone(),
+            Value::Function { .. } => ExtensionSet::new(), // no extensions reqd to load Hugr (only to run)
+            Value::Tuple { vs } => vs
+                .iter()
+                .map(Value::extension_reqs)
+                .fold(ExtensionSet::new(), |a, b| a.union(&b)),
```
ss2165, line 128: this could be worth defining a function/method for?

PR #738, opened 2023-12-13
docs: Spec clarifications

cqc-alec:
A couple of small clarifications suggested by Will.

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1272,6 +1272,7 @@ The new hugr is then derived as follows:
     For each child $C$ of $P$ (in order), if $C \in S$, redirect the hierarchy edge $P \rightarrow C$ to
     target the next node in $N$. Stop if there are no more nodes in $N$.
     Add any remaining nodes in $N$ to the end of $P$'s list of children.
+    (Reindexing may be necessary after this step.)
 6.  For each node $(n, b = B(n))$ and for each child $m$ of $b$, replace the
     hierarchy edge from $b$ to $m$ with a hierarchy edge from the new copy of
     $n$ to $m$ (preserving the order).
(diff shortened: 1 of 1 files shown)


specification/hugr.md
```diff
@@ -1234,7 +1234,9 @@ Note that considering all three $\mu$ lists together,
    - the `TgtNode` + `TgtPos`s of all `NewEdgeSpec`s with `EdgeKind` == `Value` will be unique
    - and similarly for `EdgeKind` == `Static`
 
-The well-formedness requirements of Hugr imply that $\mu\_\textrm{inp}$ and $\mu\_\textrm{out}$ may only contain `NewEdgeSpec`s with certain `EdgeKind`s, depending on $P$:
+The well-formedness requirements of Hugr imply that $\mu\_\textrm{inp}$,
```
acl-cqc, line 1237: Yes.

specification/hugr.md
```diff
@@ -1259,6 +1261,7 @@ The new hugr is then derived as follows:
     For each child $C$ of $P$ (in order), if $C \in S$, redirect the hierarchy edge $P \rightarrow C$ to
     target the next node in $N$. Stop if there are no more nodes in $N$.
     Add any remaining nodes in $N$ to the end of $P$'s list of children.
+    (Reindexing may be necessary after this step.)
```
acl-cqc, line 1264: Not sure what reindexing this means. The new nodes $N$ are already the *copies made in $\Gamma$*, so that is the new indices. And things like "input/output nodes are first/second children" describes the ordering of the hierarchy edges not the *node* indices.

Reindexing might be necessary after step 7 ??

acl-cqc (commented):
Thanks @alec (and Will - is that @willsimmons1465 ?) - Happy for the first one to go in, but can you explain the second?

specification/hugr.md
cqc-alec: Ah I think that may be what I meant...

cqc-alec: ... however I am struggling to see why it would be necessary there. @willsimmons1465 could you clarify this please, it made sense when we discussed it but I seem to have forgotten now?

cqc-alec: Forgot they had to be densely indexed. OK will do that.

acl-cqc (approved):
Looks good, thanks @cqc-alec

PR #739, opened 2023-12-13
Infer extension constraints on FuncDefn; remove many ExtensionSets from builder

acl-cqc:
Follows #734. Big step towards #702 but maybe not the last word?
* Builder methods `create_with_io` and `dfg_builder` no longer take an ExtensionSet, nor set extensions of the Input/Output node
* Instead, extension inference now correctly understands the constraint between Input and Output for FuncDefn (the missing case before)
* Many test changes...e.g. tests looking for *validation* (rather than inference) failures must construct Hugrs that do not need inference (i.e. have all annotations already set), which often now means *not using the builder*

### DIFF
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -97,8 +97,7 @@ pub trait Container {
             ExtensionSet::new(),
         ))?;
 
-        let db =
-            DFGBuilder::create_with_io(self.hugr_mut(), f_node, body, Some(ExtensionSet::new()))?;
+        let db = DFGBuilder::create_with_io(self.hugr_mut(), f_node, body)?;
         Ok(FunctionBuilder::from_dfg_builder(db))
     }
 
@@ -303,10 +302,10 @@ pub trait Dataflow: Container {
         let op = ops::DFG {
             signature: signature.clone(),
         };
-        let nodetype = NodeType::new(op, input_extensions.clone());
+        let nodetype = NodeType::new(op, input_extensions);
         let (dfg_n, _) = add_node_with_wires(self, nodetype, input_wires.into_iter().collect())?;
 
-        DFGBuilder::create_with_io(self.hugr_mut(), dfg_n, signature, input_extensions)
+        DFGBuilder::create_with_io(self.hugr_mut(), dfg_n, signature)
     }
 
     /// Return a builder for a [`crate::ops::CFG`] node,
--- a/src/builder/build_traits.rs
+++ b/src/builder/build_traits.rs
@@ -296,13 +296,12 @@ pub trait Dataflow: Container {
     fn dfg_builder(
         &mut self,
         signature: FunctionType,
-        input_extensions: Option<ExtensionSet>,
         input_wires: impl IntoIterator<Item = Wire>,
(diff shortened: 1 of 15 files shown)


src/extension/infer/test.rs
```diff
@@ -154,36 +155,25 @@ fn plus() -> Result<(), InferExtensionError> {
 }
 
 #[test]
-// This generates a solution that causes validation to fail
```
acl-cqc, line 157: This comment was incorrect (it doesn't generate a solution - the error is CantInfer). But the body/subject of this test has largely switched with the identically-named test in validation.

src/extension/infer/test.rs
```diff
@@ -996,6 +986,7 @@ fn funcdefn_signature_mismatch() -> Result<(), Box<dyn Error>> {
         result,
         Err(ValidationError::CantInfer(
             InferExtensionError::MismatchedConcreteWithLocations { .. }
+                | InferExtensionError::EdgeMismatch(ExtensionError::SrcExceedsTgtExtensions { .. })
```
acl-cqc, line 989: This one is a bit nondeterministic as to which error is reported

src/hugr/views/sibling_subgraph.rs
```diff
@@ -762,7 +763,12 @@ mod tests {
 
     fn build_3not_hugr() -> Result<(Hugr, Node), BuildError> {
         let mut mod_builder = ModuleBuilder::new();
-        let func = mod_builder.declare("test", FunctionType::new_endo(type_row![BOOL_T]).into())?;
+        let func = mod_builder.declare(
+            "test",
+            FunctionType::new_endo(type_row![BOOL_T])
```
acl-cqc, line 768: That this delta wasn't needed immediately following #734 is a kind of "bug" fixed by this PR

src/hugr/views/sibling_subgraph.rs
```diff
@@ -781,7 +787,9 @@ mod tests {
         let mut mod_builder = ModuleBuilder::new();
         let func = mod_builder.declare(
             "test",
-            FunctionType::new(type_row![BOOL_T], type_row![BOOL_T]).into(),
+            FunctionType::new_endo(type_row![BOOL_T])
```
acl-cqc, line 790: ...and similarly

src/builder/build_traits.rs
```diff
@@ -297,16 +296,15 @@ pub trait Dataflow: Container {
...
         &mut self,
         signature: FunctionType,
-        input_extensions: Option<ExtensionSet>,
         input_wires: impl IntoIterator<Item = Wire>,
     ) -> Result<DFGBuilder<&mut Hugr>, BuildError> {
         let op = ops::DFG {
             signature: signature.clone(),
         };
-        let nodetype = NodeType::new(op, input_extensions.clone());
+        let nodetype = NodeType::new_auto(op);
```
acl-cqc, line 304: The choice of new_auto here is about the only non-obvious thing in the PR. ("pure" for `FuncDefn` but "open" for any other DFG.) I think the next PR might even remove new_auto and that could be a good point to say #702 is finished

acl-cqc:
#702 closed and this long superceded...(#1142 / #1183)

PR #740, opened 2023-12-13
feat!: use ConvexChecker trait

lmondada:
A new portgraph release is needed for this.


aborgna-q (approved):
LGTM

I just pushed a change to use the latest portgraph instead of a git dependency

PR #741, opened 2023-12-14
docs: Spec updates

cqc-alec:
Following Will's review.

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1350,8 +1350,8 @@ wiring its predecessor to its successor.
 
 Insert an Order edge from `n0` to `n1` where `n0` and `n1` are distinct
 siblings in a DSG such that there is no path in the DSG from `n1` to
-`n0`. If there is already an order edge from `n0` to `n1` this does
-nothing (but is not an error).
+`n0`. (Thus acyclicity is preserved.) If there is already an order edge from
+`n0` to `n1` this does nothing (but is not an error).
 
 ###### `RemoveOrder`
 
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1356,8 +1356,8 @@ siblings in a DSG such that there is no path in the DSG from `n1` to
 ###### `RemoveOrder`
 
 Given nodes `n0` and `n1`, if there is an Order edge from `n0` to `n1`,
-remove it. (If there is an intergraph edge from `n0` to a descendent of
-`n1`, this invalidates the hugr. TODO should this be an error?)
+remove it. (If there is a non-local edge from `n0` to a descendent of `n1`, this
+invalidates the hugr. TODO should this be an error?)
 
 ##### Insertion and removal of const loads
 
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1396,7 +1396,7 @@ nodes.
 
 The most basic case – replacing a convex set of Op nodes in a DSG with
 another graph of Op nodes having the same signature – is implemented by
-having T map everything to the parent node, and bot(G) is empty.
+`SimpleReplace`.
 
 If one of the nodes in the region is a complex container node that we
 wish to preserve in the replacement without doing a deep copy, we can
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -408,8 +408,6 @@ Some normalizations are possible:
   - If the entry node has only one successor and that successor is the
     exit node, the CFG node itself can be removed.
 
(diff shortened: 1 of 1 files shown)


ss2165:
couldn't a `passthrough<T>` just be a normal basic block that wires inputs to outputs?

cqc-alec:
> couldn't a `passthrough<T>` just be a normal basic block that wires inputs to outputs?

Yes that's the idea, but I needed a name for it (it isn't a new node type).

ss2165:
Ah ok, but just to double check `identity<T>` is a node type right?

cqc-alec:
> Ah ok, but just to double check `identity<T>` is a node type right?

Hmm, yes. This is confusing I agree. Will try to make clearer.

specification/hugr.md
```diff
@@ -1191,7 +1194,18 @@ The new hugr is then derived as follows:
...
 ###### `Replace`
 
-This is the general subgraph-replacement method.
+This is the general subgraph-replacement method. Intuitively, it takes a set of
+sibling nodes to remove and replace with a new set of nodes. The new set of
+nodes is itself a HUGR with some "holes" (edges and nodes that get "filled in"
+by the `Replace` operation). To fully specify the operation, some further data
+are needed:
+
+  - The replacement may containg container nodes with no children, which replace
```
acl-cqc, line 1203: `containg` -> `contain` or perhaps `include`. However I'm less sure about these "replace"ing removed container nodes. They do WRT the replaced node's children, but not necessarily in any other way (e.g. they may have different parents, siblings, incoming/outgoing dataflow/controlflow edges to the "replaced" container nodes).

specification/hugr.md
```diff
@@ -408,6 +408,9 @@ Some normalizations are possible:
   - If the entry node has only one successor and that successor is the
     exit node, the CFG node itself can be removed.
 
+A `DFB` that simply wires its inputs to its outputs, having no effect on the
```
acl-cqc, line 411: Is this a new node type? I think it's just a special case of an existing one, so might be worth clarifying that.
* If it is indeed a special case, then it'll still have Input and Output children with dataflow edges between those
* But if it's a special kind of node then maybe it has no children. This would be more like Identity, and that'd be nice for uniformity wrt (Remove/Insert)Identity. Note this would be a lot like an Exit node, so maybe we can even *remove* the Exit node type by requiring that the second child of a CFG node be a passthrough node (even if not necessarily the only one)...???

specification/hugr.md
acl-cqc: How about:
```
 - The replacement may include container nodes with no children, which adopt the children of removed container nodes and prevent those children being removed
```
? - note we haven't said (yet / in the summary) that children of removed nodes are themselves removed

specification/hugr.md
```diff
@@ -1371,7 +1386,7 @@ nodes.
 
 The most basic case – replacing a convex set of Op nodes in a DSG with
 another graph of Op nodes having the same signature – is implemented by
-having T map everything to the parent node, and bot(G) is empty.
+`SimpleReplace`.
```
acl-cqc, line 1389: Hehe! Yes :)

specification/hugr.md
```diff
@@ -1191,7 +1194,18 @@ The new hugr is then derived as follows:
 
 ###### `Replace`
 
-This is the general subgraph-replacement method.
+This is the general subgraph-replacement method. Intuitively, it takes a set of
```
acl-cqc, line 1197: Great idea to provide an intuition here :)

acl-cqc (commented):
Thanks @alec - in general, like it :), some thoughts about `passthrough` and "adoptions"

specification/hugr.md
cqc-alec: So the intention was not to define a new node type here, but maybe there is an argument for doing so, especially if it allows us to remove the `Exit` type. Hmm, that seems like a big change though.

acl-cqc: Conversely, maybe `identity<T>` is not required. It could just be a `DFG` node with, much like a "passthrough" DFB, all out-ports of the Input node wired to all in-ports of the Output node....

Perhaps the thing to do is to break the passthrough/identity/etc. change out (if it's gonna be big one way or another) and then everything else here seems uncontroversial...

cqc-alec: https://github.com/CQCL/hugr/issues/763

acl-cqc (approved):
All good by me, thanks @cqc-alec. Just remove the node about passthrough in the description before you merge :)

PR #749, opened 2023-12-18
fix: case node should not have an external signature

### LINKED ISSUE #750: Case nodes have unconnected ports
as identified by @mark-koch 

The serialized hugr in this file:
[graph (1).hugr.zip](https://github.com/CQCL/hugr/files/13706416/graph.1.hugr.zip)

fails validation when deserialized because case nodes are reporting that they have unconnected ports (they shouldn't have any)

ss2165:
Previously it was incorrectly reporting the internal signature as the node signature.

Fixes #750 

Required further fixes to `replace.rs`:
* test was looking for an error that should have been hidden behind another (that was not previously reported because of #750). Look for the latter, not the former.
* Fix node indices not being correctly reported from `apply`

### DIFF
--- a/src/extension/infer/test.rs
+++ b/src/extension/infer/test.rs
@@ -11,7 +11,7 @@ use crate::extension::{prelude::PRELUDE_REGISTRY, ExtensionSet};
 use crate::hugr::{validate::ValidationError, Hugr, HugrMut, HugrView, NodeType};
 use crate::macros::const_extension_ids;
 use crate::ops::custom::{ExternalOp, OpaqueOp};
-use crate::ops::{self, dataflow::IOTrait, handle::NodeHandle, OpTrait};
+use crate::ops::{self, dataflow::IOTrait, handle::NodeHandle};
 use crate::ops::{LeafOp, OpType};
 
 use crate::type_row;
@@ -314,12 +314,8 @@ fn test_conditional_inference() -> Result<(), Box<dyn Error>> {
         first_ext: ExtensionId,
         second_ext: ExtensionId,
     ) -> Result<Node, Box<dyn Error>> {
-        let [case, case_in, case_out] = create_with_io(
-            hugr,
-            conditional_node,
-            op.clone(),
-            Into::<OpType>::into(op).dataflow_signature().unwrap(),
-        )?;
+        let [case, case_in, case_out] =
+            create_with_io(hugr, conditional_node, op.clone(), op.inner_signature())?;
 
         let lift1 = hugr.add_node_with_parent(
             case,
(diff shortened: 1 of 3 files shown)


ss2165:
@acl-cqc This has induced a replace test to fail at two points: 
https://github.com/CQCL/hugr/blob/c48fe7add21a29ecb131f97c469b7050a5813e31/src/hugr/rewrite/replace.rs#L684

and 
https://github.com/CQCL/hugr/blob/c48fe7add21a29ecb131f97c469b7050a5813e31/src/hugr/rewrite/replace.rs#L818-L824

The errors do not match the expected ones. I am having a hard time figuring out whether this is an actual regression or just that the expected error needs updating. Any pointers are appreciated.

Other tests are fortunately happy.

acl-cqc:
> @acl-cqc This has induced a replace test to fail at two points:
> 
> https://github.com/CQCL/hugr/blob/c48fe7add21a29ecb131f97c469b7050a5813e31/src/hugr/rewrite/replace.rs#L684
> 
> and
> 
> https://github.com/CQCL/hugr/blob/c48fe7add21a29ecb131f97c469b7050a5813e31/src/hugr/rewrite/replace.rs#L818-L824
> 
> The errors do not match the expected ones. I am having a hard time figuring out whether this is an actual regression or just that the expected error needs updating. Any pointers are appreciated.
> 
> Other tests are fortunately happy.

Neither is obvious, but the error in `verify_apply` definitely suggests a bug - the two methods are reporting different errors, albeit, the same kind of error but at different node indices. I guess one may not be translating indices (as happens when the replacement subgraph is inserted) but the other is. That *both* are giving a different error than the expected one, I'm less sure - I can have a look when I'm back in a couple of days

acl-cqc:
Ok, the failure in `verify_apply` is fixed by branch `fix/replace-bad-edge-nums`. Of course, really we should have a test of *that* case, i.e. returning such a `BadEdgeKind`.

Your change to `Case` indeed makes the current failure such a test (e.g. if we changed `test_invalid` to look for BadEdgeKind). *However* we are also missing a test of what that bit was *supposed* to test, i.e. a Replace that specifies a new out-edge in `mu_out` that does not correspond to an existing edge. That'll take a bit more test-writing....

acl-cqc:
General thought on PR: `inner_signature` is good, but should this be a method specific to `Case`, does the same not apply to `FuncDefn` and also `BasicBlock::DFG` i.e. any DataflowContainer? (Even `DFG` - just in that case only, `inner_signature` is the same as `dataflow_signature`)

acl-cqc (approved):
Well a bunch of problems from my commits ;-) that `clippy --fix` should handle! But otherwise looks good, we should raise at least the issue that `inner_signature` is providing a case missing from HugrView's `get_function_type`, and possibly also that we've lost coverage of a Replace error (I mean, if we are surveying test coverage thoroughly+regularly enough then maybe that's redundant but)

PR #754, opened 2023-12-20
feat: DataflowParent trait for getting inner signatures

ss2165:


src/builder/cfg.rs
```diff
@@ -248,18 +242,9 @@ impl<B: AsMut<Hugr> + AsRef<Hugr>> BlockBuilder<B> {
...
-        inputs: TypeRow,
-    ) -> Result<Self, BuildError> {
-        // The node outputs a TupleSum before the data outputs of the block node
-        let tuple_sum_type = Type::new_tuple_sum(tuple_sum_rows);
-        let mut node_outputs = vec![tuple_sum_type];
-        node_outputs.extend_from_slice(&other_outputs);
-        let signature = FunctionType::new(inputs, TypeRow::from(node_outputs));
+    fn create(base: B, block_n: Node) -> Result<Self, BuildError> {
+        let block_op = base.get_optype(block_n).as_basic_block().unwrap();
+        let signature = block_op.inner_signature();
```
acl-cqc, line 247: I was a bit suspicious how this change fitted in at first but it's a nice result!

src/types/check.rs
```diff
@@ -45,6 +49,22 @@ pub enum ConstTypeError {
...
+fn check_ts(v: &Hugr, t: &PolyFuncType) -> bool {
+    // exact signature equality, in future this may need to be
+    // relaxed to be compatibility checks between the signatures.
+    let root_op = v.get_optype(v.root());
+    if let OpType::FuncDecl(FuncDecl { signature, .. })
+    | OpType::FuncDefn(FuncDefn { signature, .. }) = root_op
+    {
+        signature == t
+    } else {
+        v.get_function_type().is_some_and(|ft| {
```
acl-cqc, line 61:
```suggestion
        v.get_function_type().is_some_and(|ft| &PolyFuncType::from(ft) == t)
```

src/ops/controlflow.rs
```diff
@@ -145,6 +145,26 @@ impl StaticTag for BasicBlock {
...
+                tuple_sum_rows,
+                ..
+            } => {
+                // The node outputs a TupleSum before the data outputs of the block node
+                let tuple_sum_type = Type::new_tuple_sum(tuple_sum_rows.clone());
+                let mut node_outputs = vec![tuple_sum_type];
+                node_outputs.extend_from_slice(other_outputs);
+                FunctionType::new(inputs.clone(), TypeRow::from(node_outputs))
+            }
+            BasicBlock::Exit { cfg_outputs } => FunctionType::new(type_row![], cfg_outputs.clone()),
```
acl-cqc, line 163: This is the biggest wart here. When does a node have an inner_signature? I thought it'd be when it had Input + Output children, but Exit does not....

I'm not sure if this justifies declaring `inner_signature(&self) -> Option<FunctionType>`. If one did that, you could even put it in `OpTrait` with a default returning None, which'd save that big switch in `get_function_type`. What do you think?

acl-cqc (commented):
Mostly - love it :). I am wondering if the issue with BasicBlock::Exit suggests an alternative route, however....?

src/ops/controlflow.rs
ss2165: I had the same conflict when doing it - I think my preference is actually to break up the basic block enum and move the shared functionality to a block trait.

Happy to do as a follow up or here

acl-cqc: Yes, I think that's an improvement, and there are some other improvements that enables, so we could go that way. We could (also, independently) just drop the trait itself, as we never use it (i.e. as a bound).

I do still wonder about the default-None, though - the approach here means that for each relevant OpType we have to *both* implement the trait *and* add an extra case to `get_function_type`, whereas by defaulting to None you only have to say that an OpType is/has a function-type in one place...

src/ops.rs
```diff
@@ -314,6 +314,37 @@ pub trait OpTrait {
...
+    }
+}
+
+impl<T: DataflowParent> OpParent for T {
+    fn inner_function_type(&self) -> Option<FunctionType> {
+        Some(DataflowParent::inner_signature(self))
+    }
+}
+
+impl OpParent for Module {}
```
acl-cqc, line 333: As I understand it, `enum_dispatch` requires that we implement OpParent (somehow) for basically anything that we use as an Op, so although these are a minor annoyance, there's little opportunity for error here. (Right?)

ss2165: right, missing one will cause an error

src/types/check.rs
```diff
@@ -45,6 +49,20 @@ pub enum ConstTypeError {
     CustomCheckFail(#[from] CustomCheckFailure),
 }
 
+fn check_ts(v: &Hugr, t: &PolyFuncType) -> bool {
```
acl-cqc, line 52: Let's have a better name - "type_sig_equal" or something like that perhaps

ss2165:
Closing in favour of #782

PR #755, opened 2023-12-21
refactor!: use enum op traits for floats + conversions

ss2165:
BREAKING CHANGES: extension() function replaced with EXTENSION static ref for float_ops and conversions

### DIFF
--- a/src/std_extensions/arithmetic/float_ops.rs
+++ b/src/std_extensions/arithmetic/float_ops.rs
@@ -1,103 +1,119 @@
 //! Basic floating-point operations.
 
-use crate::{
-    extension::{ExtensionId, ExtensionSet},
-    type_row,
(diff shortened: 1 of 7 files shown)


src/std_extensions/arithmetic/float_ops.rs
```diff
@@ -1,103 +1,119 @@
...
+            fge => "\"greater than or equal\"",
+            fmax => "maximum",
+            fmin => "minimum",
+            fadd => "addition",
+            fsub => "subtraction",
+            fneg => "negation",
+            fabs => "absolute value",
+            fmul => "multiplication",
+            fdiv => "division",
+            ffloor => "floot",
```
croyzor, line 80: `floot`

croyzor (approved):

PR #757, opened 2023-12-22
feat: implement RemoveConst and RemoveConstIgnore

ss2165:
as per spec

refactor!: allow Into<Const> for builder.add_const

BREAKING_CHANGES: existing CustomConst.into() calls will error


src/hugr/rewrite/consts.rs
```diff
@@ -0,0 +1,231 @@
...
+
+/// Remove a [`crate::ops::LoadConstant`] node with no outputs.
+#[derive(Debug, Clone)]
+pub struct RemoveConstIgnore(pub Node);
+
+/// Error from an [`RemoveConstIgnore`] operation.
+#[derive(Debug, Clone, Error, PartialEq, Eq)]
+pub enum RemoveConstIgnoreError {
+    /// Invalid node.
+    #[error("Node is invalid (either not in HUGR or not LoadConst).")]
```
acl-cqc, line 23:
```suggestion
    #[error("Node is invalid (either not in HUGR or not LoadConstant).")]
```

src/hugr/rewrite/consts.rs
```diff
@@ -0,0 +1,231 @@
...
+};
+#[rustversion::since(1.75)] // uses impl in return position
+use itertools::Itertools;
+use thiserror::Error;
+
+use super::Rewrite;
+
+/// Remove a [`crate::ops::LoadConstant`] node with no outputs.
+#[derive(Debug, Clone)]
+pub struct RemoveConstIgnore(pub Node);
```
acl-cqc, line 17: We should really rename this to `RemoveLoadConstant` rather than this "ignore" malarky?

src/hugr/rewrite/consts.rs
```diff
@@ -0,0 +1,231 @@
...
+    hugr::{HugrError, HugrMut},
+    HugrView, Node,
+};
+#[rustversion::since(1.75)] // uses impl in return position
+use itertools::Itertools;
+use thiserror::Error;
+
+use super::Rewrite;
+
+/// Remove a [`crate::ops::LoadConstant`] node with no outputs.
```
acl-cqc, line 15: nit: "with no outputs" - it has an outport; the condition is that there is nothing connected i.e. consuming it (right??). "No successors" or "no consumers"?

src/hugr/rewrite/consts.rs
```diff
@@ -0,0 +1,231 @@
...
+#[derive(Debug, Clone, Error, PartialEq, Eq)]
+pub enum RemoveConstIgnoreError {
+    /// Invalid node.
+    #[error("Node is invalid (either not in HUGR or not LoadConst).")]
+    InvalidNode(Node),
+    /// Node in use.
+    #[error("Node: {0:?} has non-zero outgoing connections.")]
+    ValueUsed(Node),
+    /// Not connected to a Const.
+    #[error("Node: {0:?} is not connected to a Const node.")]
```
acl-cqc, line 29: nit: this does raise the question of, if this has happened, what the heck are you supposed to do with it ;-). I guess you have to remind to a state where you can avoid getting into that mess. Consider defining `ApplyResult = Option<Node>` and then `return None` in such cases?

src/hugr/rewrite/consts.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    fn verify(&self, h: &impl HugrView) -> Result<(), Self::Error> {
+        let node = self.0;
+
+        if (!h.contains_node(node)) || (!h.get_optype(node).is_load_constant()) {
+            return Err(RemoveConstIgnoreError::InvalidNode(node));
+        }
+
+        if h.out_value_types(node)
+            .next()
```
acl-cqc, line 55: How about `h.out_value_types(node).exactly_one().unwrap()`? Aren't we allowed to assume the input Hugr validates?

acl-cqc: Or, just use `h.output_neighbours().next().is_some()` ?

src/hugr/rewrite/consts.rs
```diff
@@ -0,0 +1,231 @@
...
+            return Err(RemoveConstError::ValueUsed(node));
+        }
+
+        Ok(())
+    }
+
+    fn apply(self, h: &mut impl HugrMut) -> Result<Self::ApplyResult, Self::Error> {
+        self.verify(h)?;
+        let node = self.0;
+        let source = h
```
acl-cqc, line 126: rename to `parent`? Or something, but I don't see how it`s `source`

src/hugr/rewrite/consts.rs
```diff
@@ -0,0 +1,231 @@
...
+        let tup = dfg_build.add_dataflow_op(
+            LeafOp::MakeTuple {
+                tys: type_row![USIZE_T, USIZE_T],
+            },
+            [load_1, load_2],
+        )?;
+        dfg_build.finish_sub_container()?;
+
+        let mut h = build.finish_prelude_hugr()?;
+        assert_eq!(h.node_count(), 8);
```
acl-cqc, line 172: You could change this to a Function-rooted Hugr to reduce counts by 1. Would be nice to comment why there are so many (Module, Function, Input, Output, Const, LoadConstant*2, Tuple)...

src/hugr/rewrite/consts.rs
```diff
@@ -0,0 +1,231 @@
...
+            prelude::{ConstUsize, USIZE_T},
+            PRELUDE_REGISTRY,
+        },
+        hugr::HugrMut,
+        ops::{handle::NodeHandle, LeafOp},
+        type_row,
+        types::FunctionType,
+    };
+    #[test]
+    fn test_const_remove() -> Result<(), Box<dyn std::error::Error>> {
```
acl-cqc, line 155: Great, thorough, test :)

acl-cqc (approved):
A few nits/minor suggestions (the renaming might want to be another PR involving the spec as well), and I've only looked at the bits from the original PR *before* #758 was merged in, but good stuff, thanks. (Sorry for slow review!)

src/hugr/rewrite/consts.rs
acl-cqc: Don't we necessarily fail validation if this is the case? It is too much to ask every rewrite to handle and report every error that would be detected by validate().

src/hugr/rewrite/consts.rs
ss2165: I believe things connected by order edges would should up there as well and I want to make sure I get the value edge.

src/hugr/rewrite/consts.rs
ss2165: Trying to test that non-local const->load edges work ok
Will add comment

src/hugr/rewrite/consts.rs
ss2165: made issue #767

src/hugr/rewrite/consts.rs
ss2165: yes I removed this error case

PR #758, opened 2023-12-22
feat: constant folding implemented for core and float extension

### LINKED ISSUE #711: Constant folding pass
Some kind of way of constant folding over extension constants.

ss2165:
Closes #711 
CI will pass after updating MSRV to 1.75 (from end of year)

based on #757


src/std_extensions/arithmetic/int_ops.rs
```diff
@@ -11,6 +11,7 @@ use crate::ops::OpName;
 use crate::type_row;
 use crate::types::{FunctionType, PolyFuncType};
 use crate::utils::collect_array;
+
```
lmondada, line 14: Out of curiosity: are you following some style guide regarding newlines here? Would like to know for my own coding if there is a recommended style.

src/std_extensions/arithmetic/float_ops/fold.rs
```diff
@@ -0,0 +1,124 @@
+use crate::{
```
lmondada, line 1: Consider naming this file `const_fold.rs` to match the extension name?

src/extension/op_def.rs
```diff
@@ -307,6 +308,9 @@ pub struct OpDef {
     // can only treat them as opaque/black-box ops.
     #[serde(flatten)]
     lower_funcs: Vec<LowerFunc>,
+
+    #[serde(skip)]
+    constant_folder: Box<dyn ConstFold>,
```
lmondada, line 313: Can you add a short docstring here?

src/algorithm/const_fold.rs
```diff
@@ -0,0 +1,302 @@
...
+
+            if let Value::Tuple { vs } = c.value() {
+                if let TypeEnum::Tuple(tys) = c.const_type().as_type_enum() {
+                    return out_row(tys.iter().zip(vs.iter()).map(|(t, v)| {
+                        Const::new(v.clone(), t.clone())
+                            .expect("types should already have been checked")
+                    }));
+                }
+            }
+            None // could panic
```
lmondada, line 66: I am not sure the comment is correct, i.e. what would `panic` and how returning None would help. I suppose you mean that the `let if` clauses might fail?

src/algorithm/const_fold.rs
```diff
@@ -0,0 +1,302 @@
...
+            break;
+        }
+        for (replace, removes) in rewrites {
+            h.apply_rewrite(replace).unwrap();
+            for rem in removes {
+                if let Ok(const_node) = h.apply_rewrite(rem) {
+                    // if the LoadConst was removed, try removing the Const too.
+                    if h.apply_rewrite(RemoveConst(const_node)).is_err() {
+                        // const cannot be removed - no problem
+                        continue;
```
lmondada, line 208: I find it surprising that this is not removed by the rewrite itself. Maybe add a comment about this in the docstring of `find_consts`?

lmondada (approved):
Thanks Seyon, this will be super useful!

src/algorithm/const_fold.rs
ss2165: i mean that `None` shouldn't ever be returned here, so it might be clearer to panic

PR #761, opened 2023-12-28
chore!: hike MSRV to 1.75

lmondada:


lmondada:
Looks like `dtolnay/rust-toolchain` does not support 1.75 yet.

aborgna-q (approved):
:tada: 

This should be running now. (The rust-toolchain repo only gets new commits for unplanned patch releases)

Issue #743, opened 2023-12-15
Remove single source requirement for DSGs


ss2165:
from @cqc-alec :

> The spec currently states that "In a data dependency subgraph, a valid ordering of operations can be achieved by topologically sorting the nodes starting from Input with respect to the Value and Order edges." I vote we drop this requirement, since it places unnecessary constraints on execution order (there's no reason why a LoadConstant operation shouldn't be executed before knowing the inputs to a DFG). If efficiency of topsort is a big concern we can find other ways to optimize it.

cqc-alec:
I believe this was fixed in the spec by https://github.com/CQCL/hugr/pull/762 , but the implementation still needs updating (to not add the unnecessary Order edges).

acl-cqc:
Do we actually have any implementation code adding such Order edges? I check e.g. https://github.com/CQCL/hugr/blob/968c8b03cb35478a05474d2e4f7d27f3c5f3adc5/src/builder/build_traits.rs#L73-L77 and https://github.com/CQCL/hugr/blob/968c8b03cb35478a05474d2e4f7d27f3c5f3adc5/src/builder/build_traits.rs#L352-L370 in the builder, and don't see anything; and we haven't implemented InsertConst(Ignore) rewrites...

cqc-alec:
I may be wrong about that, but my assumption was based on comments @peter-campora made a few weeks ago where he was surprised by the presence of some Order edges.

cqc-alec:
I can't find anywhere in the code where we do this either.

cqc-alec:
OK! So can we close this issue?

ss2165:
Looks like it, the validation code does not depend on it either https://github.com/CQCL/hugr/blob/0c910cf50ef5ef09065a47d39fff0d7d41f2b4c9/src/hugr/validate.rs#L376-L377

PR #762, opened 2024-01-02
docs: [spec] Remove references to causal cone and Order edges from Input

acl-cqc:
I believe the question of causal cones was definitively resolved by #468, and the codebase seems clear on that point - however, I just stumbled across this I-believe-erroneous reference in the spec. (FWIW, searching through the spec for "causal" has only one other occurrence, which appears to explicitly say we don't need this rule.)

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -267,10 +267,7 @@ the following basic dataflow operations are available (in addition to the
 
   - `Input/Output`: input/output nodes, the outputs of `Input` node are
     the inputs to the function, and the inputs to `Output` are the
-    outputs of the function. In a data dependency subgraph, a valid
-    ordering of operations can be achieved by topologically sorting the
-    nodes starting from `Input` with respect to the Value and Order
-    edges.
+    outputs of the function.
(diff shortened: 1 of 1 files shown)


cqc-alec (commented):
This relates to #743 .

For self-consistency there are a couple of other sections of text that should be changed at the same time:

* Just above here, the statement "In a data dependency subgraph, a valid ordering of operations can be achieved by topologically sorting the nodes starting from Input with respect to the Value and Order edges." is no longer necessarily true and should be removed.
* In the description of `InsertConstIgnore`, the instruction to "Also add an Order edge from the Input node under P to n." should be removed.

(I think the implementation also needs updating, but that can be done in a separate PR.)

acl-cqc:
Good spots, thanks @cqc-alec. I wanted to rejig the statement about topsort to refer to DSGs but realized it was waaay before we'd even defined that....so moved it and have tidied up a few more. Thankfully I don't think we have actually implemented `InsertConstIgnore` in code yet!

cqc-alec (approved):

PR #768, opened 2024-01-03
refactor: simplify removeconstignore verify

ss2165:
leftover review comment:
https://github.com/CQCL/hugr/pull/757#discussion_r1439399609


acl-cqc (approved):
Yeah ok. It's not the simplification that I thought but is more defensive and looks good to me, ta.

PR #772, opened 2024-01-03
refactor!: unwrap BasicBlock enum

ss2165:
BREAKING_CHANGES: DFB and Exit are now standalone structs. A level of nesting in serialization is also removed for these operations.

### DIFF
--- a/src/ops/controlflow.rs
+++ b/src/ops/controlflow.rs
@@ -224,7 +224,7 @@ impl OpTrait for Exit {
 
 /// Functionality shared by DFB and Exit CFG block types.
 pub trait BasicBlock {
-    /// The input signature of the contained dataflow graph.
+    /// The input dataflow signature of the CFG block.
     fn dataflow_input(&self) -> &TypeRow;
 }
 
--- a/src/ops/controlflow.rs
+++ b/src/ops/controlflow.rs
@@ -115,70 +115,60 @@ impl DataflowOpTrait for CFG {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
-#[serde(tag = "block")]
-/// Basic block ops - nodes valid in control flow graphs.
+/// A CFG basic block node. The signature is that of the internal Dataflow graph.
 #[allow(missing_docs)]
-pub enum BasicBlock {
-    /// A CFG basic block node. The signature is that of the internal Dataflow graph.
-    DFB {
-        inputs: TypeRow,
-        other_outputs: TypeRow,
-        tuple_sum_rows: Vec<TypeRow>,
-        extension_delta: ExtensionSet,
-    },
-    /// The single exit node of the CFG, has no children,
-    /// stores the types of the CFG node output.
-    Exit { cfg_outputs: TypeRow },
+pub struct DFB {
+    pub inputs: TypeRow,
+    pub other_outputs: TypeRow,
+    pub tuple_sum_rows: Vec<TypeRow>,
+    pub extension_delta: ExtensionSet,
 }
 
-impl OpName for BasicBlock {
-    /// The name of the operation.
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+/// The single exit node of the CFG, has no children,
+/// stores the types of the CFG node output.
+pub struct Exit {
+    /// Output type row of the CFG.
+    pub cfg_outputs: TypeRow,
+}
+
+impl OpName for DFB {
     fn name(&self) -> SmolStr {
-        match self {
-            BasicBlock::DFB { .. } => "DFB".into(),
-            BasicBlock::Exit { .. } => "Exit".into(),
-        }
+        "DFB".into()
     }
 }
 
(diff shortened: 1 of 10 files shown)


src/ops/validate.rs
```diff
@@ -309,20 +305,26 @@ impl ValidateOp for BasicBlock {
...
+    }
+}
+
+impl ValidateOp for Exit {
+    /// Returns the set of allowed parent operation types.
+    fn validity_flags(&self) -> OpValidityFlags {
+        Default::default()
+    }
+
+    /// Validate the ordered list of children.
```
acl-cqc, line 322: This is the default impl, so can skip

src/ops/validate.rs
```diff
@@ -309,20 +305,26 @@ impl ValidateOp for BasicBlock {
...
+        let node_outputs: TypeRow = [&[tuple_sum_type], self.other_outputs.as_ref()]
+            .concat()
+            .into();
+        validate_io_nodes(&self.inputs, &node_outputs, "basic block graph", children)
+    }
+}
+
+impl ValidateOp for Exit {
+    /// Returns the set of allowed parent operation types.
+    fn validity_flags(&self) -> OpValidityFlags {
```
acl-cqc, line 318: This is *also* the default impl, so can skip

acl-cqc:
Ok, so the cunning followup which might make this `DataflowParent` trait (currently unused) really fly, is to provide an
```
impl<T: DataflowParent> ValidateOp for T {...}
```

noting that all impls (for such T) have the same `validity_flags` and I *think* that the differences in `validate_op_children` are captured exacttly by `inner_signature`....

acl-cqc:
My other thought is nomenclature. How about `DFB` -> `BasicBlock` and `Exit` -> `ExitBlock`? (DFB is very similar to DFG, I admit, but the one character that's different doesn't quite seem right)

ss2165:
Yeah those new names seem a bit nicer - but that's definitely a broader spec-inclusive issue

src/ops/controlflow.rs
```diff
@@ -190,43 +180,73 @@ impl OpTrait for BasicBlock {
     }
 
     fn dataflow_signature(&self) -> Option<FunctionType> {
-        Some(match self {
-            BasicBlock::DFB {
-                extension_delta, ..
-            } => FunctionType::new(type_row![], type_row![]).with_extension_delta(extension_delta),
-            BasicBlock::Exit { .. } => FunctionType::new(type_row![], type_row![]),
-        })
+        Some(
```
acl-cqc, line 183: Hmmm, I highly suspect this should be `None` (but instead we should implement `fn extension_delta` which by default delegates to this). However, that's not really part of this PR

src/ops/validate.rs
```diff
@@ -345,7 +263,7 @@ impl ValidateOp for super::Case {
         children: impl DoubleEndedIterator<Item = (NodeIndex, &'a OpType)>,
     ) -> Result<(), ChildrenValidationError> {
         let sig = self.inner_signature();
-        validate_io_nodes(&sig.input, &sig.output, "Conditional", children)
+        validate_io_nodes(&sig.input, &sig.output, "DataflowParent", children)
```
acl-cqc, line 266: I guess there is no handy way to get a name? "DataflowParent" seems fine if nothing springs to mind

ss2165: we can add a dependency on the `OpName` trait and use that but this function
(and everything subsequently including errors) expects static strings so its
just a big annoying change to make.

ss2165: have gone with `std::any::type_name` instead

src/ops/controlflow.rs
```diff
@@ -190,43 +180,73 @@ impl OpTrait for BasicBlock {
...
+        match dir {
+            Direction::Incoming => 1,
+            Direction::Outgoing => 0,
         }
     }
+}
 
+/// Functionality shared by DFB and Exit CFG block types.
+pub trait BasicBlock {
+    /// The input signature of the contained dataflow graph.
```
acl-cqc, line 227: The dataflow signature of the values input to the block? (Exit blocks do not have a "contained dataflow graph")

acl-cqc:
Ok, so - on the naming front: the spec says DFB and Exit are the two types of Basic Block; this PR keeps DFB and Exit but largely drops the notion of "Basic Block"! So I think it would be fine to do some name changes in the code (with appropriate references to the spec - `struct DataflowBlock` is (in the spec) the dataflow subtype of Basic Block, and `struct ExitBlock` is (in the spec) the exit subtype of Basic Block).

But, generally, I'm happy to fold this into #754 and then follow up with a renaming PR. (The *ideal* would probably be to unwrap basicblock and rename in a first PR and then do #754 after, but I'm not going to make you do that!)

acl-cqc:
(I think we should not put #754 in without this, because of the rather dubious inner-signature for exit blocks)

src/ops/validate.rs
acl-cqc: using `T`! Nice :-)

ss2165:
I've taken your suggestion and reordered the PRs, this one is being closed in favour of #781

PR #782, opened 2024-01-04
feat: DataflowParent trait for getting inner signatures

ss2165:
use trait to common up validation and `get_function_type`
Introduces new OpType level `OpParent` trait to optionally retrieve inner function type.

### DIFF
--- a/src/hugr/views.rs
+++ b/src/hugr/views.rs
@@ -24,10 +24,10 @@ use portgraph::{multiportgraph, LinkView, MultiPortGraph, PortView};
 
 use super::{Hugr, HugrError, NodeMetadata, NodeMetadataMap, NodeType, DEFAULT_NODETYPE};
 use crate::ops::handle::NodeHandle;
-use crate::ops::{FuncDecl, FuncDefn, OpName, OpTag, OpTrait, OpType, DFG};
+use crate::ops::{OpName, OpParent, OpTag, OpTrait, OpType};
 
 use crate::types::Type;
-use crate::types::{EdgeKind, FunctionType, PolyFuncType};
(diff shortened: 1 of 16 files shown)


src/hugr/views/descendants.rs
```diff
@@ -261,12 +261,12 @@ pub(super) mod test {
 
         assert_eq!(
             region.get_function_type(),
-            Some(FunctionType::new(type_row![NAT, QB], type_row![NAT, QB]).into())
+            Some(FunctionType::new(type_row![NAT, QB], type_row![NAT, QB]))
```
acl-cqc, line 264:
```suggestion
            Some(FunctionType::new_endo(type_row![NAT, QB]))
```

src/hugr/views.rs
```diff
@@ -333,14 +333,9 @@ pub trait HugrView: sealed::HugrInternals {
 
     /// For function-like HUGRs (DFG, FuncDefn, FuncDecl), report the function
     /// type. Otherwise return None.
-    fn get_function_type(&self) -> Option<PolyFuncType> {
+    fn get_function_type(&self) -> Option<FunctionType> {
```
acl-cqc, line 336: In the docstring - I prefer the wording that you've used in a couple of places, "signature of the inner dataflow (sibling) graph" (or "inner function type", but I think the wording with "signature" is good?)

acl-cqc (approved):
Yes, great, thanks :). `OpParent` and `DataflowParent` make sense.

I guess a future nice-to-have might be to tie the `trait DataflowParent` together with `OpTag::DataflowParent` in some way, but I'm not really sure how, so looks good to me.

(A couple of very minor suggestions.)

PR #783, opened 2024-01-05
fix!: Normalize input/output value/static/other ports in `OpType`

### LINKED ISSUE #777: Inconsistent `OpTrait::other_{input,output}` and `OpType::static_{i,o}_port` behaviour.
The `Const` operation defines its `OpTrait::other_output` as the `EdgeKind::Static` port with the const data.
https://github.com/CQCL/hugr/blob/968c8b03cb35478a05474d2e4f7d27f3c5f3adc5/src/ops/constant.rs#L108-L110

This port is then exposed by `OpType::other_output_port`.

On the other hand, `OpType::static_output_port` uses different logic to define the static output for `Const`;
https://github.com/CQCL/hugr/blob/968c8b03cb35478a05474d2e4f7d27f3c5f3adc5/src/ops.rs#L196-L201

This same situation happens with `FuncDecl` and `FuncDef`.

`static_output_port` is rarely used, so we haven't seen problems yet, but it makes the logic inconsistent.

---

If we compare the corresponding methods on the incoming side; we have `Call` and `LoadConstant` with hard-coded static inputs:
https://github.com/CQCL/hugr/blob/968c8b03cb35478a05474d2e4f7d27f3c5f3adc5/src/ops.rs#L188-L194

Both use `DataflowOpTrait`'s default `other_port`; a `StateOrder`.
https://github.com/CQCL/hugr/blob/968c8b03cb35478a05474d2e4f7d27f3c5f3adc5/src/ops/dataflow.rs#L20-L22

`static_input_port` has multiple references in the code, where it gets used as a separate thing from `other_input_port`.

---

One would expect the same behaviour between input and output, so I consider this inconsistency a bug.
I'd propose changing the `output_port`s to mimic the input behaviour.

### LINKED ISSUE #778: Incorrect `Call` input ports in docs.
The `Call` operation is defined as:
https://github.com/CQCL/hugr/blob/968c8b03cb35478a05474d2e4f7d27f3c5f3adc5/src/ops/dataflow.rs#L133-L137

However, it's dataflow input is just the called function's signature
https://github.com/CQCL/hugr/blob/968c8b03cb35478a05474d2e4f7d27f3c5f3adc5/src/ops/dataflow.rs#L152-L154

And only after those ports comes the `static_port` with the called function (hard-coded in `OpType::other_static_input`.
https://github.com/CQCL/hugr/blob/968c8b03cb35478a05474d2e4f7d27f3c5f3adc5/src/ops.rs#L187-L194

### LINKED ISSUE #779: Serialisation roundtrip breaks `LoadConst`
When running this simple test case that encodes/decodes a simple load const operation,
```rust
    #[test]
    fn constants() -> Result<(), Box<dyn std::error::Error>> {
        let mut builder = DFGBuilder::new(FunctionType::new(vec![], vec![FLOAT64_TYPE])).unwrap();
        let w = builder.add_load_const(ConstF64::new(0.5))?;
        let hugr = builder.finish_hugr_with_outputs([w], &FLOAT_OPS_REGISTRY)?;

        let ser = serde_json::to_string(&hugr)?;
        let deser = serde_json::from_str(&ser)?;
        assert_eq!(hugr, deser);
        Ok(())
    }
```
the extracted edge between the const definition and load gets an invalid target.

See the dot render:

| Original | Decoded |
| ------------- | ------------- |
| image | image |

This seems to be due to some inconsistent logic in the encoder/decoder, related to #777.

aborgna-q:
Closes #777.

Now `OpTrait` clearly separates `other_port` from `static_port`, and `OpType` has a uniform interface for `value` ports (from the signature), `static` ports (contstants), and `other` ports.

Fixes #779. The bug was caused by the encoder ignoring the constant input port offset, so the decoder reconnected the edge to the `StateOrder` port instead. Now we use the proper OpType methods.

Fixes #778.

### DIFF
--- a/src/ops.rs
+++ b/src/ops.rs
@@ -139,8 +139,8 @@ impl OpType {
     /// Returns the edge kind for the given port.
     ///
     /// The result may be a value port, a static port, or a non-dataflow port.
-    /// See [`OpType::value_port_kind`], [`OpType::static_port_kind`], and
-    /// [`OpType::dataflow_signature`].
+    /// See [`OpType::dataflow_signature`], [`OpType::static_port_kind`], and
+    /// [`OpType::other_port_kind`].
     pub fn port_kind(&self, port: impl Into<Port>) -> Option<EdgeKind> {
         let signature = self.dataflow_signature().unwrap_or_default();
         let port: Port = port.into();
--- a/src/ops.rs
+++ b/src/ops.rs
@@ -10,7 +10,7 @@ pub mod module;
 pub mod tag;
 pub mod validate;
 use crate::extension::ExtensionSet;
-use crate::types::{EdgeKind, FunctionType, Type};
+use crate::types::{EdgeKind, FunctionType};
 use crate::{Direction, OutgoingPort, Port};
 use crate::{IncomingPort, PortIndex};
 use paste::paste;
@@ -109,11 +109,12 @@ impl Default for OpType {
 }
 
 impl OpType {
-    /// The edge kind for the non-dataflow or constant ports of the
-    /// operation, not described by the signature.
+    /// The edge kind for the non-dataflow ports of the operation, not described
+    /// by the signature.
     ///
-    /// If not None, a single extra multiport of that kind will be present on
-    /// the given direction.
+    /// If not None, a single extra port of that kind will be present on
+    /// the given direction after any dataflow or constant ports.
+    #[inline]
     pub fn other_port_kind(&self, dir: Direction) -> Option<EdgeKind> {
         match dir {
             Direction::Incoming => self.other_input(),
@@ -121,21 +122,46 @@ impl OpType {
         }
     }
 
+    /// The edge kind for the constant ports of the operation, not described by
+    /// the dataflow signature.
+    ///
+    /// If not None, an extra input port of that kind will be present on the
+    /// given direction after any dataflow ports and before any
+    /// [`OpType::other_port_kind`] ports.
+    #[inline]
+    pub fn static_port_kind(&self, dir: Direction) -> Option<EdgeKind> {
+        match dir {
+            Direction::Incoming => self.static_input(),
+            Direction::Outgoing => self.static_output(),
+        }
+    }
+
     /// Returns the edge kind for the given port.
+    ///
+    /// The result may be a value port, a static port, or a non-dataflow port.
+    /// See [`OpType::value_port_kind`], [`OpType::static_port_kind`], and
+    /// [`OpType::dataflow_signature`].
     pub fn port_kind(&self, port: impl Into<Port>) -> Option<EdgeKind> {
         let signature = self.dataflow_signature().unwrap_or_default();
         let port: Port = port.into();
-        let port_as_in = port.as_incoming().ok();
         let dir = port.direction();
-
         let port_count = signature.port_count(dir);
+
+        // Dataflow ports
         if port.index() < port_count {
-            signature.port_type(port).cloned().map(EdgeKind::Value)
-        } else if port_as_in.is_some() && port_as_in == self.static_input_port() {
-            Some(EdgeKind::Static(static_in_type(self)))
-        } else {
-            self.other_port_kind(dir)
+            return signature.port_type(port).cloned().map(EdgeKind::Value);
         }
+
+        // Constant port
+        let static_kind = self.static_port_kind(dir);
+        if port.index() == port_count {
+            if let Some(kind) = static_kind {
+                return Some(kind);
+            }
+        }
+
+        // Non-dataflow ports
+        self.other_port_kind(dir)
     }
 
     /// The non-dataflow port for the operation, not described by the signature.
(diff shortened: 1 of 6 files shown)


src/ops/dataflow.rs
```diff
@@ -25,9 +26,30 @@ pub(crate) trait DataflowOpTrait {
...
+    #[inline]
+    fn static_input(&self) -> Option<EdgeKind> {
+        None
+    }
+
+    /// The edge kind for a single constant output of the operation, not
+    /// described by the dataflow signature.
+    ///
+    /// If not None, an extra output port of that kind will be present after the
+    /// dataflow input ports and before any [`DataflowOpTrait::other_output`] ports.
```
ss2165, line 48:
```suggestion
    /// dataflow output ports and before any [`DataflowOpTrait::other_output`] ports.
```

ss2165 (commented):
I personally think static_input and static_output don't belong in `OpTrait`. They only ever apply to a few specific ops, and therefore only need to be implemented there. Perhaps two new traits `StaticInput` (implemented by `LoadConst` and `Call`) and `StaticOutput` (implemented by `FuncDefn`, `FuncDecl` and `Const`)? 

This will induce some more pattern matching on these operations (or just their tags) in top-level op interfaces and I actually think that adds clarity (but probably more code admittedly). As an alternative, the PR #782 introduces such a trait then uses a blanket implementation at the top level.

aborgna-q:
Pattern matching based on a tag wouldn't work if the trait is only implemented on some ops, we'd still have to hard-code a pattern match on the operations in `ops.rs` (that's something I was trying to avoid).

#782 could be an alternative, though since it needs to be implemented everywhere I'm not sure what we gain vs the default `OpTrait` method?

ss2165:
> we'd still have to hard-code a pattern match on the operations in ops.rs (that's something I was trying to avoid

if that's the only place that's done, is that such a bad thing? 
In general I'd like to move towards a pattern where we cast ops to what we expect them to be (`as_X`, `is_X`) and then use concrete methods on those that always return (rather than option) as it makes what the code is doing a lot more obvious, rather than chase a trail of trait implementations.

src/ops/dataflow.rs
```diff
@@ -25,9 +26,30 @@ pub(crate) trait DataflowOpTrait {
...
+
+    /// The edge kind for a single constant output of the operation, not
+    /// described by the dataflow signature.
+    ///
+    /// If not None, an extra output port of that kind will be present after the
+    /// dataflow output ports and before any [`DataflowOpTrait::other_output`] ports.
+    #[inline]
+    fn static_output(&self) -> Option<EdgeKind> {
+        None
+    }
```
ss2165, line 52: no dataflow operations have a static output right, so this method doesn't need to exist

aborgna-q:
The problem is that this kind of operations are also needed to generically list all nodes of a node. Otherwise the static ports are hidden, not in the `dataflow_signature` nor in the `other_port`.

Listing the different ports of a node then requires the user to be aware of this, and maybe pattern matching on the specific ops...

Explicitly saying "these are the three varieties of ports you may encounter, here are the accessors" is a lot more straightforward.

ss2165:
> Explicitly saying "these are the three varieties of ports you may encounter, here are the accessors" is a lot more straightforward.

I'm convinced!

src/ops.rs
```diff
@@ -109,33 +109,59 @@ impl Default for OpType {
...
+    /// the given direction after any dataflow or constant ports.
+    #[inline]
     pub fn other_port_kind(&self, dir: Direction) -> Option<EdgeKind> {
         match dir {
             Direction::Incoming => self.other_input(),
             Direction::Outgoing => self.other_output(),
         }
     }
 
+    /// The edge kind for the constant ports of the operation, not described by
```
ss2165, line 125:
```suggestion
    /// The edge kind for the static ports of the operation, not described by
```

src/ops.rs
```diff
@@ -155,84 +181,92 @@ impl OpType {
...
 
+    /// The non-dataflow input port for the operation, not described by the signature.
+    /// See `[OpType::other_port]`.
+    #[inline]
+    pub fn other_input_port(&self) -> Option<IncomingPort> {
+        self.other_port(Direction::Incoming)
+            .map(|p| p.as_incoming().unwrap())
+    }
+
+    /// The non-dataflow input port for the operation, not described by the signature.
```
ss2165, line 192:
```suggestion
    /// The non-dataflow output port for the operation, not described by the signature.
```

src/ops.rs
```diff
@@ -155,84 +181,92 @@ impl OpType {
...
-    pub fn static_output_port(&self) -> Option<OutgoingPort> {
-        OpTag::StaticOutput
-            .is_superset(self.tag())
-            .then_some(0.into())
-    }
-
     /// Returns the number of ports for the given direction.
+    #[inline]
     pub fn port_count(&self, dir: Direction) -> usize {
+        let static_input = self.static_port_kind(dir).is_some() as usize;
```
ss2165, line 246: maybe rename to something like `has_static_port`

src/ops/dataflow.rs
```diff
@@ -161,6 +182,21 @@ impl Call {
...
+    /// # use hugr::extension::prelude::QB_T;
+    /// let signature = FunctionType::new(vec![QB_T, QB_T], vec![QB_T, QB_T]);
+    /// let call = Call { signature };
+    /// let op = OpType::Call(call.clone());
+    /// assert_eq!(op.static_input_port(), Some(call.called_function_port()));
+    /// ```
+    ///
+    /// [`OpType::static_input_port`]: crate::ops::OpType::static_input_port
     #[inline]
     pub fn called_function_port(&self) -> IncomingPort {
```
ss2165, line 201: do we still need this method (or the matching one for Const for that matter)?

ss2165 (changes requested):

src/ops/dataflow.rs
aborgna-q: They are needed if we use the specific `Call`/`LoadConstant` ops directly, as `static_port` is only implemented for `OpType`.

We _could_ add that method to `DataflowOpTrait` too?
Though here I'd say that your comment applies; in that if you're at this level you are dealing with specific operations.
(Alternatively we can always `Into<OpType>` and call those methods)

PR #786, opened 2024-01-05
refactor: Put extension inference behind a feature gate

### LINKED ISSUE #784: feature flag for extension inference and validation
Extension inference and validation are not complete - there is work yet to do to be able to infer lift nodes (if we need them at all) and to avoid annotations in DFG etc. see #426. For release 0.1 just put the inference and validation behind an experimental feature flag (and record that in the readme - that it doesn't work). CI need not test with this flag on. Existing tests should pass, and  new tests can be written more simply that won't pass if the flag is turned on.

croyzor:
Add a feature flag "extension_inference", without which `Hugr::infer_extensions` is replaced by a dummy method, and `ValidationContext` doesn't use a `ExtensionValidator`, and tests that require extension inference to be done are disabled.
Our CI runs tests and benchmarks once with default features, then again with all features, so extension inference is still being tested by CI. The workflow has only been altered to tell `clippy` to run with all features as well.

Resolves #784

### DIFF
--- a/src/extension/infer/test.rs
+++ b/src/extension/infer/test.rs
@@ -1,6 +1,7 @@
 use std::error::Error;
 
 use super::*;
+#[cfg(feature = "extension_inference")]
 use crate::builder::test::closed_dfg_root_hugr;
 use crate::builder::{
     Container, DFGBuilder, Dataflow, DataflowHugr, DataflowSubContainer, HugrBuilder, ModuleBuilder,
@@ -8,10 +9,14 @@ use crate::builder::{
 use crate::extension::prelude::QB_T;
 use crate::extension::ExtensionId;
 use crate::extension::{prelude::PRELUDE_REGISTRY, ExtensionSet};
-use crate::hugr::{validate::ValidationError, Hugr, HugrMut, HugrView, NodeType};
+#[cfg(feature = "extension_inference")]
+use crate::hugr::validate::ValidationError;
+use crate::hugr::{Hugr, HugrMut, HugrView, NodeType};
 use crate::macros::const_extension_ids;
 use crate::ops::custom::{ExternalOp, OpaqueOp};
-use crate::ops::{self, dataflow::IOTrait, handle::NodeHandle};
+#[cfg(feature = "extension_inference")]
+use crate::ops::handle::NodeHandle;
+use crate::ops::{self, dataflow::IOTrait};
 use crate::ops::{LeafOp, OpType};
 
 use crate::type_row;
@@ -153,6 +158,7 @@ fn plus() -> Result<(), InferExtensionError> {
     Ok(())
 }
 
+#[cfg(feature = "extension_inference")]
 #[test]
 // This generates a solution that causes validation to fail
 // because of a missing lift node
@@ -214,6 +220,7 @@ fn open_variables() -> Result<(), InferExtensionError> {
     Ok(())
 }
 
+#[cfg(feature = "extension_inference")]
 #[test]
 // Infer the extensions on a child node with no inputs
 fn dangling_src() -> Result<(), Box<dyn Error>> {
@@ -305,6 +312,7 @@ fn create_with_io(
     Ok([node, input, output])
 }
 
+#[cfg(feature = "extension_inference")]
 #[test]
 fn test_conditional_inference() -> Result<(), Box<dyn Error>> {
     fn build_case(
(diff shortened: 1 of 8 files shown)


croyzor:
Todo (not necessarily in this PR): Make a feature branch for extension inference; PRs and commits to which run the CI with this feature switched on.

src/hugr/validate/test.rs
```diff
@@ -404,6 +409,7 @@ fn test_ext_edge() -> Result<(), HugrError> {
     Ok(())
 }
 
+#[cfg(feature = "extension_inference")]
```
ss2165, line 412: this function should be usable without inference

ss2165 (changes requested):

src/extension/infer/test.rs
```diff
@@ -214,6 +220,7 @@ fn open_variables() -> Result<(), InferExtensionError> {
     Ok(())
 }
 
+#[cfg(feature = "extension_inference")]
```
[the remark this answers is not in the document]
croyzor: It's not all the tests in the module!

src/hugr/validate/test.rs
```diff
@@ -71,6 +74,7 @@ fn add_df_children(b: &mut Hugr, parent: Node, copies: usize) -> (Node, Node, No
 /// Intended to be used to populate a BasicBlock node in a CFG.
 ///
 /// Returns the node indices of each of the operations.
+#[cfg(feature = "extension_inference")]
```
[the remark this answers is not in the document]
croyzor: Because it's only used by tests which are gated. I could alternatively `allow(dead_code)`?

src/hugr/validate/test.rs
croyzor: Yeah this is another dead code warning gambit

src/hugr.rs
```diff
@@ -345,13 +354,20 @@ pub enum HugrError {
 #[cfg(test)]
 mod test {
     use super::{Hugr, HugrView};
+    #[cfg(feature = "extension_inference")]
```
acl-cqc, line 357: Looks like a call for another
```
#[cfg(feature = "extension_inference")]
use crate::{
  builder::test::closed_dfg_root_hugr,
  extension::ExtensionSet,
  hugr::HugrMut,
....
```
combination here

src/hugr/validate/test.rs
ss2165: it would be better to add a test that uses it (without needing inference)

acl-cqc (commented):
Generally looks good. I am a bit surprised not to see a `#[cfg(feature="extension_inference")]` on `struct ExtensionValidator` in src/extension/validate.rs.

src/hugr/validate.rs
```diff
@@ -72,6 +77,7 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
         Self {
             hugr,
             dominators: HashMap::new(),
+            #[cfg(feature = "extension_inference")]
```
acl-cqc, line 80: Alternatively perhaps leave this ExtensionValidator here but make its methods not do anything?

src/hugr/validate.rs
```diff
@@ -163,6 +169,7 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
 
         // FuncDefns have no resources since they're static nodes, but the
         // functions they define can have any extension delta.
+        #[cfg(feature = "extension_inference")]
```
acl-cqc, line 172: (e.g. here `validate_io_extensions` would just be a no-op)

src/hugr/validate.rs
ss2165: this is nicer 👍

src/hugr/validate/test.rs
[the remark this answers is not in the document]
aborgna-q: You could add a nested
```rust
#[cfg(feature = "extension_inference")]
mod ext_iface_test {
  use super::*;
  ...
}
```

with all the conditional imports and tests.

croyzor:
I'm not sure whether to adjust the CI code coverage to run with "--all-features" or not

src/extension/validate.rs
```diff
@@ -120,33 +120,37 @@ impl ExtensionValidator {
...
-            };
-            if parent_output_extensions != output_extensions {
-                return Err(ExtensionError::ParentIOExtensionMismatch {
-                    parent,
-                    parent_extensions: parent_output_extensions.clone(),
-                    child: output,
-                    child_extensions: output_extensions.clone(),
-                });
-            };
+        #[cfg(feature = "extension_inference")]
```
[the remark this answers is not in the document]
croyzor: Updated! I wasn't sure what stance we were going to take on whether coverage should represent default hugr or include inference by default

ss2165 (approved):

PR #791, opened 2024-01-08
docs: Update spec wrt. polymorphism

### LINKED ISSUE #790: Update spec wrt. Polymorphism
At present the codebase allows FunctionType's to be polymorphic, so there is the possibility of polymorphic lambdas being around as first-class values, and *FuncDefn*s (only) can declare TypeParams (perhaps a Const node containing a Hugr rooted at a FuncDefn could produce a value of such a polymorphic type?).

Although #285 argues for simplifying this (so Consts would only contain functions of monomorphic type, and only Static edges *from FuncDefn nodes* would be able to carry polymorphic types), we haven't done that yet, so that'll be another spec update if we do it...

acl-cqc:
* Move "Operation Extensionsibility" to after Type System (and up one level), rename to "Extension System"
* Add Polymorphism section inside Type System: any `Function` type may be polymorphic
* Clarify working of operations, including adding Appendix 3 with full details binary `compute_signature`
* Add `OpaqueOp` as a dataflow node operation
 
fixes #790

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -226,7 +226,8 @@ operations have value ports, but some have Static or other
 edges. The following operations are *only* valid as immediate children of a
 `Module` node.
 
-  - `FuncDecl`: an external function declaration. The name of the function 
+  - `FuncDecl`: an external function declaration. The name of the function,
+    a list of type parameters (TypeParams, see [Type System](#type-system))
     and function attributes (relevant for compilation)
     define the node weight. The node has an outgoing `Static<Function>`
     edge for each use of the function. The function name is used at link time to
@@ -269,8 +270,13 @@ the following basic dataflow operations are available (in addition to the
     the inputs to the function, and the inputs to `Output` are the
     outputs of the function.
   - `Call`: Call a statically defined function. There is an incoming
-    `Static<Function>` edge to specify the graph being called. The
-    signature of the node (defined by its incoming and outgoing `Value` edges) matches the function being called.
+    `Static<Function>` edge to specify the graph being called. The `Call`
+    node specifies any type arguments to the function in the node weight,
+    and the signature of the node (defined by its incoming and outgoing `Value` edges)
+    matches the (type-instantiated) function being called.
+  - `TypeApply`: has a `Value<Function>` input, whose type is polymorphic (i.e. declares some type parameters);
+    the node specifies some type arguments (matching those parameters) in the node weight;
+    and there is  a `Value<Function>` output (corresponding to the type instantiation of the input)
   - `LoadConstant<T>`: has an incoming `Static<T>` edge, where `T` is a `CopyableType`, and a
     `Value<T>` output, used to load a static constant into the local
     dataflow graph.
@@ -792,17 +798,58 @@ Extensions ::= (Extension)* -- a set, not a list
 
 Type ::= Tuple(#) -- fixed-arity, heterogeneous components 
        | Sum(#)   -- disjoint union of other types, ??tagged by unsigned int??
-       | Function[Extensions](#, #) -- monomorphic
-       | Opaque(Name, TypeArgs) -- a (instantiation of a) custom type defined by an extension
+       | Opaque(Name, [TypeArg]) -- a (instantiation of a) custom type defined by an extension
+       | Function(TypeParams, #, #, Extensions) -- polymorphic with type parameters, function arguments + results
+       | Variable -- refers to a TypeParam bound by the nearest enclosing FuncDefn node, or an enclosing Function Type
 ```
-<!--      Function(TypeParams, #, #, Extensions) -- polymorphic, so move TypeParams section here
-#       | Variable -- refers to a TypeParam bound by an enclosing Graph-->
-
 The majority of types will be Opaque ones defined by extensions including the [standard library](#standard-library). However a number of types can be constructed using only the core type constructors: for example the empty tuple type, aka `unit`, with exactly one instance (so 0 bits of data); the empty sum, with no instances; the empty Function type (taking no arguments and producing no results - `void -> void`); and compositions thereof.
 
 Types representing functions are generally ``CopyableType``, but not ``EqType``. (It is undecidable whether two functions produce the same result for all possible inputs, or similarly whether one computation graph can be rewritten into another by semantic-preserving rewrites).
 
 Tuples and Sums are ``CopyableType`` (respectively, ``EqType``) if all their components are; they are also fixed-size if their components are.
+
+### Polymorphism
+
+`Function` types are polymorphic: they may declare a number of type parameters of kinds as follows:
+
+```haskell
+TypeParam ::= Type(Any|Copyable|Eq)
+            | BoundedUSize(u64|) -- note optional bound
+            | Extensions
+            | List(TypeParam) -- homogenous, any sized
+            | Tuple([TypeParam]) -- heterogenous, fixed size
+            | Opaque(Name, [TypeArg]) -- e.g. Opaque("Array", [5, Opaque("usize", [])])
+```
+(We write `[Foo]` to indicate a list of Foo's.)
+
+For each such TypeParam, the body of the FunctionType (input, output, and extensions - see [Extension Tracking](#extension-tracking))
+may contain "type variables" referring to that TypeParam, i.e. the binder. (The type variable is typically a type, but
+not necessarily, depending upon the TypeParam.)
+
+When the `Function` is `Call`ed, TypeArgs appropriate for the TypeParams must be provided by the Call:
+
+```
+TypeArg ::= Type(Type)
+          | BoundedUSize(u64)
+          | Extensions(Extensions)
+          | Sequence([TypeArg])
+          | Opaque(Value)
+          | Variable -- refers to an enclosing TypeParam (binder)
+```
+
+A `Sequence` argument is appropriate for a parameter of kind either `List` or `Tuple`.
+For example, a Function declaring a `TypeParam::Opaque("Array", [5, TypeArg::Type(Type::Opaque("usize"))])`
+means that any `Call` to it must statically provide a *value* that is an array of 5 `usize`s;
+or a Function declaring a `TypeParam::BoundedUSize(5)` and a `TypeParam::Type(Any)` requires two TypeArgs,
+firstly a non-negative integer less than 5, secondly a type (which might be from an extension, e.g. `usize`).
+
+Given TypeArgs, the body of the `Function` type can be converted to a monomorphic signature by substitution,
+placing the provided TypeArg into any place in the body referring to the corresponding type variable.
+This is infallible as long as the TypeArgs match the declared TypeParams, which can be checked statically.
+
+Note that when within a `Function` type, TypeParams of kind `List`, `Tuple` or `Opaque` can only be used
+to instantiate polymorphic type declarations in extensions; these TypeParams have other uses in defining
+extension ops. See [Operation Extensibility](#operation-extensibility).
 
 ### Extension Tracking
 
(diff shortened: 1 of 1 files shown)


specification/hugr.md
```diff
@@ -269,10 +270,15 @@ the following basic dataflow operations are available (in addition to the
     the inputs to the function, and the inputs to `Output` are the
     outputs of the function.
   - `Call`: Call a statically defined function. There is an incoming
-    `Static<Function>` edge to specify the graph being called. The
-    signature of the node (defined by its incoming and outgoing `Value` edges) matches the function being called.
+    `Static<Function>` edge to specify the graph being called. The `Call`
+    node specifies any type arguments to the function in the node weight,
+    and the signature of the node (defined by its incoming and outgoing `Value` edges)
+    matches the (type-instantiated) function being called.
+  - `TypeApply`: has a `Value<Function>` input, whose type is polymorphic (i.e. declares some type parameters);
```
acl-cqc, line 277: This is likely to go following resolution of #285, but I could put in more detail?

specification/hugr.md
```diff
@@ -712,10 +718,222 @@ flowchart
     extension, taking a graph argument; and `run_circuit` will return the
     same way.
 
+### Extensible metadata
```
acl-cqc, line 721: This is an unhelpful diff - this section has not changed

specification/hugr.md
```diff
@@ -772,22 +991,25 @@ Ultimately though, we cannot avoid the "stringly" type problem if we
...
-in a HUGR will provide for each a static-constant "type argument": a value that in many
-cases will be a type. These type arguments are processed by the type scheme interpreter
-or the `compute_signature` implementation to compute the type of that operation node.
+polymorphic in type. Each OpDef *either*:
+
+1. Provides a `Function` type, as per [Type System](#type-system), which may declare TypeParams; *or*
+2. The extension may self-register binary code (e.g. a Rust trait) providing a function
+   `compute_signature` that fallibly computes a `Function` type given some type arguments.
+   The operation declares the arguments required by the `compute_signature` function as a list
+   of TypeParams (notably including `TypeParam::List`, `Tuple` and `Opaque`);
```
acl-cqc, line 1004: Hmmm, this "notably" implies that they may not be included elsewhere, but such TypeParams can be declared anywhere. Really the point is that the binary `compute_signature` is the (only) code that actually gets to *process* these funky TypeArgs.

specification/hugr.md
```diff
@@ -908,21 +1130,7 @@ extensions:
...
-Opaque ::= string<[TypeArgs]>
-
-TypeArgs ::= Type(Type) | BoundedUSize(u64) | Extensions | List([TypeArg]) | Tuple([TypeArg])
-
-Type ::= Name<[TypeArg]>
-```
-(We write `[Foo]` to indicate a list of Foo's; and omit `<>` where the contents is the empty list).
-
-To use an OpDef as an Op, or a TypeDef as a type, the user must provide a type argument for each type param in the def: a type in the appropriate class, a bounded usize, a set of extensions, a list or tuple of arguments.
+<!-- To use an OpDef as an Op, or a TypeDef as a type, the user must provide a type argument for each type param in the def: a type in the appropriate class, a bounded usize, a set of extensions, a list or tuple of arguments. -->
```
acl-cqc, line 1133: Think this is redundant? Oh, maybe about TypeDef's??

specification/hugr.md
```diff
@@ -712,10 +718,223 @@ flowchart
...
+graph-based control flow into a CFG node that a quantum computer could
+run, which removes the `HigherOrder` extension requirement.
+
+```
+precompute :: Function[](Function[Quantum,HigherOrder](Array(5, Qubit), (ms: Array(5, Qubit), results: Array(5, Bit))),
+                                         Function[Quantum](Array(5, Qubit), (ms: Array(5, Qubit), results: Array(5, Bit))))
+```
 
-#### Goals and constraints
+## Operation Extensibility
```
acl-cqc, line 935: Hmmmm, this (as previously) includes *type* extensibility too

specification/hugr.md
```diff
@@ -269,10 +270,15 @@ the following basic dataflow operations are available (in addition to the
...
   - `Call`: Call a statically defined function. There is an incoming
-    `Static<Function>` edge to specify the graph being called. The
-    signature of the node (defined by its incoming and outgoing `Value` edges) matches the function being called.
+    `Static<Function>` edge to specify the graph being called. The `Call`
+    node specifies any type arguments to the function in the node weight,
+    and the signature of the node (defined by its incoming and outgoing `Value` edges)
+    matches the (type-instantiated) function being called.
+  - `TypeApply`: has a `Value<Function>` input, whose type is polymorphic (i.e. declares some type parameters);
+    the node specifies some type arguments (matching those parameters) in the node weight;
+    and there is  a `Value<Function>` output (corresponding to the type instantiation of the input)
```
ss2165, line 279:
```suggestion
    and there is  a `Value<Function>` output (corresponding to the type instantiation of the input).
```

specification/hugr.md
```diff
@@ -794,18 +800,59 @@ Extensions ::= (Extension)* -- a set, not a list
 
 Type ::= Tuple(#) -- fixed-arity, heterogeneous components 
        | Sum(#)   -- disjoint union of other types, ??tagged by unsigned int??
-       | Function[Extensions](#, #) -- monomorphic
-       | Opaque(Name, TypeArgs) -- a (instantiation of a) custom type defined by an extension
+       | Opaque(Name, [TypeArg]) -- a (instantiation of a) custom type defined by an extension
+       | Function(TypeParams, #, #, Extensions) -- polymorphic with type parameters, function arguments + results
```
ss2165, line 804: and extensions?

specification/hugr.md
```diff
@@ -948,18 +995,21 @@ at runtime. In many cases this is desirable.
 
 To strike a balance then, every extension provides YAML that declares its opaque
```
ss2165, line 996: I don't think we mean just YAML here do we?

specification/hugr.md
ss2165: yes i think it might be worth being clearer about typedef polymorphism

ss2165 (commented):
looks good overall

specification/hugr.md
acl-cqc: Yes, the YAML seems less and less relevant!! Wilco (but keeping the "example" YAML)...

specification/hugr.md
acl-cqc: with mention of "delta"

specification/hugr.md
acl-cqc: Added "or equivalent Rust structs"

specification/hugr.md
acl-cqc: Added quite a bit above in the Extension System section

acl-cqc:
Lord. I thought the merge would be straightforward but forgot that I'd moved a section and git seems to be handling that terribly. Really hoping I have done the merge right!....

specification/hugr.md
```diff
@@ -710,9 +714,227 @@ flowchart
...
+
+```haskell
+TypeParam ::= Type(Any|Copyable|Eq)
+            | BoundedUSize(u64|) -- note optional bound
+            | Extensions
+            | List(TypeParam) -- homogenous, any sized
+            | Tuple([TypeParam]) -- heterogenous, fixed size
+            | Opaque(Name, [TypeArg]) -- e.g. Opaque("Array", [5, Opaque("usize", [])])
+```
+(We write `[Foo]` to indicate a list of Foo's.)
```
acl-cqc, line 822: This belongs with `Type ::=` above, as we have `[TypeArg]` there

specification/hugr.md
```diff
@@ -710,9 +714,227 @@ flowchart
...
+
+For each such TypeParam, the body of the FunctionType (input, output, and extensions - see [Extension Tracking](#extension-tracking))
+may contain "type variables" referring to that TypeParam, i.e. the binder. (The type variable is typically a type, but
+not necessarily, depending upon the TypeParam.)
+
+When the `Function` is `Call`ed, TypeArgs appropriate for the TypeParams must be provided by the Call:
```
acl-cqc, line 828: Hmmm. `Function` types appear on `Static` edges from `Call`s to `FuncDefn`/`FuncDecl`, so this needs rewording

specification/hugr.md
```diff
@@ -768,39 +990,62 @@ Ultimately though, we cannot avoid the "stringly" type problem if we
...
+For example, the TypeDef for `array` in the prelude declares two TypeParams: a `BoundedUSize`
+(the array length) and a `Type`. Any valid instantiation (e.g. `array<5, usize>`) is a type.
+Much the same applies for OpDef's that provide a `Function` type.
+
+However, for OpDefs providing binary `compute_signature`,
+* each node will provide TypeArgs for *both* the binary function *and* the `Function` type that returns
+* the operation is only valid if `compute_signature` succeeds, and returns a `Function` type
+  into whose TypeParams the *remaining* TypeArgs (in the node) fit. Note that this means
+  the binary function may use the values of its TypeArgs---specifically including `List`, `Tuple` and
+  `Opaque`---to determine the structure of the returned `Function` type such that the latter's own TypeParams
```
ss2165, line 1021:
```suggestion
  the binary function may use the values of its TypeArgs--specifically including `List`, `Tuple` and
  `Opaque`--to determine the structure of the returned `Function` type such that the latter's own TypeParams
```

specification/hugr.md
```diff
@@ -710,9 +714,227 @@ flowchart
...
+For example, a Function declaring a `TypeParam::Opaque("Array", [5, TypeArg::Type(Type::Opaque("usize"))])`
+means that any `Call` to it must statically provide a *value* that is an array of 5 `usize`s;
+or a Function declaring a `TypeParam::BoundedUSize(5)` and a `TypeParam::Type(Any)` requires two TypeArgs,
+firstly a non-negative integer less than 5, secondly a type (which might be from an extension, e.g. `usize`).
+
+Given TypeArgs, the body of the `Function` type can be converted to a monomorphic signature by substitution,
+placing the provided TypeArg into any place in the body referring to the corresponding type variable.
+This is infallible as long as the TypeArgs match the declared TypeParams, which can be checked statically.
+
+Note that when within a `Function` type, TypeParams of kind `List`, `Tuple` or `Opaque` can only be used
```
acl-cqc, line 849: Unless we want to allow polymorphic types like
```
forall I : List(Type(Any)), O : List(Type(Any)).
  Function(I -> O) -> ...
```
i.e. allowing List(Type) to be a *row* of types. Not that hard to implement, depending a bit on whether we want to allow row concatenation, etc. - `Function(I -> O) -> Function(I,I, int -> O, O) `, say ???

specification/hugr.md
```diff
@@ -768,39 +990,62 @@ Ultimately though, we cannot avoid the "stringly" type problem if we
...
-2. The extension self-registers binary code (e.g. a Rust trait) providing a function
-   `compute_signature` that computes the type.
-
-Each OpDef may declare named type parameters---if so then the individual operation nodes
-in a HUGR will provide for each a static-constant "type argument": a value that in many
-cases will be a type. These type arguments are processed by the type scheme interpreter
-or the `compute_signature` implementation to compute the type of that operation node.
+### Extension Implementation
+
+To strike a balance then, every extension provides YAML (or equivalent Rust structs)
```
ss2165, line 995: could refer explicitly to the "Declarative Format" section below

specification/hugr.md
ss2165: looks like we can revisit this after 787 then?

specification/hugr.md
acl-cqc: ?? Searching Google for "parenthetic dashes" confirms that these should be em dashes, which is `---` ?

specification/hugr.md
```diff
@@ -768,39 +990,62 @@ Ultimately though, we cannot avoid the "stringly" type problem if we
...
+However, for OpDefs providing binary `compute_signature`,
+* each node will provide TypeArgs for *both* the binary function *and* the `Function` type that returns
+* the operation is only valid if `compute_signature` succeeds, and returns a `Function` type
+  into whose TypeParams the *remaining* TypeArgs (in the node) fit. Note that this means
+  the binary function may use the values of its TypeArgs---specifically including `List`, `Tuple` and
+  `Opaque`---to determine the structure of the returned `Function` type such that the latter's own TypeParams
+  depend upon the values of the TypeArgs passed into the binary.
+* the TypeArgs passed to `compute_signature` may *not* refer to any type variables declared by
+  ancestor nodes in the Hugr (specifically, may *not* use variables declared by the
+  enclosing FuncDefn) i.e. these must be static constants unaffected by substitution.
```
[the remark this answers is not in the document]
acl-cqc: Definitely not sure we need it all here - I think it's good to have it in the spec, but great idea to move it out, so in appendix 3

specification/hugr.md
acl-cqc: I put the note about Rust structs there too.

acl-cqc:
Ok, moved details into appendix, and - hang on, how did we not have this already, did I miss something - added `OpaqueOp` as a dataflow op

ss2165 (approved):
Great!

PR #792, opened 2024-01-08
fix: No dataflow_signature for block types

acl-cqc:
Basic blocks should not have a `dataflow_signature`, they have no dataflow ports! But, for DataflowBlock, implement `extension_delta` instead (and fix replace.rs test correspondingly)

### DIFF
--- a/src/ops/controlflow.rs
+++ b/src/ops/controlflow.rs
@@ -4,7 +4,7 @@ use smol_str::SmolStr;
 
 use crate::extension::ExtensionSet;
 use crate::types::{EdgeKind, FunctionType, Type, TypeRow};
-use crate::{type_row, Direction};
+use crate::Direction;
 
 use super::dataflow::{DataflowOpTrait, DataflowParent};
 use super::OpTag;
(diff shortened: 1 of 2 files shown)


src/ops/controlflow.rs
```diff
@@ -179,10 +179,8 @@ impl OpTrait for DataflowBlock {
         Some(EdgeKind::ControlFlow)
     }
 
-    fn dataflow_signature(&self) -> Option<FunctionType> {
-        Some(
-            FunctionType::new(type_row![], type_row![]).with_extension_delta(&self.extension_delta),
-        )
+    fn extension_delta(&self) -> ExtensionSet {
```
acl-cqc, line 182: The interesting thing here is that the default `extension_delta` did not delegate to `dataflow_signature` but just returned none, which AFAICS meant that extension inference was previously *ignoring* the extension delta for basic blocks: https://github.com/CQCL/hugr/blob/16dce1edb86374fbbe73fae648acdb030bd279c0/src/extension/infer.rs#L320

PR #793, opened 2024-01-08
feat: constant folding for logic extension

### LINKED ISSUE #775: Constant folding for logic extension

ss2165:
Closes #775

### DIFF
--- a/src/std_extensions/logic.rs
+++ b/src/std_extensions/logic.rs
@@ -3,6 +3,7 @@
 use strum_macros::{EnumIter, EnumString, IntoStaticStr};
 
 use crate::{
+    algorithm::const_fold::sorted_consts,
     extension::{
         prelude::BOOL_T,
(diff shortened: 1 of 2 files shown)


src/std_extensions/logic.rs
```diff
@@ -171,6 +187,33 @@ impl MakeRegisteredOp for NotOp {
...
+    let false_val = ops::Const::false_val();
+    let inps: Option<Vec<bool>> = sorted_consts(consts)
+        .into_iter()
+        .map(|c| {
+            if c == &true_val {
+                Some(true)
+            } else if c == &false_val {
+                Some(false)
+            } else {
+                None
```
cqc-alec, line 201: This must be correct (since it compiles), but it looks like it's making a vector of optional bools rather than an optional vector of bools... what am I missing?

cqc-alec (approved):

src/std_extensions/logic.rs
cqc-alec: Ah OK!

PR #795, opened 2024-01-09
feat: constant folding for list operations

ss2165:
+ utility enums and structs for dealing with list ops and types

Closes #774

PR #796, opened 2024-01-09
chore: turn off extension inference testing

ss2165:


.github/workflows/ci.yml
```diff
@@ -102,9 +102,9 @@ jobs:
       - name: Run tests with coverage instrumentation
         run: |
             cargo llvm-cov clean --workspace
-            cargo llvm-cov --all-features --doctests
+            cargo llvm-cov --doctests
       - name: Generate coverage report
-        run: cargo llvm-cov --all-features report --codecov --output-path coverage.json
```
croyzor, line 107: Is deleting this necessary?

.github/workflows/ci.yml
```diff
@@ -84,8 +84,8 @@ jobs:
         run: cargo test --verbose --no-default-features
       - name: Build with all features
         run: cargo test --verbose --all-features --no-run
-      - name: Tests with all features
-        run: cargo test --verbose --all-features
+      # - name: Tests with all features
+      #   run: cargo test --verbose --all-features
```
aborgna-q, line 88: Replace this with a note about why we don't run the tests
```suggestion
      # Note: The `extension_inference` feature is experimental, we don't expect tests to succeed.
```

.github/workflows/ci.yml
```diff
@@ -102,9 +102,9 @@ jobs:
       - name: Run tests with coverage instrumentation
         run: |
             cargo llvm-cov clean --workspace
-            cargo llvm-cov --all-features --doctests
+            cargo llvm-cov --doctests
```
aborgna-q, line 105: This will bring the coverage down by a large margin.
It may be nice to spread
```rust
#![cfg_attr(coverage, feature(no_coverage))]
```
around.

(Note: That'll make `llvm-cov` run only on nightly, but that should be ok).

aborgna-q (commented):
It's a bit annoying having failing tests in `HEAD`.
Perhaps we could conditionally `#[ignore]` them instead? 

We'd then be able to run the tests with 
```sh
cargo test -- --include-ignored
```
when needed.

If we keep this PR as-is, make sure to remove the `--all-features` flags from `.github/pre-commit` too.

ss2165:
Closing in favour of skipping failing inference tests: [ac78815](https://github.com/CQCL/hugr/pull/795/commits/ac78815a55d373f20849bb0cde84a238834ae46f)

PR #798, opened 2024-01-09
chore: Only test all features on branch names containing "inference"

croyzor:
A PR to rival #796


croyzor:
If GH actions weren't broken right now, I'd expect this PR to run with "--all-tests" then I could open another with a title not containing "inference" to compare/test

PR #800, opened 2024-01-09
docs: mention experimental inference in readme

ss2165:


README.md
```diff
@@ -27,6 +27,13 @@ The library crate is called `hugr`.
 
 Please read the [API documentation here][].
 
+## Features
```
aborgna-q, line 30: Perhaps `Experimental Features` instead, to clearly mark this. (Otherwise the _experimental_ label is a bit hidden inside the description).

I wonder if we should do like `clap`, and explicitly name the feature `unstable_extension_inference`;
https://docs.rs/clap/latest/clap/_features/index.html

aborgna-q (approved):

README.md
ss2165: worth bearing in mind if we add any more features i think

PR #807, opened 2024-01-15
fix: include schema rather than read file

ss2165:
otherwise fails miri checks by opening file

running workflow manually to check


aborgna-q (approved):
Nice workaround. The check is still running, but this should work.

PR #827, opened 2024-01-30
feat!: impls of CustomConst must be able to report their type

doug-q:
We remove `trait KnownTypeConst` as it is now redundant. We move the body of `check_known_type` into a new default implementation of `CustomConst::check_custom_type`.

### DIFF
--- a/src/std_extensions/collections.rs
+++ b/src/std_extensions/collections.rs
@@ -1,5 +1,6 @@
 //! List type and operations.
 
+use itertools::Itertools;
 use lazy_static::lazy_static;
 use serde::{Deserialize, Serialize};
 use smol_str::SmolStr;
@@ -36,8 +37,8 @@ pub struct ListValue(Vec<Value>, Type);
 impl ListValue {
     /// Create a new [CustomConst] for a list of values of type `typ`.
     /// That all values ore of type `typ` is not checked here.
-    pub fn new(typ: Type, contents: Vec<Value>) -> Self {
-        Self(contents, typ)
+    pub fn new(typ: Type, contents: impl IntoIterator<Item = Value>) -> Self {
+        Self(contents.into_iter().collect_vec(), typ)
     }
 
     /// Create a new [CustomConst] for an empty list of values of type `typ`.
@@ -52,7 +53,7 @@ impl CustomConst for ListValue {
         SmolStr::new_inline("list")
     }
 
-    fn typ(&self) -> CustomType {
+    fn custom_type(&self) -> CustomType {
         let list_type_def = EXTENSION.get_type(&LIST_TYPENAME).unwrap();
         list_type_def
             .instantiate(vec![Into::<TypeArg>::into(self.1.clone())])
--- a/src/std_extensions/collections.rs
+++ b/src/std_extensions/collections.rs
@@ -31,13 +31,13 @@ pub const EXTENSION_NAME: ExtensionId = ExtensionId::new_unchecked("Collections"
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 /// Dynamically sized list of values, all of the same type.
-pub struct ListValue(Vec<Value>);
+pub struct ListValue(Vec<Value>, Type);
 
 impl ListValue {
     /// Create a new [CustomConst] for a list of values.
     /// (The caller will need these to all be of the same type, but that is not checked here.)
-    pub fn new(contents: Vec<Value>) -> Self {
-        Self(contents)
+    pub fn new(t: Type, contents: Vec<Value>) -> Self {
+        Self(contents, t)
     }
 }
 
@@ -45,6 +45,13 @@ impl ListValue {
 impl CustomConst for ListValue {
     fn name(&self) -> SmolStr {
         SmolStr::new_inline("list")
+    }
+
+    fn typ(&self) -> CustomType {
+        let list_type_def = EXTENSION.get_type(&LIST_TYPENAME).unwrap();
+        list_type_def
+            .instantiate(vec![Into::<TypeArg>::into(self.1.clone())])
+            .unwrap()
     }
 
     fn check_custom_type(&self, typ: &CustomType) -> Result<(), CustomCheckFailure> {
@@ -96,20 +103,16 @@ impl ConstFold for PopFold {
         let list: &ListValue = list.get_custom_value().expect("Should be list value.");
         let mut list = list.clone();
         let elem = list.0.pop()?; // empty list fails to evaluate "pop"
-        let list = make_list_const(list, type_args);
+        let list = make_list_const(list);
         let elem = ops::Const::new(elem, ty.clone()).unwrap();
 
         Some(vec![(0.into(), list), (1.into(), elem)])
     }
 }
 
-pub(crate) fn make_list_const(list: ListValue, type_args: &[TypeArg]) -> ops::Const {
-    let list_type_def = EXTENSION.get_type(&LIST_TYPENAME).unwrap();
-    ops::Const::new(
-        list.into(),
-        Type::new_extension(list_type_def.instantiate(type_args).unwrap()),
-    )
-    .unwrap()
+pub(crate) fn make_list_const(list: ListValue) -> ops::Const {
+    let t = list.typ();
+    ops::Const::new(list.into(), Type::new_extension(t)).unwrap()
 }
 
 struct PushFold;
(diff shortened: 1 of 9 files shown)


doug-q:
Tests pass, so it seems serializing `CustomConst` isn't currently tested.

TODO: `check_custom_type` should fail if `self.typ() != typ` in most cases (notable excluding `CustomTestValue`)

doug-q:
I've removed the serialization stuff, this is just about CustomConst getting `typ`

src/std_extensions/collections.rs
```diff
@@ -31,13 +31,13 @@ pub const EXTENSION_NAME: ExtensionId = ExtensionId::new_unchecked("Collections"
...
 /// Dynamically sized list of values, all of the same type.
-pub struct ListValue(Vec<Value>);
+pub struct ListValue(Vec<Value>, Type);
 
 impl ListValue {
     /// Create a new [CustomConst] for a list of values.
     /// (The caller will need these to all be of the same type, but that is not checked here.)
-    pub fn new(contents: Vec<Value>) -> Self {
-        Self(contents)
+    pub fn new(t: Type, contents: Vec<Value>) -> Self {
```
acl-cqc, line 39: I would move in favour of two constructors:
* `new_empty` that takes just a Type
* `new` that takes a Vec<Value> and not a type, and extracts the type from the values (checking that they are all the same, I guess - I think it's reasonable to panic if they are not)

src/values.rs
```diff
@@ -257,21 +263,21 @@ pub(crate) mod test {
...
     use crate::type_row;
-    use crate::types::{FunctionType, Type, TypeBound};
+    use crate::types::{FunctionType, Type};
 
     #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 
     /// A custom constant value used in testing that purports to be an instance
     /// of a custom type with a specific type bound.
-    pub(crate) struct CustomTestValue(pub TypeBound, pub ExtensionSet);
+    pub(crate) struct CustomTestValue(pub CustomType);
```
acl-cqc, line 272: Update comment on previous line

src/std_extensions/collections.rs
```diff
@@ -96,35 +103,31 @@ impl ConstFold for PopFold {
...
 }
 
-pub(crate) fn make_list_const(list: ListValue, type_args: &[TypeArg]) -> ops::Const {
-    let list_type_def = EXTENSION.get_type(&LIST_TYPENAME).unwrap();
-    ops::Const::new(
-        list.into(),
-        Type::new_extension(list_type_def.instantiate(type_args).unwrap()),
-    )
-    .unwrap()
+pub(crate) fn make_list_const(list: ListValue) -> ops::Const {
```
acl-cqc, line 113: I think you can change the impl block in src/ops/constant.rs:
```
impl<T> From<T> for Const
where
    T: KnownTypeConst + CustomConst,
```
to just `impl <T: CustomConst> From<T> for Const`, at which point this `make_list_const` is just `Const::from` so can be removed

acl-cqc (commented):
Looks good, but I wonder if there is really any use-case left for KnownTypeConst? Perhaps we should just remove it?

src/std_extensions/collections.rs
doug-q: We do not seem to have an interface to get the type of a `Value`, although we can do it now. I will add.

doug-q: Hmm, not so easy. a `Value::Function` can be done via `hugr.get_function_signature().unwrap()`, although I'm not sure this is a good idea.
`Value::Sum` does not know the full type.

I'm going to leave this for now, so your second bullet will not be satisfied. i.e. no checking they are all the same. Lmk if you disagree.

acl-cqc:
> > Looks good, but I wonder if there is really any use-case left for KnownTypeConst? Perhaps we should just remove it?
> 
> Very good point. Will do. Do you think we need to modify `check_custom_type` or it's caller `Type::check_type`?

Hmmm, this is a bit tricky, and relates to your observation [earlier](https://github.com/CQCL/hugr/pull/827#discussion_r1475886150) that Values don't know their type. A *Const* is a combination of a Value and a Type (so a Const is a Value that *does* know its type). But note a CustomConst is a Value (so it should be called CustomValue or maybe ExtensionValue?), but does know its type. Hmmm.

So the really-elaborate "fix" would be to make Const an Enum - either a CustomConst, or a Value+Type; and then remove Value::Extension. I haven't worked this through but it sounds quite invasive and I'm not sure I'd bother.

The *least* thing I think you should do however is to remove the implementations of `fn check_custom_type` which now just duplicate the trait-default one.

The intermediate thing - see what you think - would indeed be to change `check_custom_type` so that it doesn't take a Type parameter. (One could almost call it `validate()` but I think `check_custom_type` is not too bad - short for check_customconst_type. *Maybe* one day we might even pass in an ExtensionRegistry too!) Then the caller (`Type::check_type(Value)`....so that really should be `check_value`, or else be a member of Value, but that's a minor detail) can check the CustomConst's `.typ()` is what's expected and then call the validate method.

If you want to rename CustomConst -> ExtensionValue, and Type::check_type to Type::check_value or Value::check_type, I'd suggest that be a followup PR.

src/values.rs
```diff
@@ -154,14 +154,27 @@ pub trait CustomConst:
...
 
     /// Compare two constants for equality, using downcasting and comparing the definitions.
     // Can't derive PartialEq for trait objects
     fn equal_consts(&self, _other: &dyn CustomConst) -> bool {
         // false unless overloaded
         false
     }
+
+    /// report the type
+    fn typ(&self) -> CustomType;
```
acl-cqc, line 177: How about `get_type` or `custom_type` ?

src/std_extensions/collections.rs
```diff
@@ -31,13 +31,18 @@ pub const EXTENSION_NAME: ExtensionId = ExtensionId::new_unchecked("Collections"
...
+pub struct ListValue(Vec<Value>, Type);
 
 impl ListValue {
-    /// Create a new [CustomConst] for a list of values.
-    /// (The caller will need these to all be of the same type, but that is not checked here.)
-    pub fn new(contents: Vec<Value>) -> Self {
-        Self(contents)
+    /// Create a new [CustomConst] for a list of values of type `typ`.
+    /// That all values ore of type `typ` is not checked here.
+    pub fn new(typ: Type, contents: Vec<Value>) -> Self {
```
acl-cqc, line 39: Good point that `Value`s don't allow to deduce the type, so we can't really do anything else here. A slight tweak would be to `contents: impl IntoInterator<Value>` allowing callers to omit the `vec!`.

Being slightly wilder, one could have a version taking no Type but a Vec<Const>, but I'm happy to skip that ;-)

src/std_extensions/collections.rs
```diff
@@ -47,6 +52,13 @@ impl CustomConst for ListValue {
         SmolStr::new_inline("list")
     }
 
+    fn typ(&self) -> CustomType {
+        let list_type_def = EXTENSION.get_type(&LIST_TYPENAME).unwrap();
```
acl-cqc, line 56: You could use `list_type` (toplevel in collections.rs) here

acl-cqc (approved):
Thanks @doug-q and sorry this is turning out a bit more complex than expected! I think you are pretty much there though, I've suggested a few renames and cleanups so feel free to ping me if unsure but I think you can do the rest! :)

src/std_extensions/collections.rs
acl-cqc: No, I mean isn't this whole method just `list_type(self.0.clone())` ?

doug-q:
Thanks @acl-cqc, great comments. I've made all your suggested fixes, the `check_custom_type` change of name (to `validate` and signature is in a separate commit. I do think it's better, in particular `check_custom_type` was looking very strange next to `custom_type`. Note the loss of custom error messages for `ConstIntS` and `ConstIntU`.

src/std_extensions/collections.rs
doug-q: No, `list_type` returns a `Type` and we need a `CustomType`.

acl-cqc: Ah ok yes you are right, sorry for noise!

acl-cqc (approved):
Great, thanks @doug-q.

I have one slight reservation about `.validate()` which is that it doesn't actually happen at validation time (I think), rather at Const-construction time. But even that is a decision that we could revisit, and it still seems the most descriptive name re. what it does (rather than what it's used for), and I am in favour of naming things after what they do (not what they are used for)...so I'm happy with `validate`.

PR #833, opened 2024-02-08
feat: Minimal implementation for YAML extensions

aborgna-q:
This PR add a method `hugr::extensions::declarative::load_extensions` that dynamically loads a set of extensions defined as YAML onto a registry.

The code is mostly comprised of struct definitions to match the human-readable serialisation format described in the spec, and some methods to translate them into the internal hugr definitions.

There's a myriad of TODOs that should be addressed in future PRs, including:
- Most parametric things (operations, type bounds, number of ports in a signature, ...).
- Lowering functions, operations with non explicit signatures.
- Resolving the signature types.
  - The syntax for describing these is not defined in the spec, so currently there's just a couple of basic hard-coded types used for testing: "Q" and "USize".
  
Here's an example of a supported definition:
```yaml
imports: [prelude]

extensions:
- name: SimpleExt
  types:
  - name: MyType
    description: A simple type with no parameters
  operations:
  - name: MyOperation
    description: A simple operation with no inputs nor outputs
    signature:
      inputs: []
      outputs: []
  - name: AnotherOperation
    description: An operation from 2 qubits to 2 qubits
    signature:
        inputs: [["Target", Q], ["Control", Q, 1]]
        outputs: [[null, Q, 2]]
```

### DIFF
--- a/src/extension/declarative.rs
+++ b/src/extension/declarative.rs
@@ -9,18 +9,37 @@
 mod ops;
 mod types;
 
-use super::ExtensionId;
+use std::fs::File;
+
+use crate::Extension;
+
+use super::{ExtensionBuildError, ExtensionId, ExtensionRegistry, ExtensionSet, SignatureError};
 use ops::OperationDeclaration;
 use types::TypeDeclaration;
 
 use serde::{Deserialize, Serialize};
+
+/// Load a set of extensions from a YAML string.
+pub fn load_extensions(yaml: &str) -> Result<ExtensionRegistry, ExtensionDeclarationError> {
+    let ext: ExtensionSetDeclaration = serde_yaml::from_str(yaml)?;
+    ext.make_registry()
+}
+
+/// Load a set of extensions from a file.
+pub fn load_extensions_file(
+    path: &std::path::Path,
+) -> Result<ExtensionRegistry, ExtensionDeclarationError> {
+    let file = File::open(path)?;
+    let ext: ExtensionSetDeclaration = serde_yaml::from_reader(file)?;
+    ext.make_registry()
+}
 
 /// A set of declarative extension definitions with some metadata.
 ///
 /// These are normally contained in a single YAML file.
 //
 // TODO: More metadata, "namespace"?
-#[derive(Debug, Clone, Serialize, Deserialize, Hash, PartialEq, Eq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 struct ExtensionSetDeclaration {
     /// A set of extension definitions.
     //
@@ -30,11 +49,11 @@ struct ExtensionSetDeclaration {
     /// Optional.
     #[serde(default)]
     #[serde(skip_serializing_if = "crate::utils::is_default")]
-    imports: Vec<ExtensionId>,
+    imports: ExtensionSet,
 }
 
 /// A declarative extension definition.
-#[derive(Debug, Clone, Serialize, Deserialize, Hash, PartialEq, Eq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 struct ExtensionDeclaration {
     /// The name of the extension.
     name: ExtensionId,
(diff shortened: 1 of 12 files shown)


src/extension/declarative/types.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// The parameter type.
+    type_name: TypeName,
+}
+
+impl TypeParamDeclaration {
+    /// Create a [`TypeParam`] from this declaration.
+    ///
+    /// Resolves any type ids using both the current extension and any other in `scope`.
+    ///
+    /// TODO: Only non-parametric opaque types are supported for now.
```
ss2165, line 123: what's the reason for starting with this? If I were to pick just one I feel like
I'd pick TypeParam::Type to support rather than TypeParam::Opaque. Simpler and
more commonly used no?

src/extension/declarative/signature.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// The outputs of the operation.
+    outputs: Vec<SignaturePortDeclaration>,
+    /// A set of extensions invoked while running this operation.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "crate::utils::is_default")]
+    extensions: ExtensionSet,
+}
+
+impl SignatureDeclaration {
+    pub fn make_signature(
```
ss2165, line 40: nit: docstring

src/extension/declarative/signature.rs
```diff
@@ -0,0 +1,257 @@
...
+
+        let poly_func = PolyFuncType::new(op_params, body);
+        Ok(SignatureFunc::TypeScheme(CustomValidator::from_polyfunc(
+            poly_func,
+        )))
+    }
+
+    /// Create a type row from a list of port declarations.
+    fn make_type_row(
+        &self,
```
ss2165, line 61: unecessary - can be associated method or function

src/extension/declarative/signature.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// The number of repetitions for this port definition.
+    repeat: PortRepetitionDeclaration,
+}
+
+impl SignaturePortDeclaration {
+    /// Return an iterator with the types for this port declaration.
+    ///
+    /// Only a fixed number of repetitions is supported for now.
+    ///
+    /// TODO: We may need to use custom signature functions if `repeat` depends on a variable.
```
ss2165, line 101: this should be covered by the type scheme sub-language, which isn't within the
scope of this PR. I would remove the "repeat variable number of times" in this
PR entirely. Syntactic sugar for fixed repeat is handy.

src/extension/declarative/signature.rs
```diff
@@ -0,0 +1,257 @@
...
+    fn make_types(
+        &self,
+        ext: &Extension,
+        scope: &ExtensionSet,
+        registry: &ExtensionRegistry,
```
ss2165, line 106: these parameters are passed in to things a lot - maybe worth defining a struct
to hold them? `DeclarationContext` perhaps?

src/extension/declarative/signature.rs
```diff
@@ -0,0 +1,257 @@
...
+impl Default for PortRepetitionDeclaration {
+    fn default() -> Self {
+        PortRepetitionDeclaration::Count(1)
+    }
+}
+
+/// A type declaration used in signatures.
+///
+/// TODO: The spec definition is more complex than just a type identifier,
+/// we should be able to support expressions like:
```
ss2165, line 197: yes - this would form a part of the type scheme definition language (the
fragment for types)

src/extension/declarative/ops.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::Extension;
+
+use super::signature::SignatureDeclaration;
+use super::ExtensionDeclarationError;
+
+/// A declarative operation definition.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub(super) struct OperationDeclaration {
+    /// The identifier the operation.
+    name: SmolStr,
```
ss2165, line 27: `OpName` like `TypeName`?

src/extension/declarative/ops.rs
```diff
@@ -0,0 +1,115 @@
...
+/// [`TypeParam`]: crate::types::type_param::TypeParam
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize)]
+pub(super) struct ParamDeclaration(
+    /// TODO: Store a [`TypeParam`], and implement custom parsers.
+    ///
+    /// [`TypeParam`]: crate::types::type_param::TypeParam
+    String,
+);
+
+/// Reference to a binary lowering function.
```
ss2165, line 104: too ill defined I think, just remove references to, maybe leave one top level
TODO about fleshing out lowering

src/extension/declarative.rs
```diff
@@ -0,0 +1,293 @@
...
+    )]
+    MissingSignature {
+        /// The extension containing the operation.
+        ext: ExtensionId,
+        /// The operation with no signature.
+        op: SmolStr,
+    },
+    /// Since currently we hard-code some basic types, we require the prelude to be in scope when using them.
+    ///
+    /// TODO: Don't require this.
```
ss2165, line 205: ? the prelude is expected to be in scope right?

ss2165 (commented):
Nicely done! 🪓

src/extension/declarative.rs
aborgna-q: Right. I auto-included it.

src/extension/declarative/ops.rs
aborgna-q: `OpName` is a trait :/

I wanted to give wrappers for op, type, and parameter names, and stop using (smol)stringly-typed names everywhere.

But that's out of the scope of this PR, so I'll open an issue.

src/extension/declarative/ops.rs
aborgna-q: I put an explicit error if the field is present.
Otherwise it would be confusing for someone copying from the spec example.

src/extension/declarative/signature.rs
aborgna-q: As with lowering, I'll leave the parser support for it as long as it's still in the spec and raise an explicit error in the unsupported cases, so the user knows it's not a syntax error.

src/extension/declarative/types.rs
aborgna-q: I went with supporting the first example.

Having any type parameter is currently rejected when translating the declaration, so this was mostly a check to make sure things can be defined correctly.

We'll probably want to support both once we re-enable type params.

src/extension/op_def.rs
```diff
@@ -445,13 +445,29 @@ impl Extension {
...
+        self.add_op_with_misc(name, description, signature_func, Default::default())
+    }
+
+    /// Add an operation definition to the extension. Must be a type scheme
+    /// (defined by a [`PolyFuncType`]), a type scheme along with binary
+    /// validation for type arguments ([`CustomValidator`]), or a custom binary
+    /// function for computing the signature given type arguments (`impl [CustomSignatureFunc]`).
+    ///
+    /// Includes a `misc` field for additional data that can be associated with the operation.
+    pub fn add_op_with_misc(
```
ss2165, line 458: this is what the `add_misc` method on `OpDef` is for, can you use that instead of adding a method?

ss2165 (changes requested):
just one change request, otherwise good to go 👍

PR #839, opened 2024-02-20
feat: Arbitrary types in the yaml extension definition

aborgna-q:
The current yaml extension definitions only support hard-coded types in the operation signatures.

This PR leaves the existing `Q` and `USize` hardcoded values, but tries to dereference the type name from the extensions in scopes otherwise.

drive-by: Extend the module doc with an example on how to load the extension.

### DIFF
--- a/src/extension/declarative/signature.rs
+++ b/src/extension/declarative/signature.rs
@@ -174,10 +174,7 @@ impl TypeDeclaration {
         ctx: DeclarationContext<'_>,
         _op_params: &[TypeParam],
     ) -> Result<CustomType, ExtensionDeclarationError> {
-        // The prelude is always in scope.
-        debug_assert!(ctx.scope.contains(&PRELUDE_ID));
-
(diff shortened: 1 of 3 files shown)


src/extension/declarative/signature.rs
```diff
@@ -177,27 +177,53 @@ impl TypeDeclaration {
...
-                    ext: ext.name().clone(),
-                    ty: self.0.clone(),
-                })
-            }
-        }
-        .ok_or(ExtensionDeclarationError::UnknownType {
-            ext: ext.name().clone(),
-            ty: self.0.clone(),
-        })?;
+        let Some(op_def) = self.resolve_type(ext, ctx) else {
```
ss2165, line 180:
```suggestion
        let Some(type_def) = self.resolve_type(ext, ctx) else {
```

ss2165 (approved):
just a variable name change suggestion

PR #849, opened 2024-02-27
docs(spec): Add note on serialized hugr node order

croyzor:


specification/hugr.md
```diff
@@ -1556,8 +1556,8 @@ pseudocode, though we advocate MessagePack format in practice (see
 [Serialization implementation](schema/serialization.md)).
 Note in particular that hierarchical relationships
 have a special encoding outside `edges`, as a field `parent`
-in a node definition.
-The unique root node of the HUGR reports itself as the parent.
+in a node definition. Nodes are identified by their position in the `nodes`
```
croyzor, line 1559: This is already said below, but I think it bears repeating because it's not super intuitive. I'd be glad to hear a rewording that makes it clearer though

mark-koch: The wording sounds good to me. Maybe we should also clarify that the position is zero-based?

croyzor: Done :+1:

mark-koch (approved):

PR #851, opened 2024-02-27
ci: Compute coverage with and without experimental features

### LINKED ISSUE #799: Submit both `--all-features` and `--no-default--features` coverage to codecov
The `extension_inference` feature is experimental and may cause errors on otherwise correct code.
Hence we skip some tests when running with `--all-features` (see #795).
This means that those tests do not count for the coverage report sent to codecov.

To fix it we can submit both reports with/without extensions. Codecov then combines them into a single coverage metric.

aborgna-q:
Closes #799


croyzor:
What is this getting us? It looks like we're generating the same report with `--all-features`?

aborgna-q:
There are some tests that get disabled when using the `extension_inference` feature.
Before this, those test where completely ignored in the report (as they don't even get compiled).

E.g.

Test before: https://app.codecov.io/gh/CQCL/hugr/commit/ff5f1179b82d8c6b85e7559a49967ac66468a07a/blob/src/algorithm/const_fold.rs#L335
Test after: https://app.codecov.io/gh/CQCL/hugr/pull/851/blob/src/algorithm/const_fold.rs#L335

croyzor:
Ah, I see. Thanks for explaining!

croyzor (approved):

PR #856, opened 2024-03-04
feat: Add impl TryFrom<&OpType> for ops::*

doug-q:


src/ops.rs
```diff
@@ -62,18 +62,25 @@ macro_rules! impl_op_ref_try_into {
...
                 pub fn [<as_ $sname:snake>](&self) -> Option<&$Op> {
-                    if let OpType::$Op(l) = self {
-                        Some(l)
-                    } else {
-                        None
-                    }
+                    TryInto::<&$Op>::try_into(self).ok()
                 }
 
                 #[doc = "If is an instance of `" $Op "`."]
```
aborgna-q, line 68: Small drive-by
```suggestion
                #[doc = "Returns `true` if the operation is an instance of `" $Op "`."]
```

aborgna-q (approved):

PR #867, opened 2024-03-06
feat: Ancilla support in CircuitBuilder

### LINKED ISSUE #854: Add/discard ancillae in `CircuitBuilder`
So we can build circuits like
image

aborgna-q:
Adds `add_ancilla` and `discard_ancilla` methods to the circuit builder, to modify the list of tracked wires.
Also adds extra info in the error messages.

Closes #854.

### DIFF
--- a/quantinuum-hugr/src/builder/circuit.rs
+++ b/quantinuum-hugr/src/builder/circuit.rs
@@ -338,7 +338,7 @@ mod test {
     #[test]
     fn circuit_builder_errors() {
         let _build_res = build_main(
-            FunctionType::new(type_row![QB, QB], type_row![QB, QB]).into(),
+            FunctionType::new_endo(type_row![QB, QB]).into(),
             |mut f_build| {
                 let mut circ = f_build.as_circuit(f_build.input_wires());
                 let [q0, q1] = circ.tracked_units_arr();
--- a/quantinuum-hugr/src/builder/circuit.rs
+++ b/quantinuum-hugr/src/builder/circuit.rs
@@ -219,9 +219,7 @@ mod test {
     use super::*;
     use cool_asserts::assert_matches;
 
-    use crate::utils::test_quantum_extension::{
-        cx_gate, cz_gate, h_gate, measure, q_alloc, q_discard,
-    };
+    use crate::utils::test_quantum_extension::{cx_gate, h_gate, measure, q_alloc, q_discard};
     use crate::{
         builder::{
             test::{build_main, NAT, QB},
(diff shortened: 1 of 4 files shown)


aborgna-q:
Alternatively, we could just have two functions
```haskell
track_wire : Wire -> usize
untrack_wire : usize -> Wire
```
and manually call them after/before allocating or discarding,
```rust
let ancilla: Wire = circ.append_with_outputs(q_alloc(), [])?[0];
let ancilla: usize = circ.track_wire(ancilla);

...

let ancilla: Wire = circ.untrack_wire(ancilla);
circ.append_and_consume(q_discard(), [ancilla]);
```

ss2165 (changes requested):
I much prefer the track/untrack methods - keeps it generic without getting too qubit-y

Will do detailed review after that change if you agree!

aborgna-q:
Done, I dropped the composable calls in favour of track/untrack.

quantinuum-hugr/src/builder/circuit.rs
```diff
@@ -208,4 +294,54 @@ mod test {
...
+            FunctionType::new(type_row![QB], type_row![QB]).into(),
+            |mut f_build| {
+                let mut circ = f_build.as_circuit(f_build.input_wires());
+                assert_eq!(circ.n_wires(), 1);
+
+                let [q0] = circ.tracked_units_arr();
+                let [ancilla] = circ
+                    .append_with_outputs::<CircuitUnit>(q_alloc(), [])?
+                    .try_into()
+                    .expect("Expected a single ancilla wire");
```
ss2165, line 310: Maybe we should have a panicking `as_arr()` utility on `Outputs` too? Can be follow up

quantinuum-hugr/src/utils.rs
```diff
@@ -85,6 +85,10 @@ pub(crate) mod test_quantum_extension {
             )
             .unwrap();
 
+        extension
+            .add_op(SmolStr::new_inline("CZ"), "CZ".into(), two_qb_func())
+            .unwrap();
+
```
ss2165, line 91: worth adding this rather than just using CX?

ss2165 (commented):
I think worth adding a quick test of the two error modes `InvalidWireIndex` and `MismatchedLinearInputs`

quantinuum-hugr/src/utils.rs
aborgna-q: I wanted to match the motivating example in the test, but we can't do it without also adding a classically controlled operation.

I'll drop this.

quantinuum-hugr/src/builder/circuit.rs
aborgna-q: https://github.com/CQCL/hugr/pull/871

quantinuum-hugr/src/builder/circuit.rs
```diff
@@ -344,4 +334,46 @@ mod test {
 
         assert_matches!(build_res, Ok(_));
     }
+
+    #[test]
+    fn circuit_builder_errors() {
+        let _build_res = build_main(
+            FunctionType::new(type_row![QB, QB], type_row![QB, QB]).into(),
```
ss2165, line 341: nit: can use `new_endo`

ss2165 (approved):

PR #875, opened 2024-03-12
feat: Make various data publicly accessible

doug-q:


quantinuum-hugr/src/types/type_param.rs
```diff
@@ -246,6 +251,12 @@ impl TypeArg {
     }
 }
 
+impl TypeArgVariable {
```
doug-q, line 254: There's no docstring here, why isn't CI catching this?

aborgna-q: Uh, seems I missed setting `lints.workspace = true` in the package's Cargo.toml when I created the workspace...

I can fix it in a separate PR, or you can add
```toml
[lints]
workspace = true
```
in `quantinuum-hugr/Cargo.toml`.

aborgna-q (approved):
:+1:

Just add the missing doc.

quantinuum-hugr/src/types/type_param.rs
aborgna-q: See #876

PR #868, opened 2024-03-06
chore: Add a rust workspace

aborgna-q:
Moves the main crate to a directory.
This change does not alter the public package.

Closes #865

PR #876, opened 2024-03-12
chore: Re-enable missing-docs lint

aborgna-q:
#868 didn't actually enable the lints in `quantinuum-hugr` when the config was moved to the workspace.

### DIFF
--- a/quantinuum-hugr/Cargo.toml
+++ b/quantinuum-hugr/Cargo.toml
@@ -13,6 +13,9 @@ description = "Quantinuum's Hierarchical Unified Graph Representation"
 keywords = ["Quantum", "Quantinuum"]
 categories = ["compilers"]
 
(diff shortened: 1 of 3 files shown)


doug-q:
I think the `#[allow(missing_docs)]` is dubious. I agree that the existing constructors are ok to not document their fields because they are quasi-documented via the error messages. However these annotations will apply to newly added constuctors. 

perhaps adding
`#[allow(missing_docs)] // all fields are mentioned in error message` per-constructor would be better?

Happy to approve if you think this would be too verbose

aborgna-q:
I wanted to avoid the boilerplate; but I agree that variants should always have docs.
I moved the attributes to the variants (without the comment though, as that's even more verbose).

doug-q (approved):

PR #880, opened 2024-03-14
feat!: Return the type of FuncDecl in `HugrView::get_function_type`

aborgna-q:
`get_function_type` was confusing, as it didn't specify that it only worked for dataflow containers, and required manually checking for function declarations on the side.

BREAKING CHANGE: `HugrView::get_function_type` now returns a `PolyFuncType`.

### DIFF
--- a/quantinuum-hugr/src/hugr/views.rs
+++ b/quantinuum-hugr/src/hugr/views.rs
@@ -327,13 +327,28 @@ pub trait HugrView: sealed::HugrInternals {
         }
     }
 
+    /// Returns the function type defined by this dataflow HUGR.
+    ///
+    /// If the root of the Hugr is a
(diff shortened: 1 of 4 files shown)


ss2165:
We discussed this a lot in https://github.com/CQCL/hugr/pull/754 and the follow up #782 

I still think that getting a FunctionType when applicable rather than a PolyFuncType is useful, and the latter only applies to two node types. I would perhaps suggest a non-breaking additional function that does what you want.

aborgna-q:
Could the previous `get_function_type` be renamed to `get_df_function_type` or something similar?
I had to spend some time tracking down the function calls to see why the `FuncDecl` match was necessary.

#881 will use this generic version in multiple places.

ss2165:
yeah I don't mind that

croyzor (approved):

PR #881, opened 2024-03-14
feat!: Merge `Value` into `Const`

### LINKED ISSUE #874: Get rid of `Value`
`Const`s are defined as `Values` plus a concrete type.

The difference between both types is confusing, specially at the API level.
In practice, most of the times that `Value` is used we do already know their `Type`.

This issue proposes either making `Value` private, or merging it directly with `Const`.

aborgna-q:
`Const` was a struct with a `Value` + a defined type. In practice, the only `Value` that couldn't report its full type was `Sum`.

This PR merges the two structs, adding an explicit `sum_type` to the sum variant.

Also:

- Removes `Type::check_type`, as this is now unnecessary. 
- Splits and clarifies the type check and constant errors;
  `SumTypeError` is not a `ConstTypeError` is not a `CustomCheckFailure`.
  - Adds contexts to those errors, so we get better messages.
- `CustomConst` is now accessible from `::ops::custom` instead of `::values`.

This change breaks the serialisation format.

Closes #874.

BREAKING CHANGE: Removed `Value`, merged into `Const`

### DIFF
--- a/quantinuum-hugr/src/algorithm/const_fold.rs
+++ b/quantinuum-hugr/src/algorithm/const_fold.rs
@@ -4,6 +4,7 @@ use std::collections::{BTreeSet, HashMap};
 
 use itertools::Itertools;
 
+use crate::types::SumType;
 use crate::{
     builder::{DFGBuilder, Dataflow, DataflowHugr},
     extension::{ConstFoldResult, ExtensionRegistry},
@@ -14,8 +15,7 @@ use crate::{
     },
     ops::{Const, LeafOp},
     type_row,
-    types::{FunctionType, Type, TypeEnum},
-    values::Value,
+    types::FunctionType,
     Hugr, HugrView, IncomingPort, Node, SimpleReplacement,
 };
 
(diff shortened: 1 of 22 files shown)


aborgna-q:
The PR is ready, but I'll try to split it a bit

aborgna-q:
Added an ad-hoc `get_const_function_type` in lieu of #880, while we decide on that.

doug-q:
I think the serialisation schema should be changing because the serialisation of const nodes has changed? 

We have already bumped the serialisation version to V1 since the last release, so it is fine to change the V1 schema. If we hadn't bumped since the last release this would necessitate bumping.

aborgna-q:
Yes, here's the new encoding for an (extension) `usize`:

```json
{
  "v": "Extension",
  "c": [
    {
      "c": "ConstUsize",
      "value": 257
    }
  ]
}
```

And for a `[float, usize] + []` const:

```json
{
  "v": "Sum",
  "tag": 0,
  "values": [
    {
      "v": "Extension",
      "c": [
        {
          "c": "ConstUsize",
          "value": 257
        }
      ]
    },
    {
      "v": "Extension",
      "c": [
        {
          "c": "CustomSerialized",
          "typ": {
            "t": "Opaque",
            "extension": "arithmetic.float.types",
            "id": "float64",
            "args": [],
            "bound": "C"
          },
          "value": 5.1,
          "extensions": [
            "arithmetic.float.types"
          ]
        }
      ]
    }
  ],
  "typ": {
    "s": "General",
    "rows": [
      [
        {
          "t": "I"
        },
        {
          "t": "Opaque",
          "extension": "arithmetic.float.types",
          "id": "float64",
          "args": [],
          "bound": "C"
        }
      ],
      []
    ]
  }
}
```

aborgna-q:
I updated the serialization schema manually (merging the json generated by https://github.com/CQCL/guppylang/pull/172 with the local changes).
Hopefully we can improve this process in the future (see https://github.com/CQCL/guppylang/issues/173).

doug-q (approved):
Awesome, just a few nits

Issue #709, opened 2023-11-21
Polymorphic top-level Hugrs


acl-cqc:
This seems a good thing to have, and could potentially simplify many tests (which pull in some extension just so we have a concrete leaf type - they could just use a type variable instead).

We could store the binders *in* the Hugr, but then we should be careful that Hugrs rooted at FuncDefn/FuncDecl don't have any binders themselves.

The simplest/least option would just be to pass the decls (`[TypeParam]`) into the top-level `Hugr::validate()` (and friends `update_and_validate` etc.....).

And, follow-up (if we have not by then decided to go back to having only `FunctionType` in the type hierarchy and only monomorphic lambdas) - allow storing the decls alongside the Hugr in a `Const` too.

acl-cqc:
So, probably this is some kind of a wrapper around (a Hugr and some TypeParams), that one can `validate` (-> calls `Hugr::validate` passing in those TypeParams) and other things.
* Should include an instantiate method that (given TypeArgs correct for *all* its TypeParams) produces a non-polymorphic Hugr. This is itself a bunch of work. (Partial instantiation probably not required)
* Given #904, unclear whether you can have a "polymorphic top level hugr" (with some TypeParams) whose root node is a FuncDefn (binding more TypeParams). If so, then we retain DeBruijn even after #904; if not, then DeBruijn disappears with #904 (as nested FuncDefns can't refer to their enclosing binders, so we never have more than one set of TypeParams in scope)

acl-cqc:
This'll also give us a way to do inference separately for subgraphs and thus a better solution to inserting a subgraph than that of `test_validate_with_closure` from #884

acl-cqc:
It is just possible that maybe we could treat FuncDefn as the *only* binder with TypeParams. We would still need a method to apply a substitution across the Hugr (i.e. `instantiate` a FuncDefn-rooted Hugr to a monomorphic, e.g. DFG-rooted, Hugr)

PR #884, opened 2024-03-18
refactor: Combine ExtensionSolutions (no separate closure)

acl-cqc:
* `infer::infer_extensions` returns only a combined solution (for previously-open locations), after variables instantiated
* `Hugr::infer_extensions` writes (all parts of) the solution into place *and* returns it
* `validate_with_extension_closure` left in-place, with test demonstrating usage w/ sub-DFGs
* This should open the way (in future PRs) to changing implementation of `infer::infer_extensions`


acl-cqc:
IIUC coverage is not reduced by the PR, just that the lines affected by the PR have lower-than-average coverage anyway, and AFAICS that is because of `assert_matches!`, so I think this is OK.

quantinuum-hugr/src/extension/infer/test.rs
```diff
@@ -795,6 +794,28 @@ fn test_cfg_loops() -> Result<(), Box<dyn Error>> {
...
+        ExtensionSet::singleton(&B),
+    )?;
+    assert_matches!(
+        hugr.validate(&PRELUDE_REGISTRY),
+        Err(ValidationError::ExtensionError(_))
+    );
+
+    let immut = hugr.clone();
+    let soln = infer_extensions(&immut)?;
+    immut.validate_with_extension_closure(soln, &PRELUDE_REGISTRY)?;
```
croyzor, line 812: Looks like this should just be `immut.validate(soln, &PRELUDE_REGISTRY)`?

acl-cqc: ?? `validate` takes only one argument, the registry.

croyzor:
Casting my mind back, I think the motivation for returning the open solution is that we might want to run extension inference on a subgraph before we add it to the main graph. I'm not really sure when this would be the case -- I think the case is that when doing replacements, we run inference on the complete graph before and after, so it's probably unnecessary

acl-cqc:
> Casting my mind back, I think the motivation for returning the open solution is that we might want to run extension inference on a subgraph before we add it to the main graph. I'm not really sure when this would be the case -- I think the case is that when doing replacements, we run inference on the complete graph before and after, so it's probably unnecessary

Ah-ha, I see. So the flow would have been:
1. Run inference on the subgraph. Write into the subgraph only the solutions that do not depend on the subgraph inputs.
2. Substitute the subgraph into the main graph.
3. Rerun inference on the main graph in order to figure out the solutions that did depend on the subgraph inputs.

Noting that if in step 1 we wrote in all the solutions, including those dependent on the subgraph inputs, then step 3 would fail because the inserted graph would have bad solutions (e.g. emptyset) that conflict.

This PR removes the ability to write *only* those solutions into the subgraph. However, the flow is largely intact if one does not write *any* solutions into the subgraph before insertion. (You can still validate the subgraph using the full set of solutions, if required.)

Close enough? I think I can realize that in a test.

acl-cqc:
@croyzor added a much more thorough test which I hope covers that use-case (as well as we can - there'll be a better, but separate, solution if we do #709). I had to make `infer_extensions(&Hugr)` return only the *new* solutions rather than existing ones to get around assertions in `gather_extensions`. Can you take another look?

croyzor (approved):
Nice, let's update this once can give a better answer for polymorphic hugrs!

PR #893, opened 2024-03-25
chore: Show new contributors in the changelog

aborgna-q:
Imports part of the github template from `git-cliff`, so the changelog mentions the new contributors to the repo.
See https://git-cliff.org/docs/integration/github.

E.g.:
```
## Unreleased (XXXX-XX-XX)

New Contributors:

* @Cobord made their first contribution in [#889](https://github.com/CQCL/hugr/pull/889)
* @qartik made their first contribution in [#843](https://github.com/CQCL/hugr/pull/843)

…
```

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -63,7 +63,7 @@ filter_unconventional = true
 # process each line of a commit as an individual commit
 split_commits = false
 # regex for preprocessing the commit messages
-commit_preprocessors = [{ pattern = '\((\w+\s)?#([0-9]+)\)', replace = "" }]
+commit_preprocessors = [{ pattern = '\(#[0-9]+\)', replace = "" }]
 # regex for parsing and grouping commits
 commit_parsers = [
   { message = "^feat", group = "Features" },
--- a/cliff.toml
+++ b/cliff.toml
@@ -5,6 +5,11 @@
 # Configuration options are organized into tables and keys.
 # See documentation for more information on available options.
 
+[remote.github]
+owner = "CQCL"
+repo = "hugr"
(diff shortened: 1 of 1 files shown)


cliff.toml
```diff
@@ -38,9 +63,7 @@ filter_unconventional = true
 # process each line of a commit as an individual commit
 split_commits = false
 # regex for preprocessing the commit messages
-commit_preprocessors = [
-  { pattern = '\((\w+\s)?#([0-9]+)\)', replace = "([#${2}](https://github.com/CQCL/portgraph/issues/${2}))"}, # replace issue numbers
-]
+commit_preprocessors = [{ pattern = '\((\w+\s)?#([0-9]+)\)', replace = "" }]
```
croyzor, line 66: What is this trying to achieve? It looks like it's trying to delete things of the form "fixes #123"?

aborgna-q: The goal is to strip all the prefixes which show the PR number from the commit title.I removed the replacement part since the cliff+github writes that link reading the metadata, which seems more robust than the hardcoded link.

As for how we where generating correct links in the changelog even though here it linked to `portgraph`, I'm not quite sure.
We don't call `git-cliff` directly, but use `release-plz` which does a lot of pre-processing and collects each relevant commit individually before generating the changelog delta programatically. It may be ignoring the `commit_preprocessors` section since it's not using `git-cliff`'s CLI.

This config I submit here produces the correct result locally, but I'll have to review what gets generated by release-plz and submit a fix PR if needed.

croyzor (approved):

PR #901, opened 2024-03-27
fix: polymorphic calls

acl-cqc:
* Store the polymorphic type of the called function in Call node, use for the static input
* Also store the typeargs, and the cache of the instantiation

BREAKING CHANGE: `Call` nodes now constructed via `try_new` function

### DIFF
--- a/quantinuum-hugr/src/builder/build_traits.rs
+++ b/quantinuum-hugr/src/builder/build_traits.rs
@@ -576,8 +576,7 @@ pub trait Dataflow: Container {
                 })
             }
         };
-        let signature = type_scheme.instantiate(type_args, exts)?;
-        let op: OpType = ops::Call { signature }.into();
+        let op: OpType = ops::Call::try_new(type_scheme, type_args, exts)?.into();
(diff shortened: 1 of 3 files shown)


acl-cqc:
The thing about that is that TypeApply works only on Value edges, not Static ones.

When we simplify to "no polymorphic lambdas", TypeApply will disappear - but Call will retain some of its essence (as here).

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________
From: doug-q ***@***.***>
Sent: Thursday, March 28, 2024 9:45:59 AM
To: CQCL/hugr ***@***.***>
Cc: Alan Lawrence ***@***.***>; Author ***@***.***>
Subject: [EXT] Re: [CQCL/hugr] fix: polymorphic calls (PR #901)


CAUTION: This email originated from outside of the organization. Do not click links or open attachments unless you recognize the sender and know the content is safe.


An alternative:

  *   Call remains as is. i.e. it stores only a FunctionType.
  *   The builder inserts a LeafOp::TypeApply if necessary

The advantage of this is simpler call node and keeping type application localised to TypeApply

PR #903, opened 2024-04-02
[WIP] Row Variables, the hard (but good?) way

acl-cqc:
* Add RowVarOrType, used instead of Type in places where row variables are permitted
* Add TypeRowVarLen (via common TypeRowBase<T>), FuncTypeVarLen (via common FuncTypeBase<T>) and PolyFuncVarLen(via common PolyFuncBase<T>).
* Function types in the type hierarchy can have variable arity; OpDefs can have variable arity; nodes (even with type variables) must have fixed arity
* A few awkward cases of extra `.into()` or non-working `Default::default()` might be better
* TypeRowElem seems badly placed and/or named - ideas welcome? There are also visibility issues.
* Not nearly enough tests

Is it worth it??
* It's a big change
* But, once all the relevant classes/types had been defined (the three pairs listed above in naming) pretty much everything did just drop neatly into place.
* However, +400 lines!! And that's without enough tests....
* See what you think....the hacky version #804 might be more appealing now!!


acl-cqc:
> I prefer #804 to this. Although I do agree that the approach here is more correct, it complicates the api to no benefit for uses that are "downstream" from any polymorphism.

Fair. TBH I think the main thing about this way is that I am a lot more confident that it is (mostly) correct!...

acl-cqc:
Let's go with #804

PR #891, opened 2024-03-25
feat: Replace `Tuple` with unary sums

aborgna-q:
Closes #864.

Updates the `Display` impl of `Sum` as part of this change, so
```mermaid
graph LR
    subgraph 0 ["(0) CFG"]
        direction LR
        subgraph 2 ["(2) DataflowBlock"]
            direction LR
            3["(3) Input"]
            3--"0:0<br>usize"-->5
            4["(4) Output"]
            5["(5) Tag"]
            5--"0:0<br>Sum(General([TypeRow { types: [Type(Extension(CustomType { extension: IdentList(#quot;prelude#quot;), id: #quot;usize#quot;, args: [], bound: Eq }), Eq)] }, TypeRow { types: [Type(Extension(CustomType { extension: IdentList(#quot;prelude#quot;), id: #quot;usize#quot;, args: [], bound: Eq }), Eq)] }]))"-->4
        end
        2-."0:0".->6
        2-."1:0".->1
        1["(1) ExitBlock"]
        subgraph 6 ["(6) DataflowBlock"]
            direction LR
            7["(7) Input"]
            7--"0:1<br>usize"-->8
            8["(8) Output"]
            9["(9) const:sum:{tag:0, vals:[]}"]
            9--"0:0<br>Sum(UnitSum(1))"-->10
            10["(10) LoadConstant"]
            10--"0:0<br>Sum(UnitSum(1))"-->8
        end
        6-."0:0".->1
    end
```
is now rendered as
```mermaid
graph LR
    subgraph 0 ["(0) CFG"]
        direction LR
        subgraph 2 ["(2) DataflowBlock"]
            direction LR
            3["(3) Input"]
            3--"0:0<br>usize"-->5
            4["(4) Output"]
            5["(5) Tag"]
            5--"0:0<br>[usize]+[usize]"-->4
        end
        2-."0:0".->6
        2-."1:0".->1
        1["(1) ExitBlock"]
        subgraph 6 ["(6) DataflowBlock"]
            direction LR
            7["(7) Input"]
            7--"0:1<br>usize"-->8
            8["(8) Output"]
            9["(9) const:sum:{tag:0, vals:[]}"]
            9--"0:0<br>[]"-->10
            10["(10) LoadConstant"]
            10--"0:0<br>[]"-->8
        end
        6-."0:0".->1
    end
```
And tuples appear directly as `[usize, usize, ...]`.
We were incorrectly using the debug formatter for the internal types.

[...]

PR #906, opened 2024-04-02
feat: No polymorphic closures

### LINKED ISSUE #904: No polymorphic closures
* Only FuncDefn (and FuncDecl) may declare binders
* `Type` hierarchy includes monomorphic function types only, not polymorphic ones.
* Two distinct `EdgeKind`s: one for values/closures (from Consts to their uses, including CallIndirect), i.e. no polymorphism; and one for functions (carries a polymorphic function type from a FuncDecl/Defn to a Call, only).
* Thus, no `TypeApply`, but `Call` does type application, as per #901
* Should simplify substitution logic as we no longer need to substitute polymorphic types in place of type variables, and don't need to substitute under binders (binders are only at the outermost level - function or opdef)

acl-cqc:
* Type::Function now stores only a FunctionType, not a PolyFuncType
* PolyFuncType remains, for OpDef's and FuncDefn/Decl's
* EdgeKind::Static now replaced by EdgeKind::Const (a type) EdgeKind::Static (a PolyFuncType)
* Remove LeafOp::TypeApply, repurpose validation code onto Call
* Thus, progressively remove all `impl Substitution`s except for `struct SubstValues`, which can become Substitution
* Update spec, introducing "Static" and "Dataflow" edge kinds as broader classes of the other edge kinds, and polymorphic "type schemes" vs monomorphic "Function types".
* Update serialization schema and add roundtrip test of a Noop operating on a value of function type

fixes #904

This should enable resolving #788 and related capture/closure issues if we forbid edges into a FuncDefn from outside (@doug-q)

BREAKING CHANGE: EdgeKind::{Static -> Const}, add new EdgeKind::Function, Type contains only monomorphic functions, remove TypeApply.


quantinuum-hugr/src/hugr/validate.rs
```diff
@@ -419,10 +422,19 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
 
         let is_static = match from_optype.port_kind(from_offset).unwrap() {
             EdgeKind::Static(typ) => {
-                if !(OpTag::Const.is_superset(from_optype.tag())
-                    || OpTag::Function.is_superset(from_optype.tag()))
-                {
-                    return Err(InterGraphEdgeError::InvalidConstSrc {
+                if !OpTag::Const.is_superset(from_optype.tag()) {
```
acl-cqc, line 425: TBH I think this check is more of an assertion about the spec, than something we really need to check at runtime; there should be no other way to construct nodes with such out-ports! So happy to remove both these checks and the relevant error variants

quantinuum-hugr/src/types.rs
```diff
@@ -42,6 +43,8 @@ pub enum EdgeKind {
     Value(Type),
     /// A reference to a static value definition.
     Static(Type),
+    /// A (perhaps-polymorphic) function (TLD)
+    Function(PolyFuncType),
```
acl-cqc, line 47: Renaming Static+Function to StaticValue and StaticFunction would be a possibility, or renaming Static edges to Const (their only sources) - they were once called ConstE edges...

quantinuum-hugr/src/hugr/validate.rs
```diff
@@ -419,16 +421,18 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
 
         let is_static = match from_optype.port_kind(from_offset).unwrap() {
             EdgeKind::Static(typ) => {
-                if !(OpTag::Const.is_superset(from_optype.tag())
-                    || OpTag::Function.is_superset(from_optype.tag()))
-                {
-                    return Err(InterGraphEdgeError::InvalidConstSrc {
+                let reqd_optype = match typ {
```
acl-cqc, line 424: TBH I think this check is more of an assertion about the spec, than something we really need to check at runtime; there is no other way to construct nodes with such out-ports, and as such I don't think it's even possible to test the error path here. So I'd be happy to remove both these checks and the relevant error variant....?

quantinuum-hugr/src/types.rs
```diff
@@ -40,8 +52,8 @@ pub enum EdgeKind {
     ControlFlow,
     /// Data edges of a DDG region, also known as "wires".
     Value(Type),
-    /// A reference to a static value definition.
-    Static(Type),
+    /// A reference to a static value definition - Const or Function
+    Static(StaticEdgeData),
```
acl-cqc, line 56: This is the bit I'm least sure about. The other possibility is to have two distinct EdgeKinds (i.e. one new) - ideally I'd call them Const (rather than Static) and Function, or perhaps StaticValue and StaticFunction. We even had ConstE edges in the spec at one point....

quantinuum-hugr/src/hugr/validate.rs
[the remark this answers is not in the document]
acl-cqc: Done!

acl-cqc:
Ok, thanks @cqc-alec I think that gave me the direction I needed.
* Gone with Const and Function. Possibly we should rename Const as we now have both Const nodes and Const edges, but since the source of every such edge is such a node, maybe this is OK. In the spec I've defined a class of Static edges as including both of those, and hence have used "Static" interchangeably (read, inconsistently :wink ) with "Const or Function".
* Also a bunch of refactoring/cleaning up via a new "EdgeKind::is_static()" and other tweaks in validate.rs
* I have left NewEdgeKind in the Replace rewrite with a "Static" option, largely because there is no *need* to change it. Could go for Const/Function to be consistent; it strikes me the whole static/const/function distinction here is a bit unnecessary as each of those NewEdgeKind's also gives the index of the port which completely identifies the EdgeKind!!
* Also removed all references to "deBruijn" as that's only relevant when one has nested binders and we don't anymore

ss2165:
quite possible that test coverage wasn't there. I'm out of date on the serialization I think, tagging in @aborgna-q

aborgna-q:
My changes only made sure that the schema we are checking against is up-to-date, but the tests remain quite spotty.

@doug-q was working on adding more coverage over all serialisable hugrs.

specification/hugr.md
```diff
@@ -123,18 +123,23 @@ carry an edge weight:
   ordering. They have no edge weight.
 - `Value` edges carry typed data at runtime. They have a *port* at each end, associated
   with the source and target nodes. They have an `AnyType`as an edge weight.
-- `Static` edges are similar to `Value` edges but carry static data (knowable at
-  compilation time). They have a `CopyableType` as an edge weight.
+- `Const` edges are similar to `Value` edges but carry static data (knowable at
+  compilation time). These has as edge weight a `CopyableType`.
```
cqc-alec, line 127:
```suggestion
  compilation time). These have as edge weight a `CopyableType`.
```

specification/hugr.md
```diff
@@ -123,18 +123,23 @@ carry an edge weight:
...
+  that (unlike values) may be polymorphic---see [Polymorphism](#polymorphism).
 - `ControlFlow` edges represent possible flows of control from one part of the
   program to another. They have no edge weight.
 - `Hierarchy` edges express the relationship between container nodes and their
   children. They have no edge weight.
 
-`Value` and `Static` edges are sometimes referred to as *dataflow* edges.
+It is useful to introduce some terms for broader classes of edge:
+* *Static* edges are the union of the `Const` and `Function` edges
+* *dataflow* edges are the union of `Value` and Static (thus, `Value`, `Const` and `Function`)
```
cqc-alec, line 137:
```suggestion
* *Static* edges are the union of the `Const` and `Function` edges.
* *Dataflow* edges are the union of `Value` and Static (thus, `Value`, `Const` and `Function`).
```

specification/hugr.md
```diff
@@ -998,19 +1002,20 @@ at runtime. In many cases this is desirable.
...
-1. Provides a `Function` type, as per [Type System](#type-system), which may declare TypeParams; *or*
+1. Provides a polymorphic type scheme, as per [Type System](#type-system), which may declare TypeParams;
+   values (TypeArgs) provided for those params will be substituted in. *Or*
 2. The extension may self-register binary code (e.g. a Rust trait) providing a function
-   `compute_signature` that fallibly computes a `Function` type given some type arguments.
+   `compute_signature` that fallibly computes a (perhaps-polymorphic) type scheme given some initial type arguments.
    The operation declares the arguments required by the `compute_signature` function as a list
-   of TypeParams; if this successfully returns a `Function` type, then that may then require
-   additional TypeParams.
+   of TypeParams; if this successfully returns a typ scheme, that may have additional TypeParams
```
cqc-alec, line 1017:
```suggestion
   of TypeParams; if this successfully returns a type scheme, that may have additional TypeParams
```

specification/hugr.md
```diff
@@ -1981,26 +1987,26 @@ one". For example, "1, ✱" means "one edge in, any number out".
...
+| Root           | 0, 0    | 0, 0    | 0, 0    | 0, 0       | 0, 0          | 0, ✱        |          |
+| `FuncDefn`     | 0, 0    | 0, 0    | 0, 0    | 0, *       | 0, 0          | 1, +        | DSG      |
+| `FuncDecl`     | 0, 0    | 0, 0    | 0, 0    | 0, *       | 0, 0          | 1, 0        |          |
+| `AliasDefn`    | 0, 0    | 0, 0    | 0, 0    | 0, 0       | 0, 0          | 1, 0        |          |
+| `AliasDecl`    | 0, 0    | 0, 0    | 0, 0    | 0, 0       | 0, 0          | 1, 0        |          |
+| `Const`        | 0, 0    | 0, 0    | 0, ✱    | 0, 0       | 0, 0          | 1, 0        |          |
+| `LoadConstant` | 0, 1    | +, ✱    | 1, 0    | 0, 0       | 0, 0          | 1, 0        |          |
+| `Input`        | 0, ✱    | 0, ✱    | 0, 0    | 0, 0       | 0, 0          | 1, 0        |          |
+| `Output`       | ✱, 0    | ✱, 0    | 0, 0    | 0, 0       | 0, 0          | 1, 0        |          |
+| `LeafOp`       | ✱, ✱    | ✱, ✱    | ✱, 0    | *, 0       | 0, 0          | 1, 0        |          |
```
cqc-alec, line 2001: Can a `LeafOp` have `Function` inputs?

specification/hugr.md
```diff
@@ -823,34 +821,36 @@ TypeParam ::= Type(Any|Copyable|Eq)
...
           | BoundedUSize(u64)
           | Extensions(Extensions)
-          | Sequence([TypeArg])
+          | Sequence([TypeArg]) -- fits either a List or Tuple TypeParam
           | Opaque(Value)
-          | Variable -- refers to an enclosing TypeParam (binder)
+          | Variable -- refers to an enclosing TypeParam (binder) of non-Type kind
 ```
 
+Note that both TypeArgs of kind both Type and Extensions may also include variables.
```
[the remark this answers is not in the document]
acl-cqc: Good, but I realized these lines were just repeating the `--` comments in the fragment above, so went with the comments

specification/hugr.md
acl-cqc: Yes, I guess so. Quite a few existing uses of "dataflow edges" thus also capitalized.

specification/hugr.md
acl-cqc: Ha (thank you!) - maybe I meant a tip scheme, for dividing gratuities between team members

specification/hugr.md
acl-cqc: Good spot! At the time I wasn't really sure what a LeafOp was, and thought it might include some of the other rows such as LoadConstant or Call, hence `Static: *,0`. However I now see that LeafOp should be MakeTuple, UnpackTuple, Tag, Lift, Noop and extension ops, so indeed, no Const *or* Function inputs *or* outputs. Done, thanks :-)

acl-cqc:
Thanks, Alec!

I've added a new roundtrip serialization test with an op on function values, which test was passing on main but failing to satisfy the schema here; and then updated the schema so it passes. So I think everything I originally wanted to include is now in here.

@aborgna-q serial_hugr.py was introduced at `v1` in #888 but the version was left at `v1` in intervening PRs e.g. #891. I am therefore assuming that I should not be bumping that to `v2` here?

aborgna-q:
> @aborgna-q serial_hugr.py was introduced at `v1` in #888 but the version was left at `v1` in intervening PRs e.g. #891. I am therefore assuming that I should not be bumping that to `v2` here?

Correct. We still haven't released any new (non-alpha) version with the v1 schema, so no need to bump it.

specification/hugr.md
cqc-alec: Actually I was only suggesting to make each bullet here a full sentence, hence the capitalization. But capitalizing Dataflow more generally seems reasonable.

specification/schema/hugr_schema_v1.json
cqc-alec, line 1: I notice that the `description` of `PolyFuncType` is identical to that of `FunctionType`. Perhaps this was a copy-and-paste error?

cqc-alec (approved):

specification/schema/hugr_schema_v1.json
acl-cqc: Good spot, these were autogenerated from the python which has its own docstrings independent of the rust. I've done some updating of all of these.

PR #922, opened 2024-04-11
feat!: Flatten `LeafOp`

aborgna-q:
Moves the variants of `LeafOp` into `OpType`. Closes #817.

It seems `Lift` was missing from the schema, and in its place there was a `TypeApplication` struct. I replaced it with lift.

As an aside, the `CustomOp -> ExternalOp -> {ExtensionOp, OpaqueOp}` chain seems unnecessary. We should be able to merge `CustomOp` and `ExternalOp` into one, and get rid of multiple `as_ref`s in the code.
This doesn't modify the schema, so I'll push it as a separate PR.

PR #923, opened 2024-04-11
refactor!: Merge `CustomOp` and `ExternalOp`.

aborgna-q:
Followup of #922 .

Drops `ExternalOp` and makes `CustomOp` an enum with two boxed variants for `OpaqueOp` and `ExtensionOp`.
The schema remains the same.


hugr/src/hugr/serialize.rs
```diff
@@ -349,10 +349,7 @@ pub mod test {
             let old_op = h_canon.get_optype(node);
             if let OpType::CustomOp(new_op) = new_op {
                 if let OpType::CustomOp(old_op) = old_op {
-                    assert_eq!(
-                        new_op.as_ref().clone().into_opaque(),
-                        old_op.as_ref().clone().into_opaque()
-                    );
+                    assert_eq!(new_op.clone().into_opaque(), old_op.clone().into_opaque());
```
ss2165, line 352: not doing the "into_opaque" here could address the lack of coverage of PartialEq in the other PR. Maybe that change should be in this PR? 

I do worry if that PartialEq implementation is a good idea though, seems likely to induce hard to find bugs.

ss2165 (approved):
nice 👍

hugr/src/hugr/serialize.rs
aborgna-q: The main idea of the `PartialEq` impl is that we can replace this hacky test with just `==`
https://github.com/CQCL/hugr/pull/923/commits/993629457b564cca1eec475f7b42ae779ad71463

In general, a serialization roundtrip should preserve equality. Resolving / unresolving a custom operation is an internal operation that doesn't impact the semantics of the hugr.

ss2165: ok I'm convinced, but scared

aborgna-q: Perhaps I should add some `is_opaque`/`is_extension` methods?

PR #924, opened 2024-04-12
docs: Modify specification of `trunc_u` and `trunc_s`.

### LINKED ISSUE #771: Fix definitions of `trunc_u` and `trunc_s`
The current definitions are:

- `trunc_u<N>`: float to unsigned int. Returns an error when the float is non-finite or cannot be exactly stored in N bits.
- `trunc_s<N>`: float to signed int. Returns an error when the float is non-finite or cannot be exactly stored in N bits.

The implication is that the float must be _exactly_ representable in the target integer type. This differs from [WASM](https://webassembly.github.io/spec/core/bikeshed/#-hrefop-trunc-umathrmtruncmathsfu_m-n-z), [Rust](https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast), and Python, all of which round towards zero before doing the conversion. We should probably do the same.

We should also clarify the effect of `trunc_u` applied to negative values. (Seems consistent to return an error if the rounded-to-zero value is negative.)

cqc-alec:
Turns out the implementation in the constant-folding code is already in line with this modified specification, so this is not a breaking change.

Closes #771 .

### DIFF
--- a/specification/hugr.md
+++ b/specification/hugr.md
@@ -1798,8 +1798,8 @@ Conversions between integers and floats:
 
 | Name           | Inputs    | Outputs                  | Meaning               |
 | -------------- | --------- | ------------------------ | --------------------- |
-| `trunc_u<N>`   | `float64` | `Sum(#(int<N>), #(ErrorType))` | float to unsigned int. Returns an error when the float is non-finite or cannot be exactly stored in N bits. |
-| `trunc_s<N>`   | `float64` | `Sum(#(int<N>), #(ErrorType))` | float to signed int. Returns an error when the float is non-finite or cannot be exactly stored in N bits. |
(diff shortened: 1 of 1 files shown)


ss2165 (approved):
thanks!

cqc-alec:
Hmm I may have been a bit hasty -- it looks like rust (at least by default, since 1.45) performs "float-type `as` integer-type" conversions according to a bunch of rules that include sending non-finite and `NaN` values to well-defined values, so the implementation is not quite in line with the spec. https://github.com/rust-lang/rust/pull/71269
In two minds as to which way to go.

cqc-alec:
Was misreading the code. There is a check for finiteness. The only edge case remaining is casting a negative float to an unsigned; this will result in 0, which I think is adequately described by "rounding towards zero".

PR #925, opened 2024-04-12
docs: add logo svg to readme and spec

ss2165:
Using vector with no size specified but github seems to scale it to something sensible 🤷 

line by line source tracking on logo definition woo


aborgna-q (approved):
:people_hugging:

PR #939, opened 2024-04-16
test: Add a test of instantiating an extension set

croyzor:


hugr/src/extension/op_def.rs
```diff
@@ -635,4 +635,34 @@ mod test {
         );
         Ok(())
     }
+
+    #[test]
+    fn instantiate_extension_delta() -> Result<(), Box<dyn std::error::Error>> {
+        use crate::extension::prelude::{BOOL_T, PRELUDE_REGISTRY};
+
+        let mut e = Extension::new(EXT_ID);
+        let ext_id = e.name().clone();
```
acl-cqc, line 644: nit: can't you just use `EXT_ID` rather than `ext_id` and avoid this?

hugr/src/extension/op_def.rs
```diff
@@ -635,4 +635,34 @@ mod test {
...
+
+        let es = ExtensionSet::singleton(&ext_id);
+        let params: Vec<TypeParam> = vec![TypeParam::Extensions];
+        let args = [TypeArg::Extensions { es: es.clone() }];
+        let mut db_set = ExtensionSet::new();
```
acl-cqc, line 649: nit: can you use `let db_set = ExtensionSet::type_var(0)`? Or, include a non-variable too (so `let mut ... = ExtensionSet::singleton` followed by `insert_type_var`) and check we take the union?

hugr/src/extension/op_def.rs
```diff
@@ -635,4 +635,34 @@ mod test {
...
+        db_set.insert_type_var(0);
+        let fun_ty = FunctionType::new_endo(vec![BOOL_T]).with_extension_delta(db_set);
+
+        let exp_fun_ty = FunctionType::new_endo(vec![BOOL_T]).with_extension_delta(es);
```
acl-cqc, line 653: optional/super-nit: I think I'd finish making `def` before you start on what you're going to do with `def` (so move `let es` and `let exp_fun_ty` after `let def`)

acl-cqc (approved):
A couple of suggestions for shortening / possibly extending but looks good, thanks @croyzor !

PR #940, opened 2024-04-16
feat: `new` methods for leaf ops

aborgna-q:
The struct definitions are now `non_exhaustive`, so they cannot be build with the `Noop {ty: ..}` syntax outside of this crate.

### DIFF
--- a/hugr/src/ops/leaf.rs
+++ b/hugr/src/ops/leaf.rs
@@ -18,6 +18,13 @@ pub struct Noop {
     pub ty: Type,
 }
 
+impl Noop {
+    /// Create a new Noop operation.
+    pub fn new(ty: Type) -> Self {
+        Self { ty }
+    }
+}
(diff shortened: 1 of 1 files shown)


croyzor:
What are we anticipating adding to these structs in the future?

aborgna-q:
Probably nothing, but marking them as `non_exhaustive` lets us add private fields without breaking semver if we need to.
(We could also add public fields, but I think that's less probable).

croyzor (approved):

qartik:
@aborgna-q Could a new release be made that includes this change, perhaps a 0.3.1?

Our upgrade to hugr 0.3 is currently broken as `new` is missing in 0.3.

aborgna-q:
This is now published in `hugr 0.3.1`.
https://crates.io/crates/hugr/0.3.1

PR #948, opened 2024-04-18
fix: serialization round-trips

doug-q:
We add a check during validation, conditioned on `cfg(test)`, that the hugr in question successfully round-trips.

### DIFF
--- a/hugr/src/hugr/serialize.rs
+++ b/hugr/src/hugr/serialize.rs
@@ -337,6 +337,12 @@ pub mod test {
     pub fn check_hugr_schema_roundtrip(hugr: &Hugr) -> Hugr {
         check_hugr_roundtrip(hugr, true)
     }
+
+    /// Serialize and deserialize a HUGR, and check that the result is the same as the original.
+    ///
+    /// If `check_schema` is true, checks the serialized json against the in-tree schema.
+    ///
+    /// Returns the deserialized HUGR.
     pub fn check_hugr_roundtrip(hugr: &Hugr, check_schema: bool) -> Hugr {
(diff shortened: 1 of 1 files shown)


hugr/src/hugr/serialize.rs
```diff
@@ -330,8 +333,11 @@ pub mod test {
     /// Serialize and deserialize a HUGR, and check that the result is the same as the original.
```
aborgna-q, line 333:
```suggestion
    /// Serialize and deserialize a HUGR, and check that the result is the same as the original.
    /// Checks the serialized json against the in-tree schema.
```

hugr/src/hugr/serialize.rs
```diff
@@ -330,8 +333,11 @@ pub mod test {
...
     ///
     /// Returns the deserialized HUGR.
-    pub fn check_hugr_roundtrip(hugr: &Hugr) -> Hugr {
-        let new_hugr: Hugr = ser_roundtrip_validate(hugr, Some(&SCHEMA));
+    pub fn check_hugr_schema_roundtrip(hugr: &Hugr) -> Hugr {
+        check_hugr_roundtrip(hugr, true)
+    }
+    pub fn check_hugr_roundtrip(hugr: &Hugr, check_schema: bool) -> Hugr {
```
aborgna-q, line 339:
```suggestion

    /// Serialize and deserialize a HUGR, and check that the result is the same as the original.
    ///
    /// If `check_schema` is true, checks the serialized json against the in-tree schema.
    ///
    /// Returns the deserialized HUGR.
    pub fn check_hugr_roundtrip(hugr: &Hugr, check_schema: bool) -> Hugr {
```

aborgna-q (approved):

PR #952, opened 2024-04-18
feat!: Add `non_exhaustive` to various enums

aborgna-q:
Mainly to `std_extensions` definitions and error enums.

This missing caused #942 to be a breaking change, since it adds new variants to the `FloatOps` and `IntOpDef` extensions. See the failing semver-checks in #928.

### DIFF
--- a/hugr/src/types.rs
+++ b/hugr/src/types.rs
@@ -115,6 +115,7 @@ pub(crate) fn least_upper_bound(mut tags: impl Iterator<Item = TypeBound>) -> Ty
 
 #[derive(Clone, PartialEq, Debug, Eq, Serialize, Deserialize)]
 #[serde(tag = "s")]
+#[non_exhaustive]
 /// Representation of a Sum type.
 /// Either store the types of the variants, or in the special (but common) case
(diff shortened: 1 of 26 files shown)


hugr/src/types.rs
```diff
@@ -115,6 +115,7 @@ pub(crate) fn least_upper_bound(mut tags: impl Iterator<Item = TypeBound>) -> Ty
 
 #[derive(Clone, PartialEq, Debug, Eq, Serialize, Deserialize)]
 #[serde(tag = "s")]
+#[non_exhaustive]
```
cqc-alec, line 118: This is the only one I'm not sure about; would we ever want other variants here?

cqc-alec (approved):

hugr/src/types.rs
aborgna-q: The current `Unit` variant is just an optimisation for a common case.
Users are not expected to match directly on this enum, but rather to use the uniform interface.

We talked at some point about other special cases we may want to add. Those should probably not require a breaking change if they are only performance-related.

PR #956, opened 2024-04-19
feat: merge basic blocks algorithm

### LINKED ISSUE #561: Merge basic blocks
i.e. when there is an edge from a BB with only one successor, to a BB with only one predecessor, the two can be combined. As a first step this would produce one BB containing two DFG nodes one of which consumes all outputs of the other, but then 2*InlineDFG operations to flatten them should remove false dependencies and enable additional optimizations.

Might require implementing both Replacement and/or InlineDFG rewrites...

acl-cqc:
Any basic block with a single successor that has no other predecessors can be merged. (So long as the successor is not the exit block; nor the entry block, as that implicitly has another predecessor).

Implement via Replace (2 BBs - > one BB containing 2 DFGs, which adopt the original BB's children) and 2*InlineDFG.

Closes #561.

### DIFF
--- a/hugr/src/algorithm/merge_bbs.rs
+++ b/hugr/src/algorithm/merge_bbs.rs
@@ -216,12 +216,21 @@ mod test {
             h.output_neighbours(entry).collect::<HashSet<_>>(),
             HashSet::from([no_b2, exit])
         );
+        // Check the Noop's are in the right blocks
         let [n_op1, n_op2] = h
             .nodes()
             .filter(|n| matches!(h.get_optype(*n), OpType::Noop(_)))
             .collect::<Vec<_>>()
             .try_into()
             .unwrap();
+        let (n_op1, n_op2) = if h.get_parent(n_op1) == Some(entry) {
+            (n_op1, n_op2)
+        } else {
+            (n_op2, n_op1)
+        };
+        assert_eq!(h.get_parent(n_op1), Some(entry));
+        assert_eq!(h.get_parent(n_op2), Some(no_b2));
+        // And the Noop in the entry block is consumed by the custom Test op
         let tst = h
             .nodes()
             .filter(|n| matches!(h.get_optype(*n), OpType::CustomOp(_)))
@@ -229,17 +238,7 @@ mod test {
             .ok()
             .unwrap();
         assert_eq!(h.get_parent(tst), Some(entry));
-        let (nop_entry, nop2) = if h.get_parent(n_op1) == Some(entry) {
-            (n_op1, n_op2)
-        } else {
-            assert_eq!(h.get_parent(n_op2), Some(entry));
-            (n_op2, n_op1)
-        };
-        assert_eq!(h.get_parent(nop2), Some(no_b2));
-        assert_eq!(
-            h.output_neighbours(nop_entry).collect::<Vec<_>>(),
-            vec![tst]
-        );
+        assert_eq!(h.output_neighbours(n_op1).collect::<Vec<_>>(), vec![tst]);
         Ok(())
     }
 }
--- a/hugr/src/algorithm/merge_bbs.rs
+++ b/hugr/src/algorithm/merge_bbs.rs
@@ -88,7 +88,7 @@ fn mk_rep(
             signature: succ_sig.clone(),
         },
     );
-    for (i, _) in succ_sig.output.into_iter().enumerate() {
+    for (i, _) in succ_sig.output.iter().enumerate() {
         replacement.connect(dfg2, i, output, i)
     }
 
@@ -102,10 +102,10 @@ fn mk_rep(
     );
     replacement.connect(dfg1, 0, unp, 0);
     let other_start = tuple_elems.len();
-    for (i, _) in tuple_elems.into_iter().enumerate() {
+    for (i, _) in tuple_elems.iter().enumerate() {
         replacement.connect(unp, i, dfg2, i)
     }
-    for (i, _) in pred_ty.other_outputs.into_iter().enumerate() {
+    for (i, _) in pred_ty.other_outputs.iter().enumerate() {
         replacement.connect(dfg1, i + 1, dfg2, i + other_start)
     }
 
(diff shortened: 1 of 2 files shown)


hugr/src/algorithm/merge_bbs.rs
```diff
@@ -0,0 +1,245 @@
...
+    use crate::{const_extension_ids, type_row, Extension, HugrView};
+
+    use super::merge_basic_blocks;
+
+    const_extension_ids! {
+        const EXT_ID: ExtensionId = "TestExt";
+    }
+    #[test]
+    fn in_loop() -> Result<(), Box<dyn std::error::Error>> {
+        /* -> Noop1 -----> Test -> Exit       -> Noop1AndTest --> Exit
```
acl-cqc, line 164: I feel this shape is a reasonable test, so am not tooo worried about adding more control-flow structures. However it'd definitely be good to add more differently-*typed* blocks e.g. with different numbers of inputs/outputs, values in the branch-control predicate or not, etc.

Cobord:
Is there anything that would want prevention of making basic blocks too big? I had this where if the merged would be too big (option input for what that limit is) for a pass to do quickly enough (because only had smaller basic blocks in mind when doing that one) it would just leave them as they were.

acl-cqc:
> Is there anything that would want prevention of making basic blocks too big? I had this where if the merged would be too big (option input for what that limit is) for a pass to do quickly enough (because only had smaller basic blocks in mind when doing that one) it would just leave them as they were.

Hi @Cobord, thanks for the interest! :)

Hmmm, well, that's quite a complex question. So in general, we favour bringing more code together into fewer basic blocks as this will make it easier to find more opportunities for other optimisations. If a particular pass is inefficient (supralinear or quadratic in BB-size, say) then that could be an issue. However, firstly note that the existing basic block boundaries are not necessarily the best boundaries - blocks are dataflow graph with parallelism, so we could break the block on any cut across. So, it'd be better for optimizations to pick their own arbitrary divisions. The argument for not merging, thus, would be if we wanted to keep the divisions between BBs that existed, presumably because we thought those were good divisions - suggesting that we've already done some deliberate partitioning; and we would only want to do such deliberate partitioning *ahead of* the pass needing it, if there were multiple passes (or repeat instances of the same pass) that want this partitioning.

I think we are a long way from that being a problem!

A second reason for merging basic blocks is that it'll simplify the step of detecting structured control flow (conditionals and tail loops) i.e. these passes will not need to deal with single-pred-single-succ chains. Hence, the solution I'd prefer (if we do want to avoid blocks getting too big) would be to skip the InlineDFGs step, rather than leave BBs unmerged. One could pass in a filter-function `should_inline_DFGs: impl Fn<BasicBlockID> -> bool`, say, that defaults to always-true. I think that's something we could add (straightforwardly) in a follow-up PR if needed, though.

acl-cqc:
> I think we should put this into a new `hugr-passes` crate. This doesn't need any private access into `hugr`; putting passes into another crate ensures it stays clean and ensures we don't do anything naughty.

I am generally for this, but it's not quite as straightforward as it sounds - for starters, we should take as much of the `algorithm` module as we can into `hugr_passes` as well, including `nest_cfgs` in particular (we might not want to move `const_fold` as the standard extensions in the main Hugr crate needs to depend upon that and we can't have cyclic dependencies). However the test code for `nest_cfgs` is also used by the `OutlineCfg` rewrite and that's not easily separated (and its a lot to duplicate). *Secondly* we would need to make `trait HugrMut` pub, and while moving this stuff out is exactly the kind of reason why we *should* make HugrMut `pub`, this is definitely something to do in a separate refactoring PR. I'm having a go at this now, but I think there is enough "meat" in this merge-bbs PR to start looking at it 
as it stands (regardless of the order in which merging and moving happens).....

cqc-alec:
> > I think we should put this into a new `hugr-passes` crate. This doesn't need any private access into `hugr`; putting passes into another crate ensures it stays clean and ensures we don't do anything naughty.
> 
> I am generally for this, but it's not quite as straightforward as it sounds - for starters, we should take as much of the `algorithm` module as we can into `hugr_passes` as well, including `nest_cfgs` in particular (we might not want to move `const_fold` as the standard extensions in the main Hugr crate needs to depend upon that and we can't have cyclic dependencies). However the test code for `nest_cfgs` is also used by the `OutlineCfg` rewrite and that's not easily separated (and its a lot to duplicate). _Secondly_ we would need to make `trait HugrMut` pub, and while moving this stuff out is exactly the kind of reason why we _should_ make HugrMut `pub`, this is definitely something to do in a separate refactoring PR. I'm having a go at this now, but I think there is enough "meat" in this merge-bbs PR to start looking at it as it stands (regardless of the order in which merging and moving happens).....

The main issue seems to be that a lot of _test_ code for non-pass functionality depends on passes that are currently implemented in HUGR. While this is convenient, we could probably organize things differently. Some of that test code (including the tests that use `constant_fold_pass`) could just as well live in the `hugr-passes` crate.

But I agree that this needn't wait for that refactor.

hugr/src/algorithm/merge_bbs.rs
```diff
@@ -0,0 +1,399 @@
...
+            type_row![QB_T, USIZE_T],
+            vec![type_row![]],
+            ExtensionSet::new(),
+            res_t.clone().into(),
+        )?;
+        let [q, u] = bb3.input_wires_arr();
+        let tst = bb3.add_dataflow_op(tst_op, [q, u])?;
+        let pred = lifted_unary_unit_sum(&mut bb3);
+        let bb3 = bb3.finish_with_outputs(pred, tst.outputs())?;
+        println!("ALAN bb1 {bb1:?} bb2 {bb2:?} bb3 {bb3:?}");
```
cqc-alec, line 366: Looks like a debug line left in by mistake?

cqc-alec (approved):

hugr/src/algorithm/merge_bbs.rs
acl-cqc: Oops, thanks!

PR #958, opened 2024-04-22
fix: Schema issues

### LINKED ISSUE #941: Call serialisation schema is broken
Following #906, the `Call` op is now defined as

https://github.com/CQCL/hugr/blob/d69198eb57bf77f32538e1ba8de1f308815a067d/hugr/src/ops/dataflow.rs#L154-L159

However, the corresponding pydantic model doesn't match up:

https://github.com/CQCL/hugr/blob/d69198eb57bf77f32538e1ba8de1f308815a067d/hugr-py/src/hugr/serialization/ops.py#L246-L264

Also, the `assert isinstance(fun_ty, PolyFuncType)` will always fail since `fun_ty` is a `Type`, which can never contain a `PolyFuncType`.

mark-koch:
Fixes several schema issue. Closes #941


mark-koch:
LGTM, thank you! 👍 

(I can't approve because I made the PR)

PR #959, opened 2024-04-22
refactor!: Replace SmolStr identifiers with wrapper types.

### LINKED ISSUE #834: Stop using SmolStr directly for identifiers
We currently use `SmolStr` for operation names, type names, parameter names, ...

This makes the types easy to mix around and produces confusing signatures.
We should avoid this stringly-typed situation by adding wrappers (like the `TypeName` introduced in #833) and making sure that they are used everywhere.

Note that `CustomConcrete` mixes type and op names, so we'll need some extra logic there.
https://github.com/CQCL/hugr/blob/6787b1aaeb7132cc37eb5b6ae8e64a684c8b98dd/src/extension.rs#L178-L182

aborgna-q:
Uses [`string_newtype`](https://github.com/ABorgna/string-newtype) to define `TypeName`, `OpName`, and `ValueName` types, to avoid using stringly-typed methods everywhere.
Also defines `TypeNameSlice`, `OpNameSlice`, `ValueNameSlice`. These that act like `str` to a `String`, or lke `Path` to a `PathBuf`; unsized slice types used through references. 

- Renames the `OpName` trait to `NamedOp`. Now `NamedOp::name` returns an `OpName` value. 

Closes #834.

As an example of why this is useful, I caught some cases where `ExtensionBuildError` was being built with the wrong variant due to mixing up the different identifiers, since everything used `SmolStr`.

### DIFF
--- a/hugr/src/extension.rs
+++ b/hugr/src/extension.rs
@@ -11,15 +11,14 @@ use std::sync::Arc;
 
 use thiserror::Error;
 
-use crate::core::Identifier;
 use crate::hugr::IdentList;
-use crate::ops::constant::ValueName;
+use crate::ops::constant::{ValueName, ValueNameSlice};
 use crate::ops::custom::{ExtensionOp, OpaqueOp};
-use crate::ops::{self, OpName};
+use crate::ops::{self, OpName, OpNameSlice};
 use crate::types::type_param::{check_type_args, TypeArgError};
 use crate::types::type_param::{TypeArg, TypeParam};
-use crate::types::FunctionType;
 use crate::types::{check_typevar_decl, CustomType, Substitution, TypeBound, TypeName};
+use crate::types::{FunctionType, TypeNameSlice};
 
 #[allow(dead_code)]
 mod infer;
@@ -178,7 +177,7 @@ pub enum SignatureError {
 
 /// Concrete instantiations of types and operations defined in extensions.
 trait CustomConcrete {
-    type Identifier: Identifier;
+    type Identifier;
     /// A generic identifier to the element.
     ///
     /// This may either refer to a [`TypeName`] or an [`OpName`].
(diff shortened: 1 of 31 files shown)


ss2165 (commented):
I like the results - especially catching those bugs in the errors. But it seems to be a lot of overhead - a whole other crate to depend on, and three new pub definitions per thing (including a marker type that will need to be  hidden from docs). If there aren't any obvious simplifications to be made I guess we can go ahead.

aborgna-q:
The new dependency is mostly for implementing all the required traits, it's a lot of boilerplate just to avoid having bounds on the Marker type. I could include that here, but it's a lot of stuff irrelevant to this crate.

We could drop the `*Slice` definitions and just use references to the owned types as parameters. It's less efficient, but may not make a big difference.
It may make some callsites more verbose though, as they'll require `.clone()`s.

hugr/src/ops.rs
```diff
@@ -301,12 +300,22 @@ macro_rules! impl_op_name {
...
+/// Marker for the [`OpName`] wrapper.
+#[doc(hidden)]
+pub enum OpNameMarker {}
+
+/// A unique identifier for a operation.
+pub type OpName = string_newtype::SmolStrBuf<OpNameMarker>;
+
+/// Slice of a [`OpName`] operation identifier.
+pub type OpNameSlice = string_newtype::SmolStrRef<OpNameMarker>;
+
```
aborgna-q, line 312: @doug-q Requesting a second opinion here. This looks a bit boilerplate-y/confusing, but it may reduce the amount of code on the callsites.

An alternative is to drop `OpNameSlice` and just use references to cloned `OpName`s for parameters.

hugr/src/extension.rs
```diff
@@ -177,18 +177,22 @@ pub enum SignatureError {
 
 /// Concrete instantiations of types and operations defined in extensions.
 trait CustomConcrete {
+    /// The identifier type for the concrete object.
+    type Identifier;
```
aborgna-q, line 181: I left this associated type even though it always gets instantiated with `SmolStr`.

It doesn't change the way we use this trait but documents a bit better the behaviour.

aborgna-q (commented):
I dropped the `string-newtype` dependency and changed the identifier to aliases of `SmolStr`.
The diff hasn't changed much aside from the identifier definitions.

ss2165 (approved):
thanks for the changes 👍

PR #967, opened 2024-04-23
feat!: bring back Value

doug-q:
This change is primarily motivated by the serialisation schema. With Const and Value merged the recursive Const type was an Op , so was serialised with "op" and "parent" fields. These fields were only relevant for "root" Consts.

Now, Const is an OpType, while Value is not. 

An unrelated change is included: The serialisation schema for SumType now consists of the correct list of TypeRows.

Note that serialisation still does not match the schema, in particular SumTypes are still not properly handled.

BEGIN_COMMIT_OVERRIDE
feat!: Revert the removal of `Value` (#967)

BREAKING CHANGE: Many uses of `Const` now use `Value`.

END_COMMIT_OVERRIDE

### DIFF
--- a/specification/schema/hugr_schema_v1.json
+++ b/specification/schema/hugr_schema_v1.json
@@ -274,7 +274,6 @@
             "required": [
                 "op",
                 "parent",
-                "input_extensions",
                 "v"
             ],
             "title": "Const",


hugr-py/src/hugr/serialization/ops.py
```diff
@@ -142,14 +136,29 @@ class Config:
...
+class Const(BaseOp):
+    """A Const operation definition."""
+
+    op: Literal["Const"] = "Const"
+    v: Value = Field()
+
+    class Config:
+        # Needed to avoid random '\n's in the pydantic description
+        json_schema_extra = {
+            "required": ["op", "parent", "input_extensions", "v"],
```
aborgna-q, line 160: I don't think this should be required
```suggestion
            "required": ["op", "parent", "v"],
```

aborgna-q (approved):

acl-cqc (commented):
Nice! Thanks @doug-q. Should we remove builder method `add_load_const` now? All the uses I see are passing `Value`s so can become `add_load_value`.

PR #968, opened 2024-04-24
fix!: serialisation schema

doug-q:
BEGIN_COMMIT_OVERRIDE
fix: Serialisation for `Type`, `PolyFuncType`, and `Value` (#968)
END_COMMIT_OVERRIDE

### DIFF
--- a/scripts/generate_schema.py
+++ b/scripts/generate_schema.py
@@ -9,6 +9,7 @@ usage: python generate_schema.py [<OUT_DIR>]
 
 import json
 import sys
+from typing import Type
 from pathlib import Path
 
 from pydantic import TypeAdapter
@@ -17,7 +18,9 @@ from hugr.serialization import SerialHugr
 from hugr.serialization.testing_hugr import TestingHugr
 
 
-def write_schema(out_dir: Path, name_prefix: str, schema):
+def write_schema(
+    out_dir: Path, name_prefix: str, schema: Type[SerialHugr] | Type[TestingHugr]
+):
     version = schema.get_version()
     filename = f"{name_prefix}_{version}.json"
     path = out_dir / filename


devenv.nix
```diff
@@ -11,6 +11,7 @@ in
     pkgs.llvmPackages_16.libllvm
     # cargo-llvm-cov is currently marked broken on nixpkgs unstable
     pkgs-stable.cargo-llvm-cov
+    pkgs.ruff
```
ss2165: isn't this part of the poetry dependencies?

hugr-py/src/hugr/serialization/hugrtype.py
```diff
@@ -0,0 +1,20 @@
...
+class HugrType(BaseModel):
+    """A serializable representation of a Hugr Type. Intended for testing only."""
+
+    typ: Type
+    v: Literal["v1"] = "v1"
+
+    @classmethod
+    def get_version(cls) -> str:
+        """Return the version of the schema."""
+        return cls(typ=Type(USize())).v
```
ss2165: should `v` be a class attribute? Is that something pydantic can handle?

scripts/generate_schema.py
```diff
@@ -7,13 +7,23 @@
...
 import json
 import sys
 from pathlib import Path
 
 from pydantic import TypeAdapter
 
 from hugr.serialization import SerialHugr
+from hugr.serialization.hugrtype import HugrType
+
+def write_schema(out_dir: Path, name_prefix: str, schema):
```
ss2165, line 20: `schema: Type[BaseModel]` ?

hugr/src/hugr/serialize.rs
```diff
@@ -26,16 +26,21 @@ use super::{HugrMut, HugrView};
 /// will try to deserialize them in order.
 #[derive(Serialize, Deserialize)]
 #[serde(tag = "version", rename_all = "lowercase")]
-enum Versioned {
+enum Versioned<T> {
```
ss2165, line 29: why is this generic? it implies a new generic parameter will need to be added for each version of the payload.

hugr/src/hugr/serialize/test.rs
```diff
@@ -67,6 +82,7 @@ pub fn ser_roundtrip_validate<T: Serialize + serde::de::DeserializeOwned>(
                 println!("Validation error: {}", error);
                 println!("Instance path: {}", error.instance_path);
             }
+            dbg!(s);
```
ss2165, line 85: remove?

hugr/src/hugr/serialize.rs
ss2165: From below it looks like this is generic to allow versioned payloads for different types. Maybe call the type parameter `V1` for clarity.

hugr/src/hugr/serialize/test.rs
```diff
@@ -312,3 +328,80 @@ fn serialize_types_roundtrip() {
...
+
+#[rstest]
+#[case(BOOL_T)]
+#[case(USIZE_T)]
+#[case(INT_TYPES[2].clone())]
+#[case(Type::new_alias(crate::ops::AliasDecl::new("t", TypeBound::Any)))]
+#[case(Type::new_var_use(2, TypeBound::Copyable))]
+#[case(Type::new_tuple(type_row![BOOL_T,QB_T]))]
+#[case(Type::new_sum([type_row![BOOL_T,QB_T], type_row![Type::new_unit_sum(4)]]))]
+#[case(Type::new_function(FunctionType::new_endo(type_row![QB_T,BOOL_T,USIZE_T])))]
```
ss2165, line 344: can we use hypothesis/proptest for at least some of these?

ss2165 (changes requested):
Looks like this gets us to a much better place. I'm a bit wary of the combined test-only schema and pydantic model with lots of optional fields. Is this done for convenience or is there something preventing us testing the individual schema?

hugr-py/src/hugr/serialization/hugrtype.py
ss2165: *version

scripts/generate_schema.py
ss2165: in which case you need to constrain it to a type that has that classmethod. Either by defining a common baseclass or protocol for `SerialHugr, TestingHugr` etc. to share or using a `Type[SerialHugr] | Type[TestingHugr]` annotation here

ss2165 (approved):

PR #973, opened 2024-04-25
ci: run ci on all pull_requests

doug-q:


aborgna-q:
Add `pr-title.yml` too

.github/workflows/pr-title.yml
```diff
@@ -2,8 +2,6 @@ name: Check Conventional Commits format
 
 on:
   pull_request_target:
-    branches:
```
cqc-alec, line 5: Is this necessary? Not convinced, but if so, we should clarify what the conventions are for PRs to other branches -- e.g. are they supposed to signal fixes or breaking changes to the branch?

cqc-alec (approved):

PR #974, opened 2024-04-25
fix!: Combine `ConstIntU` and `ConstIntS`

### LINKED ISSUE #970: `ConstIntU` and `ConstIntS` should be the same type
There is only one integer type for signed and unsigned integers; signedness is a property of the operations. Currently we have separate implementations of `ConstIntU` and `ConstIntS`, which makes constant-folding sequences of mixed signed and unsigned operations impossible.

We should have a single `ConstInt` type, with two methods of construction (and possibly a third from bitstrings if we want).

cqc-alec:
Fixes #970 .

Drive-by: change name of integer constant to show width rather than log width, e.g. `u32(1000)` not `u5(1000)`.

(Note that the name always shows the unsigned interpretation. We could make it show both, but I'm not sure it's worth it.)

I assume this will break serialization. Do I need to change the serialization version?


ss2165 (approved):
I don't think this PR needs to worry about bumping serialization version

PR #977, opened 2024-04-26
fix: disable serialisation tests when miri is active

doug-q:
We move the `serialize_types_roundtrip` test to put it under the new cfg flag


aborgna-q (approved):
👍

PR #981, opened 2024-04-30
test: proptests for round trip serialisation of `Type`s and `Op`s.

doug-q:
Note that I make no attempt to generate "valid" types. In particular this means that type variables will be grossly inconsistent. It would be possible to improve this by passing a type environment through all the arbitrary instances in `Parameters`, although it's not clear to me just how far that would take us. 

We can't use `prop_recursive` for mutually recursive types, so I have reimplemented it using `Parameters`.

### DIFF
--- a/hugr/Cargo.toml
+++ b/hugr/Cargo.toml
@@ -52,9 +52,6 @@ delegate = "0.12.0"
 paste = "1.0"
 strum = "0.26.1"
 strum_macros = "0.26.1"
-proptest = { version =  "1.4.0" }
-proptest-derive = { version =  "0.4.0"}
-regex-syntax = { version =  "0.8.3"}
 
 [dev-dependencies]
 criterion = { version = "0.5.1", features = ["html_reports"] }
@@ -64,6 +61,9 @@ urlencoding = "2.1.2"
 cool_asserts = "2.0.3"
 insta = { workspace = true, features = ["yaml"] }
 jsonschema = "0.18.0"
+proptest = { version =  "1.4.0" }
+proptest-derive = { version =  "0.4.0"}
+regex-syntax = { version =  "0.8.3"}
 
 [[bench]]
 name = "bench_main"


hugr/Cargo.toml
```diff
@@ -52,6 +54,9 @@ delegate = "0.12.0"
 paste = "1.0"
 strum = "0.26.1"
 strum_macros = "0.26.1"
+proptest = { version =  "1.4.0", optional = true }
```
acl-cqc, line 57: Presumably these are used only if testing, so can they not go in `dev-dependencies`?

hugr/src/hugr/ident.rs
```diff
@@ -6,8 +6,12 @@ use regex::Regex;
 use smol_str::SmolStr;
 use thiserror::Error;
 
+pub static PATH_COMPONENT_REGEX_STR: &str = r"[\w--\d]\w*";
```
acl-cqc, line 9:
```suggestion
pub static PATH_COMPONENT_REGEX_STR: &str = r"[\w_\d]\w*";
```
Not sure if we allow `-`s anywhere, but I'd be surprised if we wanted path components to start with `-`.

Hmmm. We probably *do* allow such, though! So maybe just `-_`

hugr/src/hugr/ident.rs
```diff
@@ -23,6 +27,7 @@ lazy_static! {
     serde::Deserialize,
 )]
 /// A non-empty dot-separated list of valid identifiers
+
```
acl-cqc, line 30: nit: newline belongs before comment, not after

hugr/src/hugr/serialize/test.rs
```diff
@@ -64,6 +62,25 @@ include_schema!(
     "../../../../specification/schema/testing_hugr_schema_strict_v1.json"
 );
 
+macro_rules! impl_sertesting_from {
```
acl-cqc, line 65: oh, I was kinda hoping we wouldn't need this anymore! Ah well

hugr/src/hugr/serialize/test.rs
```diff
@@ -349,79 +367,34 @@ fn serialize_types_roundtrip() {
...
-#[case(ops::Call::try_new(polyfunctype1(), [TypeArg::BoundedNat{n: 1}, TypeArg::Extensions{ es: ExtensionSet::singleton(&PRELUDE_ID)} ], &EMPTY_REG).unwrap())]
-#[case(ops::CallIndirect { signature : FunctionType::new_endo(type_row![BOOL_T]) })]
-fn roundtrip_optype(#[case] optype: impl Into<OpType> + std::fmt::Debug) {
-    check_testing_roundtrip(NodeSer {
-        parent: portgraph::NodeIndex::new(0).into(),
-        input_extensions: None,
-        op: optype.into(),
-    });
+        #[test]
+        fn prop_roundtrip_optype(op in ((0..(std::u32::MAX / 2) as usize). prop_map(|x| portgraph::NodeIndex::new(x).into()), any::<Option<ExtensionSet>>(), any::<OpType>()).prop_map(|(parent, input_extensions, op)| NodeSer { parent, input_extensions, op })) {
```
acl-cqc, line 396: This is a little hard to read...is it possible to outline this `NodeSer`-factory? I think the overall is `fn prop_roundtrip_optype(op in <thing>)` which is presumably a proptest-specific construct, but maybe that can be an `Iterator<NodeSer>` or something? (A shame to have to construct `NodeSer`s really - ideally `prop_roundtrip_optype` should take an Optype, but I guess that doesn't fit the serialization format?)

hugr/src/lib.rs
```diff
@@ -136,6 +136,7 @@
 // Unstable check, may cause false positives.
 // https://github.com/rust-lang/rust-clippy/issues/5112
 #![warn(clippy::debug_assert_with_mut_call)]
+#![recursion_limit = "512"]
```
acl-cqc, line 139: Comment? Or can you restrict this to (when using) proptest?

hugr/src/ops/constant.rs
```diff
@@ -547,4 +555,99 @@ mod test {
...
+
+        impl Arbitrary for Value {
+            type Parameters = ();
+            type Strategy = BoxedStrategy<Self>;
+            fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+                use ::proptest::collection::vec;
+                let leaf_strat = prop_oneof![
+                    any::<ExtensionValue>().prop_map(|e| Self::Extension { e }),
+                    prop_oneof![
+                        // TODO we need an example of each legal root, in particular FuncDe{fn,cl}
```
acl-cqc, line 619: I reckon in future we will have `impl Arbitrary for Hugr` - nothing says it has to be able to generate *any* Hugr, but we could cover a good cross-section - but very happy not to do that in this PR

hugr/src/proptest.rs
```diff
@@ -0,0 +1,122 @@
...
+    static ref ANY_STRING: SBoxedStrategy<String> = {
+        use proptest::string::string_regex;
+        prop_oneof![
+            string_regex(r"[[:alpha:]]{0,3}").unwrap(),
+            string_regex(r"[[:alpha:]]*").unwrap(),
+            string_regex(r".*").unwrap(),
+        ].sboxed()
+    };
+
+    static ref ANY_SERDE_YAML_VALUE_LEAF: SBoxedStrategy<serde_yaml::Value> = {
```
acl-cqc, line 64: whoa, hang on. Surely we should not be constructing YAML this way. Arbitrary Rust object, serialize, validate against schema, deserialize, check equality ...?

hugr/src/types.rs
```diff
@@ -448,8 +449,61 @@ pub(crate) fn check_typevar_decl(
...
+
+    #[cfg(feature = "proptest")]
+    mod proptest {
+        use crate::ops::AliasDecl;
+        use crate::proptest::TypeDepth;
+        use crate::types::custom::test::proptest::CustomTypeArbitraryParameters;
+        use crate::types::{CustomType, FunctionType, SumType, TypeBound, TypeEnum, TypeRow};
+        use ::proptest::prelude::*;
+        impl Arbitrary for super::TypeEnum {
+            type Parameters = TypeDepth;
```
acl-cqc, line 467: Presumably you've done this because constructing arbitrary instances without enforcing a depth limit leaves to overly-massive/deep structures??

I wonder if an easier way might be just to *probabilistically* instruct the framework to prefer the leaf constructors over those with more nested components. Maybe that doesn't, or is hard to tune to, give sufficient coverage of the nested components!

acl-cqc: Can you perhaps use the `prop_recursive` thing that appears to have a limit on depth and a target total size? Might that be a much easier way of getting structures of the shape you want?

justfile
```diff
@@ -14,10 +14,13 @@ check:
 
 # Run all the tests.
 test language="[rust|python]" : (_run_lang language \
-        "HUGR_TEST_SCHEMA=\"1\" cargo test --all-features" \
+        "HUGR_TEST_SCHEMA=\"1\" cargo test --features extension_inference" \
         "poetry run pytest"
     )
 
+proptest:
+    cargo test --all-features
```
acl-cqc, line 22: R u sure you want to turn `extension_inference` on here too? ;-)

I guess this makes sense but I'm not 100% sure about the naming

hugr/src/hugr/serialize/test.rs
```diff
@@ -349,79 +367,34 @@ fn serialize_types_roundtrip() {
...
+    use super::check_testing_roundtrip;
+    use crate::extension::ExtensionSet;
+    use crate::ops::{OpType, Value};
+    use crate::types::{PolyFuncType, Type};
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn prop_roundtrip_type(t:  Type) {
+            check_testing_roundtrip(t)
```
acl-cqc, line 382: Yay :). I was a bit confused when I looked at the wealth of methods with `roundtrip` in the name (it might help to rename `ser_roundtrip_validate` to `ser_schema_roundtrip` as it does not call `validate`), but this one looks to do what we want :-). You could remove the `allow_unused` on its declaration now!

acl-cqc (commented):
Looks good to me, many thanks @doug-q . It's a shame the code to come up with instances has to be so tricky but I think that's an inevitable result of using Rust rather than Haskell here ;-). My main thoughts are that
* I don't really understand the details of the `prop_recursive` macros....
* but maybe that's an easier way than having `TypeDepth` ??

acl-cqc:
> We can't use `prop_recursive` for mutually recursive types, so I have reimplemented it using `Parameters`.
> 
> The logic for when to `descend` is not written down here(TODO), one wants to do so as little as possible while guaranteeing you do it at least once between recursing from one type into itself. `Type` being a glorfied newtype around `TypeEnum` is very useful here: `descend` in Type and, because nothing else has a `TypeEnum`, you don't have to descend anywhere else in the `Type` "heirarchy".

....but maybe I haven't understood this. You mean that `prop_recursive` only works for Rust types that directly reference themselves, rather than *via* other structures (e.g. SumType)? Ok, boo... [are we sure we can't get round this by defining a big factory method for a Type that constructs Sums, Customs, etc. all in the same factory?]

I was going to say that Hugr doesn't have recursive Types, but realize now that that's not useful...

The random type variables is fine as long as we don't `validate` the Types, and I agree we don't need to, so long as we check for equality against the original (which I think we do).

ss2165:
The feature flag seems useful for dependencies to optionally use hugrs in their own proptests?

hugr/Cargo.toml
acl-cqc: Ah got it. So these are the `"dep: xxx"` on the RHS of `proptest` in `[features]`

hugr/Cargo.toml
```diff
@@ -22,7 +22,9 @@ bench = false
 path = "src/lib.rs"
 
 [features]
+default = ["proptest"]
```
acl-cqc, line 25: Is it possible to turn the proptest feature on only when `#[cfg(test)]` is true, maybe?

acl-cqc: Also pondering how well this agrees with what's in `justfile`

justfile
```diff
@@ -15,10 +15,14 @@ check:
...
 # Run all the tests.
 test language="[rust|python]" : (_run_lang language \
-        "HUGR_TEST_SCHEMA=\"1\" cargo test --all-features" \
+        "HUGR_TEST_SCHEMA=\"1\" cargo test --features extension_inference" \
         "poetry run pytest"
     )
 
+# Run all proptests
+proptest:
+    cargo test --feature proptest '::proptest'
```
acl-cqc, line 24: `--feature proptest` is on by default, no? So the `'::proptest'` here causes the tests to execute when `cargo test` (/ `cargo test --feature proptest`, which should be equivalent) does not?

acl-cqc (approved):
No blockers, the proptest package doesn't give the easiest code to follow but I think I got there in the end. A few suggestions/places where comments might help. Also you should really global replace "heirarchy" with "hierarchy", the former is not a word in English ("heir" is about passing on your possessions when you die, "hierarchy" comes from Greek "hieros" = "ruler")...

* RecursionDepth might be a good basis for adding legal typeparams, etc., in the future :). Are we sure we don't want to add a TypeBound in there (subsuming CustomTypeParameters)?

* Those --feature flags are doing what we expect, right? `cargo test` on its own does not execute proptests?

acl-cqc:
I think keeping proptests non-`pub` is sensible, the instances have significant "gotchas" like not being well-formed, we will need to consider carefully what we say when/if they become `pub`

PR #982, opened 2024-05-01
test!: test roundtrip serialisation against strict + lax schema

doug-q:
We also include some ops in testing, and apply required fixes.

Note that we intend to replace this `rstest::case` style testing with `proptest`.

BEGIN_COMMIT_OVERRIDE
test: Roundtrip serialisation against both strict and lax schemas (#982)
END_COMMIT_OVERRIDE


cqc-alec (approved):
It would be good to add some documentation, probably in `DEVELOPMENT.md`, of how and when to use the script.

PR #988, opened 2024-05-01
ci: Move file change filter definitions to a file

aborgna-q:
Cleanup PR. Centralises the file filter used by `dorny/paths-filter` when detecting changes in a single `.yml`.

### DIFF
--- a/.github/workflows/ci-py.yml
+++ b/.github/workflows/ci-py.yml
@@ -31,11 +31,7 @@ jobs:
     - uses: dorny/paths-filter@v3
       id: filter
       with:
-        filters: |
-          python:
-            - 'hugr-py/**'
-            - 'pyproject.toml'
-            - 'specification/schema/**'
+        filters: .github/change-filters.yml
 
   check:
     needs: changes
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-py.yml
```diff
@@ -31,11 +31,7 @@ jobs:
     - uses: dorny/paths-filter@v3
       id: filter
       with:
-        filters: |
-          python:
-            - 'hugr-py/**'
-            - 'pyproject.toml'
-            - 'specification/schema/**'
+        filters: .github/change-filters.yml
```
croyzor, line 34: Do we not need to specify whether we want the rust filter or the python filter here?

aborgna-q: It computes both, an we use the one we need. It's not an expensive operation so it doesn't change much.

croyzor: So after this PR, we'll run both `ci-py` and `ci-rs` when relevant python or rust files are changed, but we wont run either when other files change. Is that right?

aborgna-q: Yes, that's the same behaviour we have already.

The job steps check for
```yaml
if: ${{ needs.changes.outputs.python == 'true' }}
```
or
```yaml
if: ${{ needs.changes.outputs.rust == 'true' }}
```
to see if they should run or be skipped

croyzor (approved):

PR #997, opened 2024-05-03
fix!: serialisation fixes

doug-q:
These are the serialisation fixes from the `proptest` branch.

BEGIN_COMMIT_OVERRIDE
fix: Serialization for `Op`s (#997)
END_COMMIT_OVERRIDE

### DIFF
--- a/hugr/src/ops/constant.rs
+++ b/hugr/src/ops/constant.rs
@@ -113,13 +113,6 @@ impl PartialEq for ExtensionValue {
     }
 }
 
-impl ExtensionValue {
-    /// TODO
-    pub fn new(konst: impl CustomConst) -> Self {
-        Self(Box::new(konst))
-    }
-}
-
 /// Struct for custom type check fails.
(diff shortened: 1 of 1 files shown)


hugr/src/types/type_param.rs
```diff
@@ -270,6 +272,7 @@ impl TypeArgVariable {
 pub struct CustomTypeArg {
     /// The type of the constant.
     /// (Exact matches only - the constant is exactly this type.)
+    #[serde(flatten)]
     pub typ: CustomType,
     /// Serialized representation.
     pub value: serde_yaml::Value,
```
mark-koch, line 278: What's the reason for flattening this?

mark-koch (approved):
The schema changes look good to me 👍

The proptest code also looks reasonable, but I'm not really an expert. You might want to send it to someone else for better feedback

mark-koch:
Mhmm the `TaggedOpaque` is not great. I think I still prefer it slightly over the previous version though, but not feeling strongly. I'd be fine with either version

PR #999, opened 2024-05-03
test: serialisation round trip testing for `OpDef`

doug-q:
We also add coverage for roundtripping `CustomSerialized` which was omitted from the previous PR.


hugr-py/src/hugr/serialization/ops.py
```diff
@@ -70,42 +70,55 @@ class FuncDecl(BaseOp):
...
+
+    @classmethod
+    def set_model_config(cls, config: ConfigDict):
+        cls.model_config = config.copy()
+        cls.model_config["extra"] = "allow"
+
+    class Config:
+        json_schema_extra = {
+            "required": ["c"],
+        }
```
mark-koch, line 84: I'm a bit confused with `ConfiguredBaseModel`, why do we need this?

I would have expected something like

```python
class CustomConst(BaseMode, extra=Extra.allow):
   c: str
```

mark-koch (commented):
The schema changes look good to me, I'm just a bit confused with the introduction of `ConfiguredBaseModel` in #982

hugr/src/extension/op_def.rs
```diff
@@ -662,4 +731,82 @@ mod test {
...
+                op_def::LowerFunc, CustomValidator, ExtensionId, ExtensionSet, OpDef, SignatureFunc,
+            },
+            types::PolyFuncType,
+        };
+
+        impl Arbitrary for SignatureFunc {
+            type Parameters = ();
+            type Strategy = BoxedStrategy<Self>;
+            fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+                // TODO there is also  SignatureFunc::CustomFunc, but for now
```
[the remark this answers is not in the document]
acl-cqc: Which means that if the recipient has the same struct+trait code available, he'll get back an instance of the same struct, and this'll work fine (and since for a roundtrip the "recipient" *is* the "sender", that'll be fine). Of course that can't work if the recipient doesn't have the code, but nvm, that's not really relevant here :)

hugr/src/extension/op_def.rs
```diff
@@ -484,6 +491,65 @@ mod test {
...
+                    .collect_vec()
+            };
+
+            extension == other_extension
+                && name == other_name
+                && description == other_description
+                && misc == other_misc
+                && get_sig(signature_func) == get_sig(other_signature_func)
+                && get_lower_funcs(lower_funcs) == get_lower_funcs(other_lower_funcs)
+                && constant_folder.is_none()
```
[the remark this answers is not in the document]
acl-cqc: Did you mean to do something here other than thumbs-up ?

hugr/src/extension/op_def.rs
```diff
@@ -484,6 +491,65 @@ mod test {
...
+                    poly_func,
+                    validate: _,
+                }) => Some(poly_func.clone()),
+                SignatureFunc::CustomFunc(_) => None,
+            };
+
+            let get_lower_funcs = |lfs: &Vec<LowerFunc>| {
+                lfs.iter()
+                    .map(|lf| match lf {
+                        // as with get_sig above, this should break if the heirarchy
```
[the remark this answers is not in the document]
acl-cqc: Doesn't look resolved - this was just a spelling fix

hugr-core/src/std_extensions/arithmetic/int_types.rs
```diff
@@ -276,4 +276,29 @@ mod test {
...
+        use super::{ConstInt, LOG_WIDTH_MAX};
+        use ::proptest::prelude::*;
+        impl Arbitrary for ConstInt {
+            type Parameters = ();
+            type Strategy = BoxedStrategy<Self>;
+            fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+                let signed_strat = (..=LOG_WIDTH_MAX).prop_flat_map(|log_width| {
+                    use i64;
+                    let max_val = (2u64.pow(log_width as u32) / 2) as i64;
+                    let min_val = -max_val - 1;
```
acl-cqc, line 290: Hang on. So let's take `log_width == 5`. Then `max_val = 2u64.pow(5)/2` i.e. `16`. `min_val` is therefore `-17`. You're then creating a value in range `-17..16`...

I'd not expect `Self::new_s` to be happy if you generated `-17` or `16` though. But maybe you're not hitting this? I tried copying your `max_val` and `min_val` computation into a separate test and that lead me tothink my interpretation was correct...

(So, consider: rename `max_val` to `max_valp1`; then `let min_val = -max_valp1`; then use `min_val..<max_valp1`)

acl-cqc (approved):
Generally great ;-) (so yes that's even better than last time!). I think there's a bug in ConstInt64, and some minor suggestions of renames etc. for readability but looks good to me. Thanks @doug-q !

hugr-core/src/extension/op_def.rs
```diff
@@ -474,12 +474,14 @@ impl Extension {
 }
 
 #[cfg(test)]
-mod test {
+pub mod test {
```
acl-cqc, line 477: super-nit: `pub(super)` might have less possibility of misleading the reader

hugr-core/src/extension/op_def.rs
```diff
@@ -494,6 +496,87 @@ mod test {
         const EXT_ID: ExtensionId = "MyExt";
     }
 
+    #[derive(serde::Serialize, serde::Deserialize, Debug)]
+    pub struct SimpleOpDef(OpDef);
+
+    impl SimpleOpDef {
+        pub fn new(op_def: OpDef) -> Self {
+            assert!(op_def.constant_folder.is_none());
```
acl-cqc, line 504: Nice! I was going to comment before, that `SimpleOpDef` was actually `OpDefEqualityWrapper` and the "simple" only came from the `impl Arbitrary for`, but this enforces simplicity, like it :)

hugr-core/src/extension/op_def.rs
```diff
@@ -494,6 +496,73 @@ mod test {
...
+
+    impl From<SimpleOpDef> for OpDef {
+        fn from(value: SimpleOpDef) -> Self {
+            value.0
+        }
+    }
+
+    impl PartialEq for SimpleOpDef {
+        fn eq(&self, other: &Self) -> bool {
+            let OpDef {
```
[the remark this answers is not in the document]
acl-cqc: Fair!

hugr-core/src/ops/constant.rs
```diff
@@ -635,51 +635,38 @@ mod test {
...
             fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
-                use proptest::collection::vec;
-                let signed_strat = (..=LOG_WIDTH_MAX).prop_flat_map(|log_width| {
-                    use i64;
-                    let max_val = (2u64.pow(log_width as u32) / 2) as i64;
-                    let min_val = -max_val - 1;
-                    (min_val..=max_val).prop_map(move |v| {
-                        OpaqueValue::new(
-                            ConstInt::new_s(log_width, v).expect("guaranteed to be in bounds"),
+                // We intentionally do not include `ConstF4` because it does not
```
acl-cqc, line 648:
```suggestion
                // We intentionally do not include `ConstF64` because it does not
```

hugr-core/src/ops/constant.rs
```diff
@@ -635,51 +635,38 @@ mod test {
...
+                prop_oneof![
+                    any::<ConstInt>().prop_map_into(),
+                    any::<CustomSerialized>().prop_map_into()
+                ]
+                .prop_recursive(
+                    3,  // No more than 3 branch levels deep
+                    32, // Target around 32 total elements
+                    3,  // Each collection is up to 3 elements long
+                    |child_strat| {
+                        (Type::any_non_row_var(), vec(child_strat, 0..3)).prop_map(
```
acl-cqc, line 659: That bit of renaming helps a lot here I think, thanks :)

hugr-core/src/ops/constant/custom.rs
```diff
@@ -476,4 +476,43 @@ mod test {
...
+            type Parameters = ();
+            type Strategy = BoxedStrategy<Self>;
+            fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+                let typ = any::<Type>();
+                let extensions = any::<ExtensionSet>();
+                // here we manually construct a serialized `dyn CustomConst`.
+                // The "c" and "v" come from the `typetag::serde` annotation on
+                // `trait CustomConst`.
+                // TODO This is not ideal, if we were to randomly
+                // generate a valid tag(e.g. "c" = "ConstInt") then things will
```
acl-cqc, line 500:
```suggestion
                // generate a valid tag (e.g. "ConstInt") then things will
```
...in that the `"c" =` is not part of the tag. But the idea is that we are relying on the random `any_string` *not* matching any handler for a recognized type, right?

Perhaps change `randomly` on previous line to `accidentally`, and/or `...then the handler for that type will interpret the "v" and fail`

hugr-core/src/std_extensions/arithmetic/int_types.rs
```diff
@@ -276,4 +276,62 @@ mod test {
...
+                });
+
+                prop_oneof![unsigned_strat, signed_strat].boxed()
+            }
+        }
+
+        fn any_signed_int_with_log_width() -> impl Strategy<Value = (u8, i64)> {
+            (..=LOG_WIDTH_MAX).prop_flat_map(|log_width| {
+                let width = 2u64.pow(log_width as u32);
+                let max_val = ((1u64 << (width - 1)) - 1u64) as i64;
```
acl-cqc, line 304: 👍

acl-cqc (approved):

PR #1002, opened 2024-05-07
feat: Add From impls for TypeArg

doug-q:


hugr/src/types/type_param.rs
```diff
@@ -136,6 +136,30 @@ impl From<Type> for TypeArg {
     }
 }
 
+impl From<CustomTypeArg> for TypeArg {
+    fn from(arg: CustomTypeArg) -> Self {
+        Self::Opaque { arg }
+    }
+}
+
+impl From<TypeArgVariable> for TypeArg {
```
acl-cqc, line 145: Originally IIRC the plan was for this struct to be somewhat opaque, and one would use `TypeArg::new_var_use` to construct one. So do we really want/need this?

Other than that, all looks good to me, thanks! :-)

acl-cqc (approved):
Looks good - I like the `impl Into<...>` argument types too :)

PR #1004, opened 2024-05-07
test: Use assert_matches! less

acl-cqc:
There are downsides to `assert_matches!`:
* Extra dependency/import
* Reduced linting - see several cases here where converting to `assert!(matches!(...))` lead to `clippy` noticing that the match could be converted to `.is_ok()` or `.is_some()`
* Increased chance of accidents, see #1006 (preliminary to this)

Hence, this PR adopts the policy of "use `assert_matches` only when it's useful". Specifically, that means only when we can use
```
assert_matches!(SUBJECT, PATTERN => assert!(EXPR))
```
(and while we're at it, where `EXPR` is an equality, use `assert_eq!`), rather than
```
assert!(matches!(SUBJECT, PATTERN if EXPR))
```
because the former gives better error messages in case of failure.

(This is quite a small advantage, and noting the increased chance of accidents, one could also say - let's not use `assert_matches` at all....I'd be happy to do that, too.)

### DIFF
--- a/hugr/src/hugr/rewrite/outline_cfg.rs
+++ b/hugr/src/hugr/rewrite/outline_cfg.rs
@@ -280,7 +280,7 @@ mod test {
         h.validate(&PRELUDE_REGISTRY).unwrap();
         let backup = h.clone();
         let r = h.apply_rewrite(OutlineCfg::new([merge, tail]));
-        assert_matches!(r, Err(OutlineCfgError::MultipleExitEdges(_, _)));
+        assert!(matches!(r, Err(OutlineCfgError::MultipleExitEdges(_, _))));
         assert_eq!(h, backup);
 
         let [left, right]: [Node; 2] = h.output_neighbours(head).collect_vec().try_into().unwrap();
--- a/hugr/src/builder/cfg.rs
+++ b/hugr/src/builder/cfg.rs
@@ -405,7 +405,6 @@ pub(crate) mod test {
     use crate::hugr::validate::InterGraphEdgeError;
     use crate::hugr::ValidationError;
     use crate::{builder::test::NAT, type_row};
-    use cool_asserts::assert_matches;
 
     use super::*;
     #[test]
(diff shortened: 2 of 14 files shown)


hugr/src/hugr/rewrite/outline_cfg.rs
```diff
@@ -280,16 +280,16 @@ mod test {
...
         let backup = h.clone();
         let r = h.apply_rewrite(OutlineCfg::new([merge, tail]));
-        assert_matches!(r, Err(OutlineCfgError::MultipleExitEdges(_, _)));
+        assert!(matches!(r, Err(OutlineCfgError::MultipleExitEdges(_, _))));
         assert_eq!(h, backup);
 
         let [left, right]: [Node; 2] = h.output_neighbours(head).collect_vec().try_into().unwrap();
         let r = h.apply_rewrite(OutlineCfg::new([left, right, head]));
-        assert_matches!(r, Err(OutlineCfgError::MultipleExitNodes(a,b)) => HashSet::from([a,b]) == HashSet::from_iter([left, right, head]));
+        assert_matches!(r, Err(OutlineCfgError::MultipleExitNodes(a,b)) => assert_eq!(HashSet::from([a,b]), HashSet::from_iter([left, right])));
```
acl-cqc, line 288: although there were several other cases where the bool was being computed and dropped, this was the only one where the bool was false...

aborgna-q:
The value of `assert_matches` is in the error messages. Compare
```rust
let x = None;
assert!(matches!(x, Some(42)));
// thread 'test::test_assert' panicked at src/lib.rs:14:9:
// assertion failed: matches!(x, Some(42))
// note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
with
```rust
let x = None;
assert_matches!(x, Some(42));
// thread 'test::test_matches' panicked at src/lib.rs:8:9:
// assertion failed at src/lib.rs:8: `(value does not match pattern)`
//   value: None
// pattern: Some(42)
```
In this case, matching against a pattern instead of doing `.is_ok()` is the intended behaviour.

I agree that adding statements on the matched branches is error-prone. I'm happy with banning those.

aborgna-q:
I'm closing this PR since seem to have agreed on the status quo.
Feel free to reopen it if you want to continue the discussion.

PR #1006, opened 2024-05-07
test: fix some bad assert_matches

acl-cqc:
Turns out that `assert_matches!(subject, pattern => STMT)` executes STMT and drops the result....

We have several cases where STMT returns a bool and we were expecting this to be true, but in fact assert_matches doesn't check the bool. So, change these to `=> assert_eq!`

Indeed, there was one case here where the bool was actually false - fix that assert....

### DIFF
--- a/hugr/src/hugr/rewrite/outline_cfg.rs
+++ b/hugr/src/hugr/rewrite/outline_cfg.rs
@@ -285,7 +285,7 @@ mod test {
 
         let [left, right]: [Node; 2] = h.output_neighbours(head).collect_vec().try_into().unwrap();
         let r = h.apply_rewrite(OutlineCfg::new([left, right, head]));
-        assert_matches!(r, Err(OutlineCfgError::MultipleExitNodes(a,b)) => HashSet::from([a,b]) == HashSet::from_iter([left, right, head]));
+        assert_matches!(r, Err(OutlineCfgError::MultipleExitNodes(a,b)) => assert_eq!(HashSet::from([a,b]), HashSet::from_iter([left, right])));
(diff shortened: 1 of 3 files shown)


hugr/src/hugr/rewrite/outline_cfg.rs
```diff
@@ -285,11 +285,11 @@ mod test {
 
         let [left, right]: [Node; 2] = h.output_neighbours(head).collect_vec().try_into().unwrap();
         let r = h.apply_rewrite(OutlineCfg::new([left, right, head]));
-        assert_matches!(r, Err(OutlineCfgError::MultipleExitNodes(a,b)) => HashSet::from([a,b]) == HashSet::from_iter([left, right, head]));
+        assert_matches!(r, Err(OutlineCfgError::MultipleExitNodes(a,b)) => assert_eq!(HashSet::from([a,b]), HashSet::from_iter([left, right])));
```
acl-cqc, line 288: Erm, ooops

hugr/src/hugr/rewrite/outline_cfg.rs
```diff
@@ -285,11 +285,11 @@ mod test {
 
         let [left, right]: [Node; 2] = h.output_neighbours(head).collect_vec().try_into().unwrap();
         let r = h.apply_rewrite(OutlineCfg::new([left, right, head]));
-        assert_matches!(r, Err(OutlineCfgError::MultipleExitNodes(a,b)) => HashSet::from([a,b]) == HashSet::from_iter([left, right, head]));
```
acl-cqc, line 288: Erm, ooops

PR #1007, opened 2024-05-07
test: Reorg OutlineCfg/nest_cfgs tests so hugr doesn't depend on algorithm

acl-cqc:
This is in preparation for moving `algorithm` out to a separate crate.
* Rewrite tests in `outline_cfg.rs` to use a single CFG, specific to that test
* Add a test where the outlined cfg has multiple incoming edges (to the same node)
* And another where there is an incoming edge and also the entry node (another implicit edge, making this illegal)
* Tidy the left-behind test code in `nest_cfg.rs`
   * Now that the OutlineCfg tests don't depend, `build_cond_then_loop_cfg(bool)` is only called with `false`, so fold
   * Also inline/remove a few other "helpers" whose interface contract is more confusing than helpful
   * And, e.g., use `FunctionType::new_endo` (and remove now-unnecessary `type_row!`)

### DIFF
--- a/hugr/src/algorithm/nest_cfgs.rs
+++ b/hugr/src/algorithm/nest_cfgs.rs
@@ -930,20 +930,19 @@ pub(crate) mod test {
         )?;
         let (split, merge) = build_if_then_else_merge(cfg_builder, &pred_const, &const_unit)?;
 
-        let (head, tail) = if separate_headers {
+        let head = if separate_headers {
             let head = n_identity(
                 cfg_builder
                     .simple_block_builder(FunctionType::new(type_row![NAT], type_row![NAT]), 1)?,
                 &const_unit,
             )?;
-            let tail = build_loop_from_header(cfg_builder, &pred_const, head)?;
             cfg_builder.branch(&head, 0, &split)?;
-            (head, tail)
+            head
         } else {
             // Combine loop header with split.
-            let tail = build_loop_from_header(cfg_builder, &pred_const, split)?;
-            (split, tail)
+            split
         };
+        let tail = build_loop_from_header(cfg_builder, &pred_const, head)?;
         cfg_builder.branch(&merge, 0, &tail)?;
 
         let exit = cfg_builder.exit_block();
--- a/hugr/src/algorithm/nest_cfgs.rs
+++ b/hugr/src/algorithm/nest_cfgs.rs
@@ -865,7 +865,7 @@ pub(crate) mod test {
     }
 
     // Result is merge and tail; loop header is (merge, if separate==true; unique successor of merge, if separate==false)
-    pub fn build_cond_then_loop_cfg(
+    fn build_cond_then_loop_cfg(
         separate: bool,
     ) -> Result<(Hugr, BasicBlockID, BasicBlockID), BuildError> {
         let mut cfg_builder = CFGBuilder::new(FunctionType::new(type_row![NAT], type_row![NAT]))?;
--- a/hugr/src/algorithm/nest_cfgs.rs
+++ b/hugr/src/algorithm/nest_cfgs.rs
@@ -615,7 +615,12 @@ pub(crate) mod test {
         )?;
         let (split, merge) = build_if_then_else_merge(&mut cfg_builder, &pred_const, &const_unit)?;
         cfg_builder.branch(&entry, 0, &split)?;
-        let (head, tail) = build_loop(&mut cfg_builder, &pred_const, &const_unit)?;
+        let head = n_identity(
+            cfg_builder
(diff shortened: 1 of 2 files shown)


hugr/src/algorithm/nest_cfgs.rs
```diff
@@ -855,39 +860,12 @@ pub(crate) mod test {
         Ok(merge)
     }
 
-    // Returns loop tail - caller must link header to tail, and provide 0th successor of tail
```
acl-cqc, line 858: I think it was these awkward interfaces "returns x, caller must do y", which are fairly essential for building "bits of CFG", that lead me to drop them; it is hard to tell that the callsites are correct without jumping around a lot

hugr/src/hugr/rewrite/outline_cfg.rs
```diff
@@ -356,57 +410,72 @@ mod test {
...
+
         for n in other_blocks {
-            assert_eq!(depth(&h, n), 1);
+            assert_eq!(h.get_parent(n), Some(cfg))
         }
-        for n in blocks_to_move {
-            assert_eq!(h.get_parent(n).unwrap(), new_cfg);
+        assert_eq!(h.get_parent(new_block), Some(cfg));
+        assert!(h.get_optype(new_block).is_dataflow_block());
+        let b = h.base_hugr(); // To cope with `h`` potentially being a SiblingMut
```
mark-koch, line 471:
```suggestion
        let b = h.base_hugr(); // To cope with `h` potentially being a SiblingMut
```

hugr/src/hugr/rewrite/outline_cfg.rs
```diff
@@ -248,106 +249,159 @@ pub enum OutlineCfgError {
...
     use cool_asserts::assert_matches;
     use itertools::Itertools;
 
     use super::{OutlineCfg, OutlineCfgError};
 
+    //      /-> left --\
+    // entry            > merge -> head -> tail -> exit
+    //      \-> right -/             \-<--<-/
+    // Result is Hugr plus merge and tail blocks
+    fn build_cond_then_loop_cfg() -> Result<(Hugr, BasicBlockID, BasicBlockID), BuildError> {
```
mark-koch, line 273: Lots of test lines below try to recover `entry, left, right, head, exit` from the `merge, tail` returned by this function. You could consider building a struct instead which I think would make the tests below less noisy:
```rust
//      /-> left --\
// entry            > merge -> head -> tail -> exit
//      \-> right -/             \-<--<-/
struct CondThenLoopCFG {
   h: Hugr,
   entry: BasicBlockId,
   left: BasicBlockId,
   right: BasicBlockId,
   ...
}

impl CondThenLoopCFG {
   fn new() -> Result<Self, BuildError> {
      ...
   }
}

```

hugr/src/algorithm/nest_cfgs.rs
mark-koch: I think I agree, inlining these is a better move 👍

hugr/src/hugr/rewrite/outline_cfg.rs
acl-cqc: Yes, great idea, a side benefit of using the same example for many tests

hugr/src/hugr/rewrite/outline_cfg.rs
```diff
@@ -248,106 +249,159 @@ pub enum OutlineCfgError {
...
         let r = h.apply_rewrite(OutlineCfg::new([left, right, merge]));
-        assert_matches!(r, Err(OutlineCfgError::MultipleEntryNodes(a,b)) => assert_eq!(HashSet::from([a,b]), HashSet::from([left, right])));
+        assert_matches!(r, Err(OutlineCfgError::MultipleEntryNodes(a,b))
+            => assert_eq!(HashSet::from([a,b]), HashSet::from([left, right])));
         assert_eq!(h, backup);
+
+        // The entry node implicitly has an extra incoming edge
+        let r = h.apply_rewrite(OutlineCfg::new([entry, left, merge]));
+        assert_matches!(r, Err(OutlineCfgError::MultipleEntryNodes(a,b))
+            => assert_eq!(HashSet::from([a,b]), HashSet::from([entry, merge])));
```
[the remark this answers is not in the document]
acl-cqc: Great spot on both counts, thanks

hugr/src/hugr/rewrite/outline_cfg.rs
acl-cqc: Done with an rstest fixture, the extra benefit of which is slight, but perhaps non-zero

mark-koch (approved):
Nice, LGTM 👍

PR #1008, opened 2024-05-08
refactor: Run validation before extension inference

### LINKED ISSUE #943: refactor: Validate hugrs before extension inference is run
### Problem:
When we add extension parameters to extension inference, as in #713, we may run into the situation where the type variables don't actually make sense, causing inference to fail with an opaque error message. We check that type variables make sense during validation, but currently this happens *after* extension inference.

### Proposed solution:
- Separate extension validation from the rest of validation
- Run general validation *before* extension inference is run
- Then, run the extension validation
- Down the line, we can think about merging extension inference with validation

Downside: This means traversing the graph for validation twice rather than just once

croyzor:
Resolves #943

### DIFF
--- a/hugr/src/builder.rs
+++ b/hugr/src/builder.rs
@@ -78,7 +78,7 @@
 //! }?;
 //!
 //! // The built HUGR is always valid.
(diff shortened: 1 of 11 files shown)


croyzor:
Superceded by #1011

PR #1010, opened 2024-05-08
test: Fix unconnected ports in extension test

croyzor:
While trying to separate out extension validation from the rest of validation, I found that the graph here was invalid! The extension error gets thrown before it starts complaining about all of the unconnected outputs.


acl-cqc (approved):
Yes, the errors look obvious in hindsight, and this looks better :-). Thanks!

PR #1011, opened 2024-05-08
refactor: Separate extension validation from the rest

### LINKED ISSUE #943: refactor: Validate hugrs before extension inference is run

croyzor:
A further refactor to validation. Note that the CI will fail without #1010 because this change raises the hugr validity errors before the extension errors. This seems like the win #943 is meant to give us 😁 

Resolves #943

### DIFF
--- a/hugr/src/hugr.rs
+++ b/hugr/src/hugr.rs
@@ -8,6 +8,8 @@ pub mod serialize;
 pub mod validate;
 pub mod views;
 
+#[cfg(feature = "extension_inference")]
+use std::collections::HashMap;
 use std::collections::VecDeque;
 use std::iter;
 
@@ -198,7 +200,8 @@ impl Hugr {
         resolve_extension_ops(self, extension_registry)?;
         self.validate_no_extensions(extension_registry)?;
         self.infer_extensions()?;
-        self.validate(extension_registry)?;
+        #[cfg(feature = "infer_extensions")]
+        self.validate_extensions(HashMap::new())?;
         Ok(())
     }
 
--- a/hugr/src/hugr.rs
+++ b/hugr/src/hugr.rs
@@ -196,8 +196,9 @@ impl Hugr {
         extension_registry: &ExtensionRegistry,
     ) -> Result<(), ValidationError> {
         resolve_extension_ops(self, extension_registry)?;
+        self.validate(extension_registry, false)?;
         self.infer_extensions()?;
-        self.validate(extension_registry)?;
+        self.validate(extension_registry, true)?;
         Ok(())
     }
 
--- a/hugr/src/hugr.rs
+++ b/hugr/src/hugr.rs
@@ -8,6 +8,8 @@ pub mod serialize;
 pub mod validate;
 pub mod views;
 
+#[cfg(feature = "extension_inference")]
+use std::collections::HashMap;
 use std::collections::VecDeque;
 use std::iter;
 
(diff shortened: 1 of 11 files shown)


hugr/src/hugr/validate.rs
```diff
@@ -51,15 +48,52 @@ impl Hugr {
...
     pub fn validate_with_extension_closure(
         &self,
         closure: ExtensionSolution,
         extension_registry: &ExtensionRegistry,
     ) -> Result<(), ValidationError> {
-        let mut validator = ValidationContext::new(self, closure, extension_registry);
-        validator.validate()
+        let mut validator = ValidationContext::new(self, extension_registry);
+        validator.validate()?;
+        #[cfg(feature = "extension_inference")]
```
acl-cqc, line 94: Is it plausible to move this #[cfg] onto the `fn validate_with_extension_closure` itself?

hugr/src/hugr.rs
```diff
@@ -196,6 +196,7 @@ impl Hugr {
         extension_registry: &ExtensionRegistry,
     ) -> Result<(), ValidationError> {
         resolve_extension_ops(self, extension_registry)?;
+        self.validate_no_extensions(extension_registry)?;
```
acl-cqc, line 199: What I'd expect to see here is - so far so good exactly this - but after `validate_no_extensions` can we condition on `#[cfg(feature="extension_inference")]` and then do
* If extension inference is on, infer extensions, and then validate *just* extensions (there's no need to repeat the rest of validation, is there?)
* If extension inference is off....nothing more to do since we've validated everything else?

acl-cqc (commented):
Nice work! I was fearing this would be much more involved, i.e. that you'd have to dive into `validate_subtree` and split that out. But I think you are right and we do not need to after all :-).

Can we go a little further, and change `fn validate` to
* Call `validate_no_extensions`, if extension_inference is off
* Call `validate_with_extension_closure`, if it's on?

hugr/src/hugr/validate.rs
acl-cqc: Hmmmm, *or*, leave this as a way of validating extensions *even if `extension_inference` is disabled*? So then the feature-flag doesn't disable any validation, but just changes what the default `validate` and `update_validate` do?

croyzor: I've left this as is, updating `Hugr::validate` to point to different things depending on the flag status instead

hugr/src/hugr.rs
```diff
@@ -196,8 +196,10 @@ impl Hugr {
...
         self.infer_extensions()?;
-        self.validate(extension_registry)?;
+        #[cfg(feature = "infer_extensions")]
```
acl-cqc, line 201: Should we put this around the line above too (in a `{...}` block)? And then drop the `#cfg` gating inside `Hugr.infer_extensions`? (So the latter does its stuff regardless of the feature, and we just avoid calling it unless the feature is on - or unless you *really* want to?)   Does that work?

hugr/src/hugr/validate.rs
```diff
@@ -47,19 +44,68 @@ impl Hugr {
     /// variables.
     /// TODO: Add a version of validation which allows for open extension
     /// variables (see github issue #457)
+    #[cfg(feature = "extension_inference")]
     pub fn validate(&self, extension_registry: &ExtensionRegistry) -> Result<(), ValidationError> {
         self.validate_with_extension_closure(HashMap::new(), extension_registry)
     }
 
+    /// Check the validity of the HUGR, disregarding extension requirements.
+    #[cfg(not(feature = "extension_inference"))]
```
acl-cqc, line 53: Nice! :)

Maybe even common-up the signature and put the `#[cfg]` just inside the method body?

acl-cqc (approved):
Sweet! Looks good to me, and agreed about the win :)

PR #1013, opened 2024-05-08
fix!: OpDef serialisation

doug-q:
BREAKING CHANGE:
* OpRef serialisation schema

BEGIN_COMMIT_OVERRIDE
fix: `OpDef` serialization (#1013)

BREAKING CHANGE: New serialisation schema
END_COMMIT_OVERRIDE

### DIFF
--- a/hugr/src/algorithm.rs
+++ b/hugr/src/algorithm.rs
@@ -2,4 +2,5 @@
 
 pub mod const_fold;
 mod half_node;
+pub mod merge_bbs;
 pub mod nest_cfgs;
--- a/hugr/src/algorithm/merge_bbs.rs
+++ b/hugr/src/algorithm/merge_bbs.rs
@@ -1,0 +1,398 @@
+//! Merge BBs along control-flow edges where the source BB has no other successors
+//! and the target BB has no other predecessors.
+use std::collections::HashMap;
+
+use itertools::Itertools;
+
+use crate::hugr::rewrite::inline_dfg::InlineDFG;
+use crate::hugr::rewrite::replace::{NewEdgeKind, NewEdgeSpec, Replacement};
+use crate::hugr::{HugrMut, RootTagged};
+use crate::ops::handle::CfgID;
+use crate::ops::leaf::UnpackTuple;
+use crate::ops::{DataflowBlock, DataflowParent, Input, Output, DFG};
+use crate::{Hugr, HugrView, Node};
+
+/// Merge any basic blocks that are direct children of the specified CFG
+/// i.e. where a basic block B has a single successor B' whose only predecessor
+/// is B, B and B' can be combined.
+pub fn merge_basic_blocks(cfg: &mut impl HugrMut<RootHandle = CfgID>) {
+    let mut worklist = cfg.nodes().collect::<Vec<_>>();
+    while let Some(n) = worklist.pop() {
+        // Consider merging n with its successor
+        let Ok(succ) = cfg.output_neighbours(n).exactly_one() else {
+            continue;
+        };
+        if cfg.input_neighbours(succ).count() != 1 {
+            continue;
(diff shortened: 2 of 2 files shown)


hugr-py/src/hugr/serialization/ops.py
```diff
@@ -218,6 +209,11 @@ def insert_port_types(self, in_types: TypeRow, out_types: TypeRow) -> None:
         assert len(in_types) == 0
         self.types = list(out_types)
 
+    # class Config:
+    #     json_schema_extra = {
+    #         "required": ["op", "parent", "types"]
+    #     }
+
```
mark-koch, line 216:
```suggestion
```

hugr-py/src/hugr/serialization/ops.py
```diff
@@ -501,24 +497,29 @@ class OpType(RootModel):
...
     description: str  # Human readable description of the operation.
-    inputs: list[tuple[str | None, Type]]
-    outputs: list[tuple[str | None, Type]]
-    misc: dict[str, Any]  # Miscellaneous data associated with the operation.
-    def_: str | None = Field(
-        ..., alias="def"
-    )  # (YAML?)-encoded definition of the operation.
-    extension_reqs: ExtensionSet  # Resources required to execute this operation.
+    misc: Optional[dict[str, Any]] = None
+    signature: Optional[PolyFuncType] = None
```
mark-koch, line 521:
```suggestion
    misc: dict[str, Any] | None = None
    signature: PolyFuncType | None = None
```

mark-koch (approved):

hugr-py/src/hugr/serialization/ops.py
mark-koch: The remaining code usues `|` so it's more consistent. Also https://docs.astral.sh/ruff/rules/non-pep604-annotation/ so I'm confused that Ruff didn't reformat it itself

hugr/src/extension/op_def.rs
```diff
@@ -495,7 +502,10 @@ mod test {
         let type_scheme = PolyFuncType::new(vec![TP], FunctionType::new_endo(vec![list_of_var]));
 
         let def = e.add_op(OP_NAME, "desc".into(), type_scheme)?;
-        def.add_lower_func(LowerFunc::FixedHugr(ExtensionSet::new(), Hugr::default()));
+        def.add_lower_func(LowerFunc::FixedHugr {
+            extensions: ExtensionSet::new(),
+            hugr: Hugr::default(),
```
acl-cqc, line 507: If this is the test I should look at - no this doesn't look good! This is a Hugr that we can splice in in place of an Op (Node), so should be a DFG-rooted thing with the same inputs/outputs as specified in the OpDef's type_scheme....

hugr/src/extension/op_def.rs
```diff
@@ -265,11 +265,19 @@ impl Debug for SignatureFunc {
...
 /// that implements the operation using a set of other extensions.
 #[derive(serde::Deserialize, serde::Serialize)]
+#[serde(untagged)]
 pub enum LowerFunc {
     /// Lowering to a fixed Hugr. Since this cannot depend upon the [TypeArg]s,
     /// this will generally only be applicable if the [OpDef] has no [TypeParam]s.
-    #[serde(rename = "hugr")]
-    FixedHugr(ExtensionSet, Hugr),
+    FixedHugr {
+        /// The extensions required by the [`Hugr`]
```
acl-cqc, line 273: Hmmmm, perhaps to think about in the future, but if we insist that the `Hugr` is DFG-rooted, then we can extract the ExtensionSet from the root node's `extension_reqs`, so I'm not sure we really need this...

acl-cqc (approved):
Thanks!

PR #1015, opened 2024-05-09
chore: add devenv option to not install git hooks

doug-q:
Tested with and without a `devenv.local.nix` with 
```
{
  hugr.installGitHooks = false;
}
```

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -1,52 +1,61 @@
-{ pkgs, lib, inputs, ... }:
+{ pkgs, lib, inputs, config, ... }:
 let
   pkgs-stable = import inputs.nixpkgs-stable { system = pkgs.stdenv.system; };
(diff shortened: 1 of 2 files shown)


aborgna-q (commented):
I'm on the fence on whether we should run `poetry install` directly in the devenv config, instead of adding complexity to the justfile.

- Is there any usecase where we'd run `just install false` manually? The parameter gets confusing there.
- Is it easy to do `poetry install` conditionally on the devenv.nix config?

aborgna-q:
We could have a
```makefile
install:
    poetry install
```
recipe for the non-hooks variant, but that's just renaming things without adding much functionality.

aborgna-q (approved):
I prefer it to the parameter version. Let's go with this 👍

PR #1022, opened 2024-05-10
fix: set `[build-system]` in `hugr-py/pyproject.toml`

doug-q:
This is required for the sdist to work


hugr-py/pyproject.toml
```diff
@@ -13,7 +13,7 @@ classifiers = [
     "Topic :: Scientific/Engineering",
 ]
 name = "hugr"
-version = "0.1.0"
+version = "0.2.0-pre"
```
cqc-alec, line 16: Should maybe separate this into another commit?

cqc-alec (approved):

PR #1024, opened 2024-05-10
feat!: Allow panic operation to have any input and output wires

### LINKED ISSUE #915: Introduce Panic op
Currently `panic` is a function defined in prelude. Change it to an op which can have any inputs and any outputs. 

This allows linear values to be consumed by `panic` in error code-paths, as well as provides vales to satisfy any local obligations(i.e. you have to wire something to the output node of parent of panic).

cqc-alec:
Closes #915 .

BREAKING CHANGE: `PANIC_OP_ID` now instantiated with expected input and output types.

### DIFF
--- a/hugr/src/extension/prelude.rs
+++ b/hugr/src/extension/prelude.rs
@@ -183,7 +183,16 @@ pub fn array_type(size: impl Into<TypeArg>, element_ty: Type) -> Type {
 
 /// Name of the operation in the prelude for creating new arrays.
 pub const NEW_ARRAY_OP_ID: OpName = OpName::new_inline("new_array");
+
 /// Name of the prelude panic operation.
+///
+/// This operation can have any input and any output wires; it is instantiated
+/// with two [TypeArg::Sequence]s representing these. The first input to the
+/// operation is always an error type; the remaining inputs correspond to the
+/// first sequence of types in its instantiation; the outputs correspond to the
+/// second sequence of types in its instantiation. Note that the inputs and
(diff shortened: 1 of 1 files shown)


hugr/src/extension/prelude.rs
```diff
@@ -41,6 +41,49 @@ impl SignatureFromArgs for ArrayOpCustom {
     }
 }
 
+fn list_of_type() -> TypeParam {
```
croyzor, line 44: I think it's better to inline this inside `static_params`

croyzor (approved):
It might be slightly better in terms of future-proofing to have this op take tuples of inputs and outputs, instead of rows. My reasoning is that in the future, when we have row kinded type params, we should be able to define this op without needing `compute_signature`. 

I.e. `inputs : Row. outputs : Row. panic : Error, Tuple(inputs) -> Tuple(outputs)`

If we define it in terms of tuples now, there's less friction if we rewrite it this way

cqc-alec:
> It might be slightly better in terms of future-proofing to have this op take tuples of inputs and outputs, instead of rows. My reasoning is that in the future, when we have row kinded type params, we should be able to define this op without needing `compute_signature`.
> 
> I.e. `inputs : Row. outputs : Row. panic : Error, Tuple(inputs) -> Tuple(outputs)`
> 
> If we define it in terms of tuples now, there's less friction if we rewrite it this way

Good point, though it would make some usages a bit less convenient. I will make a new issue for this.

Issue #996, opened 2024-05-02
`constant_fold_pass()` panics in `apply_rewrite()`


cqc-alec:
For a moderately complex but fully-foldable DFG, `const_fold_pass()` panics, reporting an invalid port.

Test code to reproduce:

```rust
    #[test]
    fn test_folding_pass() {
        // pseudocode:
        //
        // x0 := 3.0
        // x1 := 4.0
        // x2 := fne(x0, x1); // true
        // x3 := flt(x0, x1); // true
        // x4 := and(x2, x3); // true
        // x5 := -10.0
        // x6 := flt(x0, x5) // false
        // x7 := or(x4, x6) // true
        // output x7
        let mut build =
            DFGBuilder::new(FunctionType::new(type_row![], vec![BOOL_T.into()])).unwrap();
        let x0 = build.add_load_const(Value::extension(ConstF64::new(3.0)));
        let x1 = build.add_load_const(Value::extension(ConstF64::new(4.0)));
        let x2 = build.add_dataflow_op(FloatOps::fne, [x0, x1]).unwrap();
        let x3 = build.add_dataflow_op(FloatOps::flt, [x0, x1]).unwrap();
        let x4 = build
            .add_dataflow_op(
                NaryLogic::And.with_n_inputs(2),
                x2.outputs().chain(x3.outputs()),
            )
            .unwrap();
        let x5 = build.add_load_const(Value::extension(ConstF64::new(-10.0)));
        let x6 = build.add_dataflow_op(FloatOps::flt, [x0, x5]).unwrap();
        let x7 = build
            .add_dataflow_op(
                NaryLogic::Or.with_n_inputs(2),
                x4.outputs().chain(x6.outputs()),
            )
            .unwrap();
        let reg = ExtensionRegistry::try_new([
            PRELUDE.to_owned(),
            logic::EXTENSION.to_owned(),
            arithmetic::float_types::EXTENSION.to_owned(),
        ])
        .unwrap();
        let mut h = build.finish_hugr_with_outputs(x7.outputs(), &reg).unwrap();
        constant_fold_pass(&mut h, &reg);
        let expected = Value::true_val();
        assert_fully_folded(&h, &expected);
    }
```

Output with `RUST_BACKTRACE=1`:

```
failures:

---- algorithm::const_fold::test::test_folding_pass stdout ----
thread 'algorithm::const_fold::test::test_folding_pass' panicked at hugr/src/hugr/hugrmut.rs:301:9:
[...]

PR #1026, opened 2024-05-13
fix: NaryLogicOp constant folding

doug-q:
This fixes the test and stops the panic in #996, although the underlying problem remains.


cqc-alec (approved):
Thanks!

PR #1035, opened 2024-05-14
feat: Add `ValidationLevel` tooling and apply to `constant_fold_pass`

doug-q:


hugr/src/algorithm/verify.rs
```diff
@@ -0,0 +1,102 @@
...
+};
+
+#[derive(Debug, Clone, Copy, Ord, Eq, PartialOrd, PartialEq)]
+/// A type for running [HugrMut] algorithms with verification.
+///
+/// Provides [VerifyLevel::run_verified_pass] to invoke a closure with pre and
+/// post verification.
+///
+/// The default level is `None` because verification can be expensive.
+pub enum VerifyLevel {
```
ss2165, line 19: is there a clear distinction between "validation" and "verification"? Should this be called "ValidateLevel" for consistency?

ss2165 (commented):
This looks nice, just a query about naming consistency

hugr/src/algorithm/verify.rs
ss2165: That makes sense, though I worry people would expect some kind of formal verification on passes. I think for something similar in tket1 we used the term "audit" though I don't like that much either.

PR #1041, opened 2024-05-14
chore: add codeowners file

ss2165:
Once merged I can require review from codeowners for main branch protection

which will auto assign team members

We can make this more granular in future

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,6 @@
+# for syntax see https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax
+
+# lower the rule, higher the precedence
+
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+# for syntax see https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax
+
+# lower the rule, higher the precedence
+
+# base rule, hugrverse team is codeowner for whole repo
+* @CQCL/hugrverse
```
aborgna-q, line 6: nitpick, so the "lower" comment is where new things should go.
```suggestion
# base rule, hugrverse team is codeowner for whole repo
* @CQCL/hugrverse

# Add specific rules below.
# Lower the rule, higher the precedence
```

aborgna-q (approved):
👍

PR #1046, opened 2024-05-15
chore: fix undefined env var in justfile

doug-q:


justfile
```diff
@@ -6,8 +6,7 @@ help:
 # setting up the pre-commit hooks.
 setup:
     poetry install
-    # setup git hooks unless environment variable HUGR_JUST_INHIBIT_GIT_HOOKS is non-empty
-    [[ -z $HUGR_JUST_INHIBIT_GIT_HOOKS ]] && poetry run pre-commit install -t pre-commit || true
+    -[[ -z "${HUGR_JUST_INHIBIT_GIT_HOOKS:-}" ]] && poetry run pre-commit install -t pre-commit
```
croyzor, line 9: Is the `-` at the start of the line intentional? What does it do?

croyzor: Okay I see - it ignores errors. Equivalent to the previous `... || true`

croyzor (approved):

ss2165 (approved):
works!

PR #1051, opened 2024-05-15
test: ignore tests which depend on typetag when cfg(miri)

### LINKED ISSUE #1042: Miri errors on `CustomSerialized` tests.
This started with commit c45e6fc67334768ea55c4bd5223af0b7b0cc47ec
See workflow error: https://github.com/CQCL/hugr/actions/runs/9020320833/job/24785216331

All four errors are similar:
```
thread 'ops::constant::custom::test::test_custom_serialized_try_from::case_1' panicked at hugr/src/ops/constant/custom.rs:398:9:
assertion `left == right` failed
  left: ConstUsize(12)
 right: CustomSerialized { [encoded ConstUsize(12)...] }
 ```

This is probably related to a failing `CustomConst` [equality check](
https://github.com/CQCL/hugr/blob/19bac6210aa8c495679bd7c783751e9cde744c61/hugr/src/ops/constant/custom.rs#L86-L90)

Without debugging the problem I suspect it may have to do with typetag.
Since `miri` does not execute run-before-main code, `typetag` does not register the known `CustomConst` variants and so a serialization roundtrip of a boxed `CustomSerialized` does not get downcasted during deserialization.
https://github.com/CQCL/hugr/blob/19bac6210aa8c495679bd7c783751e9cde744c61/hugr/src/ops/constant/custom.rs#L249-L250
This may be exposing some missing logic in the equality checks, as we are comparing a `ConstUsize` with its serialized version instead of `ConstUsize` vs `ConstUsize` as in the non-miri tests.

doug-q:
fixes #1042

### DIFF
--- a/hugr/src/ops/constant/custom.rs
+++ b/hugr/src/ops/constant/custom.rs
@@ -353,6 +353,7 @@ mod test {
     }
 
     #[rstest]
+    #[cfg_attr(miri, ignore = "miri is incompatible with the typetag crate")]
     #[case(scce_usize())]
     #[case(scce_list())]
     fn test_custom_serialized_try_from<
@@ -438,6 +439,7 @@ mod test {
     }
 
     #[rstest]
+    #[cfg_attr(miri, ignore = "miri is incompatible with the typetag crate")]
     #[case(example_custom_serialized())]
     #[case(example_nested_custom_serialized())]
     fn test_try_from_custom_serialized_recursive<CC: CustomConst + PartialEq>(


croyzor (approved):

aborgna-q:
At this rate we could as well remove the miri test.
`hugr` does not have unsafe code, and most interesting tests will need life-before-main in one way or the other.

The tests here served mostly as a check for the underlying portgraph dependency, but that lib is quite stable and we have tests over there.

PR #1061, opened 2024-05-15
feat!: Disallow nonlocal value edges into FuncDefn's

### LINKED ISSUE #936: No edges into FuncDefn's
* Prevent even 'ext' and 'dom' nonlocal edges from entering into FuncDefn's from outside
* Needs both spec updates, and code in validate.rs

acl-cqc:
Closes #936 

Test for `ext` edges only in this PR. (Would be good to have one for `dom` edges too but that can follow.)

BREAKING CHANGE: Hugr's with nonlocal ("Inter-Graph") edges that enter FuncDefns will now fail validation

### DIFF
--- a/hugr/src/hugr/validate.rs
+++ b/hugr/src/hugr/validate.rs
@@ -480,24 +480,23 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
         // any such, then that is an error, but we report that only if the dom/ext
         // relation was otherwise ok (an error about an edge "entering" some ancestor
         // node could be misleading if the source isn't where it's expected)
-        let mut func_entered = None;
-        let found = |func_entered| match func_entered {
-            None => Ok(()),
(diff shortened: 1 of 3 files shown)


mark-koch (approved):
LGTM 👍

acl-cqc:
I fiddled the order of things so that if you have an edge that enters just a FuncDefn, you'll get the "no edge into funcdefn" error; previously you'd have got the "missing order edge" error (and you can't actually add the order edge to the funcdefn, as you would need to do to get the "no edge into funcdefn" error, because funcdefns can't receive order edges).

The test for that can't be constructed with the builder (because the builder always tries to add the order edge, and fails), so would have to duplicate most of `test_ext_edge`. Which doesn't feel worth it to check we've slightly-improved an error message :(. I'm hoping this way is acceptable for being simpler anyway :)...

PR #1068, opened 2024-05-16
ci: publish to pypi on releases or workflow_dispatch for tags

aborgna-q:
Fixes merging release-please PRs not triggering the pypi release workflow.
  This is due to workflows not triggering other workflows when using the default `GITHUB_TOKEN`.
  
The pypi release workflow ran when a tag matching `hugr-py-v**` was pushed by anyone. This was a bit dangerous and prone to fat-fingering.
Now the workflow is triggered either by new releases with a matching tag (as created by release-please), or by manually triggering the workflow on a tag (so we can still trigger it if needed).
This should be harder to trigger by accident.

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -1,12 +1,22 @@
 name: Build and publish python wheels
+# Builds and publishes the wheels on pypi.
+#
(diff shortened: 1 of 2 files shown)


croyzor (approved):
LGTM

PR #1071, opened 2024-05-17
chore: Declare ci_run flag in build.rs

### LINKED ISSUE #1067: Deal with "unexpected `cfg` condition name: `ci_run`
We are getting warnings on nightly because of:

https://blog.rust-lang.org/2024/05/06/check-cfg.html

To silence these we must do one of:
* remove every instance of `#[cfg(ci_run)]`
* add a build.rs that configures expected `cfg`s.
* disable the warning

aborgna-q:
Rust 1.80 will require pre-declaring configuration flags that we want to set via RUSTFLAGS.
See the [blog post](https://blog.rust-lang.org/2024/05/06/check-cfg.html).

Fixes #1067.

### DIFF
--- a/hugr/build.rs
+++ b/hugr/build.rs
@@ -1,0 +1,7 @@
+//! Build script for the `hugr` crate.
+
+fn main() {
+    // We use a `ci_run` RUSTFLAG to indicate that we are running a CI check,
+    // so we can reject debug code using some tools defined in `utils.rs`.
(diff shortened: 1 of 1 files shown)


croyzor:
Can't this be defined in a `Cargo.toml` file?

aborgna-q:
From the blogpost:
> There is currently no way to expect a custom cfg other than with cargo::rustc-check-cfg in a build.rs.

Hopefully they add the option at some point

PR #1072, opened 2024-05-17
ci: Improve times by splitting stable test runs

### LINKED ISSUE #1064: ci: Consider running `cargo test --all-features` and `cargo test `--no-features` in parallel.
This could give some speedup

aborgna-q:
Parallelises running stable tests with `--no-features` and `-all-features`.
This should reduce the critical path latency for the required checks.

I'm not doing this for the other compiler versions, as the delay is not that critical there and splitting workers pollutes the cache.

Closes #1064.


croyzor (approved):
Nice 👍

aborgna-q:
Looks roughly like a 30% improvement :)

PR #1079, opened 2024-05-20
feat!: Remove `PartialEq` impl for `ConstF64`

doug-q:
Comparing floats for equality is notoriously difficult. So we don't.

Note that we change the serialisation round-tripping checks a bit, because `Const` nodes containt `CustomConst`s that do not implement `equals_const` will break equality checks on the Hugrs.

BREAKING CHANGE: Remove `PartialEq` impl for `ConstF64`


hugr/src/std_extensions/arithmetic/float_types.rs
```diff
@@ -110,8 +110,5 @@ mod test {
         assert_eq!(const_f64_1.value(), 1.0);
         assert_eq!(*const_f64_2, 2.0);
         assert_eq!(const_f64_1.name(), "f64(1)");
-        assert!(const_f64_1.equal_consts(&ConstF64::new(1.0)));
```
cqc-alec, line 113: Maybe we should keep (but negate) this test, for coverage.

cqc-alec (approved):

PR #1084, opened 2024-05-20
fix(py):  get rid of pydantic config deprecation warnings

ss2165:
add a test that triggers the warnings

https://docs.pydantic.dev/2.7/migration/#changes-to-config


aborgna-q (approved):
👍 
It's nice to start having tests on the python side.

PR #1085, opened 2024-05-20
chore: correct __version__ in hugr.py

doug-q:


release-please-config.json
```diff
@@ -12,7 +12,11 @@
             "package-name": "hugr",
             "include-component-in-tag": true,
             "draft": false,
-            "prerelease": false
+            "prerelease": false,
+            "extra-files": [{
+                "type": "generic",
```
croyzor, line 17: I think you can get rid of this line and have release-please interpret __init__ as a python file based on its extension?

croyzor: I just meant the `type` field, not the whole `extra-files` bit!

croyzor: But it seems to work... I'm a bit surpised!

croyzor (approved):

PR #1087, opened 2024-05-20
refactor!: No Ports in TypeRow

acl-cqc:
TypeRow is not just used for node input/output signatures, but also for elements of tuples/sums, and FunctionTypes that are not part of Nodes. So, using Port to index into it seems odd...

BREAKING CHANGE: get() and get_mut() now take only usize, so must call PortIndex::index() on the argument first


aborgna-q (approved):
Note that using `usize` was valid before (as it implements `PortIndex`),
but I agree that it's better to separate the ports and rows.

PR #1096, opened 2024-05-22
feat: hugr binary cli tool

### LINKED ISSUE #1095: HUGR validation CLI

ss2165:
Closes #1095

currently validates against std extensions 

can read from stdin with `echo "<json>" | hugr -` 

binary depends on optional feature to limit dependencies, but not put in sub crate to allow `cargo install hugr`

add integration testing of binary behaviour
based on https://docs.rs/assert_cmd/2.0.14/assert_cmd/cmd/struct.Command.html#method.write_stdin

### DIFF
--- a/hugr/Cargo.toml
+++ b/hugr/Cargo.toml
@@ -67,6 +67,9 @@ jsonschema = "0.18.0"
 proptest = { version = "1.4.0" }
 proptest-derive = { version = "0.4.0" }
 regex-syntax = { version = "0.8.3" }
+assert_cmd = "2.0.14"
+predicates = "3.1.0"
+assert_fs = "1.1.1"
 
 [[bench]]
 name = "bench_main"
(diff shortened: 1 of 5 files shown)


aborgna-q:
Does the cli need access to non-public stuff? It would be cleaner to have it as a separate subcrate instead;

Add a `hugr-cli` dir with the relevant code and a `Cargo.toml` with
```toml
[[bin]]
name = "hugr"
path = "src/main.rs"
```

This way we don't need to add a new feature that's useless to lib users.

aborgna-q:
After some discussion; if we want to be able to do `cargo install hugr` (instead of `hugr-cli`) then it's right that we'll need to define the binary there.

I still think the library should go in a subcrate. The `cli` feature here could be renamed to `_cli` so it doesn't appear in `docs.rs` and users that need the args definitions can depend on `hugr-cli` directly.

We can merge this PR as-is, and I'll propose the subcrate in a separate issue.

Issue #486, opened 2023-09-01
Python bindings for builder interface


ss2165:
Will be necessary for python based HUGR transformations

aborgna-q:
Waiting on the pytket-future API
https://github.com/CQCL/tket2/issues/308

ss2165:
The feature is now complete enough to do https://github.com/CQCL/guppylang/issues/257 so closing, follow ups are listed in description above.

PR #1098, opened 2024-05-22
feat(hugr-py): python hugr builder

ss2165:
First steps of #486

- Uses cli tool to validate.
- Just covers DFG building for now but should be extensible in principle.
- Op types are declared explicitly, no "inference".
- `Hugr` does some light stable index maintenance, but rewriting is not prioritised as a use-case. 
- Both inline nested graph building and full graph insertion are expected use cases so are demonstrated.


hugr-py/src/hugr/utils.py
```diff
@@ -0,0 +1,53 @@
...
+    def items(self) -> ItemsView[L, R]:
+        return self.fwd.items()
+
+    def get_left(self, key: R) -> L | None:
+        return self.bck.get(key)
+
+    def get_right(self, key: L) -> R | None:
+        return self.fwd.get(key)
+
+    def insert_left(self, key: L, value: R) -> None:
```
cqc-alec, line 39: There is no check here that `value` isn't already the value of another key. This could lead to inconsistent bimaps. (Similarly for `insert_right`.)

hugr-py/tests/test_hugr_build.py
```diff
@@ -0,0 +1,176 @@
...
+    a, b = h1.add_op(DummyOp(sops.UnpackTuple(parent=-1, tys=row)), [mt])[0, 1]
+    h1.set_outputs([a, b])
+
+    assert h.hugr.to_serial() == h1.hugr.to_serial()
+
+
+def test_multi_out():
+    h = Dfg([INT_T] * 2, [INT_T] * 2)
+    a, b = h.inputs()
+    a, b = h.add_op(DIV_OP, [a, b])[:2]
```
cqc-alec, line 130: Could we just write `[:]` here (assuming we can't omit the indexing)?

hugr-py/tests/test_hugr_build.py
```diff
@@ -0,0 +1,176 @@
...
+        id="int",
+        args=[ARG_5],
+        bound=stys.TypeBound.Eq,
+    )
+)
+
+NOT_OP = DummyOp(
+    # TODO get from YAML
+    sops.CustomOp(
+        parent=-1,
```
cqc-alec, line 23: Would probably friendlier for users not to need to say `parent = -1`; is it avoidable?

hugr-py/tests/test_hugr_build.py
```diff
@@ -0,0 +1,177 @@
...
+    (a,) = h.inputs()
+    h.set_outputs([a, a])
+
+    _validate(h.hugr)
+
+
+def test_add_op():
+    h = Dfg.endo([BOOL_T])
+    (a,) = h.inputs()
+    nt = h.add_op(NOT_OP, [a])
```
cqc-alec, line 102: Did you consider a varargs style (`add_op(op, *args)`) so that you could write `add_op(NOT_OP, a)` here?

cqc-alec (commented):
Looks good, I mainly focused on the test code and left a few comments.

hugr-py/tests/test_hugr_build.py
ss2165: we don't know in general how many ports it has, in fact without a stop value an error is raised

hugr-py/tests/test_hugr_build.py
```diff
@@ -0,0 +1,176 @@
...
+    h.set_outputs([nt])
+
+    _validate(h.hugr)
+
+
+def test_tuple():
+    row = [BOOL_T, QB_T]
+    h = Dfg.endo(row)
+    a, b = h.inputs()
+    t = h.make_tuple([a, b], row)
```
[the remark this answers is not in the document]
ss2165: it is annoying - unfortunately to avoid it generically we need to either add an interface for querying ops for their signature, or take guppy's approach of tracking types where possible while building. I prefer the former approach because it's less flaky but it is quite a lot of extra code, and liable to many edge cases for polymorphic/generic ops.

hugr-py/src/hugr/utils.py
```diff
@@ -0,0 +1,53 @@
...
+        self.fwd[key] = value
+        self.bck[value] = key
+
+    def insert_right(self, key: R, value: L) -> None:
+        self.bck[key] = value
+        self.fwd[value] = key
+
+    def delete_left(self, key: L) -> None:
+        del self.bck[self.fwd[key]]
```
aborgna-q, line 48: Needs a check for `key in self.fwd`

hugr-py/src/hugr/hugr.py
```diff
@@ -0,0 +1,313 @@
...
+    sub_offset: int = 0
+
+
+@dataclass(frozen=True, eq=True, order=True)
+class InPort(Port):
+    pass
+
+
+class ToPort(Protocol):
+    def to_port(self) -> "OutPort": ...
```
aborgna-q, line 26: What is the definition of a `ToPort`? It's implemented for `OutPort` and `Node`, but not for `Port` nor `InPort`.

hugr-py/src/hugr/utils.py
ss2165: won't it just overwrite that entry (like a dict would) ?

hugr-py/tests/test_hugr_build.py
ss2165: annoyingly python is unwilling to just stop an iterator based on how many values it needs to unpack in to - instead it complains that the iterator wants to keep going

hugr-py/src/hugr/hugr.py
```diff
@@ -0,0 +1,313 @@
...
+    def num_out_ports(self, node: Node) -> int:
+        return len(self.out_ports(node))
+
+    def in_ports(self, node: Node) -> Collection[InPort]:
+        # can be optimised by caching number of ports
+        # or by guaranteeing that all ports are contiguous
+        return [p for p in self._links.bck if p.node == node]
+
+    def out_ports(self, node: Node) -> Collection[OutPort]:
+        return [p for p in self._links.fwd if p.node == node]
```
aborgna-q, line 193: Can we have a `Direction` enum, and and methods that choose input/output depending on the parameter?
Those ended up being quite useful on the rust implementation.

hugr-py/src/hugr/hugr.py
```diff
@@ -0,0 +1,313 @@
...
+            ],
+        )
+
+    @classmethod
+    def from_serial(cls, serial: SerialHugr) -> "Hugr":
+        raise NotImplementedError
+
+
+@dataclass()
+class Dfg:
```
aborgna-q, line 231: Move this to another file?

hugr-py/tests/test_hugr_build.py
```diff
@@ -0,0 +1,177 @@
...
+        op_name="idivmod_u",
+        signature=stys.FunctionType(input=[INT_T] * 2, output=[INT_T] * 2),
+        args=[ARG_5, ARG_5],
+    )
+)
+
+
+def _validate(h: Hugr, mermaid: bool = False):
+    # cmd = ["cargo", "run", "--features", "cli", "--"]
+    cmd = ["./target/debug/hugr"]
```
aborgna-q, line 43: This assumes we've built a debug version of our rust tool in this same environment... I see that failing unexpectedly.

Can't we require `hugr` to either be available on PATH, or be defined in an env var?

hugr-py/src/hugr/hugr.py
aborgna-q, line 231: Should we think here about a commands-based DFG construction similar to tket's?
(E.g. `dfg.add(CNot(target=a, control=b))`)

aborgna-q (commented):
Looks clean!

hugr-py/src/hugr/hugr.py
ss2165: should really be `ToOutPort`, used so you can use a node in place of the single out port of the node (common for single return ops)

hugr-py/tests/test_hugr_build.py
ss2165: yeah sorry this is local testing state, I like the env var option

hugr-py/src/hugr/utils.py
ss2165: would raise a KeyError if it's not?

aborgna-q: Uh, right 👍

hugr-py/src/hugr/hugr.py
ss2165: see https://github.com/CQCL/hugr/pull/1098/commits/de22b5a646046296d6c7080819e7eceff69fad4d...

hugr-py/src/hugr/utils.py
cqc-alec: It will overwrite the entry in `self.bck` but not the entry in `self.fwd`. So a subsequent call to `get_right(other_key)` will still return `value`.

mark-koch (commented):
I think so far, the API in its current form should be usable by Guppy.

Below are some thoughts:

* Having to specify all types by hand is a bit annoying but should be fine for Guppy since we have figured out all types once we do Hugr generation.
* Still, if I had to use to builder to create a Hugr by hand, I would probably get annoyed by the verbosity (e.g. having to specify all element types of a tuple I want to unpack)
* Guppy won't be able to use the `DFG` pattern, were a subgraph is constructed independently and then inserted into a larger graph since we might want to insert non-local edges. I assume we can still use `add_op` to add a DFG op and then build the contents? Maybe we want an `add_dfg` method on `Hugr` as well?
* How will the builder handle Const-/Function-/State-/CF-edges? This was one of the main pain points of the Guppy Hugr builder and I think this might complicate the API quite a bit.
* Is it the builder's job to ensure that nodes are serialised in the correct order (e.g. input/output are the first children of a DFG)? This becomes relevant if we want to nest graphs by hand and add our own IO nodes.
* Will the builder take care of inserting state edges when creating a non-local edge?

ss2165:
@mark-koch 
> Guppy won't be able to use the DFG pattern, were a subgraph is constructed independently and then inserted into a larger graph since we might want to insert non-local edges. 

If I understand correctly, this test: https://github.com/CQCL/hugr/blob/600f18d56f2086a7869cccef5fde2b42262426a4/hugr-py/tests/test_hugr_build.py#L225 shows what you are after? (also answers your other question about state order, auto insertion is possible but not sure if good idea yet)

mark-koch:
Ah yes, sorry I must have missed that. That's what I want 👍

hugr-py/src/hugr/hugr.py
```diff
@@ -0,0 +1,435 @@
...
+
+@dataclass(frozen=True, eq=True, order=True)
+class Node(ToPort):
+    idx: int
+    _num_out_ports: int | None = field(default=None, compare=False)
+
+    @overload
+    def __getitem__(self, index: int) -> OutPort: ...
+    @overload
+    def __getitem__(self, index: slice) -> Iterable[OutPort]: ...
```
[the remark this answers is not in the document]
ss2165: I think should be `Iterator` yes

I figured since it is backed by a `range` it might as well be lazy like `range`

hugr-py/src/hugr/hugr.py
```diff
@@ -0,0 +1,435 @@
...
+        dfg.input_node = mapping[dfg.input_node]
+        dfg.output_node = mapping[dfg.output_node]
+        dfg.root = mapping[dfg.root]
+
+        return dfg
+
+    def set_outputs(self, *args: ToPort) -> None:
+        self._wire_up(self.output_node, args)
+
+    def make_tuple(self, tys: Sequence[Type], *args: ToPort) -> Node:
```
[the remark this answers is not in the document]
ss2165: I like the use of Commands, will do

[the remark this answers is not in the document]
ss2165: resolving as outdated

hugr-py/src/hugr/utils.py
ss2165: should be fixed now

hugr-py/src/hugr/_hugr.py
```diff
@@ -0,0 +1,464 @@
...
+
+class Direction(Enum):
+    INCOMING = 0
+    OUTGOING = 1
+
+
+@dataclass(frozen=True, eq=True, order=True)
+class _Port:
+    node: Node
+    offset: int
```
aborgna-q, line 34:
```suggestion
@dataclass(frozen=True, eq=True, order=True)
class _Port:
    node: Node
    offset: int
    direction: ClassVar[Direction]
```

hugr-py/src/hugr/ops.py
```diff
@@ -0,0 +1,135 @@
...
+from typing import Generic, TypeVar, TYPE_CHECKING
+from hugr.serialization.ops import BaseOp, OpType as SerialOp
+import hugr.serialization.ops as sops
+import hugr.serialization.tys as tys
+
+if TYPE_CHECKING:
+    from hugr._hugr import Hugr, Node
+
+
+class Op(ABC):
```
aborgna-q, line 14: Is there no way to unify this with the protocol in `_hugr.py`?
Can't we define the protocol here, and include it from there?

aborgna-q (approved):
Looks good (if I don't look at the missing docstrings).

It's annoying how all the definitions have to be bunched in the same file to avoid cyclic dependencies, it'd be nice to split them somehow.

hugr-py/src/hugr/ops.py
ss2165: accidental file include, gone now

Issue #294, opened 2023-07-26
`hugr-internals` crate


aborgna-q:
Move the internal code there, and re-export public things from `hugr`.

This would be an escape hatch for other dependents requiring low-level access, but we shouldn't provide any stability guarantees there.

See:
- https://crates.io/crates/pin-project-internal
- https://crates.io/crates/serde_derive_internals

PR #1100, opened 2024-05-23
feat!: Move passes from `algorithms` into a separate crate

### LINKED ISSUE #1000: Make separate `hugr-passes` crate, depending on `hugr`
Already there are some passes implemented in the `hugr` crate that should be moved out, such as `constant_fold_pass` and `transform_all_cfgs`, and more are in the pipeline: #978 #945 #818 #561 . These should be moved into a separate crate, to avoid bloating the `hugr` crate and ensure that we have an API suitable for users to write their own passes.

cqc-alec:
Closes #1000 .

Also:
- Moved some tests that were making use of these passes (or their test code) from the `hugr` crate into the `hugr-passes` crate.
- Moved some functions from test code into the `utils` module and made it public.
- Made `Hugr::new()` public. I thought I could avoid this by using the `CFGBuilder` but doing so caused some extension-inference tests to fail, for reasons I couldn't fathom.

Do we want to update the `release-plz` configuration? Not sure exactly what to do here.

BREAKING CHANGE: The `algorithms` module is removed from `hugr`. It's functions are now in a separate `hugr-passes` crate.


hugr/src/lib.rs
```diff
@@ -140,7 +140,6 @@
 // https://github.com/proptest-rs/proptest/issues/447
 #![cfg_attr(test, allow(non_local_definitions))]
 
-pub mod algorithm;
```
aborgna-q, line 143: Once we do #294 we should re-export algorithms here.

Normally we sub-crates are used for specific components of the library and the main library re-exports all non-internal things, but that will have to wait for now.

hugr-passes/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "hugr-passes"
+version = "0.1.0"
+edition = "2021"
```
aborgna-q, line 4:
```suggestion
version = "0.4.0"
edition = { workspace = true }
rust-version = { workspace = true }

license = { workspace = true }
readme = "README.md"
documentation = "https://docs.rs/hugr-passes/"
homepage = { workspace = true }
repository = { workspace = true }
description = "Compiler passes for Quantinuum's HUGR"
keywords = ["Quantum", "Quantinuum"]
categories = ["compilers"]
```
Let's use the same major/minor version for all packages, to avoid confusion.

hugr-passes/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[dependencies]
+hugr = { path = "../hugr" }
```
aborgna-q, line 7:
```suggestion
hugr = { path = "../hugr", version = "0.4.0 }
```
crates.io requires published packages to depend only on other published packages.

Cargo will use the path for local development but ignore it when publishing.

hugr-passes/Cargo.toml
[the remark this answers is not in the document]
aborgna-q: Alternatively we could just remove the `package` sections and default to `changelog_update = true`.
See here a project that does this with a big workspace; their `release-plz.toml` is one line.
https://github.com/trillium-rs/trillium/blob/main/release-plz.toml

aborgna-q:
Ah, and two more tasks:

- Add the to-be-created `hugr-passes/CHANGELOG.md` to `.github/CODEOWNERS`
- Add the new directory to the rust section in `.github/change-filters.yml`

hugr-passes/Cargo.toml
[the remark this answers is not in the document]
cqc-alec: Will add it manually for now, until it's clear that we want this default.

hugr-passes/Cargo.toml
cqc-alec: In that case I can add `version` under `[workspace.package]` in the top-level `Cargo.toml` and use `{workspace = true }` in both packages?

cqc-alec: Do we want to have to make a new release of `hugr` whenever we do a new release of `hugr-passes`? This shouldn't be necessary.

aborgna-q: I was thinking about having a matching minor number (0.x), even if the patch version varies.
But it may get desynchronised too, so I guess it's fine to leave it as 0.1.0.

The releases are done synchronously, but only for packages with changes.
See [this PR](https://www.github.com/trillium-rs/trillium/pull/639) for an example on how this looks.

hugr-passes/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+This project is licensed under Apache License, Version 2.0 ([LICENSE][] or http://www.apache.org/licenses/LICENSE-2.0).
+
+  [API documentation here]: https://docs.rs/hugr/
+  [build_status]: https://github.com/CQCL/hugr/actions/workflows/ci-rs.yml/badge.svg?branch=main
+  [msrv]: https://img.shields.io/badge/rust-1.75.0%2B-blue.svg
+  [crates]: https://img.shields.io/crates/v/hugr-passes
+  [codecov]: https://img.shields.io/codecov/c/gh/CQCL/hugr?logo=codecov
+  [LICENSE]: https://github.com/CQCL/hugr/blob/main/LICENCE
+  [CHANGELOG]: https://github.com/CQCL/hugr/blob/main/CHANGELOG.md
```
aborgna-q, line 58:
```suggestion
  [API documentation here]: https://docs.rs/hugr-passes/
  [build_status]: https://github.com/CQCL/hugr/actions/workflows/ci-rs.yml/badge.svg?branch=main
  [msrv]: https://img.shields.io/badge/rust-1.75.0%2B-blue.svg
  [crates]: https://img.shields.io/crates/v/hugr-passes
  [codecov]: https://img.shields.io/codecov/c/gh/CQCL/hugr?logo=codecov
  [LICENSE]: https://github.com/CQCL/hugr/blob/main/LICENCE
  [CHANGELOG]: https://github.com/CQCL/hugr/blob/main/hugr-passes/CHANGELOG.md
```

cqc-alec: Oops, thanks.

cqc-alec:
> Just noticed that the top-level `CHANGELOG.md` is a symlink to `hugr/CHANGELOG.md`. Should we delete the symlink?

Would then need to update the link in the `hugr/README.md`.

aborgna-q:
Yes! We should also replace the README symlink with a root README explaining the different crates, but that can be done later

hugr-passes/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "hugr-passes"
+version = "0.1.0"
+edition = "2021"
```
aborgna-q, line 4: Last one
```suggestion
edition = { workspace = true }
```

aborgna-q:
Btw, the token used by `release-plz` does not have permission for publishing new crates on crates.io, so the first publish must be manual.
(And then we'll have to add the new crate to the crates.io token's list)

PR #1103, opened 2024-05-23
ci: Disable release-plz publishing hugr-passes for now

aborgna-q:
This is mainly to avoid fails on CI due to `hugr-passes` trying to use the published `hugr 0.4.0`

https://github.com/CQCL/hugr/actions/runs/9208802531/job/25331815962#step:4:325

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -27,3 +27,6 @@ name = "hugr-passes"
 
 # Enable the changelog for this package
 changelog_update = true
+
+# Disabled until the first version is manually published
+publish = false


cqc-alec (approved):

aborgna-q:
Uh, I disabled publishing to crates.io but not creating a release.
So it went and did a release for this "new version" https://github.com/CQCL/hugr/releases/tag/hugr-passes-v0.1.0.

Thankfully it should be harmless, we can `cargo publish` it once the next `hugr` is published.

PR #1108, opened 2024-05-24
refactor: Add a `hugr-core` crate

### LINKED ISSUE #294: `hugr-internals` crate

aborgna-q:
Moves all the code from `hugr` into a `hugr-core` subcrate.
This way, `hugr-passes` (and any other new subcrate) can depend on the core definitions and get reexported in `hugr`.

This PR does not. Change any definition or visibility in the code, as the renamed files are noisy enough.
In particular, having `hugr-core` will let us unseal `HugrInternals` and `HugrMutInternals` so they can be used in `tket2`, but that's work for another PR.

As a bonus, #1100 is no longer a breaking change since we can re-export the library without a cyclic dependency.

Closes #294


hugr-core/src/types.rs
```diff
@@ -260,15 +260,15 @@ impl TypeEnum {
 ///
 /// Examples:
 /// ```
-/// # use hugr::types::{Type, TypeBound};
-/// # use hugr::type_row;
+/// # use hugr_core::types::{Type, TypeBound};
```
ss2165, line 263: examples should not mention `hugr_core`

ss2165 (changes requested):

hugr-core/src/types.rs
aborgna-q: Fixed in 0f23521f212afe234506f9750f3eabf0ef7b4f45

hugr-core/src/builder.rs
```diff
@@ -26,11 +26,11 @@
...
 //!
 //! ```rust
-//! # use hugr::Hugr;
-//! # use hugr::builder::{BuildError, BuildHandle, Container, DFGBuilder, Dataflow, DataflowHugr, ModuleBuilder, DataflowSubContainer, HugrBuilder};
-//! use hugr::extension::prelude::BOOL_T;
-//! use hugr::std_extensions::logic::{NotOp, LOGIC_REG};
-//! use hugr::types::FunctionType;
+//! # use hugr_core::Hugr;
+//! # use hugr_core::builder::{BuildError, BuildHandle, Container, DFGBuilder, Dataflow, DataflowHugr, ModuleBuilder, DataflowSubContainer, HugrBuilder};
+//! use hugr_core::extension::prelude::BOOL_T;
```
ss2165, line 31: should these have `#` in front of them

hugr-core/src/lib.rs
```diff
@@ -0,0 +1,163 @@
+//! Extensible, graph-based program representation with first-class support for linear types.
```
ss2165, line 1: is this a copy paste of the lib docs? another thing to go out of date potentially?

hugr-core/src/builder.rs
aborgna-q: I missed module-level docs when searching for examples...

I think these give a bit of context to the example below, so it's fine to have them explicit.
I changed all `hugr_core::` uses for `hugr::`, to avoid further confusions.

hugr/src/lib.rs
```diff
@@ -132,14 +132,6 @@
 //! println!("{}", serialized);
 //! ```
 
-// Unstable check, may cause false positives.
-// https://github.com/rust-lang/rust-clippy/issues/5112
-#![warn(clippy::debug_assert_with_mut_call)]
-// proptest-derive generates many of these warnings.
-// https://github.com/rust-lang/rust/issues/120363
-// https://github.com/proptest-rs/proptest/issues/447
-#![cfg_attr(test, allow(non_local_definitions))]
```
ss2165, line 141: this isn't needed any more?

ss2165 (approved):
👍

hugr/src/lib.rs
aborgna-q: It's only needed in -core, where we define the proptests

PR #1114, opened 2024-05-28
feat: Make internals of int ops and the "int" CustomType more public.

### LINKED ISSUE #1113: Allow introspection of int ops

doug-q:
closes #1113.


mark-koch (approved):
LGTM 👍

PR #1116, opened 2024-05-28
feat(cli): Add verbosity, return `Hugr` from `run`.

doug-q:


aborgna-q:
> Is adding the dependency for this overkill?

Seems to be a wrapper over [`log`](https://crates.io/crates/log) with a helper to add the clap parameter.

It's behind the binary dependency, so it shouldn't affect other things.
And it's part of the `clap-rs` project, so it should be fine.

hugr/src/cli.rs
```diff
@@ -38,17 +43,23 @@ pub const VALID_PRINT: &str = "HUGR valid!";
...
         if self.mermaid {
             println!("{}", hugr.mermaid_string());
         }
 
         if !self.no_validate {
             hugr.update_validate(registry)?;
-
-            println!("{}", VALID_PRINT);
+            if self.verbosity(Level::Info) {
+                eprintln!("{}", VALID_PRINT);
```
ss2165, line 55: why send this to stderr? Is it more appropriate than stdout?

hugr/tests/cli.rs
```diff
@@ -91,3 +100,21 @@ fn test_bad_json(mut cmd: Command) {
...
+        .failure()
+        .stderr(contains("Error parsing input").not());
+}
+
+#[rstest]
+fn test_valid_silent(test_hugr_string: String, mut cmd: Command) {
+    cmd.arg("-");
+    cmd.write_stdin(test_hugr_string);
+
+    cmd.assert().success().stderr(contains(VALID_PRINT).not());
```
ss2165, line 119: not sure I like the default behaviour being silent. I was emulating e.g. mypy which prints a success message (we could also use colours for the messages 👀 )

ss2165: yeah happy with default at Info

ss2165 (approved):

PR #1118, opened 2024-05-28
ci: Stop `benches` ci job from always running

aborgna-q:
The `benches` CI job likes to run even when it is not required
https://github.com/CQCL/hugr/actions/runs/9270412094

It seems the `if` statement is wrong. *Why* is it wrong escapes me.
Here I'm just copying the one from the other conditional tests, which we know works.

drive-by: Add a name to make clear it is not _running_ the benchmarks.

### DIFF
--- a/.github/workflows/ci-rs.yml
+++ b/.github/workflows/ci-rs.yml
@@ -61,8 +61,9 @@ jobs:
           RUSTDOCFLAGS: "-Dwarnings"
 
   benches:
+    name: Build benchmarks 🏋️
     needs: changes
-    if: ${{ needs.changes.outputs.rust == 'true' }} && github.event_name != 'merge_group'
+    if: ${{ needs.changes.outputs.rust == 'true' && github.event_name != 'merge_group' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4


.github/workflows/ci-rs.yml
```diff
@@ -61,8 +61,9 @@ jobs:
           RUSTDOCFLAGS: "-Dwarnings"
 
   benches:
+    name: Build benchmarks 🏋️
     needs: changes
-    if: ${{ needs.changes.outputs.rust == 'true' }} && github.event_name != 'merge_group'
+    if: ${{ needs.changes.outputs.rust == 'true' && github.event_name != 'merge_group' }}
```
cqc-alec, line 66: Can probably omit the `${{` and `}}`? But not important.

aborgna-q: I think they always enclose the full `if` expression in brackets:
https://docs.github.com/en/actions/learn-github-actions/expressions

cqc-alec: > I think they always enclose the full `if` expression in brackets: https://docs.github.com/en/actions/learn-github-actions/expressions

Seems for _most_ `if` expressions you can omit them: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

But probably good practice to keep them as there are one or two exceptions.

PR #1123, opened 2024-05-29
feat: Add `ConstExternalSymbol` to prelude

### LINKED ISSUE #233: Const-Decl
i.e. we should allow declaring that there's an external Const (in another Hugr), without giving its value here.

This may mean we need to add a "name" to each Const, paralleling {Alias,Func}{Defn,Decl}. Or, perhaps we should revisit Func{Defn,Decl} having names, and say these are metadata as they affect only the linker.

doug-q:
closes #233.

### DIFF
--- a/hugr-core/src/extension/prelude.rs
+++ b/hugr-core/src/extension/prelude.rs
@@ -2,7 +2,7 @@
 //! operations and constants.
 use lazy_static::lazy_static;
 
-use crate::ops::constant::ValueName;
+use crate::ops::constant::{CustomCheckFailure, ValueName};
 use crate::ops::{CustomOp, OpName};
 use crate::types::{SumType, TypeName};
 use crate::{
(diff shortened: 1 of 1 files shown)


hugr-core/src/extension/prelude.rs
```diff
@@ -343,6 +343,62 @@ impl CustomConst for ConstError {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+/// A structure for holding references to external symbols.
+pub struct ConstExternalSymbol {
+    /// The symbol name that this value refers to. Must be nonempty.
+    pub symbol: String,
```
aborgna-q, line 350: Should we add a type alias for this, akin to `ValueName` / `ValueNameRef` ?

aborgna-q (approved):
LGTM

PR #1124, opened 2024-05-29
ci: Pin nightly rust version

aborgna-q:
Nightly rust tests are failing, seemingly due to [rust-lang/rust#125474](https://github.com/rust-lang/rust/issues/125474).

This PR fixes the toolchain to the last working nightly until the problem is resolved.

CI fail: https://github.com/CQCL/hugr/actions/runs/9281876942/job/25538729650#step:6:856

### DIFF
--- a/.github/workflows/ci-rs.yml
+++ b/.github/workflows/ci-rs.yml
@@ -122,7 +122,9 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        rust: ['1.75', beta, nightly]
+        # Pinned nightly version until this gets resolved:
+        # https://github.com/rust-lang/rust/issues/125474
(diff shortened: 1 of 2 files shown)


cqc-alec (approved):

aborgna-q:
Uhm, the `ci-rs` workflow is not getting triggered...

aborgna-q:
I had replaced the `uses: dtolnay/rust-toolchain@nightly` line with `uses: dtolnay/rust-toolchain` and an explicit parameter for the nightly version,
but it seems that missing the branch name when specifying the action causes the workflow to silently fail (and not get triggered).

u.u

The fix was to add a `@master` prefix.

PR #1129, opened 2024-05-29
fix(hugr-core): Set initial version of hugr-core to 0.1.0

doug-q:


aborgna-q:
Yes. Initially I wasn't sure how would release-plz process it once fully configured, but it seems not to bump the versions for new crates.
So `0.1.0` should be the correct value.

Just be sure to also update the dependency versions on the other crates.

aborgna-q (approved):

PR #1135, opened 2024-05-29
ci: use binary build stage in python CI

ss2165:


.github/workflows/ci-py.yml
```diff
@@ -68,8 +71,59 @@ jobs:
...
+        uses: dtolnay/rust-toolchain@stable
+      - name: Build HUGR binary
+        run: cargo build -p hugr-cli
+      - name: Upload the binary to the artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: hugr_binary
+          path: target/debug/hugr
+  test:
+    needs: [changes, check, build_binary]
```
aborgna-q, line 97: No need to wait for the `check` step to run this one.
```suggestion
    needs: [changes, build_binary]
```

aborgna-q (approved):
LGTM 👍

aborgna-q:
Uhm. Should updates to the rust crate trigger the python tests now?

ss2165:
Ideally only changes to the serialisation should right?

aborgna-q:
Ideally yes.
If we trust the rust-side test to check that we're not breaking the cli tool then we should be OK.

ss2165:
> If we trust the rust-side test to check that we're not breaking the cli tool then we should be OK.

this should be done using @doug-q 's suggestion of a collection of "v1 HUGRs" stored as json that we validate to ensure compatibilty

PR #1136, opened 2024-05-29
refactor: remove `default_members` field in workspace Cargo.toml

ss2165:
All sub crates will be targeted 

specifically now `cargo run` will hit the cli binary


aborgna-q (approved):
Fine by me either way. We can always add `default-members` back if needed.

PR #1137, opened 2024-05-30
WIP/FYI: Subset constraints

acl-cqc:
Around 20 tests failing; some probably need updating, but there are probably still some bugs yet. (Also some internal/unit tests of the inference algorithm that don't compile now, which I've commented out.) However I hope this includes basically the new code that we need, with only "tidying" and bugfixing still to come.

And, of course, there is Lukas' suggestion, maybe we don't need `input_extensions` (or an algorithm to compute them) at all...


acl-cqc:
Think we are going for #1142

Issue #1143, opened 2024-05-31
merge_bbs creating too many copies of self-loop edge


acl-cqc:
This relates to #1144 which reveals the bug (by making the Hugr fail validation), previously the multiple-edges had been going unnoticed.

The `merge_bbs.rs` test `in_loop` with `self_loop: true` replaces two BBs with one BB that has an edge back to itself. After performing the Replacement, the new BB in the Hugr has *four* edges back to itself (all from the same port).
I've checked the `Replacement` - the `replacement` Hugr contains exactly one self-edge, `mu_out` contains exactly one edge leaving the `replacement`, `mu_inp` is empty - all as expected.

The problem appears to arise immediately after this: https://github.com/CQCL/hugr/blob/1032b7a0cc7822371177e9877632580b1ed8daba/hugr-core/src/hugr/rewrite/replace.rs#L282

i.e. before we even inspect `mu_inp` / `mu_out` etc. So I believe this is due to https://github.com/CQCL/portgraph/issues/130 but I'm not clear why we see *four* copies of the edge as from that portgraph issue I would expect to see two.

aborgna-q:
The next portgraph release should fix that. See https://github.com/CQCL/portgraph/pull/132

PR #1144, opened 2024-05-31
fix!: Validate that control-flow outputs have exactly one successor

acl-cqc:
This was missed in prior validation, and there are a few instances in our codebase, so fix them.

I also had to comment out a failing merge_bbs test, see #1143

BREAKING CHANGE: BasicBlocks with multiple successors from the same outport will now fail to validate (as they should!)


aborgna-q (approved):
Good catch 👍

PR #1147, opened 2024-06-03
fix: merge_bbs creating too many self-loop edges

### LINKED ISSUE #1143: merge_bbs creating too many copies of self-loop edge

aborgna-q:
Updates `portgraph` to include the fix in https://github.com/CQCL/portgraph/pull/132.
Uncomments the previously-failing mentioned in #1143.

Fixes #1143.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ missing_docs = "warn"
 debug_assert_with_mut_call = "warn"
 
 [workspace.dependencies]
-portgraph = { version = "0.12.0" }
+portgraph = { version = "0.12.1" }
 insta = { version = "1.34.0" }
 bitvec = "1.0.1"
 cgmath = "0.18.0"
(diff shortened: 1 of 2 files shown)


acl-cqc:
Thanks @aborgna-q ! :-)

PR #1157, opened 2024-06-05
DRAFT: datalog style constant folding solver skeleton

doug-q:


hugr-passes/src/const_fold2/datalog/test.rs
```diff
@@ -0,0 +1,215 @@
...
+    let mut machine = Machine::new();
+    let arg_pv = PartialValue::variant(1, []).join(PartialValue::variant(2,[PartialValue::variant(0,[])]));
+    machine.propolutate_out_wires([(arg_w, arg_pv)]);
+    let c = machine.run_hugr(&hugr);
+
+    let cond_r1 = machine.read_out_wire_value(&c, cond_o1).unwrap();
+    assert_eq!(cond_r1, Value::false_val());
+    assert!(machine.read_out_wire_value(&c, cond_o2).is_none());
+
+    assert!(!machine.case_reachable(&c, case1.node()));
```
acl-cqc, line 212: @doug-q why does this pass? The predicate to the conditional node is the input to the whole Hugr here, which has Sum type with three variants; nothing tells us what argument is passed into that input??

acl-cqc: oh I get it - the `arg_pv`

acl-cqc: The argument is either variant 1 or variant 2 (containing variant 0 == false)

PR #1160, opened 2024-06-05
feat(hugr-py): store children in node weight

### LINKED ISSUE #1159: Allow python HUGR to report children of a node

ss2165:
Could instead bite the bullet and iterate through every node for calculating the children of a node, but this implementation seems like a decent code complexity/runtime balance

Closes #1159

### DIFF
--- a/hugr-py/src/hugr/_hugr.py
+++ b/hugr-py/src/hugr/_hugr.py
@@ -1,27 +1,30 @@
 from __future__ import annotations
-from dataclasses import dataclass, field, replace
+
 from collections.abc import Mapping
+from dataclasses import dataclass, field, replace
 from enum import Enum
 from typing import (
+    TYPE_CHECKING,
+    ClassVar,
+    Generic,
     Iterable,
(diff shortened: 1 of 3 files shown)


hugr-py/src/hugr/_hugr.py
```diff
@@ -112,6 +115,7 @@ class NodeData:
     _num_inps: int = 0
     _num_outs: int = 0
     # TODO children field?
+    children: list[Node] = field(default_factory=list)
```
mark-koch, line 118: Do we care about the order of children? Otherwise a set might be better for deleting nodes?

mark-koch (approved):

hugr-py/src/hugr/_hugr.py
ss2165: yes, order of children is significant in HUGR in general. Deleting nodes does
not need to be efficient - HUGR rewriting efficiency is only a requirement of
the rust crate.

PR #1165, opened 2024-06-05
feat(hugr-py): automatically add state order edges for inter-graph edges

ss2165:
uses a slightly dodgy -1 offset hack

hack currently means if you query the hugr for outgoing edges the state order edges are ignored.

### DIFF
--- a/hugr-py/src/hugr/_dfg.py
+++ b/hugr-py/src/hugr/_dfg.py
@@ -76,7 +76,7 @@ class Dfg:
             node_ancestor = _ancestral_sibling(self.hugr, src.node, node)
             if node_ancestor is None:
                 raise NoSiblingAncestor(src.node.idx, node.idx)
-            if node_ancestor != src.node:
+            if node_ancestor != node:
                 self.add_state_order(src.node, node_ancestor)
             self.hugr.add_link(src, node.inp(i))
 
--- a/hugr-py/src/hugr/_dfg.py
+++ b/hugr-py/src/hugr/_dfg.py
@@ -76,3 +76,14 @@ class Dfg:
         for i, p in enumerate(ports):
             src = p.out_port()
             self.hugr.add_link(src, node.inp(i))
+
+
+def _ancestral_sibling(h: Hugr, src: Node, tgt: Node) -> Node | None:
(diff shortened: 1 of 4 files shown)


hugr-py/src/hugr/_dfg.py
```diff
@@ -67,12 +68,25 @@ def set_outputs(self, *args: Wire) -> None:
...
+                raise NoSiblingAncestor(src.node.idx, node.idx)
+            if node_ancestor != src.node:
+                self.add_state_order(src.node, node_ancestor)
             self.hugr.add_link(src, node.inp(i))
+
+
+def _ancestral_sibling(h: Hugr, src: Node, tgt: Node) -> Node | None:
+    src_parent = h[src].parent
+
+    while (tgt_parent := h[tgt].parent) is not None:
```
mark-koch, line 87: You could get away with this:
```suggestion
    while tgt_parent := h[tgt].parent:
```

hugr-py/src/hugr/_hugr.py
```diff
@@ -292,6 +292,7 @@ def _node_links(
             return
         # iterate over known offsets
         for offset in range(self.num_ports(node, direction)):
+            # TODO should this also look for -1 state order edges?
```
mark-koch, line 295: I'd prefer a second method to get the order links

mark-koch (approved):
I think this will be sufficient for Guppy but note that non-local Dom edges from the spec are not covered by this PR. Should the builder handle those as well?

Anyways that can propbably wait for another PR

ss2165:
Yeah this is not sufficient to handle Dom, my intent is to cover that with the CFG PR, because we can probably do some overloaded behaviour by detecting we are inside a basic block

hugr-py/src/hugr/_hugr.py
ss2165: just the order links or the order links along with the value?

hugr-py/src/hugr/_dfg.py
ss2165: i prefer explicit checks, python truthiness is gross

PR #1168, opened 2024-06-06
feat: CircuitBuilder::add_constant

### LINKED ISSUE #1163: `add_constant` in circuit builder
Given a `CircuitBuilder`, it's not currently possible to define new constants to pass to parametric operations.

We should define a
```rust
pub fn add_constant(&mut self, value: impl Into<Value>) -> Wire;
```
similar to `Dataflow::add_load_value`.

aborgna-q:
Closes #1163.

Blocked by #1142.


aborgna-q:
I added a test skip when extension inference is enabled as suggested by @acl-cqc, so we can now merge this into `main` directly.

@doug-q Can you take another look?

aborgna-q:
Yes, once that PR is merged we should be able to revisit this.

PR #1173, opened 2024-06-06
feat: `HugrView::extract_hugr` to extract regions into owned hugrs.

### LINKED ISSUE #1171: `HugrView::extract_hugr` to extract views into owned hugrs.
Add a utility method to extract regions of a Hugr.
We currently have `SiblingSubgraph::extract_subgraph` for subgraphs, this should be the full-region equivalent.

This can then be used when an owned Hugr is expected, such as for the target of `SimpleReplacement`/`Replacement`.

aborgna-q:
Add a `extract_hugr` method to extract regions into owned `Hugr`s.
The implementation is short-circuited for `Hugr` to avoid unnecessary clones.

Closes #1171.

### DIFF
--- a/hugr-core/src/hugr.rs
+++ b/hugr-core/src/hugr.rs
@@ -263,6 +263,12 @@ impl Hugr {
         }
     }
 
+    /// Set the root node of the hugr.
+    pub(crate) fn set_root(&mut self, root: Node) {
+        self.hierarchy.detach(self.root);
+        self.root = root.pg_index();
+    }
+
     /// Add a node to the graph.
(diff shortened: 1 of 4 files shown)


mark-koch (approved):
Looks good to me 👍 

Would it be poor Rust-taste to do this by implementing `Into`? I guess this is consistent with `extract_subgraph()`, otherwise I would have suggested `to_hugr()` or something like that

aborgna-q:
We have many ways of obtaining a hugr from a view (there's `base_hugr` already), so I prefer to keep the explicit method to disambiguate.

PR #1174, opened 2024-06-06
feat!: Make the rewrite errors more useful

aborgna-q:
Adds some context to the errors.

BREAKING CHANGE: Added more attributes to the `InvalidReplacement` and `InvalidSubgraph` error enums.

### DIFF
--- a/hugr-core/src/hugr/views/sibling_subgraph.rs
+++ b/hugr-core/src/hugr/views/sibling_subgraph.rs
@@ -21,7 +21,7 @@ use crate::builder::{Container, FunctionBuilder};
 use crate::hugr::{HugrMut, HugrView, RootTagged};
 use crate::ops::dataflow::DataflowOpTrait;
 use crate::ops::handle::{ContainerHandle, DataflowOpID};
-use crate::ops::{OpTag, OpTrait};
+use crate::ops::{NamedOp, OpTag, OpTrait, OpType};
 use crate::types::{FunctionType, Type};
 use crate::{Hugr, IncomingPort, Node, OutgoingPort, Port, SimpleReplacement};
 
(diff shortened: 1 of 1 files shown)


hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -624,18 +646,36 @@ fn has_other_edge<H: HugrView>(hugr: &H, node: Node, dir: Direction) -> bool {
...
+        /// The node ID of the root node.
+        node: Node,
+        /// The op type of the root node.
+        op: OpType,
+    },
     /// Replacement graph boundary size mismatch.
-    #[error("Replacement graph boundary size mismatch.")]
-    InvalidSignature,
+    #[error(
+        "Replacement graph boundary size mismatch. Expected {expected}, got {}.",
```
acl-cqc, line 670:
```suggestion
        "Replacement graph boundary type mismatch. Expected {expected}, got {}.",
```

Or maybe just "graph type" without the "boundary"

hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -340,13 +340,22 @@ impl SiblingSubgraph {
...
         if !OpTag::Dfg.is_superset(dfg_optype.tag()) {
-            return Err(InvalidReplacement::InvalidDataflowGraph);
+            return Err(InvalidReplacement::InvalidDataflowGraph {
+                node: rep_root,
+                op: dfg_optype.clone(),
+            });
         }
         let Some([rep_input, rep_output]) = replacement.get_io(rep_root) else {
-            return Err(InvalidReplacement::InvalidDataflowParent);
+            return Err(InvalidReplacement::InvalidDataflowParent {
```
acl-cqc, line 349: This feels a strangely named error. It's not that the optype isn't a `DataflowParent` - that's guaranteed from the previous check that it's a Dfg. (Odd that we are checking `OpTag::Dfg.is_superset(dfg_optype)` rather than `OpTag::Dfg == dfg_optype` but there.)

So, it's that the input/output are missing. Indeed that's what the `#[error]` message for `InvalidDataflowParent` says! I'd suggest renaming to, say, `MissingInputOutput`.

(Also we kind of know what the `op` is, it's an `OpType::DFG` - so this just puts the input/output *typerows* in, well I guess that's fair.)

hugr-core/src/hugr/views/sibling_subgraph.rs
acl-cqc: (And the comment on the previous line too)

hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -649,8 +689,21 @@ pub enum InvalidSubgraph {
     #[error("The subgraph is not convex.")]
     NotConvex,
     /// Not all nodes have the same parent.
-    #[error("Not a sibling subgraph.")]
-    NoSharedParent,
+    #[error(
+        "Invalid sibling subgraph. {first_node} has parent {}, but {other_node} has parent {}.",
```
acl-cqc, line 693: I quite like "not a" sibling subgraph (the definition of such is that they are all siblings!), but ok

acl-cqc (approved):
Thanks, Agustin! A couple of suggestions for improvements but I guess they are kinda problems that existed already

hugr-core/src/hugr/views/sibling_subgraph.rs
aborgna-q: I think we can replace this error with an `expect`. It can only trigger with invalid hugrs.

PR #1176, opened 2024-06-06
feat(hugr-py): define type hierarchy separate from serialized

### LINKED ISSUE #1172: Introduce python builder type hierarchy

ss2165:
Closes #1172

Changes in first commit.
Note second commit is just running the schema update, which appears to be trivial and non-breaking. 

Coverage not great because this hierarchy is comprehensive (unlike the Ops) but not all are tested. Deemed not worth adding a load of boiler plate tests for now.


zrho (approved):
Ultimately this should probably be generated so that it automatically keeps in sync with the Rust implementation and specification. But for now it's probably good to have this.

PR #1177, opened 2024-06-06
fix: Do not require matching extension_reqs when creating a replacement

### LINKED ISSUE #1175: Allow changing `extension_reqs` in `SiblingSubgraph::create_simple_replacement`.
Replacing a node with another that requires more extensions should be possible, but currently it fails with a `InvalidReplacement::InvalidSignature` error due to the mismatched `extension_reqs`.

aborgna-q:
Fixes #1175 

Signatures have three components:
- The input type row
- The output type row
- The set of extensions used by the local operations

Many interesting replacements in a graph changes the latter, but `SiblingSubgraph::create_simple_replacement` required the signatures to match exactly.


cqc-alec (approved):
Looks right to me.

PR #1187, opened 2024-06-12
chore: Unskip test now that input_extensions is no more

aborgna-q:
Removes the conditional test skip introduced in https://github.com/CQCL/hugr/pull/1168#issuecomment-2154936867, now that #1142 is merged.


acl-cqc (approved):
Thanks!

Issue #640, opened 2023-11-04
Infer extension-requirements for nested DFGs


acl-cqc:
Nodes such as DFG, CFG, BasicBlock etc. are created wth a FunctionType that specifies a delta: https://github.com/CQCL/hugr/blob/1a07cd9aca60aa75257bdb297d7442d9529765fb/src/types/signature.rs#L25

This must be specified explicitly at creation time - one cannot say "None" as one can for `input_extensions`. However it would be nicer if one could, such that the inference algorithm then infers it, which I believe is possible as follows.

(This would require some change to the FunctionType/NodeType/DFG structures to allow specifying an Option, and then mutating it when inference is complete, in discussion below)

During inference, we both 1. constraint the input+output variables to match those at the Input and Output nodes of the DFG: https://github.com/CQCL/hugr/blob/1a07cd9aca60aa75257bdb297d7442d9529765fb/src/extension/infer.rs#L283-L288 and 2. constrain the outputs to be the input plus the specified delta: https://github.com/CQCL/hugr/blob/1a07cd9aca60aa75257bdb297d7442d9529765fb/src/extension/infer.rs#L314-L318

If we simplify avoid the second constraints in such cases, the constraints with the Input/Output nodes should be sufficient to calculate both. We can then compute the delta as the difference between the two. (If this last seems dubious - indeed the delta could also include any subset of the *inputs* - then this is no worse than at present, where such a delta specified by the programmer would be accepted by inference and validation. We could instead move to a system where the Input node also has *empty* input extensions, but that's another issue.)

acl-cqc:
Thus, two possible solutions:
* Recursively infer delta from the nested subgraph first, use that in the outer graph. This won't handle functions containing recursive calls (you need the delta from the function body to infer around the call node) - user-provided annotations on functions?
* Try to generalize "Plus" constraint to merge two variables. This might magically work or break everything, not sure.

acl-cqc:
> This would require some change to the FunctionType/NodeType/DFG structures to allow specifying an Option, and then mutating it when inference is complete

So it's worth thinking about this because it could be quite disruptive. There are several possible methods, including:
[...]

PR #1195, opened 2024-06-13
feat!: Infer extension deltas for Case, Cfg, Conditional, DataflowBlock, Dfg, TailLoop 

acl-cqc:
* Add an ExtensionId TO_BE_INFERRED
* On Case, Cfg, Conditional, DataflowBlock, Dfg, TailLoop, `Hugr::infer_extensions` replaces TO_BE_INFERRED with the smallest set of ExtensionIds that's correct wrt. its child nodes (i.e., the union of the child node deltas). If there are other ExtensionIds alongside TO_BE_INFERRED these will be kept (allowing a "lower bound" to be specified for individual nodes)
* `Hugr::infer_extensions` also takes a `remove: bool` which, if true, modifies deltas of the same OpTypes that do *not* have TO_BE_INFERRED, by removing ExtensionIds. (Thus, non-inferred deltas act as upper bounds).

Relates to #640 but does not address the builder where this should be the default behaviour.

BREAKING CHANGE: ExtensionError moved from hugr::validate to hugr; Hugr::infer_extensions takes bool parameter

### DIFF
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -406,6 +406,8 @@ pub enum ExtensionBuildError {
 pub struct ExtensionSet(BTreeSet<ExtensionId>);
 
 impl ExtensionSet {
+    pub const TO_BE_INFERRED: ExtensionId = ExtensionId::new_unchecked("TO_BE_INFERRED");
+
     /// Creates a new empty extension set.
     pub const fn new() -> Self {
         Self(BTreeSet::new())
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -405,9 +405,20 @@ pub enum ExtensionBuildError {
 #[derive(Clone, Debug, Default, Hash, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct ExtensionSet(BTreeSet<ExtensionId>);
 
-impl ExtensionSet {
-    pub const TO_BE_INFERRED: ExtensionId = ExtensionId::new_unchecked("TO_BE_INFERRED");
+/// A special ExtensionId which indicates that the delta of a non-Function
+/// container node should be computed by extension inference.
+/// Usable only in non-Function container nodes
+/// ([Case], [CFG], [Conditional], [DataflowBlock], [DFG], [TailLoop])
+///
+/// [Case]: crate::ops::Case
+/// [CFG]: crate::ops::CFG
+/// [Conditional]: crate::ops::Conditional
+/// [DataflowBlock]: crate::ops::DataflowBlock
+/// [DFG]: crate::ops::DFG
+/// [TailLoop]: crate::ops::TailLoop
+pub const TO_BE_INFERRED: ExtensionId = ExtensionId::new_unchecked("TO_BE_INFERRED");
 
+impl ExtensionSet {
     /// Creates a new empty extension set.
     pub const fn new() -> Self {
         Self(BTreeSet::new())
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -406,16 +406,10 @@ pub enum ExtensionBuildError {
 pub struct ExtensionSet(BTreeSet<ExtensionId>);
 
 /// A special ExtensionId which indicates that the delta of a non-Function
-/// container node should be computed by extension inference.
-/// Usable only in non-Function container nodes
-/// ([Case], [CFG], [Conditional], [DataflowBlock], [DFG], [TailLoop])
+/// container node should be computed by extension inference. See [`infer_extensions`]
+/// which lists the container nodes to which this can be applied.
 ///
-/// [Case]: crate::ops::Case
-/// [CFG]: crate::ops::CFG
-/// [Conditional]: crate::ops::Conditional
-/// [DataflowBlock]: crate::ops::DataflowBlock
-/// [DFG]: crate::ops::DFG
-/// [TailLoop]: crate::ops::TailLoop
+/// [`infer_extensions`]: crate::hugr::Hugr::infer_extensions
 pub const TO_BE_INFERRED: ExtensionId = ExtensionId::new_unchecked(".TO_BE_INFERRED");
 
 impl ExtensionSet {
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -416,7 +416,7 @@ pub struct ExtensionSet(BTreeSet<ExtensionId>);
 /// [DataflowBlock]: crate::ops::DataflowBlock
 /// [DFG]: crate::ops::DFG
 /// [TailLoop]: crate::ops::TailLoop
-pub const TO_BE_INFERRED: ExtensionId = ExtensionId::new_unchecked("TO_BE_INFERRED");
+pub const TO_BE_INFERRED: ExtensionId = ExtensionId::new_unchecked(".TO_BE_INFERRED");
 
 impl ExtensionSet {
     /// Creates a new empty extension set.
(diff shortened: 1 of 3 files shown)


hugr-core/src/hugr.rs
```diff
@@ -240,4 +303,158 @@ mod test {
...
+            Err(ValidationError::ExtensionError(expected_err.clone()))
+        );
+        #[cfg(not(feature = "extension_inference"))]
+        assert!(val_res.is_ok());
+
+        let inf_res = h.infer_extensions(true);
+        assert_eq!(inf_res, Err(expected_err));
+    }
+
+    fn build_ext_cfg(parent: ExtensionSet) -> (Hugr, Node) {
```
ss2165, line 364:
```suggestion
    fn build_ext_dfg(parent: ExtensionSet) -> (Hugr, Node) {
```

?

hugr-core/src/hugr.rs
```diff
@@ -304,4 +303,158 @@ mod test {
...
+                type_row: ty.into(),
+                new_extension: XA,
+            },
+        );
+        h.connect(inp, 0, mid, 0);
+        h.connect(mid, 0, out, 0);
+        mid
+    }
+
+    #[rstest]
```
ss2165, line 399: i think these cases would benefit from comments next to each saying briefly why the result is expected

hugr-core/src/hugr.rs
```diff
@@ -304,4 +303,158 @@ mod test {
...
+            let expected_p = ops::Case {
+                signature: FunctionType::new_endo(ty).with_extension_delta(result.clone()),
+            };
+            assert!(h.get_optype(p) == &expected_p.into());
+            let expected_gp = ops::Conditional {
+                extension_delta: result,
+                ..root_ty
+            };
+            assert!(h.root_type() == &expected_gp.into())
+            // rest should be unchanged...
```
ss2165, line 447: worth asserting?

ss2165 (commented):
The test simplifications and the brevity of the actual infer routine are very satisfying.

Thoughts on builder interface:

- Builders that infer extensions should be default, therefore they should have the short names (probably just the set of names we currently use with explicit deltas).
- Explicit delta versions, if necessary, can have some standard naming convention (e.g. `dfg_builder_delta`)
- Could a potential simplification be just to require `FunctionType` inputs wherever possible (rather than in_types + out_types) and then make the default extension set in `FunctionType` `TO_BE_INFERRED`. Then `with_extension_delta` on `FunctionType` would overwrite the default set? Are there uses of `FunctionType` where having the default extension set be empty is important?

acl-cqc:
> Thoughts on builder interface:
> 
> * Builders that infer extensions should be default, therefore they should have the short names (probably just the set of names we currently use with explicit deltas).
> * Explicit delta versions, if necessary, can have some standard naming convention (e.g. `dfg_builder_delta`)
> * Could a potential simplification be just to require `FunctionType` inputs wherever possible (rather than in_types + out_types) and then make the default extension set in `FunctionType` `TO_BE_INFERRED`. Then `with_extension_delta` on `FunctionType` would overwrite the default set? Are there uses of `FunctionType` where having the default extension set be empty is important?

Ok, moved to #1219

hugr-core/src/extension.rs
```diff
@@ -406,6 +406,8 @@ pub enum ExtensionBuildError {
 pub struct ExtensionSet(BTreeSet<ExtensionId>);
 
 impl ExtensionSet {
+    pub const TO_BE_INFERRED: ExtensionId = ExtensionId::new_unchecked("TO_BE_INFERRED");
```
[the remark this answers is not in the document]
acl-cqc: Made an "illegal" identifier i.e. you can *only* create it via `new_unchecked` and not via `new`

hugr-core/src/hugr/validate.rs
```diff
@@ -743,6 +747,9 @@ pub enum ValidationError {
     /// There are errors in the extension deltas.
     #[error(transparent)]
     ExtensionError(#[from] ExtensionError),
+    /// A node claims to still be awaiting extension inference. Perhaps it is not acted upon by inference...
```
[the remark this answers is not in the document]
acl-cqc: Fair, updated message too.

However, just to note - any Hugr-node that contains TO_BE_INFERRED in its delta is going to either be an error (if its parent doesn't) or result in a TLD that is basically unusable (because no runtime/environment/`run_on_quantum_hw` node/etc. will support such extensions). So we *could* skip both the error and the check. Just a thought....

hugr-core/src/hugr.rs
```diff
@@ -92,15 +91,63 @@ impl Hugr {
...
         #[cfg(feature = "extension_inference")]
         {
-            self.infer_extensions()?;
+            self.infer_extensions(false)?;
             self.validate_extensions()?;
         }
         Ok(())
     }
 
     /// Leaving this here as in the future we plan for it to infer deltas
```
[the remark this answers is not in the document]
acl-cqc: Ooops, thanks

hugr-core/src/hugr.rs
```diff
@@ -92,15 +91,63 @@ impl Hugr {
...
+        }
+        fn infer(h: &mut Hugr, node: Node, remove: bool) -> Result<ExtensionSet, ExtensionError> {
+            let mut child_sets = h
+                .children(node)
+                .collect::<Vec<_>>() // Avoid borrowing h over recursive call
+                .into_iter()
+                .map(|ch| Ok((ch, infer(h, ch, remove)?)))
+                .collect::<Result<Vec<_>, _>>()?;
+
+            let Some(es) = delta_mut(h.op_types.get_mut(node.pg_index())) else {
```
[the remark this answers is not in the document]
acl-cqc: Maybe, but
1. This raises tricky questions of what to do for a Node that is not in the Hugr. What'll actually happen here is we'll fall back to DenseUnmanagedMap's behaviour of expanding to contain an OpType for that Node (index), which I suspect is not what we'd want on a (quasi-?)public method...so at the least, it's a bit more involved that it sounds
2. If we do that, we should probably remove `replace_op` as having full `&mut` access is clearly superior! (`replace_op(idx, op)` is `get_mut(idx) = op`). I guess that could follow, but obviously it's a breaking change, and I think better to do that all in another PR.
 
I think this is fine for inside `crate::hugr`. (`h` is just `self` but we're in an inner-`fn`)

hugr-core/src/hugr.rs
acl-cqc: Erm, yes, oops, no idea how that happened

hugr-core/src/hugr.rs
acl-cqc: Fair, done

hugr-core/src/hugr.rs
acl-cqc: Done

hugr-core/src/hugr.rs
```diff
@@ -240,4 +321,167 @@ mod test {
...
+        h.connect(mid, 0, out, 0);
+        mid
+    }
+
+    #[rstest]
+    // Base case success: delta inferred for parent equals grandparent.
+    #[case([XA], [TO_BE_INFERRED], true, [XA])]
+    // Success: delta inferred for parent is subset of grandparent
+    #[case([XA, XB], [TO_BE_INFERRED], true, [XA])]
+    // Base case failure: infers [XA] for parent but grandparent has disjoint et
```
ss2165, line 422:
```suggestion
    // Base case failure: infers [XA] for parent but grandparent has disjoint set
```

ss2165 (approved):
thanks!

PR #1208, opened 2024-06-19
feat(hugr-sexpr): Add s-expression library.

zrho:
This PR adds the `hugr-sexpr` library for s-expressions.

- Read s-expressions from a string.
- Pretty print s-expressions.
- Derive macros for converting between s-expressions and user defined types.

(This commit also updates the devenv lock file to the newest version, since on the `main` branch version the Rust toolchain fails to work on my machine.)


hugr-core/Cargo.toml
```diff
@@ -20,6 +20,7 @@ workspace = true
 extension_inference = []
 
 [dependencies]
+hugr-sexpr = { path = "../hugr-sexpr", version = "0.1.0", features = ["derive"] }
```
aborgna-q, line 23: Why is this included as a dependency for core? I don't see any callsites.

Cargo.toml
```diff
@@ -1,9 +1,9 @@
-[profile.release]
+# [profile.release]
 lto = "thin"
```
aborgna-q, line 2: ?

Cargo.toml
aborgna-q, line 1: There's a couple places where you should add the new crate:

- `.github/change-filters.yml`: Add the crates to the `rust` filter

- `release-plz.toml`: Add a config for both. Until we decide to release the crates, they can be marked as such:
  ```toml
  [[package]]
  name = "hugr-sexpr"
  release = true
  
  # Don't create github releases for this crate yet.
  git_tag_enable = false
  git_release_enable = false
  # Don't push releases to crates.io
  publish = false

  [[package]]
  name = "hugr-sexpr-derive"
  release = true
  
  # Don't create github releases for this crate yet.
  git_tag_enable = false
  git_release_enable = false
  # Don't push releases to crates.io
  publish = false
  ```

- Add READMEs. You can copy most of it from the existing crates.

aborgna-q (commented):
Some comments to ensure the new crates work well with the automation.

Cargo.toml
zrho: Similarly puzzled how that happened. Reverting that of course.

hugr-core/Cargo.toml
zrho: I used to test the crate by generating and pretty printing s-expressions generated from HUGRs, but then decided to do that in a later PR. This is a remnant of that.

zrho:
As a general purpose s-expression crate this probably does not belong in this repository after all.

PR #1220, opened 2024-06-24
refactor!: FunctionBuilder takes impl Into<PolyFuncType>

acl-cqc:
Currently FunctionBuilder::new and the `define_function` method both take a PolyFuncType; this PR changes that to a more-general `impl Into<PolyFuncType>`.

BREAKING CHANGE: clients calling these methods with a FunctionType's `.into()` will fail to compile with a "type annotations required" error, which can be fixed by deleting the `.into()`.

### DIFF
--- a/hugr-core/src/builder.rs
+++ b/hugr-core/src/builder.rs
@@ -42,7 +42,7 @@
 //!     let _dfg_handle = {
 //!         let mut dfg = module_builder.define_function(
 //!             "main",
-//!             FunctionType::new(vec![BOOL_T], vec![BOOL_T]).into(),
+//!             FunctionType::new(vec![BOOL_T], vec![BOOL_T]),
 //!         )?;
 //!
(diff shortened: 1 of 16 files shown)


hugr-core/src/builder.rs
```diff
@@ -59,7 +59,7 @@
 //!     let _circuit_handle = {
 //!         let mut dfg = module_builder.define_function(
 //!             "circuit",
-//!             FunctionType::new(vec![BOOL_T, BOOL_T], vec![BOOL_T, BOOL_T]).into(),
+//!             FunctionType::new_endo(vec![BOOL_T, BOOL_T]),
```
acl-cqc, line 62: `new_endo` here is a driveby but seemed a reasonable thing to demonstrate in a doctest

cqc-alec (approved):
Nice.

PR #1226, opened 2024-06-25
fix!: Ops require their own extension

### LINKED ISSUE #388: Custom operations implicitly require their extension

acl-cqc:
Includes adding new variants of `block_builder`, `entry_builder`, `simple_entry_builder` and `conditional_builder`: the default version omits the extension set parameter, the `_exts` variant takes an extra parameter (being an ExtensionSet). `simple_block_builder` is untouched (as it takes a FunctionType, so can use `endo_ft`/`inout_ft`)

We'll need similar updates to `cfg_builder`, `tail_loop_builder`, `ConditionalBuilder::new` and `TailLoopBuilder::new` but I'll leave those for another PR, there's quite enough here ;)

closes #388

BREAKING CHANGE: (1) container-node extension-deltas will need to be enlarged to include ops therein; for FuncDefn this will have to be manually specified but for other containers TO_BE_INFERRED, `endo_ft` or `inout_ft` all work. (2) `block_builder`, `entry_builder`, `simple_entry_builder` and `conditional_builder` no longer take an ExtensionSet; either drop the argument or use the `..._exts` variant.


aborgna-q (commented):
Nice! We can finally close one of the longest-standing issues.
I mostly have doc comments.

I agree with having the `_exts` suffix variants.
It's better to not muddle things with another trait that will still be set to `None` most of the time.
I'd just suggest swapping the extra argument to the end, to keep it consistent with the non-suffix ones.
- `f(*inputs, *outputs)`
- `f_exts(*inputs, *outputs, extension_delta)`

Also, it's not part of this PR but I think the `::builder::{ft1, ft2}` are are quite confusing. They are used in doc examples but their names don't say anything (they rather look like local variables).

hugr-core/src/builder/cfg.rs
```diff
@@ -251,32 +275,53 @@ impl<B: AsMut<Hugr> + AsRef<Hugr>> CFGBuilder<B> {
...
+        &mut self,
+        sum_rows: impl IntoIterator<Item = TypeRow>,
+        other_outputs: TypeRow,
+    ) -> Result<BlockBuilder<&mut Hugr>, BuildError> {
+        self.entry_builder_exts(TO_BE_INFERRED, sum_rows, other_outputs)
+    }
+
     /// Return a builder for the entry [`DataflowBlock`] child graph with `inputs`
     /// and `outputs` and the variants of the branching Sum value
     /// specified by `sum_rows`.
```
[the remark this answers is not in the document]
acl-cqc: Yes, pretty much did this, also noted that delta will be inferred on the implicit version

hugr-core/src/builder/cfg.rs
```diff
@@ -289,9 +334,23 @@ impl<B: AsMut<Hugr> + AsRef<Hugr>> CFGBuilder<B> {
...
         outputs: TypeRow,
         n_cases: usize,
-        extension_delta: ExtensionSet,
     ) -> Result<BlockBuilder<&mut Hugr>, BuildError> {
-        self.entry_builder(vec![type_row![]; n_cases], outputs, extension_delta)
+        self.entry_builder(vec![type_row![]; n_cases], outputs)
+    }
+
+    /// Return a builder for the entry [`DataflowBlock`] child graph with `inputs`
+    /// and `outputs` and a UnitSum type: a Sum of `n_cases` unit types.
```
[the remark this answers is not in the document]
acl-cqc: In fact I removed mention of non-existent parameter `inputs` from both, too. Plus cross-link.

acl-cqc:
Description cleaned up - note
1. this is much less painful than the previous attempt #734 but there is still some pain for FuncDefn's (as there was always going to be) - that's #1237.
2. Alternative considered but rejected - to keep one variant of each function, but take an `exts: impl Into<Option<ExtensionSet>>`. This would mean we'd write `None` for inference, rather than (in this PR) an extra `_exts` when we don't want inference.
   * Using a custom trait (rather than `Into<Option<ExtensionSet>>`) would allow us to provide `impl<T: Into<ExtensionSet>> OurExtensionSetTrait for T` so we can use the useful `Into<ExtensionSet>`s that we have, alongside `impl OurExtensionSetTrait for Option` so we could still write `None`.
   * Or we could just take `exts: impl Into<ExtensionSet>` and define some top-level `const INFER` of a type that impls `Into<ExtensionSet>` (producing `ExtensionSet::singleton(TO_BE_INFERRED)`). Then for inference you'd write `INFER`.
   * We expect inference to be the normal hence I think the `_exts` is a good trade, writing nothing at all beats writing `None` or `INFER`

acl-cqc:
Thanks @aborgna-q - merged with #1297 rename, added docs cross-links (I can take those out if you don't like them, but I think it's fine to put them in one way only, we want to encourage use of inference-by-default)

hugr-core/src/builder/build_traits.rs
```diff
@@ -437,24 +437,46 @@ pub trait Dataflow: Container {
         TailLoopBuilder::create_with_io(self.hugr_mut(), loop_node, &tail_loop)
     }
 
+    /// Return a builder for a [`crate::ops::Conditional`] node.
+    /// `sum_input` is a tuple of the type of the Sum
+    /// variants and the corresponding wire.
+    ///
+    /// The `other_inputs` must be an iterable over pairs of the type of the input and
+    /// the corresponding wire.
+    /// The `outputs` are the types of the outputs. Extension delta will be inferred.
```
aborgna-q, line 446:
```suggestion
    /// The `output_types` are the types of the outputs. Extension delta will be inferred.
```

hugr-core/src/builder/build_traits.rs
```diff
@@ -437,24 +437,46 @@ pub trait Dataflow: Container {
...
+    }
+
     /// Return a builder for a [`crate::ops::Conditional`] node.
     /// `sum_rows` and `sum_wire` define the type of the Sum
     /// variants and the wire carrying the Sum respectively.
     ///
     /// The `other_inputs` must be an iterable over pairs of the type of the input and
     /// the corresponding wire.
     /// The `outputs` are the types of the outputs.
+    /// `exts` is the extension delta, here explicit. [conditional_builder](Self::conditional_builder) may be used to infer.
```
aborgna-q, line 468: Using the same names as the parameters
```suggestion
    /// The `output_types` are the types of the outputs.
    /// `extension_delta` defines an explicit delta for the conditional. See [conditional_builder](Self::conditional_builder) may be used to infer it instead.
```

aborgna-q (approved):
Nice.
I don't want to get too held up on the docs wording, so feel free to merge.

hugr-core/src/builder/build_traits.rs
acl-cqc: Yes-ish

PR #1229, opened 2024-06-26
fix: cfg not validating entry/exit types

### LINKED ISSUE #1189: CFG nodes do not fail validation when their input ports do not match the input node of the entry block
see https://github.com/CQCL/guppylang/issues/242 for an example

ss2165:
Closes #1189

Failing example generated using hugr-py

### DIFF
--- a/hugr-core/src/hugr/validate/test.rs
+++ b/hugr-core/src/hugr/validate/test.rs
@@ -1,3 +1,6 @@
+use std::fs::File;
+use std::io::BufReader;
+
 use cool_asserts::assert_matches;
 use rstest::rstest;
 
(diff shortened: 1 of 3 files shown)


hugr-core/src/hugr/validate/test.rs
```diff
@@ -926,6 +929,13 @@ fn cfg_children_restrictions() {
     b.remove_node(exit2);
 
     // Change the types in the BasicBlock node to work on qubits instead of bits
+    b.replace_op(
```
ss2165, line 932: this test failed because it was getting away with not updating the cfg op beforehand

PR #1232, opened 2024-06-27
refactor: cleanups from writing docstring

ss2165:
go commit by commit
mostly things like making methods/classes internal

### DIFF
--- a/hugr-py/src/hugr/val.py
+++ b/hugr-py/src/hugr/val.py
@@ -50,13 +50,16 @@ FALSE = bool_value(False)
 
 
 @dataclass
-class Tuple(Value):
+class Tuple(Sum):
     vals: list[Value]
 
-    def type_(self) -> tys.Tuple:
-        return tys.Tuple(*(v.type_() for v in self.vals))
+    def __init__(self, *vals: Value):
+        val_list = list(vals)
+        super().__init__(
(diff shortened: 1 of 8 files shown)


hugr-py/src/hugr/val.py
```diff
@@ -50,13 +50,16 @@ def bool_value(b: bool) -> Sum:
...
-    def type_(self) -> tys.Tuple:
-        return tys.Tuple(*(v.type_() for v in self.vals))
+    def __init__(self, *vals: Value):
+        val_list = list(vals)
+        super().__init__(
+            tag=0, typ=tys.Tuple(*(v.type_() for v in val_list)), vals=val_list
+        )
 
-    def to_serial(self) -> sops.TupleValue:
+    def to_serial(self) -> sops.TupleValue:  # type: ignore[override]
```
aborgna-q, line 62: Any idea why the ignore is needed?
It seems straightforward code

aborgna-q (approved):
Nice cleanup 👍

hugr-py/src/hugr/val.py
ss2165: `sops.TupleValue` isn't an instance of `sops.SumValue` so mypy doesn't like the override of `Sum.to_serial`

ss2165: will add comment

PR #1246, opened 2024-07-01
fix: fix some typos

ss2165:
add a config file for `typos` tool https://github.com/crate-ci/typos?tab=readme-ov-file

we could consider adding this to CI/pre-commit but don't want to do a big repo-wide change right now

I used cli typos tool with `cargo install typos-cli`

### DIFF
--- a/hugr-py/src/hugr/serialization/__init__.py
+++ b/hugr-py/src/hugr/serialization/__init__.py
@@ -1,3 +1,0 @@
-from .serial_hugr import SerialHugr
-
-__all__ = ["SerialHugr"]
--- a/hugr-py/src/hugr/serialization/tys.py
+++ b/hugr-py/src/hugr/serialization/tys.py
@@ -1,23 +1,27 @@
 from __future__ import annotations
 
-from abc import ABC, abstractmethod
 import inspect
 import sys
+from abc import ABC, abstractmethod
 from enum import Enum
-from typing import Annotated, Any, Literal, Union, Mapping
+from typing import TYPE_CHECKING, Annotated, Any, Literal
 
-from hugr.utils import deser_it
 from pydantic import (
     BaseModel,
+    ConfigDict,
     Field,
     RootModel,
     ValidationError,
     ValidationInfo,
     ValidatorFunctionWrapHandler,
     WrapValidator,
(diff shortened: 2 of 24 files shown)


hugr-py/src/hugr/serialization/tys.py
```diff
@@ -475,9 +480,10 @@ def deserialize(self) -> tys.Type:
 
 def model_rebuild(
     classes: Mapping[str, type],
-    config: ConfigDict = ConfigDict(),
+    config: ConfigDict | None = None,
     **kwargs,
 ):
+    config = config or ConfigDict()
```
croyzor, line 486: This seems worse?

croyzor: Looks like it's this lint: https://docs.astral.sh/ruff/rules/function-call-in-default-argument/
It seems we can add exceptions, but I can live with this convention

ss2165: python generally doesn't like functions being called in the arguments, taking none then calling the default function inside the body is the preferred pattern

https://docs.astral.sh/ruff/rules/function-call-in-default-argument/

croyzor: By convention, I'd expect nullary constructors like `MyStruct()` to be benign and more acceptable to put here than an arbitrary function. Maybe too much c++ brain...

hugr-py/src/hugr/serialization/__init__.py
```diff
@@ -1,3 +0,0 @@
-from .serial_hugr import SerialHugr
-
-__all__ = ["SerialHugr"]
```
croyzor, line 3: I'm not quite sure what's going on here, I'm surprised that a lint would tell us to change this interface?

ss2165: ah yeah sorry, this is because some of the import changes caused a circular import problem. This was my chosen solution - we don't rexport through init anywhere else and the serialised hugr should be an internal thing now anyway

croyzor (approved):
The changes look good but the PR title is wack imo. It's mostly applying lints + a couple of typos

PR #1249, opened 2024-07-01
ci: Tell miri to ignore tests that open files.

cqc-alec:
Some tests read from files, and miri objects to this unless this flag is set.


ss2165 (commented):
Thank you, so far for these cases we've disabled individual tests for miri, e.g.

https://github.com/CQCL/hugr/blob/4806552fe072fa97b902b21e3131aede6a3eef73/hugr-core/src/extension/declarative.rs#L359

can we just keep doing that?

ss2165 (approved):
👍

PR #1251, opened 2024-07-02
style: Change "serialise" etc to "serialize" etc.

cqc-alec:
Did: `git grep -l -i 'serialis' | xargs sed -i 's/erialis/erializ/g'`

The "-ize" spelling is already used much more often, so let's be consistent.

### DIFF
--- a/hugr-cli/README.md
+++ b/hugr-cli/README.md
@@ -8,7 +8,7 @@ hugr-cli
 [![msrv][]](https://github.com/CQCL/hugr)
 [![codecov][]](https://codecov.io/gh/CQCL/hugr)
 
-`hugr` CLI tool for common tasks on serialised HUGR (e.g. validation,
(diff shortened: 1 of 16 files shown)


aborgna-q (approved):
👍 
It looks like we only used UK spelling on docs. Was all the code US-style?

We should maybe decide to use either UK or US spelling everywhere, and add it to DEVELOPMENT.md.

cqc-alec:
> 👍 It looks like we only used UK spelling on docs. Was all the code US-style?
> 
> We should maybe decide to use either UK or US spelling everywhere, and add it to DEVELOPMENT.md.

It's not quite as simple as US/UK style: there are different UK styles. See for example https://academic.oup.com/pages/authoring/books/preparing-your-manuscript/house-style#oup .

I wouldn't be opposed to adopting (and documenting) a house style, such as the OUP one, but this should perhaps be a larger conversation between software projects.

PR #1263, opened 2024-07-05
chore: Update portgraph to `0.12.2`

aborgna-q:
Includes the fix in https://github.com/CQCL/portgraph/pull/139 that closes #1197

PR #1265, opened 2024-07-05
refactor!: rename `HugrView` function type methods + simplify behaviour

### LINKED ISSUE #957: HugrView::get_function_type makes no sense
It returns the PolyFuncType of a function, but the FunctionType of any other container e.g. DFG *plus zero binders* - what is going on here?

* Used in constants, where we `try_into` the PolyFuncType *back* to a FunctionType
* And in test code
* Probably better to make it either
    * return a monomorphic function type or none (i.e. reject any FuncDecl/Defn with binders), as per the use in const; or
    * make it return a PolyFuncType for FuncDecl/fn, only (and None for anything else). It could then be used in the builder for building Call and LoadFunction nodes (which has its own code in `build_traits.rs` atm)

At the root here is the question - can you put a FuncDefn in a Constant and then LoadConstant it? Currently the rules say that you *can* (this lets you get a closure of a recursive function, which you can't do by LoadConstant'ing a DFG). But after #947 one could just add the FuncDefn to the Hugr and use LoadFunction. So maybe / do we really want to keep the FuncDefn-in-Constant just so we can have a *nested Hugr* rather than a child node?

ss2165:
Closes #957

- `mono_fn_type` now enforces function value monomorphism
- get_df_function_type -> inner_function_type so it matches the actual op function being called more obviously.
- The culprit `get_function_type` changed to `poly_func_type` which _only_ reports polymorphic type from func defn/decl

BREAKING CHANGE: 
 - rename `get_df_function_type` to `inner_function_type`
- `get_function_type` replaced with `poly_func_type` which only reports polymorphic function type for Func(Defn/Decl) rooted HUGRs

### DIFF
--- a/hugr-core/src/ops/constant.rs
+++ b/hugr-core/src/ops/constant.rs
@@ -326,14 +326,14 @@ pub enum ConstTypeError {
 
 /// Hugrs (even functions) inside Consts must be monomorphic
 fn mono_fn_type(h: &Hugr) -> Result<FunctionType, ConstTypeError> {
-    if let Some(pf) = h.get_function_type() {
-        if let Ok(ft) = pf.try_into() {
-            return Ok(ft);
-        }
-    }
-    Err(ConstTypeError::NotMonomorphicFunction {
+    let err = || ConstTypeError::NotMonomorphicFunction {
         hugr_root_type: h.root_type().clone(),
-    })
+    };
+    if let Some(pf) = h.poly_func_type() {
+        return pf.try_into().map_err(|_| err());
+    }
+
+    h.inner_function_type().ok_or_else(err)
 }
 
 impl Value {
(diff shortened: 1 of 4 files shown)


hugr-core/src/ops/constant.rs
```diff
@@ -326,14 +326,15 @@ pub enum ConstTypeError {
 
 /// Hugrs (even functions) inside Consts must be monomorphic
 fn mono_fn_type(h: &Hugr) -> Result<FunctionType, ConstTypeError> {
```
ss2165, line 328: this function just matches existing behaviour - this PR does not address the question of what a valid monomorphic function value hugr root type is

aborgna-q (approved):
LGTM

The failed test should get fixed with #1263.
Mermaid rendering was not considered a breaking chance :/

acl-cqc:
LGTM, thanks @ss2165

PR #1267, opened 2024-07-05
Set up quantinuum docs theme.

aidanCQ:
Nav is rather sparse at the moment. Worth dicussing its utility in this project.

Screenshot 2024-07-05 at 17 43 27


ss2165:
> Nav is rather sparse at the moment.

Becomes more useful if you can get to other docs? (TKET1, guppy, tket2...?)

PR #1269, opened 2024-07-08
RFC: Replace ft1/ft2 with Into<FunctionType> from tuple of TypeRow(s)

acl-cqc:
In answer to [this comment](https://github.com/CQCL/hugr/pull/1226#pullrequestreview-2158309800) that "the `ft1`/`ft2` helpers are quite confusing....they rather look like local variables"...

This probably doesn't make them any less confusing, indeed, it may make them *more* confusing, but they don't look like local variables anymore - indeed, this makes them almost disappear, so they barely look like anything anymore....

See what you think; there are a number of other variations, such as
* `impl<T: Into<TypeRow>> IntoFunctionType for T` i.e. dropping the unary-tuple. (We have to add a `pub trait IntoFunctionType` or similar to get around the foreign trait restriction but that is no big deal)
* Add `pub enum FTSpec { Endo(TypeRow), InOut(TypeRow,TypeRow) }` with a straightforward `impl From<FTSpec> for FunctionType`. The sad bit here is that `FTSpec::Endo` is a *lot* longer than `ft1` and I'd really like to keep these minimal....
* Originally I'd thought of using `impl Into<FTSpec>` for *some* type FTSpec but I haven't really got anything that makes particular sense here and/or that's really different from the `impl Into<FunctionType>` in the PR. (Also it'd seem odd, if passing an actual FunctionType with *explicit* delta, to be converting FunctionType to new-spec-type and then back to FunctionType *again*)

### DIFF
--- a/hugr-core/src/builder.rs
+++ b/hugr-core/src/builder.rs
@@ -121,16 +121,16 @@ pub use conditional::{CaseBuilder, ConditionalBuilder};
 mod circuit;
 pub use circuit::{CircuitBuildError, CircuitBuilder};
 
-/// Return a FunctionType with the same input and output types (specified)
-/// whose extension delta, when used in a non-FuncDefn container, will be inferred.
-pub fn ft1(types: impl Into<TypeRow>) -> FunctionType {
-    FunctionType::new_endo(types).with_extension_delta(TO_BE_INFERRED)
+impl<T: Into<TypeRow>> From<(T,)> for FunctionType {
+    fn from(value: (T,)) -> Self {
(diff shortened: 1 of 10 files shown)


aborgna-q:
> it may make them more confusing

Agree with that :P

What about just renaming the functions to ~`endo_typerow` and ~`in_out_typerow`?

acl-cqc:
> > it may make them more confusing
> 
> Agree with that :P
> 
> What about just renaming the functions to ~`endo_typerow` and ~`in_out_typerow`?

They make `FunctionType`s, so I could settle with `endo_fn` and `in_out_fn` ?

PR #1271, opened 2024-07-08
fix!: `ops::Module` now empty struct rather than unit struct

### LINKED ISSUE #1270: Serialisation regression in #1183
#1183 included a breaking change to the serialisation format, removing the `input_extensions` field from the operations.

This should have either been a new `v2` serialisation with backwards compatibility, or implemented so that the field gets silently ignored (if possible).

This json was generated by `guppylang 0.5.2` (which uses `hugr-py v0.2.0a1`). Since it includes the input extensions fields it produces errors when decoded with the latest `hugr` crate, but works correctly once the fields are removed.
https://gist.github.com/aborgna-q/f4d6caa1fd7f9b5288935f39df4d4918

ss2165:
Fixes #1270 without keeping the unused `input_extensions` field. 

BREAKING CHANGE: `ops::Module` no longer unit struct, must be constructed with `new()` or `default()`

### DIFF
--- a/hugr-core/src/hugr/serialize.rs
+++ b/hugr-core/src/hugr/serialize.rs
@@ -49,9 +49,6 @@ struct NodeSer {
     parent: Node,
     #[serde(flatten)]
     op: OpType,
-
-    #[serde(skip_serializing, default)]
-    input_extensions: Option<crate::extension::ExtensionSet>,
 }
 
 /// Version 1 of the HUGR serialization format.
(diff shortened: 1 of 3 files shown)


ss2165:
Converting to draft in case there is some alternative fix for: https://github.com/serde-rs/serde/issues/2775

aborgna-q (approved):
The non-breaking patch is already published, we can merge this.

PR #1275, opened 2024-07-08
fix: bring back input_extensions serialized field in rust NodeSer

### LINKED ISSUE #1270: Serialisation regression in #1183

ss2165:
(skip serializing, deserialize to None)

Closes #1270

### DIFF
--- a/hugr-core/src/hugr.rs
+++ b/hugr-core/src/hugr.rs
@@ -51,7 +51,7 @@ pub struct Hugr {
 
 impl Default for Hugr {
     fn default() -> Self {
-        Self::new(crate::ops::Module)
+        Self::new(crate::ops::Module::new())
     }
 }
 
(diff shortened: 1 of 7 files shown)


hugr-core/src/hugr.rs
```diff
@@ -51,7 +51,7 @@ pub struct Hugr {
 
 impl Default for Hugr {
     fn default() -> Self {
-        Self::new(crate::ops::Module)
+        Self::new(crate::ops::Module::new())
```
ss2165, line 54: kept this from the other PR, can remove but is harmless

aborgna-q (approved):
Thanks!

PR #1276, opened 2024-07-08
feat(hugr-py): context manager style nested building

### LINKED ISSUE #1243: feat(hugr-py): Context managers for nested building
(=especially to ensure things are done in the right order

ss2165:
Currently the context doesn't do anything much, just offers syntactical nesting to make it easier to write HUGRs

The exception being Conditional, where we use the exit as an opportunity to check all cases have been built.
This can be extended to other cases (e.g. ensure expected  outputs have been set when exiting a dfg).

Closes #1243

### DIFF
--- a/hugr-py/src/hugr/cfg.py
+++ b/hugr-py/src/hugr/cfg.py
@@ -2,8 +2,11 @@
 
 from __future__ import annotations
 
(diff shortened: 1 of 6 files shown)


cqc-alec (approved):
Nice!

PR #1283, opened 2024-07-09
feat: Make `DataflowOpTrait` public

### LINKED ISSUE #1273: Make `DataflowOpTrait` public
This is a useful trait, in particular `signature`. We should make it accessibly outside the crate.

doug-q:
Closes #1273


acl-cqc (approved):
Yes seems like a good idea to me

PR #1285, opened 2024-07-09
feat(hugr-passes): Add `force_order` pass.

### LINKED ISSUE #1282: Add a "sink" pass
We have a need to perform some operations as early as possible and some operations as late as possible.

Create a "sink" pass that is parameterised by a `rank: Fn(Node) -> i64`. 

For all pairs of operations `(o1,o2)` such that:
* they have the same dataflow parent
* neither dominates the other
* `rank(o1) > rank(o2)` 

After the pass is run `o1` dominates `o2`.

This should be achieved by adding order edges.

doug-q:
closes #1282

### DIFF
--- a/hugr-passes/src/force_order.rs
+++ b/hugr-passes/src/force_order.rs
@@ -1,5 +1,5 @@
 //! Provides [force_order], a tool for fixing the order of nodes in a Hugr.
-use std::{cmp::Reverse, collections::BinaryHeap, iter::zip};
+use std::{cmp::Reverse, collections::BinaryHeap};
 
 use hugr_core::{
     hugr::{
@@ -93,7 +93,7 @@ pub fn force_order_by_key<K: Ord>(
 /// ensures we visit lower ranked nodes before higher ranked nodes whenever the
 /// topology of the graph allows.
 #[derive(Clone)]
-struct ForceOrder<K: Ord, N: Ord, VM, F> {
+struct ForceOrder<K, N, VM, F> {
     tovisit: BinaryHeap<(Reverse<K>, N)>,
     ordered: VM,
     rank: F,


aborgna-q:
It'd be nice to do some benchmarks at some point, but we're lacking hugr generators to benchmark things in general :/

hugr-passes/src/force_order.rs
```diff
@@ -0,0 +1,291 @@
...
+            let new_nodes = g
+                .neighbors(nix)
+                .filter(|&n| {
+                    Reversed(g)
+                        .neighbors(n)
+                        .all(|b| self.ordered.is_visited(&b))
+                })
+                .collect_vec();
+
+            self.extend(new_nodes);
```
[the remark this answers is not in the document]
aborgna-q: Just `[0, 1, 2, 3]` in your example :)

```rust
#[test]
fn test_force_order_3() {
    let (mut hugr, [v0, v1, v2, v3]) = test_hugr();
    let rank_map = [(v0, 0), (v1, 1), (v2, 2), (v3, 3)].into_iter().collect();
    let topo_sort = force_order_test_impl(&mut hugr, rank_map);
    assert_eq!(vec![v0, v1, v2, v3], topo_sort);
}
```

aborgna-q:
Btw, a more flexible API could be to offer a `force_order_by_key` where instead of a rank you take a
```rust
key: impl Fn(Node) -> (impl Ord)
```
which is just a generalization of `rank`.

hugr-passes/src/force_order.rs
```diff
@@ -0,0 +1,307 @@
...
+        let sg = SiblingGraph::<Node>::try_new(hugr, dp)?;
+        let petgraph = NodeFiltered::from_fn(sg.as_petgraph(), |x| x != dp);
+        let ordered_nodes = ForceOrder::new(&petgraph, &rank)
+            .iter(&petgraph)
+            .filter(|&x| hugr.get_optype(x).tag() <= OpTag::DataflowChild)
+            .collect_vec();
+
+        // ordered_nodes must always include the input and output nodes, so the
+        // slice below is in bounds.
+        for (&n1, &n2) in zip(&ordered_nodes[..], &ordered_nodes[1..]) {
```
aborgna-q, line 67: A bit cleaner :)
```suggestion
        for (&n1, &n2) in ordered_nodes.iter().tuple_windows() {
```

hugr-passes/src/force_order.rs
```diff
@@ -0,0 +1,307 @@
...
+            let (n1_ot, n2_ot) = (hugr.get_optype(n1), hugr.get_optype(n2));
+            assert_eq!(
+                Some(EdgeKind::StateOrder),
+                n1_ot.other_port_kind(Direction::Outgoing)
+            );
+            assert_eq!(
+                Some(EdgeKind::StateOrder),
+                n2_ot.other_port_kind(Direction::Incoming)
+            );
```
aborgna-q, line 76: Some panic context
```suggestion
            assert_eq!(
                Some(EdgeKind::StateOrder),
                n1_ot.other_port_kind(Direction::Outgoing),
                "Node {n1} does not support state order edges"
            );
            assert_eq!(
                Some(EdgeKind::StateOrder),
                n2_ot.other_port_kind(Direction::Incoming),
                "Node {n2} does not support state order edges"
            );
```

hugr-passes/src/force_order.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    Ok(())
+}
+
+/// An adaption of [petgraph::visit::Topo]. We differ only in that we sort nodes
+/// by the rank function before adding them to the internal work stack. This
+/// ensures we visit lower ranked nodes before higher ranked nodes whenever the
+/// topology of the graph allows.
+#[derive(Clone)]
+struct ForceOrder<K: Ord, N: Ord, VM, F> {
```
aborgna-q, line 96: No need to repeat the bound here (it's only needed in the `impl`)
```suggestion
struct ForceOrder<K, N, VM, F> {
```

aborgna-q (approved):
👍 

Just some minor code suggestions

Issue #1287, opened 2024-07-09
feat(hugr-py): Add remaining standard extension ops and types


ss2165:
E.g. And, Or, integer and float operations.

May be worth waiting for potential code generation from extension spec work.

PR #1288, opened 2024-07-09
feat(hugr-py): move std extension types/ops in to `std` module

ss2165:
Not complete, see #1287

Quantum operations left in tests as they are not part of std extensions set (maybe they should be...?)

### DIFF
--- a/hugr-py/src/hugr/std/int.py
+++ b/hugr-py/src/hugr/std/int.py
@@ -1,0 +1,71 @@
+"""HUGR integer types and operations."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+from hugr import tys, val
+from hugr.ops import Custom
+
+
+def int_t(width: int) -> tys.Opaque:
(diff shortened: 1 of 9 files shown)


hugr-py/src/hugr/std/int.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@dataclass(frozen=True)
+class IntOps(Custom):
+    """Base class for integer operations."""
+
+    extension: tys.ExtensionId = "arithmetic.int"
+
+
+_ARG_5 = tys.BoundedNatArg(n=5)
```
aborgna-q, line 54: nit: Call this `_ARG_i32`? It's a bit more legible

aborgna-q (approved):

hugr-py/src/hugr/std/int.py
ss2165: its a generic type arg, not restricted to integers

PR #1290, opened 2024-07-10
feat(hugr-py): `AsCustomOp` protocol for user-defined custom op types.

ss2165:
Downstream users can implement this protocol to use their own convenience classes. Typically a use wants to implement:
- `to_custom`
- `__call__` to provide a useful calling signature.
- `from_custom` if different instances of the type correspond to different ops (non-singleton).

Replace existing custom operation types with this.

Follow ups:
- Similar thing for custom types.
- Optional: allow these types to register themselves with `serialization.ops.CustomOp` so they can be deserialized directly.

### DIFF
--- a/hugr-py/src/hugr/std/int.py
+++ b/hugr-py/src/hugr/std/int.py
@@ -73,13 +73,14 @@ class _DivModDef(AsCustomOp):
 
     @classmethod
     def from_custom(cls, custom: Custom) -> Self | None:
-        if not (custom.extension == OPS_EXTENSION and custom.op_name == cls.op_name):
+        if not custom.check_id(OPS_EXTENSION, "idivmod_u"):
             return None
         match custom.args:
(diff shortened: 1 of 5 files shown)


hugr-py/src/hugr/std/int.py
```diff
@@ -44,27 +50,39 @@ def to_value(self) -> val.Extension:
...
+
+    @classmethod
+    def from_custom(cls, custom: Custom) -> Self | None:
+        if not (custom.extension == OPS_EXTENSION and custom.op_name == cls.op_name):
+            return None
+        match custom.args:
+            case [tys.BoundedNatArg(n=a1), tys.BoundedNatArg(n=a2)]:
+                return cls(arg1=a1, arg2=a2)
+            case _:
+                return None
```
aborgna-q, line 82: Should this raise an error? If the name and extension match then we should always be able to extract the op

PR #1293, opened 2024-07-11
test: verify order edges

### LINKED ISSUE #1167: `LoadConstant` nodes can't have `OrderEdge`s
At present Nodes are only allowed one kind of `other_port_kind`. This seems like an artificial restriction, which good taste demands should be lifted, but "LoadConstant nodes can't have OrderEdges" does seem like a limitation.

In my mind `LoadConstant` is an operation with side effects. Memory side effects being obvious, but also `LoadConstant` of a `Value::Function` can be read as compiling that value. As such, users should be able to impose an order on two `LoadConstant` operations to control the order of their side effects.

Weaker arguments for nodes to have multiple `other_port_kind`s:
* `Case` nodes in a conditional are interpreted with respect to their index into their parent. This is ugly, one could instead demand, or allow, that they have order edges which induce a total order, and use that total order to assign them an index.
* `DataFlowBlock`s could get indices in the same was as `Case` above.

edges spec issue: https://github.com/CQCL/hugr/issues/848

### LINKED ISSUE #1284: `Call` cannot have order edges
As https://github.com/CQCL/hugr/issues/1167. This is a significantly more urgent problem.

doug-q:
fixes #1167 fixes #1284

### DIFF
--- a/hugr-core/src/builder/dataflow.rs
+++ b/hugr-core/src/builder/dataflow.rs
@@ -207,8 +207,8 @@ pub(crate) mod test {
     use crate::extension::prelude::{BOOL_T, USIZE_T};
     use crate::extension::{ExtensionId, SignatureError, EMPTY_REG, PRELUDE_REGISTRY};
     use crate::hugr::validate::InterGraphEdgeError;
-    use crate::ops::{OpTrait, Value};
     use crate::ops::{handle::NodeHandle, Lift, Noop, OpTag};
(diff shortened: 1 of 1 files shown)


ss2165:
I don't mind about keeping or removing the test, but I would suggest removing the "fixes" from the commit description. Those issues should be closed as not happening, rather than fixed. I will do that now.

ss2165 (approved):

PR #1300, opened 2024-07-15
fix!: force_order failing on Const nodes, add arg to rank.

doug-q:
BREAKING CHANGE: the `rank` argument of `force_order` takes an additional argument.

### DIFF
--- a/hugr-passes/src/force_order.rs
+++ b/hugr-passes/src/force_order.rs
@@ -36,20 +36,20 @@ use petgraph::{
 /// there is no path from `n2` to `n1` (otherwise this would invalidate `hugr`).
 /// Nodes of equal rank will be ordered arbitrarily, although that arbitrary
 /// order is deterministic.
-pub fn force_order(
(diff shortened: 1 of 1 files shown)


hugr-passes/src/force_order.rs
```diff
@@ -56,25 +56,39 @@ pub fn force_order_by_key<K: Ord>(
         .filter(|n| hugr.get_optype(*n).tag() <= OpTag::DataflowParent)
         .collect_vec();
     for dp in dataflow_parents {
+        // we filter out the input and output nodes from the topological sort
+        let Some([i, o]) = hugr.get_io(dp) else {
+            continue;
+        };
```
mark-koch, line 62: For which dataflow parent would this `else` branch trigger?

mark-koch (approved):
> I've taken the opportunity to add a &Hugr arg to rank. Rank can't hold a reference to the Hugr because force_order takes a mut reference.

Nice, I was running into that issue before 👍

PR #1302, opened 2024-07-15
ci: Check python test result on required checks check

aborgna-q:
A python test failure didn't cause the required checks to fail...

drive-by: Add workflow files to the CI change filters, so modifications trigger the checks.


aborgna-q:
I added the workflows to the change filters, so tests should run now.

PR #1317, opened 2024-07-16
fix: add op's extension to signature check in `resolve_opaque_op`

doug-q:
We also remove an unsafe `unwrap` in `resolve_opaque_op` while we are here.

### DIFF
--- a/hugr-core/src/ops/custom.rs
+++ b/hugr-core/src/ops/custom.rs
@@ -294,7 +294,7 @@ pub struct OpaqueOp {
     #[cfg_attr(test, proptest(strategy = "any_nonempty_string()"))]
     description: String, // cache in advance so description() can return &str
     args: Vec<TypeArg>,
-    // note that `signature` may not include `extension`. Thus this field must
+    // note that the `signature` field might not include `extension`. Thus this must
     // remain private, and should be accessed through
     // `DataflowOpTrait::signature`.
     signature: Signature,


hugr-core/src/ops/custom.rs
```diff
@@ -392,9 +392,14 @@ pub fn resolve_opaque_op(
                 r.name().clone(),
             ));
         };
-        let ext_op =
-            ExtensionOp::new(def.clone(), opaque.args.clone(), extension_registry).unwrap();
-        if opaque.signature != ext_op.signature {
+        let ext_op = ExtensionOp::new(def.clone(), opaque.args.clone(), extension_registry)?;
+        // ops always require their own extensions, so we do not force OpaqueOps
```
acl-cqc, line 396: The comment I imagined was "OpaqueOps serialized prior to PR #1226 did not include their own extension. Add it manually for backwards compatibility when deserializing such early Hugr's".

Of course, my version doesn't allow that not all OpaqueOps result from serializing known ExtensionOps - you can also write OpaqueOps out by hand with any extension delta you like. So I guess your comment is fine, but might be worth adding `Especially for backwards compatibility with Hugrs serialized before PR#1226`, say.

hugr-core/src/ops/custom.rs
```diff
@@ -392,9 +392,14 @@ pub fn resolve_opaque_op(
                 r.name().clone(),
             ));
         };
-        let ext_op =
-            ExtensionOp::new(def.clone(), opaque.args.clone(), extension_registry).unwrap();
-        if opaque.signature != ext_op.signature {
+        let ext_op = ExtensionOp::new(def.clone(), opaque.args.clone(), extension_registry)?;
```
acl-cqc, line 395: The change from `unwrap` to `?` here necessitates the addition of  `CustomOpError::SignatureError`? Good move :)

acl-cqc (approved):
Looks good to me, thanks @doug-q

hugr-py/src/hugr/std/float.py
```diff
@@ -7,9 +7,8 @@
 from hugr import tys, val
 
 #: HUGR 64-bit IEEE 754-2019 floating point type.
-FLOAT_EXT_ID = "arithmetic.float.types"
 FLOAT_T = tys.Opaque(
-    extension=FLOAT_EXT_ID,
+    extension="arithmetic.float.types",
```
acl-cqc, line 11: nit: I don't really see any need to revert this

hugr-core/src/hugr/rewrite/replace.rs
```diff
@@ -643,38 +643,56 @@ mod test {
...
-        let h = h.finish_hugr_with_outputs(cond.outputs(), &PRELUDE_REGISTRY)?;
+        let mut cond = h
+            .conditional_builder_exts(
+                (vec![type_row![]; 2], b),
+                [(USIZE_T, i)],
+                type_row![USIZE_T],
+                unknown_ext.clone(),
+            )
+            .unwrap();
+        let mut case1 = cond.case_builder(0).unwrap();
```
acl-cqc, line 672: Why `?` -> `unwrap` everywhere? It does make things a lot longer and look like you are changing more than you are.

Issue #1048, opened 2024-05-15
`serde_yaml::Value` does not perfectly round trip serialize. 


doug-q:
In paticular https://docs.rs/serde_yaml/latest/serde_yaml/value/struct.TaggedValue.html is deserialized as a `Mapping`. Observe this by uncommenting the generator in `crate::proptest::any_serde_value` (currently in review #981).

Ideas:
* use `serde_json::Value` for our "opaque" representation in `CustomSerialized`, `CustomTypeArg`, etc. we already use `serde_json::Value` for metadata.
* Fail to serialize `TaggedValue`s.
* Try to fix the round trip.

zrho:
What is the current state of discussion with respect to YAML? If we are moving to JSON (or at least away from YAML) anyway, there would be nothing to do for this issue.

Tagged values are a really nice feature of YAML that is unfortunately missing in JSON. Therefore JSON formats always have to make up their own encoding of tags, and there isn't a universally agreed upon way to do this. Msgpack has extensions, but they are not quite the same as tagged values, and there does not appear to be an standardised extension for tagged values. Therefore msgpack would do a similar encoding as JSON.

The Rust ecosystem of YAML libraries is in an unfortunate state. In particular, the `serde_yaml` crate is no longer maintained, marked as deprecated on crates.io and the repository is archived.

There is an additional option: Use a custom `Value` type which serialises and deserialises cleanly to JSON and msgpack, but otherwise maps more directly onto the sort of data that we deal with. I might, for instance, want to have metadata consisting of a type, or type scheme, or type row, etc.

PR #1328, opened 2024-07-19
feat!: opaque type arguments are untyped bytes.

### LINKED ISSUE #1308: `CustomTypeArg` should hold a `Value`
Currently `CustomTypeArg` holds a `serde_yaml::Value` payload. We want to change this to a `serde_json::Value` https://github.com/CQCL/hugr/issues/1048 , I suggest instead we should change it to `crate::ops::custom::Value`

ss2165:
Closes #1308

- Previously held a yaml value, this doesn't really make much sense
- If it is an opaque blob, no reason it needs a HUGR custom type attached to it?
- Now truly opaque, it is up to extension-aware tooling to use the bytes data as they see fit
- Serialised as base64 encoded string
- Utilities for common case, strings
- Conducive with adding typed "operation intrinsics/properties" as a follow-up. This would be hugr-typed values that can be attached to operations (in a semantically significant manner, unlike metadata) but perhaps are not allowed to affect the signature.
- should there be a cap on size of byte array? If so should we use `[u8; N]` over `Vec<u8>`?
- Serialisation break to be handled TODO

BREAKING_CHANGE: opaque type arguments are now untyped and hold bytes rather than a yaml value. Serialised form is base64 string.


ss2165:
Converting to draft as along the lines of: https://github.com/CQCL/hugr/issues/1308#issuecomment-2242217082
I'm not sure this is worth merging in this state.

acl-cqc (commented):
Hmmm, so I can see some good points here, but I wonder if we have gone too far in some respects....and not far enough in others ;-)
* If I understand correctly, the "value" can be arbitrary bytes, not necessarily valid UTF-8. But, when we serialize, we base64-encode, and then write that out as "utf-8" (i.e. ASCII, there are no unicode characters in base64!)
   * And, printing the byte array will fail (!) if it's not valid UTF-8. 
   * I think here we might want to go further. Particularly given second point (below). When we are dealing with a *Hugr* (potentially containing "blackbox" ops and/or types that we don't understand), we'll be seeing ports (potentially) with types like "SomeType<SomeCustomTypeInstance>". That is, the type contains a byte array value. Printing this out is *the* thing we need to do here, and I think it'll help a lot if it prints out in a vaguely-human-readable format (or any format, rather than not at all). Why not require the value to be a string (i.e. valid unicode), indeed maybe even base64-encoded, already.
* Secondly, the "untyped" bit. I think we've gone back-and-forth between CustomTypeArgs being "checkable" (type provided from outside) vs "synthesizable" (know their own type) a few times. Here we've very much switched to the former.
   * If you have the extension loaded, then you always can look up the declaration (of the TypeParam for which the TypeArg is a value) to retrieve the type of which the value is an instance. You *can*, but it's much more awkward (in code) to do than if the TypeArg just provides the type.
   * If you don't have the extension then you're even more stuffed (in practice this shouldn't be an issue as type declarations are easily passed around, right)

So I'd go with (a) keep the type in the CustomTypeArg; (b) make the extension serialize to at least `string` rather than `bytes`. I think that still achieves the critical goal of making the extension determine the serialization (and thus fixing roundtrip)....?

acl-cqc:
Also, I don't think this fixes #1308. I agree it fixes #1048, which is the blocking/urgent bug, so it might be worth taking it for that, but keeping #1308 open. However I'd not let this go in as a fix for #1308 until discussion there has resolved.

hugr-core/src/extension/declarative/types.rs
```diff
@@ -100,14 +100,11 @@ impl From<TypeDefBoundDeclaration> for TypeDefBound {
...
-/// Either a type, or a 2-element list containing a human-readable name and a type id.
+/// Only supports opaque parameters for now.
 #[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(untagged)]
 enum TypeParamDeclaration {
-    /// Just the type id.
-    Type(TypeName),
-    /// A 2-tuple containing a human-readable name and a type id.
-    WithDescription(String, TypeName),
+    Opaque,
```
acl-cqc, line 107: Rename this to String also

hugr-core/src/extension/declarative/types.rs
```diff
@@ -119,52 +116,9 @@ impl TypeParamDeclaration {
     /// TODO: The parameter description is currently ignored.
```
acl-cqc, line 116: The parameter description no longer exists, so drop the comment.

Or reinstate the description - it's harmless, could sit there as a reminder what the string is actually supposed to be, but I absolutely do *not* insist

hugr-core/src/types/type_param.rs
```diff
@@ -293,15 +281,7 @@ impl TypeArg {
                 // RowVariables are represented as TypeArg::Variable
                 ty.substitute1(t).into()
             }
-            TypeArg::BoundedNat { .. } => self.clone(), // We do not allow variables as bounds on BoundedNat's
-            TypeArg::Opaque {
-                arg: CustomTypeArg { typ, .. },
-            } => {
-                // The type must be equal to that declared (in a TypeParam) by the instantiated TypeDef,
```
acl-cqc, line 300: Good to see this headscratcher gone

hugr-py/src/hugr/tys.py
```diff
@@ -88,10 +88,8 @@ def to_serial(self) -> stys.BoundedNatParam:
 class OpaqueParam(TypeParam):
     """Opaque type parameter."""
```
acl-cqc, line 89: Update the doc at least, even if you are deciding not to rename the class. (Do we really avoid breakage by leaving it as `OpaqueParam`?? It seems odd not to rename when you *are* renaming the corresponding `OpaqueArg` -> `StringArg`)

hugr-py/src/hugr/serialization/tys.py
```diff
@@ -87,12 +87,11 @@ def deserialize(self) -> tys.BoundedNatParam:
...
 
-class OpaqueParam(BaseTypeParam):
-    tp: Literal["Opaque"] = "Opaque"
-    ty: Opaque
+class StringParam(BaseTypeParam):
+    tp: Literal["String"] = "String"
 
     def deserialize(self) -> tys.OpaqueParam:
-        return tys.OpaqueParam(ty=self.ty.deserialize())
+        return tys.OpaqueParam()
```
acl-cqc, line 94: It does seem odd not to be returning something called `tys.StringParam` here. I won't 100% insist (if there are serious migration issues) but I will strongly encourage

PR #1332, opened 2024-07-22
fix: dfg wrapper build handles incorrect output wire numbers

### LINKED ISSUE #1257: When building a TailLoop, calling `outputs_arr`, you are given an `OrderEdge` "output wire"
Cleanup + reproducer to come.

ss2165:
DFGWrapper handles incorrectly were using internal dfg output number as the outputs from the container node. Fixed for function, case and tailloop. Also fixed a test that relied on this incorrect behaviour.

Closes #1257

### DIFF
--- a/hugr-core/src/builder/handle.rs
+++ b/hugr-core/src/builder/handle.rs
@@ -1,5 +1,5 @@
 //! Handles to nodes in HUGR used during the building phase.
-use crate::ops::handle::{BasicBlockID, CaseID, DfgID, FuncID, NodeHandle, TailLoopID};
+use crate::ops::handle::{BasicBlockID, CaseID, DfgID, FuncID, NodeHandle};
 use crate::ops::OpTag;
 use crate::utils::collect_array;
 use crate::{Node, OutgoingPort, Wire};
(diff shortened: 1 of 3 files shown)


hugr-core/src/builder/handle.rs
```diff
@@ -87,17 +87,7 @@ impl From<BuildHandle<DfgID>> for BuildHandle<CaseID> {
     fn from(value: BuildHandle<DfgID>) -> Self {
         Self {
             node_handle: value.node().into(),
-            num_value_outputs: value.num_value_outputs,
-        }
-    }
-}
-
-impl From<BuildHandle<DfgID>> for BuildHandle<TailLoopID> {
```
aborgna-q, line 95: Could we keep the implementation (dropping the loop flag output)?

aborgna-q (approved):
LGTM

hugr-core/src/builder/handle.rs
ss2165: there may be some outputs embedded in the loop flag output in the "break" variant of the Sum.

PR #1347, opened 2024-07-25
feat(hugr-core): Add `impl Hash for Type`

doug-q:
Note that I have made a choice here to include `PolyFuncType` and `TypeParam`. We could exclude `TypeParam` by not including `TypeArgVariable::cached_decl` in the `Hash` (and `PartialEq`, `Eq`, which would be a change of behaviour).


aborgna-q (approved):
`cached_decl` should be alright, since the value is always present (vs. lazily computed and cached).

LGTM

PR #1352, opened 2024-07-25
fix(hugr-core)!: bump serialisation version with no upgrade path

doug-q:
Since we are breaking serialisation I edited most test cases to use v3. The single test case which I deleted was testing that we were able to l deserialise from an older version, so is now redundant. This should have been done when removing `CustomTypeArg`.

BREAKING CHANGE: All Hugrs serialised with earlier versions will fail to deserialise


hugr-core/src/hugr/serialize.rs
```diff
@@ -48,24 +49,27 @@ enum Versioned<SerHugr = SerHugrLatest> {
...
 impl<T> Versioned<T> {
     pub fn new_latest(t: T) -> Self {
-        Self::V2(t)
+        Self::V3(t)
     }
 }
 
 impl<T: DeserializeOwned> Versioned<T> {
-    fn upgrade(mut self) -> Result<T, UpgradeError> {
+    fn upgrade(#[allow(unused_mut)] mut self) -> Result<T, UpgradeError> {
```
aborgna-q, line 57: The `mut` keyword is internal. We're still taking by value, so the interface doesn't change.
```suggestion
    fn upgrade(self) -> Result<T, UpgradeError> {
```

hugr-py/src/hugr/__init__.py
```diff
@@ -9,4 +9,4 @@
 
 def get_serialisation_version() -> str:
     """Return the current version of the serialization schema."""
-    return "v2"
+    return "v3"
```
aborgna-q, line 12: I don't see any version checks on the `hugr-py` deserialisation. Should we throw an "unsupported version" error there?

It also seems like `Hugr.to_serial` is using a hard-coded "v1" tag instead of calling this function.
https://github.com/CQCL/hugr/blob/a2f34b6ea668dfb233066456e1ea85ff8c1f977c/hugr-py/src/hugr/hugr.py#L562-L568

aborgna-q (commented):
Is there any list of changes in the v3 format? The schema is the same.

hugr-py/src/hugr/__init__.py
ss2165: > It also seems like Hugr.to_serial is using a hard-coded "v1" tag instead of calling this function.

removed

ss2165:
I've bumped all the schemas and usages to a "live" version

aborgna-q (approved):
I preferred using numbered versions, even if forwards-incompatible, so we could throw an error about that instead of randomly failing.

But that may be too annoying to keep updating, so I'm ok with going with `live` versions until we stabilise the `-model`.

PR #1367, opened 2024-07-25
feat!: `Extension` requires a version

### LINKED ISSUE #1357: Add versions to serialised extensions
Extensions can be defined anywhere and may be loaded dynamically. Two extensions with the same qualified name should be distinguishable by version. When in conflict, we should prefer the newer version (with conflict warnings).

ss2165:
Closes #1357 

BREAKING CHANGE: `Extension::new` now requires a `Version` argument. `Extension::new_with_reqs` replaced with `with_reqs` which takes `self` (chain with `new`).

### DIFF
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -330,10 +330,10 @@ impl Extension {
         }
     }
 
-    /// Creates a new extension with the given name and requirements.
+    /// Extend the requirements of this extension with another set of extensions.
     pub fn with_reqs(self, extension_reqs: impl Into<ExtensionSet>) -> Self {
         Self {
-            extension_reqs: extension_reqs.into(),
+            extension_reqs: self.extension_reqs.union(extension_reqs.into()),
             ..self
         }
     }
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -58,10 +58,13 @@ impl ExtensionRegistry {
     ) -> Result<Self, ExtensionRegistryError> {
         let mut exts = BTreeMap::new();
         for ext in value.into_iter() {
+            let ext_v = ext.version().clone();
             let prev = exts.insert(ext.name.clone(), ext);
             if let Some(prev) = prev {
                 return Err(ExtensionRegistryError::AlreadyRegistered(
                     prev.name().clone(),
+                    prev.version().clone(),
+                    ext_v,
                 ));
             };
         }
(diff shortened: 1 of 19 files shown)


hugr-core/src/extension.rs
```diff
@@ -3,6 +3,7 @@
 //! TODO: YAML declaration and parsing. This should be similar to a plugin
 //! system (outside the `types` module), which also parses nested [`OpDef`]s.
 
+pub use semver::Version;
```
ss2165, line 6: pub so downstream crates don't need to add a `semver` dependency

hugr-core/src/extension.rs
```diff
@@ -82,13 +86,35 @@ impl ExtensionRegistry {
...
+                prev.get().version().clone(),
+                extension.version().clone(),
             )),
             btree_map::Entry::Vacant(ve) => Ok(ve.insert(extension)),
         }
     }
 
+    /// Registers a new extension to the registry, keeping most up to date if extension exists.
+    ///
+    /// If extension IDs match, the extension with the higher version is kept.
```
acl-cqc, line 100: I note we are not really doing anything with Semver here (it could be any totally-ordered thing), and I wonder whether this should be the place. Replacing an extension with a breaking-change-upgrade could be problematic; should we
* have a boolean flag here, allow breaking-change upgrades
* keep multiple major versions of the extension (a single minor version of each)? I think this will be too difficult because then we'd have to version every *reference* to an extension.
* Error on any breaking-change upgrade
* Do something at Hugr validation time. In which case maybe just the checks we have are sufficient, so maybe the right answer is to do nothing, i.e. leave PR as it stands....

acl-cqc (approved):
Generally great, thanks @ss2165, tho a few questions in comments.

hugr-passes/src/merge_bbs.rs
```diff
@@ -178,7 +178,7 @@ mod test {
     }
 
     fn extension() -> Extension {
-        let mut e = Extension::new(EXT_ID);
+        let mut e = Extension::new(EXT_ID, hugr_core::extension::Version::new(0, 1, 0));
```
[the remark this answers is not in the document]
ss2165: I use `semver` inside the hugr-core crate and `hugr_core::` outside

hugr-core/src/extension.rs
ss2165: Yes I think I prefer to keep the logic here simple and flexible and deal with valid
updates etc. elsewhere. 

Maybe I should add a "remove" function so the registry is actually fully flexible, and
this is just a utility for update-only changes?

ss2165: added remove

PR #1371, opened 2024-07-26
feat!: Serialised extensions

### LINKED ISSUE #1228: All extensions should have a machine readable source of truth
We need to be able to exchange extension definitions between different hugr tools. 

All extension documentation should be generated by loading the extension and `.print_docs_markdown()`ing it.

Extensions are already serialisable, with the caveat of #1057 
YAML Declarative extensions are one source of machine readable extensions, but not the only one. Extensions should be exchanged in JSON and validated against a versioned schema, just like Hugrs.

Tasks:
 - [ ] #1057 
 - [x] Define a JSON schema for extensions
 - [ ] Remove all extension docs from the spec and replace with markdown generated by extension definitions.
 - [ ] Guppy is able to load an Extension from JSON and automagically provide the ops and types

### LINKED ISSUE #1360: Serialisation schema for extensions.
Defined from pydantic models.

Validate prelude and standard extensions against schema.

### LINKED ISSUE #1361: Add "requires binary" flag to opdef signatures.
Operations in Extension declarations can state one of:
- An explicit type scheme 
- A type scheme + a flag indicating binary required for validation, in-memory extensions should point to this function.
- A flag indicating code required to compute the type scheme, in-memory extensions should point to this function.

ss2165:
Fear not the diff, mostly schema update.

- Define a Pydantic model and serialised schema for extensions.
- Update the rust `SignatureFunc` serialisation to be compatible with this. 
- Serialized extension "declarations" can state they require binary compute or validation functions. `SignatureFunc` reports this if they are missing, and it is up to the caller (typically validation) to decide how to recover from this. 

Closes #1360 
Closes #1361

Addresses parts of #1228 

Best reviewed as individual commits.

BREAKING CHANGE: `TypeDefBound` uses struct-variants for serialization. `SignatureFunc` now has variants for missing binary functions, and serializes in to a new format that indicates expected binaries.


hugr-core/src/extension/op_def.rs
```diff
@@ -117,12 +117,9 @@ pub trait CustomLowerFunc: Send + Sync {
 /// Encode a signature as `PolyFuncTypeRV` but optionally allow validating type
 /// arguments via a custom binary. The binary cannot be serialized so will be
 /// lost over a serialization round-trip.
-#[derive(serde::Deserialize, serde::Serialize)]
 pub struct CustomValidator {
-    #[serde(flatten)]
     poly_func: PolyFuncTypeRV,
-    #[serde(skip)]
-    pub(crate) validate: Box<dyn ValidateTypeArgs>,
+    pub(crate) validate: Option<Box<dyn ValidateTypeArgs>>,
```
acl-cqc, line 122: If `None` here means "no custom validation required" then agreed, but in that case let's rename (or even inline) `struct CustomValidator`

hugr-core/src/extension/op_def.rs
```diff
@@ -142,26 +139,80 @@ impl CustomValidator {
...
     /// A PolyFuncType (polymorphic function type), with optional custom
     /// validation for provided type arguments,
-    #[serde(rename = "signature")]
     PolyFuncType(CustomValidator),
-    #[serde(skip)]
     /// A custom binary which computes a polymorphic function type given values
     /// for its static type parameters.
     CustomFunc(Box<dyn CustomSignatureFunc>),
+    /// Declaration specified a custom binary but it was not provided.
+    MissingComputeFunc,
```
acl-cqc, line 159: What about a missing custom validation func? In either case I think we just have to trust the cache, maybe with a warning, so I guess if the warning is the same then we don't have to distinguish, is that the plan?

hugr-core/src/extension/op_def.rs
```diff
@@ -142,26 +139,80 @@ impl CustomValidator {
...
+mod serialize_signature_func {
+    use serde::{Deserialize, Serialize};
+
+    use super::{PolyFuncTypeRV, SignatureFunc};
+    #[derive(serde::Deserialize, serde::Serialize)]
+    struct SerSignatureFunc {
+        /// If the type scheme is available explicitly, store it.
+        signature: Option<PolyFuncTypeRV>,
+        /// Whether an associated binary function is expected.
+        /// If `signature` is `None`, a true value here indicates a custom compute function.
```
acl-cqc, line 171: Nice, and good docs, ok, that makes sense

hugr-core/src/extension/op_def.rs
```diff
@@ -142,26 +139,80 @@ impl CustomValidator {
...
+        .serialize(serializer)
+    }
+
+    pub(super) fn deserialize<'de, D>(deserializer: D) -> Result<super::SignatureFunc, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let SerSignatureFunc { signature, .. } = SerSignatureFunc::deserialize(deserializer)?;
+
+        // TODO for now, an indication of expected custom validation function is ignored.
```
acl-cqc, line 206: Ah ok. Maybe just add a bool field in SignatureFunc::MissingComputeFunc (renaming to MissingFunc and using an enum rather than bool would be better)

hugr-core/src/extension/op_def.rs
```diff
@@ -142,26 +139,80 @@ impl CustomValidator {
...
+        Ok(if let Some(sig) = signature {
```
acl-cqc, line 207: nit - this is something like `signature.map_or(MissingComputeFunc, super::SignatureFunc::From)` where the `MissingComputeFunc` is the default for `None`

hugr-core/src/extension/op_def.rs
```diff
@@ -217,11 +268,12 @@ impl From<CustomValidator> for SignatureFunc {
...
 
 impl SignatureFunc {
-    fn static_params(&self) -> &[TypeParam] {
-        match self {
+    fn static_params(&self) -> Result<&[TypeParam], SignatureError> {
+        Ok(match self {
             SignatureFunc::PolyFuncType(ts) => ts.poly_func.params(),
             SignatureFunc::CustomFunc(func) => func.static_params(),
-        }
+            SignatureFunc::MissingComputeFunc => return Err(SignatureError::MissingComputeFunc),
```
acl-cqc, line 275: if `SignatureFunc::ComputeFunc` specified which kind of function was missing, you could handle the "missing validation" by returning empty

hugr-core/src/extension/op_def.rs
```diff
@@ -217,11 +268,12 @@ impl From<CustomValidator> for SignatureFunc {
 }
 
 impl SignatureFunc {
-    fn static_params(&self) -> &[TypeParam] {
-        match self {
+    fn static_params(&self) -> Result<&[TypeParam], SignatureError> {
+        Ok(match self {
             SignatureFunc::PolyFuncType(ts) => ts.poly_func.params(),
```
acl-cqc, line 273: aaiieeee is this one even correct? "static_params" are the ones to pass to binary compute_signature, so if there isn't such, I think this should be the empty list

hugr-core/src/extension/type_def.rs
```diff
@@ -13,16 +13,18 @@ use crate::types::TypeBound;
 
 /// The type bound of a [`TypeDef`]
 #[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[serde(tag = "b")]
+#[allow(missing_docs)]
 pub enum TypeDefBound {
     /// Defined by an explicit bound.
-    Explicit(TypeBound),
+    Explicit { bound: TypeBound },
```
acl-cqc, line 20: Presumably we will be able to stop doing these annoying renames-to-support serde when we have `hugr-model` :)

ss2165:
I intend to pull out the CLI extension dumping in to a follow up PR. In this PR will test that serialisations of standard extensions validate against schema. In follow up will validate against schema when generating from CLI + store cli-generated extensions in source tree + add a test that running generation matches stored extensions (an update-extensions script for updating them).

acl-cqc (commented):
So far looks good! I haven't seen the problem(s) yet, although I was thinking perhaps adding a struct for Hugr+extensions, with a `validate` method (maybe making a `trait Validatable` would be a good start), might come first

acl-cqc:
Does a PR that just tackles "Extension+OpDef can state there is a binary compute_signature/validate function but we don't have it here" work on its own, and then enable separate/parallel follow-ups of (a) serialization, cli, .... (b) remove opaqueop?

hugr-core/src/extension/op_def.rs
ss2165: yeah I found these functions aren't actually used (therefore aren't tested...) will add an issue.

hugr-core/src/extension/op_def.rs
ss2165: I'm not sure - in the case of missing validation function do you trust the cached signature or use the type scheme to generate the signature and check against cache as you would without custom validation?

acl-cqc: Good point. Given the validation function can only say one of two things - invalid, or use the type scheme - you can try the latter (which might reject if TypeArgs don't match the TypeParams, say) and see if that matches; that might get you an error, but even if the typescheme says ok, if there's a binary validation function that you haven't got, then that still has to be a warning

hugr-core/src/extension/op_def.rs
ss2165: a rename or inline would be a breaking change, which certainly doesn't have to be
avoided but in this case I think just enough docstrings should do the trick?

hugr-py/src/hugr/serialization/ops.py
```diff
@@ -658,27 +658,6 @@ class OpType(RootModel):
...
-# --------------- OpDef ----------------
-# --------------------------------------
-
-
-class FixedHugr(ConfiguredBaseModel):
-    extensions: ExtensionSet
-    hugr: Any
-
-
-class OpDef(ConfiguredBaseModel, populate_by_name=True):
```
ss2165, line 671: moved to extension.py

hugr-core/src/extension/op_def.rs
ss2165: I've decided to capture missing validation explicitly.

hugr-core/src/extension/op_def.rs
acl-cqc: A breaking change just to people using the rust classes direct, but not to serialization (as this isn't serialized), right? I think that'd be OK, but see comment on `MissingValidateFunc` for an alternative "proposal"....

If this does affect serialization - I think the long-term plan here is to separate out a definition of serialization structure independent of this, in which case we can rename this without breaking serialization. But given this is no longer `#[serde]` I think this is not the case....

hugr-core/src/extension/op_def.rs
```diff
@@ -117,26 +118,17 @@ pub trait CustomLowerFunc: Send + Sync {
 /// Encode a signature as `PolyFuncTypeRV` but optionally allow validating type
 /// arguments via a custom binary. The binary cannot be serialized so will be
 /// lost over a serialization round-trip.
-#[derive(serde::Deserialize, serde::Serialize)]
 pub struct CustomValidator {
-    #[serde(flatten)]
     poly_func: PolyFuncTypeRV,
-    #[serde(skip)]
+    //Optional custom function for validating type arguments before returning the signature.
```
acl-cqc: This doesn't look very optional ?

hugr-core/src/extension/op_def.rs
```diff
@@ -147,21 +139,22 @@ impl CustomValidator {
...
-    // CustomSignatureFunc trait too, and replace this enum with Box<dyn CustomSignatureFunc>.
-    // However instead we treat all CustomFunc's as non-serializable.
-    /// A PolyFuncType (polymorphic function type), with optional custom
-    /// validation for provided type arguments,
-    #[serde(rename = "signature")]
-    PolyFuncType(CustomValidator),
-    #[serde(skip)]
+    /// An explicit polymorphic function type.
+    PolyFuncType(PolyFuncTypeRV),
+    /// A polymorphic function type with a custom binary for validating type arguments.
```
acl-cqc, line 145: maybe we could slightly strengthen that this is mostly the same as the previous case but PLUS a custom binary

hugr-core/src/extension/op_def.rs
acl-cqc: Right, I'd started commenting in a different direction, but making this non-optional is a different way of doing what I was going to suggest, that works for me. Just correct the comment tho :)

hugr-core/src/extension/op_def.rs
```diff
@@ -147,21 +139,22 @@ impl CustomValidator {
...
+    /// Serialized declaration specified a custom validate binary but it was not provided.
+    MissingValidateFunc(PolyFuncTypeRV),
     /// A custom binary which computes a polymorphic function type given values
     /// for its static type parameters.
     CustomFunc(Box<dyn CustomSignatureFunc>),
+    /// Serialized declaration specified a custom compute binary but it was not provided.
+    MissingComputeFunc,
 }
+
+#[derive(PartialEq, Eq, Debug)]
```
acl-cqc, line 156: Methinks we should not need `NoValidate`. One should use `Signature::PolyFuncType` instead.

hugr-core/src/extension/op_def.rs
```diff
@@ -243,10 +247,11 @@ impl SignatureFunc {
     ) -> Result<Signature, SignatureError> {
         let temp: PolyFuncTypeRV; // to keep alive
         let (pf, args) = match &self {
-            SignatureFunc::PolyFuncType(custom) => {
-                custom.validate.validate(args, def, exts)?;
+            SignatureFunc::CustomValidator(custom) => {
+                custom.validate.as_ref().validate(args, def, exts)?;
```
acl-cqc, line 250: presumably this extra `as_ref()` is needed? Code looks very similar to what we had before.

hugr-core/src/extension/op_def.rs
```diff
@@ -557,12 +575,14 @@ pub(super) mod test {
...
+                SignatureFunc::CustomValidator(CustomValidator {
                     poly_func,
                     validate: _,
-                }) => Some(poly_func.clone()),
+                })
+                | SignatureFunc::PolyFuncType(poly_func)
+                | SignatureFunc::MissingValidateFunc(poly_func) => Some(poly_func.clone()),
                 // This is ruled out by `new()` but leave it here for later.
-                SignatureFunc::CustomFunc(_) => None,
+                SignatureFunc::CustomFunc(_) | SignatureFunc::MissingComputeFunc => None,
```
acl-cqc, line 584: I suspect the comment on the previous line can go (I suspect "later" meant "when we have implemented serialization", essentially!), please clarify if I'm wrong

hugr-core/src/extension/op_def/serialize_signature_func.rs
```diff
@@ -0,0 +1,171 @@
...
+struct SerSignatureFunc {
+    /// If the type scheme is available explicitly, store it.
+    signature: Option<PolyFuncTypeRV>,
+    /// Whether an associated binary function is expected.
+    /// If `signature` is `None`, a true value here indicates a custom compute function.
+    /// If `signature` is not `None`, a true value here indicates a custom validation function.
+    binary: bool,
+}
+
+pub(super) fn serialize<S>(value: &super::SignatureFunc, serializer: S) -> Result<S::Ok, S::Error>
```
acl-cqc, line 14: can('t) you do a `#[serde(into=....)]` to avoid having to manually implement `fn serialize` ?

hugr-core/src/extension/op_def/serialize_signature_func.rs
```diff
@@ -0,0 +1,171 @@
...
+        (Some(sig), false) => Ok(sig.into()),
+        (Some(sig), true) => Ok(SignatureFunc::MissingValidateFunc(sig)),
+        (None, true) => Ok(SignatureFunc::MissingComputeFunc),
+        (None, false) => Err(serde::de::Error::custom(
+            "No signature provided and custom computation not expected.",
+        )),
+    }
+}
+#[derive(serde::Deserialize, serde::Serialize, Debug)]
+/// Wrapper we can derive serde for, to allow round-trip serialization
```
acl-cqc, line 52: Does this belong inside `mod test` ?

hugr-core/src/extension/prelude.rs
```diff
@@ -95,14 +95,14 @@ lazy_static! {
                 TypeName::new_inline("usize"),
                 vec![],
                 "usize".into(),
-                TypeDefBound::Explicit(crate::types::TypeBound::Copyable),
+                TypeDefBound::Explicit { bound: crate::types::TypeBound::Copyable },
```
acl-cqc, line 98: you might consider factory functions `copyable()` and `any`/`linear()` on TypeDefBound

hugr-core/src/hugr/serialize/test.rs
```diff
@@ -517,6 +517,18 @@ fn roundtrip_optype(#[case] optype: impl Into<OpType> + std::fmt::Debug) {
...
 }
 
+#[test]
+// test all standard extension serialisations are valid against scheme
+fn std_extensions_valid() {
+    let std_reg = crate::std_extensions::std_reg();
+    for (_, ext) in std_reg.into_iter() {
+        let val = serde_json::to_value(ext).unwrap();
+        NamedSchema::check_schemas(&val, get_schemas(true));
+        // check deserialises correctly, can't check equality because of custom binaries.
```
acl-cqc, line 527: However I think you could try reserializing the deserialized Extension to check equality with the original?
```
deserialize(serialize(ext)) != ext // because custom binaries
serialize(deserialize(serialize(ext))) == serialize(ext) // I think

File: hugr-py/pyproject.toml [hunk 19]
```diff
@@ -26,6 +26,8 @@ repository = "https://github.com/CQCL/hugr"
 [tool.poetry.dependencies]
 python = ">=3.10"
 pydantic = ">=2.7,<2.9"
+pydantic-extra-types = "^2.9.0"
+semver = "^3.0.2"
```
acl-cqc, line 30: nit: does this belong in this PR?

hugr-py/src/hugr/serialization/extension.py
```diff
@@ -0,0 +1,82 @@
...
+    types: dict[str, TypeDef]
+    values: dict[str, ExtensionValue]
+    operations: dict[str, OpDef]
+
+    @classmethod
+    def get_version(cls) -> str:
+        return get_serialisation_version()
+
+    @classmethod
+    def _pydantic_rebuild(cls, config: pd.ConfigDict | None = None, **kwargs):
```
acl-cqc, line 81: Might be good to comment why you need this for the uninitiated

hugr-py/src/hugr/serialization/ops.py
acl-cqc: Yes good call

hugr-py/pyproject.toml
acl-cqc: Ah, you added Rust semver in an earlier PR but didn't update the python, and tests added here reveal this?

hugr-core/src/extension/op_def.rs
```diff
@@ -115,13 +115,12 @@ pub trait CustomLowerFunc: Send + Sync {
     ) -> Option<Hugr>;
 }
 
-/// Encode a signature as `PolyFuncTypeRV` but optionally allow validating type
+/// Encode a signature as `PolyFuncTypeRV` but allow validating type
```
acl-cqc, line 118:
```suggestion
/// Encode a signature as [PolyFuncTypeRV] but with additional (custom) validation of type args
```

We don't just allow validating, we require it!

acl-cqc (approved):
Nice :). All major design considerations from earlier comments neatly addressed, thanks. Minor issues only, thanks!

(I think - if I'm barking up the wrong tree in places then feel free to ping me but I think you can probably take it from here)

hugr-core/src/extension/op_def/serialize_signature_func.rs
ss2165: that requires `Clone` on `SignatureFunc` which I didn't necessarily want to add all the way down.

hugr-py/pyproject.toml
ss2165: the earlier PR added versions to the rust struct, but there was not matching python model. This PR adds the python model, and with it the semver dependency for the python model.

hugr-core/src/hugr/serialize/test.rs
ss2165: yes nice thanks

hugr-core/src/extension/op_def.rs
ss2165: yes, moved it just to the test that used it

hugr-core/src/extension/op_def.rs
ss2165: wasn't needed, removed

hugr-py/src/hugr/serialization/extension.py
ss2165: removed

hugr-core/src/extension/prelude.rs
ss2165: nice idea, will do

PR #1377, opened 2024-07-29
feat: serialised standard extensions

### LINKED ISSUE #1228: All extensions should have a machine readable source of truth

ss2165:
Closes #1228 

Add cli for generating serialised extensions, and justfile/CI for checking checked in definitions match. 

Switched from `HashMap` to `BTreeMap` in `Extension` to ensure serialization happens in alphabetical order.

### DIFF
--- a/hugr-cli/src/extensions.rs
+++ b/hugr-cli/src/extensions.rs
@@ -40,7 +40,8 @@ impl ExtArgs {
 
             serde_json::to_writer_pretty(&mut file, &ext).unwrap();
 
-            // write newline, for pre-commit checks.
+            // write newline, for pre-commit end of file check that edits the file to
+            // add newlines if missing.
             file.write_all(b"\n").unwrap();
         }
     }
--- a/hugr-cli/src/extensions.rs
+++ b/hugr-cli/src/extensions.rs
@@ -1,6 +1,6 @@
 //! Dump standard extensions in serialized form.
 use clap::Parser;
-use std::path::PathBuf;
+use std::{io::Write, path::PathBuf};
 
 /// Dump the standard extensions.
 #[derive(Parser, Debug)]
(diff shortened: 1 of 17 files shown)


hugr-cli/src/extensions.rs
```diff
@@ -0,0 +1,47 @@
...
+            path.set_extension("json");
+
+            std::fs::create_dir_all(path.clone().parent().unwrap()).unwrap();
+            // file buffer
+            let mut file = std::fs::File::create(&path).unwrap();
+
+            serde_json::to_writer_pretty(&mut file, &ext).unwrap();
+
+            // write newline, for pre-commit checks.
+            file.write_all(b"\n").unwrap();
```
mark-koch, line 44: Why is this related to the pre-commit check?

justfile
```diff
@@ -55,6 +55,9 @@ update-schema:
     poetry update
     poetry run python scripts/generate_schema.py specification/schema/
 
+# Generate serialized declarations for the standard extensions and prelude.
+gen-extensions:
+    cargo build -p hugr-cli && poetry run ./target/debug/hugr gen-extensions -o specification/std_extensions
```
mark-koch, line 60: Why not `cargo run`?

mark-koch (approved):

hugr-cli/src/extensions.rs
ss2165: expanded comment

PR #1383, opened 2024-07-30
ci: move std extensions check to rust workflow

ss2165:


.github/workflows/ci-rs.yml
```diff
@@ -145,6 +152,31 @@ jobs:
       - name: Tests with all features
         run: cargo test --verbose --workspace --all-features
 
+  # Ensure that serialized extensions match rust implementation
+  std-extensions:
+    needs: [tests-stable-all-features]
```
aborgna-q, line 157: Make file diff requirement explicit?
```suggestion
    needs: [changes, tests-stable-all-features]
    if: ${{ needs.changes.outputs.rust == 'true' && github.event_name != 'merge_group' }}
```
We may want to run this on python changes too, but that'll require enabling all the required steps too.

ss2165: this should be independent of python changes, though that might change

aborgna-q (approved):

PR #1384, opened 2024-07-30
ci: Fix pr-title not running

aborgna-q:
It seems that skipping reusable workflow jobs causes inconsistent check names between `pull_request_target` and `merge_group`.
For the non-skipped PR runs we get a `local_job/remote_job` name,
but the skipped jobs in the merge queue are just called `local_job`.... (even if skipped in the called workflow instead of locally).

I updated the reusable workflow to skip individual steps instead of the whole job, so now we shouldn't have to deal with that inconsistency.

I will need to update the required check name to `local/remote`

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -14,8 +14,6 @@ on:
 
 jobs:
     check-title:
-        name: Validate Conventional Commit PR title
-        if: github.event_name == 'pull_request_target'
         uses: CQCL/hugrverse-actions/.github/workflows/pr-title.yml@main
         secrets:
             GITHUB_PAT: ${{ secrets.HUGRBOT_PAT }}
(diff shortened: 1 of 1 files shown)


aborgna-q:
For some reason the check in this PR is called `remote/remote` 🤦 .
image

Let's just try and see if it works...

aborgna-q:
...and the check is called `local/remote` on the merge queue
image

aborgna-q:
I _think_ it's because the workflow is running with the file from `main` here...

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

This would explain why the merge queue uses the renamed branch but the PR doesn't.
I removed the job name from the remote definition, and this check was still called `Validate (...)/main` even though "Validate (...)" is not defined anywhere.

Let's merge this, and hopefully new PRs will use the fixed config.

PR #1386, opened 2024-07-31
feat!: Update remaining builder methods to "infer by default"

### LINKED ISSUE #1318: Remaining bits of builder frontend for inference
- [ ] `cfg_builder` (CFGBuilder::new takes a FunctionType so no need to update),
- [ ] `tail_loop_builder`
- [ ] `ConditionalBuilder::new`
- [ ] `TailLoopBuilder::new`

+ any others not already covered in #1226 or previously

acl-cqc:
closes #1318...unless there are any others I've missed (but those taking FunctionType don't need update)

BREAKING CHANGE: `cfg_builder`, `tail_loop_builder`, `ConditionalBuilder::new` and `TailLoopBuilder::new` no longer take an ExtensionSet parameter; either remove the argument (to use extension inference) or use the `_exts` variant

### DIFF
--- a/hugr-core/src/builder/build_traits.rs
+++ b/hugr-core/src/builder/build_traits.rs
@@ -426,17 +426,39 @@ pub trait Dataflow: Container {
     /// The `inputs` must be an iterable over pairs of the type of the input and
     /// the corresponding wire.
     /// The `output_types` are the types of the outputs.
+    /// The extension delta will be inferred.
     ///
     /// # Errors
     ///
     /// This function will return an error if there is an error when building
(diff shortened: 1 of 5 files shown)


acl-cqc:
@aborgna-q manually rolling the die here as @zrho is away in Scotland but has not set himself to "Busy"...hope that's ok!

aborgna-q (approved):
Nice.

Does `BlockBuilder` not get an inferred variant?

acl-cqc:
Good spot, thanks. I will draw up a table with all the different nodes and check we really have dealt with all of them....

PR #1390, opened 2024-08-02
 feat(hugr-cli)!: move mermaid to own sub-command

ss2165:
drive-by replace `clap_stdin` with `clio` as that's that `clap` recommends, and it comes with stdout too.

BREAKING CHANGE: Cli validate command no longer has a mermaid option, use `mermaid` sub-command instead.


ss2165:
Yes perhaps, the change is because `validate --mermaid --no-validate` is just a confusing command to me.
I can common up the validation so mermaid also calls it, and add the `--no-validate` back to the mermaid command. Does that suit?

aborgna-q:
> I would not be surprised if mermaid-printing an invalid Hugr caused a panic.

It shouldn't, as long as the underlying portgraph is working correctly.
I think not having rendering fail for unrelated errors is a good thing.

ss2165:
> I can common up the validation so mermaid also calls it,

In latest commit if you want to look at it. Though broadly I agree with @aborgna-q  and would prefer to revert it.

PR #1393, opened 2024-08-02
feat(hugr-py)!: Reexport commonly used classes from the package root

aborgna-q:
I had to move `serialization_version` and delay a an `import hugr` to avoid cyclic dependencies.

drive-by: Import `Edge` / `OrderEdge` aliases from guppy

BREAKING CHANGE: Moved `hugr.get_serialization_version` to `hugr.serialization.serial_hugr.serialization_version`

### DIFF
--- a/hugr-py/src/hugr/node_port.py
+++ b/hugr-py/src/hugr/node_port.py
@@ -28,8 +28,6 @@ class Direction(Enum):
 
 NodeIdx = int
 PortOffset = int
-Edge = tuple["OutPort", "InPort"]
-OrderEdge = tuple[NodeIdx, NodeIdx]
 
 
 @dataclass(frozen=True, eq=True, order=True)
--- a/hugr-py/src/hugr/node_port.py
+++ b/hugr-py/src/hugr/node_port.py
@@ -28,6 +28,8 @@ class Direction(Enum):
 
 NodeIdx = int
 PortOffset = int
+Edge = tuple["OutPort", "InPort"]
+OrderEdge = tuple[NodeIdx, NodeIdx]
 
 
 @dataclass(frozen=True, eq=True, order=True)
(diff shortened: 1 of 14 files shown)


ss2165:
I am very wary - recommended practice now is to import from where things are defined and we risk tying our shoe laces together with circular imports

hugr-py/src/hugr/serialization/serial_hugr.py
```diff
@@ -13,8 +11,14 @@
 Port = tuple[NodeIdx, PortOffset | None]
 Edge = tuple[Port, Port]
 
+
+def serialization_version() -> str:
+    """Return the current version of the serialization schema."""
+    return "live"
```
acl-cqc, line 17: Is this really the best we can do? (Sorry - that sounds a bit rude! But I'd hope that it might be, oh, replaced with an actual version number by some part of the release process or something?? That would be fine, but if so I think it'd be good to have a comment here explaining that)

hugr-py/src/hugr/node_port.py
```diff
@@ -28,6 +28,8 @@ class Direction(Enum):
 
 NodeIdx = int
 PortOffset = int
+Edge = tuple["OutPort", "InPort"]
```
acl-cqc, line 31: Is this related? (Neither of these is re-exported)

acl-cqc (approved):
Thanks @aborgna-q  - I approve of the move :), everything else looks fine, thanks :). My complaint about `"live"` is not a new issue in this PR, but I am nonetheless curious....

hugr-py/src/hugr/serialization/serial_hugr.py
aborgna-q: I'm not really happy with the unversioned versions, but it's the "no guarantees" value we are currently used. I guess we'll revisit it once `hugr-model` is implemented...

hugr-py/src/hugr/node_port.py
aborgna-q: Not at all :P
It was part of the port of useful things from guppy, but I should probably remove it from here.

PR #1394, opened 2024-08-02
feat(hugr-py): Allow defining functions, consts, and aliases inside DFGs

aborgna-q:
Move `define_function` and `add_alias_defn` from `Module` to a common root for both it and `DFG`.
Move both `add_const` definitions to that common base.

I had to move `Function` to `dfg.py` due to circular deps, but I added a reexport to avoid breaking changes.


ss2165 (approved):
yes!

PR #1396, opened 2024-08-06
docs: Clarify CustomConst::equal_consts

acl-cqc:
This is just explaining existing behaviour, but I think it's important to explain that we are not expecting the equivalent of `Eq` on a structure that could be just an `f64`....

This does constrain impls to be symmetric and transitive, but all current ones are, and I couldn't think of any... (indeed, "either Eq or always return false" would describe all current impls, and this is still significantly more permissive than that.)

### DIFF
--- a/hugr-core/src/ops/constant/custom.rs
+++ b/hugr-core/src/ops/constant/custom.rs
@@ -75,6 +75,11 @@ pub trait CustomConst:
     /// Compare two constants for equality, using downcasting and comparing the definitions.
     ///
     /// If the type implements `PartialEq`, use [`downcast_equal_consts`] to compare the values.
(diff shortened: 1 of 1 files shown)


acl-cqc:
This is a comment-only PR, I reject `codecov` bot's opinion here :-p

aborgna-q:
Do you think there's _any_ custom type for which we wouldn't want to have transitiveness in an equality check? It seems a bit unexpected for that to happen.

acl-cqc:
I can think of crazy ideas, but I can't think of any good ones...
e.g. a Hugr value where we put "some" effort into testing equivalence?

Ok, gone with transitivity too.

PR #1397, opened 2024-08-06
feat!: Allow CustomConsts to (optionally) be hashable

acl-cqc:
* Add trait TryHash as prereq for CustomConst
* Automatically impl'd if your const impl's Hash
* Can also trivially implement (i.e. `impl TryHash for Foo { }`) to say "no, not hashable"
* Derive Hash for most consts, but not ConstF64

BREAKING CHANGE: any `impl CustomConst` will need to either `impl Hash` or `impl MaybeHash`


acl-cqc:
This is not really *needed* yet but seems like a very good thing to have for dataflow analysis and not a bad idea anyway: most constants *are* hashable!

The approach taken with MaybeHash contrasts with `equal_consts` which requires every custom-const struct that impls PartialEq to *also* override `equal_consts` using the `downcast_equal_consts` helper, avoided here. (Instead, those that *don't* implement `Hash` have to empty-implement `MaybeHash`, so it's not a free lunch, but that effort is pretty minimal and applies only in the minority of cases, whereas `downcast_equal_consts` has to be used in the *majority* of cases...

Hence, picking @aborgna-q out here as he may know of better/rustier ways here. Can we `Downcast` to see if a thing impl's `Hash` ??

hugr-core/src/ops/constant/custom.rs
```diff
@@ -90,6 +88,33 @@ pub trait CustomConst:
...
+    /// This relates with [CustomConst::equal_consts] just like [Hash] with [Eq]:
+    /// * if `x.equal_consts(y)` ==> `x.maybe_hash(s)` behaves equivalently to `y.maybe_hash(s)`
+    /// * if `x.hash(s)` behaves differently from `y.hash(s)` ==> `x.equal_consts(y) == false`
+    ///
+    /// As with [Hash], these requirements can trivially be satisfied by either
+    /// * `equal_consts` always returning `false`, or
+    /// * `maybe_hash` always behaving the same (e.g. returning `false`, as it does by default)
+    ///
+    /// Note: this uses `dyn` rather than being parametrized by `<H: Hasher>` so that we can
+    /// still use `dyn CustomConst`.
```
aborgna-q, line 105: I.e. wee need the trait to be [object safe](https://doc.rust-lang.org/reference/items/traits.html#object-safety)

hugr-core/src/ops/constant.rs
```diff
@@ -508,6 +529,17 @@ impl Value {
             None
         }
     }
+
+    /// Hashes this value, if possible. [Value::Extension]s are hashable according
+    /// to their implementation of [MaybeHash]; [Value::Function]s never are;
+    /// [Value::Sum]s are if their contents are.
+    pub fn maybe_hash<H: Hasher>(&self, st: &mut H) -> bool {
+        match self {
+            Value::Extension { e } => e.value().maybe_hash(&mut Box::new(st)),
```
aborgna-q, line 538:
```suggestion
            Value::Extension { e } => e.value().maybe_hash(&mut *st),
```

hugr-core/src/ops/constant/custom.rs
```diff
@@ -90,6 +88,33 @@ pub trait CustomConst:
...
+    fn maybe_hash(&self, _state: &mut dyn Hasher) -> bool {
+        false
+    }
+}
+
+impl<T: Hash> MaybeHash for T {
+    fn maybe_hash(&self, st: &mut dyn Hasher) -> bool {
+        Hash::hash(self, &mut Box::new(st));
```
aborgna-q, line 113:
```suggestion
    fn maybe_hash(&self, mut st: &mut dyn Hasher) -> bool {
        Hash::hash(self, &mut st);
```

hugr-core/src/ops/constant/custom.rs
```diff
@@ -253,6 +278,14 @@ impl CustomSerialized {
     }
 }
 
+impl MaybeHash for CustomSerialized {
+    fn maybe_hash(&self, state: &mut dyn Hasher) -> bool {
+        // Consistent with equality, same serialization <=> same hash.
+        self.value.to_string().hash(&mut Box::new(state));
```
aborgna-q, line 284:
```suggestion
    fn maybe_hash(&self, mut state: &mut dyn Hasher) -> bool {
        // Consistent with equality, same serialization <=> same hash.
        self.value.to_string().hash(&mut state);
```

hugr-core/src/std_extensions/collections.rs
```diff
@@ -56,6 +58,16 @@ impl ListValue {
     }
 }
 
+impl MaybeHash for ListValue {
+    fn maybe_hash(&self, st: &mut dyn Hasher) -> bool {
+        let mut b = Box::new(st);
+        maybe_hash_values(&self.0, &mut b) && {
+            self.1.hash(&mut b);
```
aborgna-q, line 65:
```suggestion
    fn maybe_hash(&self, mut st: &mut dyn Hasher) -> bool {
        maybe_hash_values(&self.0, &mut st) && {
            self.1.hash(&mut st);
```

aborgna-q (commented):
Drop all those allocations!

hugr-core/src/ops/constant.rs
acl-cqc: Nice, that one works...

hugr-core/src/ops/constant/custom.rs
acl-cqc: That gives me:
```
error[E0277]: the size for values of type `dyn Hasher` cannot be known at compilation time
   --> hugr-core/src/ops/constant/custom.rs:113:26
    |
113 |         Hash::hash(self, &mut *st);
    |         ----------       ^^^^^^^^ doesn't have a size known at compile-time
    |         |
    |         required by a bound introduced by this call
    |
    = help: the trait `Sized` is not implemented for `dyn Hasher`
```
which I admit is odd - I'm not trying to pass a Hasher, only a `&mut` to it. Am I missing something?

acl-cqc:
> There is no reflection in Rust, so there's no way to branch on whether a type implements a trait or not (not with stable code at least).
> 
> This looks like a good alternative. We can discuss whether to define a new trait vs adding another trait method to `CustomConst` (more verbose vs less clear), but they are the same solution.
> 
> I may suggest renaming it to `TryHash` instead, to follow the existing name convention (e.g. `TryFrom`).

I like `TryHash`, and indeed `try_hash`.

Also I admit I'm not really clear why we don't require PartialEq as a prereq for CustomConst. You can always implement with `return false`, which is what CustomConst will give you by default if you don't...

hugr-core/src/ops/constant/custom.rs
acl-cqc: Ah sorry I'd put in an extra `*`

aborgna-q: `*st` is trying to dereference an unsized value into the stack.
The suggestion avoids that by using `T : Hash` $\implies$ &mut T : Hash`

acl-cqc: This does look to me as if I'm passing a mutable reference to a mutable reference!

hugr-core/src/ops/constant/custom.rs
acl-cqc: Comment updated.

aborgna-q (approved):
> Patch coverage is 4.76190%

🫠

acl-cqc:
> > Patch coverage is 4.76190%
> 
> 🫠

Sorry, I really did think there was already some code hashing lists. Maybe I got confused with stuff in a previous iteration of "const_fold2" 😅 ...anyway, I've added enough to make it green, could dup much of that into collections too I suppose to get a final few percent.

PR #1398, opened 2024-08-07
feat: Add Eq op to logic extension

croyzor:


hugr-core/src/std_extensions/logic.rs
```diff
@@ -47,6 +47,13 @@ impl ConstFold for NaryLogic {
                 (res || inps.len() as u64 == num_args)
                     .then_some(vec![(0.into(), ops::Value::from_bool(res))])
             }
+            Self::Eq => {
+                let inps = read_inputs(consts)?;
+                let res = inps.clone().into_iter().reduce(|a, b| a == b)?;
+                // We can only fold to true if we have a const for all our inputs
```
acl-cqc, line 53:
```suggestion
                // If we have only some inputs, we can still fold to false, but not to true
```

hugr-core/src/std_extensions/logic.rs
```diff
@@ -331,6 +340,8 @@ pub(crate) mod test {
     #[case(NaryLogic::Or, [], false)]
     #[case(NaryLogic::Or, [false, false, true], true)]
     #[case(NaryLogic::Or, [false, false, false], false)]
+    #[case(NaryLogic::Eq, [true, true, false, true], false)]
+    #[case(NaryLogic::Eq, [false, false], true)]
```
acl-cqc, line 344: Add some couple of cases to `nary_partial_const_fold` too?

hugr-core/src/std_extensions/logic.rs
```diff
@@ -47,6 +47,13 @@ impl ConstFold for NaryLogic {
                 (res || inps.len() as u64 == num_args)
                     .then_some(vec![(0.into(), ops::Value::from_bool(res))])
             }
+            Self::Eq => {
+                let inps = read_inputs(consts)?;
+                let res = inps.clone().into_iter().reduce(|a, b| a == b)?;
```
acl-cqc, line 52:
```suggestion
                let res = inps.iter().copied().reduce(|a, b| a == b)?;
```
Annoying you can't use `bool::eq` but indeed you can't; however `bool` impls `Copy`.

acl-cqc (commented):
Thanks Craig! That is a neat observation about partial folding. Test please :)

hugr-core/src/std_extensions/logic.rs
acl-cqc: I guess this is less consistent with the others

hugr-core/src/std_extensions/logic.rs
acl-cqc: (a) good to use Copy rather than Clone when available
(b) `clone().into_iter()` probably allocates a clone of the entire Vec, `iter().copied()` doesn't

hugr-core/src/std_extensions/logic.rs
```diff
@@ -355,6 +366,9 @@ pub(crate) mod test {
     #[case(NaryLogic::And, [Some(false), None], Some(false))]
     #[case(NaryLogic::Or, [None, Some(false)], None)]
     #[case(NaryLogic::Or, [None, Some(true)], Some(true))]
+    #[case(NaryLogic::Eq, [None, Some(true), Some(true)], None)]
+    #[case(NaryLogic::Eq, [None, Some(false), Some(true)], Some(false))]
+    #[case(NaryLogic::Eq, [Some(true), Some(true), Some(true)], Some(true))]
```
acl-cqc, line 371: nit: the last doesn't need to be expressed in the more-complex `nary_partial_const_fold` format, right? Doing so is a test of the test framework, rather than the code-under-test! You could move it to `nary_const_fold` or remove it.

acl-cqc (approved):
Thanks @croyzor, looks good to me :)

PR #1399, opened 2024-08-08
feat(py): implement `iter` on `ToNode`

aborgna-q:


hugr-py/src/hugr/node_port.py
```diff
@@ -90,6 +90,13 @@ def __getitem__(
     def out_port(self) -> OutPort:
         return OutPort(self.to_node(), 0)
 
+    def outputs(self) -> Iterator[OutPort]:
+        """Returns an iterator over the output ports of this node."""
+        return self.to_node()[:]
```
croyzor, line 95: I'm not really sure how this works - it seems like `Node` doesn't have a `getitem`, so it'd defer to the one defined in `ToNode`? I'm guessing that `__getitem__` and `to_node` are also handled by some kind of magic?

aborgna-q: Ah, I see I could have written `return self[:]` directly, using the local `__getitem__`.
`to_node` just converts the object into a `Node` keeping the relevant parameters, so it behaves exactly the same when used as a `ToNode`.

croyzor (approved):

PR #1401, opened 2024-08-08
fix(py): `Hugr.__iter__` returning `NodeData | None` instead of `Node`s

aborgna-q:
`Hugr` is a `Mapping[Node, NodeData]`, so its `__iter__` should return a list of `Node`s, but it was returning a list of node data with `None` holes

### DIFF
--- a/hugr-py/src/hugr/hugr.py
+++ b/hugr-py/src/hugr/hugr.py
@@ -121,9 +121,7 @@ class Hugr(Mapping[Node, NodeData], Generic[OpVar]):
         return n
 
     def __iter__(self) -> Iterator[Node]:
-        for idx, data in enumerate(self._nodes):
-            if data is not None:
-                yield Node(idx, data._num_outs)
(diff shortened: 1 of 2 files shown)


hugr-py/src/hugr/hugr.py
```diff
@@ -131,6 +133,12 @@ def _get_typed_op(self, node: ToNode, cl: type[OpVar2]) -> OpVar2:
         assert isinstance(op, cl)
         return op
 
+    def nodes(self) -> Iterator[tuple[Node, NodeData]]:
```
ss2165, line 136: I think this is now indentical to `.items()` either remove or call `.items()` ?

aborgna-q: I think it's expected to have a `nodes` method.
I tried to use it from guppy, and mypy was suggesting `_nodes`...

ss2165 (approved):
👍

PR #1406, opened 2024-08-08
feat(py): Parametric int type helper, and arbitrary width int constants

aborgna-q:
- Adds an `int_tv` method to create parametric integer types
- Adds a `width` member to `IntVal`, so we can define constants for other integer widths

### DIFF
--- a/hugr-py/src/hugr/std/int.py
+++ b/hugr-py/src/hugr/std/int.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, ClassVar
 
 from typing_extensions import Self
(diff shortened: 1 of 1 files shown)


hugr-py/src/hugr/std/int.py
```diff
@@ -45,9 +73,10 @@ class IntVal(val.ExtensionValue):
     """Custom value for an integer."""
 
     v: int
+    width: int = field(default=5)
 
     def to_value(self) -> val.Extension:
-        return val.Extension("int", INT_T, self.v)
+        return val.Extension("int", int_t(self.width), self.v)
```
croyzor, line 79: The parameter to this is meant to bre an index, not the width?

aborgna-q: That's `int_tv` (for integers with parametric size), this is calling `int_t` (a bit higher in the file).

Sorry for the messy diff, it's just a collection of changes I needed for guppy.
Though I didn't end up needing `int_tv`. I can drop it if it seems too confusing.

croyzor: Ah I see. Definitely drop `int_tv` if it's not used!

croyzor (approved):

PR #1410, opened 2024-08-08
feat(hugr-py): add type_bound method to `Type`

### LINKED ISSUE #1365: test(hugr-py): add test for TypeBound.join()
Not covered by tests?

_Originally posted by @aborgna-q in https://github.com/CQCL/hugr/pull/1364#discussion_r1691553245_

ss2165:
Closes #1365

### DIFF
--- a/hugr-py/src/hugr/tys.py
+++ b/hugr-py/src/hugr/tys.py
@@ -39,6 +39,17 @@ class TypeArg(Protocol):
 class Type(Protocol):
     """A HUGR type."""
 
+    def type_bound(self) -> stys.TypeBound:
+        """The bound of this type.
+
+        Example:
+            >>> Tuple(Bool, Bool).type_bound()
+            <TypeBound.Copyable: 'C'>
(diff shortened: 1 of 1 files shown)


hugr-py/src/hugr/tys.py
```diff
@@ -309,6 +338,9 @@ class FunctionType(Type):
     output: TypeRow
     extension_reqs: ExtensionSet = field(default_factory=ExtensionSet)
 
+    def type_bound(self) -> TypeBound:
+        return TypeBound.Copyable
```
aborgna-q, line 342: Is this always the case? Could we have some partially-applied function with linear state?

ss2165: I just copied the spec/crate.
The justification I believe is that if the linear state can be encoded in to the function then it must be copyable, so there's no need to mark it as a linear type. It is a bit restrictive, you may wish to tell the type system to treat something as linear even if you can encode it. This is one for a discussion or issue I think.

aborgna-q (approved):

PR #1414, opened 2024-08-12
feat: Export the core data structures to a model representation

zrho:
This PR introduces a draft for the `hugr-model` data structures together with a function to export a `Hugr` into this format.

### DIFF
--- a/hugr-core/src/export.rs
+++ b/hugr-core/src/export.rs
@@ -8,10 +8,10 @@ use crate::{
         CustomType, FuncTypeBase, MaybeRV, PolyFuncTypeBase, RowVariable, SumType, TypeArg,
         TypeBase, TypeEnum,
     },
-    Direction, Hugr, HugrView, Node, Port,
+    Hugr, HugrView, Node, Port,
 };
 use hugr_model::v0 as model;
-use indexmap::IndexMap;
+use indexmap::IndexSet;
 use smol_str::ToSmolStr;
 use tinyvec::TinyVec;
 
@@ -30,7 +30,7 @@ struct Context<'a> {
     module: model::Module,
     /// Mapping from ports to edge indices.
     /// This only includes the minimum port among groups of linked ports.
-    edges: IndexMap<(Node, Port), usize>,
+    edges: IndexSet<(Node, Port)>,
 }
 
 impl<'a> Context<'a> {
@@ -38,7 +38,7 @@ impl<'a> Context<'a> {
         Self {
             hugr,
             module: model::Module::default(),
-            edges: IndexMap::new(),
+            edges: IndexSet::new(),
         }
     }
 
@@ -54,33 +54,31 @@ impl<'a> Context<'a> {
         model::NodeId(index as _)
     }
 
-    pub fn make_port(&mut self, node: Node, port: impl Into<Port>) -> model::PortId {
-        let port = port.into();
-        let index = self.module.ports.len();
-        let port_id = model::PortId(index as _);
-        let r#type = self.make_term(model::Term::Wildcard); // TODO
-        self.module.ports.push(model::Port {
-            r#type,
-            meta: Vec::new(),
-        });
-
+    /// Returns the edge id for a given port, creating a new edge if necessary.
+    ///
+    /// Any two ports that are linked will be represented by the same edge.
+    fn get_edge_id(&mut self, node: Node, port: Port) -> model::EdgeId {
         // To ensure that linked ports are represented by the same edge, we take the minimum port
         // among all the linked ports, including the one we started with.
         let linked_ports = self.hugr.linked_ports(node, port);
         let all_ports = std::iter::once((node, port)).chain(linked_ports);
         let repr = all_ports.min().unwrap();
+        let edge = self.edges.insert_full(repr).1 as _;
+        model::EdgeId(edge)
+    }
 
-        let edge_id = *self.edges.entry(repr).or_insert_with(|| {
-            let edge_id = self.module.edges.len();
-            let edge = model::Edge::default();
-            self.module.edges.push(edge);
-            edge_id
+    pub fn make_port(&mut self, node: Node, port: impl Into<Port>) -> model::PortId {
+        let port = port.into();
+        let index = self.module.ports.len();
+        let port_id = model::PortId(index as _);
+        let r#type = self.make_term(model::Term::Wildcard); // TODO
+        let edge = self.get_edge_id(node, port);
+
+        self.module.ports.push(model::Port {
+            r#type,
+            meta: Vec::new(),
+            edge,
         });
-
-        match port.direction() {
-            Direction::Incoming => self.module.edges[edge_id].inputs.push(port_id),
-            Direction::Outgoing => self.module.edges[edge_id].outputs.push(port_id),
-        }
 
         port_id
     }
--- a/hugr-core/src/export.rs
+++ b/hugr-core/src/export.rs
@@ -49,7 +49,7 @@ impl<'a> Context<'a> {
     }
 
     pub fn make_node(&mut self, node: model::Node) -> model::NodeId {
-        let index = self.module.terms.len();
+        let index = self.module.nodes.len();
         self.module.nodes.push(node);
         model::NodeId(index as _)
     }
--- a/hugr-core/src/export.rs
+++ b/hugr-core/src/export.rs
@@ -1,0 +1,417 @@
+//! Text syntax.
+use crate::{
+    extension::ExtensionSet,
+    ops::{NamedOp as _, OpType},
+    types::{
+        type_param::{TypeArgVariable, TypeParam},
+        type_row::TypeRowBase,
+        CustomType, FuncTypeBase, MaybeRV, PolyFuncTypeBase, RowVariable, SumType, TypeArg,
+        TypeBase, TypeEnum,
+    },
+    Direction, Hugr, HugrView, Node, Port,
+};
+use hugr_model::v0 as model;
+use indexmap::IndexMap;
+use smol_str::ToSmolStr;
+use tinyvec::TinyVec;
+
+/// Export a [`Hugr`] graph to its representation in the model.
+pub fn export(hugr: &Hugr) -> model::Module {
+    let mut context = Context::new(hugr);
+    hugr.root().export(&mut context);
+    context.module
+}
+
+/// State for converting a HUGR graph to its representation in the model.
+struct Context<'a> {
+    /// The HUGR graph to convert.
+    hugr: &'a Hugr,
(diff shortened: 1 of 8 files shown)


hugr-core/src/types.rs
acl-cqc, line 1: 👍

hugr-model/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+//! The data model of the HUGR intermediate representation.
+//! This crate defines data structures that capture the structure of a HUGR graph and
+//! all its associated information in a form that can be stored on disk. The data structures
```
acl-cqc, line 3: nit: does anyone still use "disk"s nowadays? :)

perhaps something like "in a form designed for simple and efficient serialization, rather than traversal or modification"??? Or we could say this is for serialization, storage *and interchange*?

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+    /// Table of [`Term`]s.
+    pub terms: Vec<Term>,
+}
+
+/// Nodes in the hugr graph.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Node {
+    /// The operation that the node performs.
+    pub operation: Operation,
+    /// Parameters that are passed to the operation.
```
acl-cqc, line 77: This is an arbitrary-length (presumed small) list of triples of TermId? Please explain here why it's a triple i.e. what each of the three elements are.

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+pub enum Term {
+    /// Standin for any term.
+    Wildcard,
+    /// The type of types.
+    Type,
+    /// The type of constraints.
+    Constraint,
+    /// A variable.
+    Var(TermVar),
+    /// A symbolic function application.
```
acl-cqc, line 252: Whoa....what's this? We are not doing dependent types, right?

acl-cqc: If the LHS is always a type scheme, and the right are TypeArg's (i.e. this is the equivalent of `ListType` where the LHS comes from an extension), then I think we should make that clear; I imagine most users will be familiar with generics but may not be coming from a type-theoretic background where they see "function application" and `MyList<T>` are the same thing...

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+    Named(term::Named),
+    /// A list, with an optional tail.
+    List(term::List),
+    /// The type of lists, given a type for the items.
+    ListType(term::ListType),
```
acl-cqc, line 257:
```suggestion
    ListType(TermId),
```
or
```suggestion
    ListType {item_type: TermId}
```

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+    use smol_str::SmolStr;
+    use tinyvec::TinyVec;
+
+    /// Named terms.
+    #[derive(Debug, Clone, PartialEq, Eq, Hash)]
+    pub struct Named {
+        /// The name of the term.
+        pub name: Symbol,
+        /// Arguments to the term.
+        pub args: TinyVec<[TermId; 3]>,
```
acl-cqc, line 294: Similar comment about triples

hugr-core/src/export.rs
```diff
@@ -0,0 +1,416 @@
...
+    pub fn new(hugr: &'a Hugr) -> Self {
+        Self {
+            hugr,
+            module: model::Module::default(),
+            edges: IndexSet::new(),
+        }
+    }
+
+    pub fn make_term(&mut self, term: model::Term) -> model::TermId {
+        let index = self.module.terms.len();
```
acl-cqc, line 46: Is this the place to do hash-consing?? (I.e. look for an existing `Term` that matches the suppiled, and if so then return that TermId rather than pushing)

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,368 @@
+//! Version 0 (unstable).
+//!
+//! # Terms
+//!
+//! Terms form a meta language that is used to describe types, parameters and metadata that
```
acl-cqc, line 5: This is a design point that has probably been addressed elsewhere, but do we really want structured metadata like this?

Ok, I can see many use cases for metadata which does contain....Hugr-defined things. But not necessarily Terms - maybe Hugrs or NodeId's or other things. And, how is *unstructured* metadata represented? (I see that a MetaItem contains exactly one Term, I haven't seen the alternative - am I looking in the wrong place?)

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+/// A port in the graph.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Port {
+    /// The id of the edge that the port is connected to.
+    /// All ports referencing the same edge are connected.
+    pub edge: EdgeId,
+
+    /// The type of the port.
+    ///
+    /// This must be a term of type `Type`.
```
acl-cqc, line 206: Also, will be the same for all `Port`s with the same `edge`

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Scheme {
+    /// The named parameters of the scheme.
+    ///
+    /// Within any term, the previous terms of the parameter list are available as variables.
+    pub params: Vec<SchemeParam>,
+    /// Constraints on the parameters of the scheme.
+    ///
+    /// All parameters are available as variables within the constraints.
+    /// The constraints must be terms of type `Constraint`.
```
acl-cqc, line 224: Right - I'm puzzled here. I don't see any `enum Term` variants that have type `Constraint`; I see only one for the type *of* Constraints, and I can't understand how that would be used. Am I missing something, looking at this the wrong way, or is this just WIP?

acl-cqc (commented):
Thanks, Lukas - a few thoughts.

Presumably if we go with a system of "constraints", in order to express the expressible parts of binary `compute_signature`, we'll also need to use the hugr-model for serializaiton of extension sets...

hugr-core/src/export.rs
```diff
@@ -0,0 +1,416 @@
...
+    /// Returns the edge id for a given port, creating a new edge if necessary.
+    ///
+    /// Any two ports that are linked will be represented by the same edge.
+    fn get_edge_id(&mut self, node: Node, port: Port) -> model::EdgeId {
+        // To ensure that linked ports are represented by the same edge, we take the minimum port
+        // among all the linked ports, including the one we started with.
+        let linked_ports = self.hugr.linked_ports(node, port);
+        let all_ports = std::iter::once((node, port)).chain(linked_ports);
+        let repr = all_ports.min().unwrap();
+        let edge = self.edges.insert_full(repr).1 as _;
```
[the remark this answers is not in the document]
zrho: `insert_full` returns a `usize`. Edge indices are `u32`. Currently this assumes that we haven't hit the (exceedingly unlikely) case that a `Hugr` has more than 2^32 nodes or ports.

hugr-core/src/export.rs
ss2165: It will be useful to associate ids with uses, i.e be able to trace problems back to the source by figuring out which instance of a term we're looking at. I'm looking in to hashconsing at serialisation time, and preserving all the ids that point to the same term. We can apply this to the model separately without changing the model behaviour I think.

hugr-model/src/v0/mod.rs
zrho: It's a vector that stores its elements inline (without a heap allocation) if it is small enough. This has no semantic significance but is a simple optimisation. `TinyVec<[u32; 3]>` is the same size as `Vec<u32>`. `TinyVec<[u32; 4]>` is bigger than `Vec<u32>`.

hugr-model/src/v0/mod.rs
zrho: This is all a big WIP.

hugr-model/src/v0/mod.rs
zrho: If you want unstructured metadata, you can still put it into a string. Structured metadata allows validation via typechecking.

hugr-model/src/lib.rs
zrho: I think it's used metaphorically, like the floppy disk symbol meaning "save" although it's most likely not being saved on a floppy disk.

hugr-model/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "hugr-model"
+version = "0.1.0"
+rust-version.workspace = true
+edition.workspace = true
+homepage.workspace = true
+repository.workspace = true
+license.workspace = true
```
aborgna-q, line 8:
```suggestion
license.workspace = true
readme = "README.md"
documentation = "https://docs.rs/hugr-model/"
description = "Serializable model for Quantinuum's HUGR"
keywords = ["Quantum", "Quantinuum"]
categories = ["compilers"]
```

aborgna-q (commented):
Some housekeeping needed when adding a new package:

- Add a README. Just copy the template from the other crates.
- Add this to `release-plz.toml`. We'll still need to do some manual deployment before the next `hugr-core` release, as `release-plz` does not have permission to create new packages.
  ```toml
  [[package]]
  name = "hugr-cli"
  release = true
  ```
  Also add it to the `changelog_include` for `hugr`
- Add the new directory it to `.github/change-filters.yml`

hugr-core/src/export.rs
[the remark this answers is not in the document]
acl-cqc: oic ok yes thanks

hugr-model/src/v0/mod.rs
acl-cqc: Ah OK yes `String` is good, right

hugr-model/src/v0/mod.rs
acl-cqc: oh, the `;3` is like the maximum size to store inline? Ok, right

hugr-model/src/v0/mod.rs
acl-cqc: A "function application" applies a function on the LHS to some argument(s) on the RHS, yes?

hugr-core/src/export.rs
ss2165: have experimented with doing hashconsing here: e4ef7be6b2c68f30c5cb8826ba623ab07b8eab5c
I think it's worthwhile.

PR #1428, opened 2024-08-14
feat(py): Add node metadata

### LINKED ISSUE #1319: Add metadata support to `hugr-py`

aborgna-q:
Adds a dictionary with metadata to the nodes.
`ToNode` now has a `metadata` property, so we can use
```python
d = Dfg(...)
d.metadata["key"] = 42

n = d.hugr.add_node(..., metadata={"something": "value"})
assert n.metadata["something"] == "value"
```

I chosed not to add a `metadata` argument to all the `add_{container}` methods in DfgBase to avoid cluttering the API.

Closes #1319

### DIFF
--- a/hugr-py/src/hugr/function.py
+++ b/hugr-py/src/hugr/function.py
@@ -57,3 +57,8 @@ class Module(_DefinitionBuilder[ops.Module]):
     def add_alias_decl(self, name: str, bound: TypeBound) -> Node:
         """Add a type alias declaration."""
         return self.hugr.add_node(ops.AliasDecl(name, bound), self.hugr.root)
+
+    @property
+    def metadata(self) -> dict[str, object]:
+        """Metadata associated with this module."""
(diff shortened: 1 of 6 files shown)


hugr-py/src/hugr/function.py
```diff
@@ -57,3 +57,8 @@ def declare_function(self, name: str, signature: PolyFuncType) -> Node:
     def add_alias_decl(self, name: str, bound: TypeBound) -> Node:
         """Add a type alias declaration."""
         return self.hugr.add_node(ops.AliasDecl(name, bound), self.hugr.root)
+
+    @property
+    def metadata(self) -> dict[str, object]:
```
aborgna-q, line 62: `Module` is the only builder that doesn't implement `ToNode`, so it needs it's own metadata getter

zrho (approved):

PR #1429, opened 2024-08-14
refactor!: flatten `CustomOp` in to `OpType`

### LINKED ISSUE #1362: Remove `OpaqueOp` (used only for serialisation)
After completeing preceding tasks in #1344 there should be no remaining uses for OpaqueOp. Demote it to only be used for op serialisation.

ss2165:
In new workflow most of the time we just want to work with extension ops, and opaque are just used for serialisation. To that end flatten `CustomOp` out.  If there are `OpaqueOp`s in the HUGR it means the extension was not available to resolve in to. Will be disallowed in follow up PR.

Relates to #1362 

BREAKING CHANGE: `CustomOp` removed, `OpType` now contains `ExtensionOp` and `OpaqueOp` directly. `CustomOpError` renamed to`OpaqueOpError`.

### DIFF
--- a/hugr-core/src/hugr/validate.rs
+++ b/hugr-core/src/hugr/validate.rs
@@ -578,7 +578,7 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
         match op_type {
             OpType::ExtensionOp(ext_op) => validate_ext(ext_op)?,
             OpType::OpaqueOp(opaque) => {
-                // ry to resolve serialized names to actual OpDefs in Extensions.
+                // Try to resolve serialized names to actual OpDefs in Extensions.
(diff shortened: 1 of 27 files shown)


hugr-core/src/hugr/validate.rs
```diff
@@ -567,36 +567,26 @@ impl<'a, 'b> ValidationContext<'a, 'b> {
         let op_type = self.hugr.get_optype(node);
         // The op_type must be defined only in terms of type variables defined outside the node
         // TODO consider turning this match into a trait method?
```
aborgna-q, line 569: Can we drop the TODO?

aborgna-q (approved):
Nice!

PR #1431, opened 2024-08-14
feat!: disallow opaque ops during validation

### LINKED ISSUE #1362: Remove `OpaqueOp` (used only for serialisation)

ss2165:
Update python to pass extensions as package when validating

Use cached signatures from opaque when resolving if missing binary compute.
Closes #1362

BREAKING CHANGE: HUGRs containing opaque operations that don't point to an extension in the registry will fail to validate. Use `Package` to pack extensions with HUGRs for serialisation.

### DIFF
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -596,7 +596,7 @@ pub mod test {
             Self::new(name, Version::new(0, 0, 0))
         }
 
-        /// Add a simple OpDef to the extension and return an extension op for it..
+        /// Add a simple OpDef to the extension and return an extension op for it.
         /// No description, no type parameters.
         pub(crate) fn simple_ext_op(
             &mut self,
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -595,6 +595,18 @@ pub mod test {
         pub(crate) fn new_test(name: ExtensionId) -> Self {
             Self::new(name, Version::new(0, 0, 0))
         }
+
+        /// Add a simple OpDef to the extension and return an extension op for it..
+        /// No description, no type parameters.
+        pub(crate) fn simple_ext_op(
+            &mut self,
+            name: &str,
+            signature: impl Into<SignatureFunc>,
+        ) -> ExtensionOp {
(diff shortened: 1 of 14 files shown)


hugr-core/src/extension.rs
```diff
@@ -595,6 +595,18 @@ pub mod test {
         pub(crate) fn new_test(name: ExtensionId) -> Self {
             Self::new(name, Version::new(0, 0, 0))
         }
+
+        /// Add a simple OpDef to the extension and return an extension op for it..
```
ss2165, line 599:
```suggestion
        /// Add a simple OpDef to the extension and return an extension op for it.
```

hugr-core/src/hugr/rewrite/replace.rs
```diff
@@ -639,12 +639,14 @@ mod test {
 
     #[test]
     fn test_invalid() {
-        let unknown_ext: ExtensionId = "unknown_ext".try_into().unwrap();
+        let mut new_ext = crate::Extension::new_test("new_ext".try_into().unwrap());
+        let ext_name = new_ext.name().clone();
+        // let unknown_ext: ExtensionId = "unknown_ext".try_into().unwrap();
```
ss2165, line 644:
```suggestion
```

hugr-py/tests/test_custom.py
```diff
@@ -14,7 +14,7 @@
 from hugr.std.logic import Not
 
 from .conftest import CX, H, Measure, Rz, validate
-from .conftest import EXTENSION as QUANTUM_EXT
+from .conftest import QUANTUM_EXT as QUANTUM_EXT
```
cqc-alec, line 17:
```suggestion
from .conftest import QUANTUM_EXT
```

cqc-alec (approved):
Should we remove `OpaqueOp` from the spec?

ss2165:
> Should we remove `OpaqueOp` from the spec?

yes, done

cqc-alec (approved):

PR #1439, opened 2024-08-16
test: add serialization benchmarks

### LINKED ISSUE #1434: Create benchmarks for current serialisation.
Use existing criterion framework to time round trip via json (including file write) of:
1. Simple HUGRs
2. Scaling HUGRs by adding gates to a circuit.

ss2165:
Criterion doesn't seem to want to tell me how much memory is used, but its not hard to just print that out (has no variance).

Headline: serialization roundtrip of a 4000 gate circuit takes 33ms, 1.5Mb size.

image

 
Closes #1434


aborgna-q (approved):
👍 
It could have gone in a separate "serialization" module, but for now LGTM.

PR #1455, opened 2024-08-21
feat!: int operations other than widen/narrow have only one width arg

ss2165:
BREAKING CHANGE: `idivmod_checked`, `idivmod`, `idiv`, `idiv_checked`, `imod`, `ishl`, `ishr`, `irotl`, `irotr` operations now only have one width argument for all inputs and outputs rather than two.


cqc-alec (approved):
Looks good. A few minor changes in "description" fields. Should we also update `hugr.md` here?

PR #1457, opened 2024-08-21
fix(py): Invalid node indexing

### LINKED ISSUE #1454: Node slice indexing with stop index 0 is broken
```python
>>> from hugr import Node
>>> n = Node(0, _num_out_ports=1)
>>> list(n[:0])
```
Expected output is `[]`, but it actually gives `[OutPort(Node(0), 0)]`.

This is because stop index 0 is evaluated as falsy in
https://github.com/CQCL/hugr/blob/95b39493e85744718e059280de2fd24d7988ee32/hugr-py/src/hugr/node_port.py#L173
We should replace this with an explicit check for `None`.

aborgna-q:
Fixes #1454, and other failing cases

### DIFF
--- a/hugr-py/src/hugr/node_port.py
+++ b/hugr-py/src/hugr/node_port.py
@@ -164,23 +164,45 @@ class Node(ToNode):
     ) -> OutPort | Iterator[OutPort]:
         match index:
             case PortOffset(index):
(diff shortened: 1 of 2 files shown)


ss2165 (approved):
Nice thanks

PR #1459, opened 2024-08-21
feat(hugr-py): use serialized extensions in python

### LINKED ISSUE #1450: Include and load serialized std extensions in hugr-py
Currently we store the rust-defined std extensions as json files in `specification/std_extensions/`. There can be loaded by the extension model defined in hugr-py's `.serialization.extension`, but are not included in the wheels.

Instead, we define some operations separately on `.std`.
We should include the json files in the wheels (compressed with zstd?) and load them when exposing the std extensions from python. We'll still need definitions for operations that require binary-computed signatures (array ops), but we should be able to keep the rest synchronized between python and rust.

ss2165:
Copies standard extensions to hugr-py source tree, as poetry won't include anything from outside the tree (not symlinks either).

Existing op/type wrappers now just point to loaded extensions.

Closes #1450

### DIFF
--- a/hugr-py/src/hugr/std/__init__.py
+++ b/hugr-py/src/hugr/std/__init__.py
@@ -1,15 +1,13 @@
 """Types and operations from the standard extension set."""
 
-from pathlib import Path
+import pkgutil
 
 from hugr.ext import Extension
 from hugr.serialization.extension import Extension as PdExtension
 
-_EXT_DIR = Path(__file__).parent / "_json_defs"
-
 
 def _load_extension(name: str) -> Extension:
     replacement = name.replace(".", "/")
-    path = _EXT_DIR / f"{replacement}.json"
-    with path.open() as f:
-        return PdExtension.model_validate_json(f.read()).deserialize()
+    json_str = pkgutil.get_data(__name__, f"_json_defs/{replacement}.json")
+    assert json_str is not None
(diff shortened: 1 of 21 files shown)


hugr-py/src/hugr/std/__init__.py
```diff
@@ -1 +1,15 @@
 """Types and operations from the standard extension set."""
+
+from pathlib import Path
+
+from hugr.ext import Extension
+from hugr.serialization.extension import Extension as PdExtension
+
+_EXT_DIR = Path(__file__).parent / "_json_defs"
```
aborgna-q, line 8: Use `pkgutil`
https://stackoverflow.com/questions/6028000/how-to-read-a-static-file-from-inside-a-python-package/58941536#58941536

hugr-py/src/hugr/std/_json_defs/README.md
aborgna-q, line 1: I wasn't sure if the jsons would be kept when building the wheels, but they are still there after running `poetry build` 👍

aborgna-q (approved):
Nice!

The `RegisteredOp` now look a bit boilerplate-y. Hopefully we can simplify those in the future.

hugr-py/src/hugr/std/_json_defs/README.md
ss2165: yes sorry should have said I checked that

PR #1460, opened 2024-08-22
refactor!: Organise the hugr-py modules

aborgna-q:
Quick breaking change:
- Move the `hugr` definition and `node_port` inside a `hugr` module
- Move all the builders to a `build` module
- Add reexports where applicable

BREAKING CHANGE: Moved all builder definitions into the `hugr.build` module. Moved `node_port` and `render` into the `hugr.hugr` module.


ss2165 (approved):
My main concern with rexporting is that any initalisation code in every re-exported submodule is run whenever the top level init is run. This would be quite bad for example for the standard extensions which load from JSON currently. I think it is acceptable for the things being re-exported here, which are just builder things mostly.

hugr-py/src/hugr/__init__.py
```diff
@@ -2,8 +2,8 @@
 representation.
 """
 
-from .hugr import Hugr
-from .node_port import Direction, InPort, Node, OutPort, Wire
+from .hugr.base import Hugr
+from .hugr.node_port import Direction, InPort, Node, OutPort, Wire
 from .ops import Op
 from .tys import Kind, Type
```
ss2165, line 8: It might be worth removing the ops and tys rexports for consistency

PR #1462, opened 2024-08-22
feat: Bring in the pure-python renderer from guppy

### LINKED ISSUE #1407: Pure-python hugr renderer
With the refactor of guppy to use the hugr builder we lost the ability to render hugrs directly in the pure-python library.
https://github.com/CQCL/guppylang/blob/7d5106cd59ad452046d0dfffd10eea9d9b617431/guppylang/hugr_builder/visualise.py

We should port that graphviz renderer to `hugr-py`, or write a mermaid renderer similar to the one in rust.

aborgna-q:
Closes #1407.

Ports @mark-koch's rendering code from guppy, since it got deleted from there in the last hugr builder update.
Adds a `Hugr.render_dot` and a `Hugr.store_dot` method.

I'm using `syrupy`—a snapshot pytest extension—for testing it. I'm not sure if we want to be this strict with the generated output though.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,16 +14,18 @@ repos:
         exclude: |
           (?x)^(
             specification/schema/.*|
-            .*.snap|
-            .*.snap.new|
+            .*\.snap|
+            .*\.snap\.new|
+            .*\.ambr|
             .release-please-manifest.json
           )$
       - id: trailing-whitespace
         exclude: |
           (?x)^(
             specification/schema/.*|
-            .*.snap|
-            .*.snap.new
+            .*\.snap|
+            .*\.ambr|
+            .*\.snap\.new
           )$
       - id: fix-byte-order-marker
       - id: mixed-line-ending
--- a/hugr-py/pyproject.toml
+++ b/hugr-py/pyproject.toml
@@ -28,6 +28,7 @@ python = ">=3.10"
 pydantic = ">=2.7,<2.9"
 pydantic-extra-types = "^2.9.0"
 semver = "^3.0.2"
+graphviz = { version = "^0.20.3" }
 
 [tool.poetry.group.docs]
 optional = true
--- a/hugr-py/src/hugr/hugr.py
+++ b/hugr-py/src/hugr/hugr.py
@@ -145,7 +145,7 @@ class Hugr(Mapping[Node, NodeData], Generic[OpVarCov]):
         """Iterator over nodes of the hugr and their data."""
         return self.items()
 
-    def links(self) -> Iterable[tuple[OutPort, InPort]]:
+    def links(self) -> Iterator[tuple[OutPort, InPort]]:
         """Iterator over all the links in the HUGR.
 
         Returns:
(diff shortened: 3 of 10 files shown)


hugr-py/src/hugr/hugr/base.py
```diff
@@ -125,6 +127,14 @@ def nodes(self) -> Iterable[tuple[Node, NodeData]]:
         """Iterator over nodes of the hugr and their data."""
         return self.items()
 
+    def links(self) -> Iterable[tuple[OutPort, InPort]]:
```
mark-koch, line 130:
```suggestion
    def links(self) -> Iterator[tuple[OutPort, InPort]]:
```

hugr-py/src/hugr/hugr/render.py
```diff
@@ -0,0 +1,280 @@
...
+    def render(self, hugr: Hugr) -> Digraph:
+        """Render a HUGR to a graphviz dot object."""
+        graph_attr = {
+            "rankdir": "",
+            "ranksep": "0.1",
+            "nodesep": "0.15",
+            "margin": "0",
+            "bgcolor": self.palette.background,
+        }
+        if not (name := hugr[hugr.root].metadata.get("name")):
```
mark-koch, line 89:
```suggestion
        if not (name := hugr[hugr.root].metadata.get("name", None)):
```

hugr-py/src/hugr/hugr/render.py
```diff
@@ -0,0 +1,280 @@
...
+        return f"{n.idx}:{self._INPUT_PREFIX}None"
+
+    def _out_order_name(self, n: Node) -> str:
+        return f"{n.idx}:{self._OUTPUT_PREFIX}None"
+
+    def _viz_node(self, node: Node, hugr: Hugr, graph: Digraph) -> None:
+        """Render a (possibly nested) node to a graphviz graph."""
+        # TODO: Port the CFG special-case rendering from guppy, and use it here
+        # when a node is a CFG node.
+        # See https://github.com/CQCL/guppylang/blob/7d5106cd59ad452046d0dfffd10eea9d9b617431/guppylang/hugr_builder/visualise.py#L250
```
mark-koch, line 202: This is purely for debugging Guppy's dataflow analysis, so probably doesn't belong here

hugr-py/src/hugr/hugr/render.py
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        label = ""
+        match kind:
+            case ValueKind(ty):
+                label = str(ty)
+                color = self.palette.edge
+            case OrderKind():
+                color = self.palette.dark
+            case ConstKind():
```
mark-koch, line 269:
```suggestion
            case ConstKind() | FunctionKind():
```

hugr-py/src/hugr/hugr/render.py
```diff
@@ -0,0 +1,280 @@
...
+                color = self.palette.const
+            case _:
```
mark-koch, line 271: Maybe explicitly match against `CFKind` and add a fallthrough case with `assert_never`. Then mypy will enforce exhaustiveness

mark-koch (approved):
Nice! 🎉

I think the snapshot testing is fine as is for now since graphviz is deterministic (??). Can always change later if it leads to problems

PR #1463, opened 2024-08-22
feat: Add missing ops

aborgna-q:
Adds some missing operations:

- `fpow`
- `fround`
- `ipow`
- `iu_to_s` / `is_to_u`. These are almost noops, but some runtimes may prefer to check and panic if the value is out-of-bounds.
- Mention that `ifrombool` / `itobool` only work with `i1` in their description.

### DIFF
--- a/hugr-core/src/std_extensions/arithmetic/float_ops.rs
+++ b/hugr-core/src/std_extensions/arithmetic/float_ops.rs
@@ -137,6 +137,9 @@ impl MakeRegisteredOp for FloatOps {
 
 #[cfg(test)]
 mod test {
+    use cgmath::AbsDiffEq;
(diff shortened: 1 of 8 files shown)


hugr-core/src/std_extensions/arithmetic/float_ops.rs
```diff
@@ -144,4 +151,47 @@ mod test {
             assert!(name.as_str().starts_with('f'));
         }
     }
+
+    #[rstest]
+    #[case::fadd(FloatOps::fadd, &[0.1, 0.2], &[0.30000000000000004])]
```
croyzor, line 156: 😁

croyzor (approved):

PR #1465, opened 2024-08-23
chore: Replace `poetry` with `uv`

aborgna-q:
Replaces the python project manager with `uv`.
For most commands that used poetry before, it just be as easy as replacing `poetry {things}` with `uv {things}`.

Some notable exeptions:

- Instead of `poetry install --with extra1,extra2`,
  use `uv sync --extra extra1 --extra extra2`.
  Note that will delete any extraneous packages/extras so the venv state is deterministic.
  Add `--inexact` if you want to replicate the old poetry behaviour.

- `poetry shell` is no longer available. The preferred workflow is to `uv run` / `uvx` things, to ensure you are always in sync with the project definition.
  If that's too annoying, it is always possible to `source .venv/bin/activate`.

Updates the justfile, so using that should be the same as before.


.github/workflows/ci-py.yml
```diff
@@ -95,34 +105,63 @@ jobs:
...
-    name: test python ${{ matrix.python-version }}
+    name: test python ${{ matrix.python-version.py }}
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
-        python-version: ['3.10', '3.12']
+        python-version:
+          - { py: '3.10', coverage: false }
+          - { py: '3.12', coverage: true }
```
mark-koch, line 115: Good idea to combine the coverage run with regular testing 👍

.pre-commit-config.yaml
```diff
@@ -36,31 +36,24 @@ repos:
 
   - repo: local
     hooks:
-      - id: poetry-install
-        name: sync poetry deps
-        description: Ensure development tools are installed.
-        entry: poetry install --sync
-        language: system
-        files: poetry.lock
-        pass_filenames: false
```
mark-koch, line 45: Do we want to `uv sync` here?

justfile
```diff
@@ -27,37 +27,32 @@ bench language="[rust|python]": (_run_lang language \
...
     )
 
 # Run unsoundness checks using miri
 miri:
     PROPTEST_DISABLE_FAILURE_PERSISTENCE=true MIRIFLAGS='-Zmiri-env-forward=PROPTEST_DISABLE_FAILURE_PERSISTENCE' cargo +nightly miri test
 
-# Load a shell with all the dependencies installed
-shell:
-    poetry shell
-
```
mark-koch, line 52: We could do `source .venv/bin/activate` here, but not sure if it's a good idea...

mark-koch (approved):
image

justfile
aborgna-q: Doesn't really work with how justfile runs its commands :/

.pre-commit-config.yaml
aborgna-q: `sync` is done automatically each time you call `uv run`.
In CI we add the extra step to have more feedback and pass `-locked` as needed, but here it should be instantaneous.

PR #1476, opened 2024-08-28
feat: Dataflow analysis framework

acl-cqc:
Forwards analysis only, but parametrized over the abstract domain, hence intended to support various applications including constant folding and potentially devirtualization, intergraph-edge-insertion, etc.

Much complexity is to do with "native" (irrespective of the underlying domain) treatment of Sum types necessary for proper understanding of control flow (e.g. conditionals, loops, CFGs).

Generalization of reachability left to #1672 .


hugr-passes/src/dataflow/test.rs
```diff
@@ -0,0 +1,310 @@
...
+
+// This allows testing creation of tuple/sum Values (only)
+impl From<Void> for Value {
+    fn from(v: Void) -> Self {
+        match v {}
+    }
+}
+
+fn pv_false() -> PartialValue<Void> {
+    PartialValue::new_variant(0, [])
```
acl-cqc, line 85: Arg! This 0 should be `prelude::FALSE_TAG` or something....??

hugr-passes/src/dataflow/test.rs
```diff
@@ -0,0 +1,310 @@
...
+}
+
+fn pv_true() -> PartialValue<Void> {
+    PartialValue::new_variant(1, [])
```
acl-cqc, line 89: Argh! This `1` should be `prelude::TRUE_TAG` or something....??

hugr-passes/src/dataflow/test.rs
```diff
@@ -0,0 +1,310 @@
...
+        Some(TailLoopTermination::NeverContinues),
+        machine.tail_loop_terminates(&hugr, tail_loop.node())
+    )
+}
+
+#[test]
+fn test_tail_loop_always_iterates() {
+    let mut builder = DFGBuilder::new(Signature::new_endo(vec![])).unwrap();
+    let r_w = builder
+        .add_load_value(Value::sum(0, [], SumType::new([type_row![], BOOL_T.into()])).unwrap());
```
acl-cqc, line 175: This `0` should be `TailLoop::CONTINUE_TAG` or *something*....??

hugr-passes/src/dataflow/test.rs
```diff
@@ -0,0 +1,310 @@
...
+        .unwrap();
+    assert_eq!(o2_r, Value::true_val());
+}
+
+#[test]
+fn test_tail_loop_never_iterates() {
+    let mut builder = DFGBuilder::new(Signature::new_endo(vec![])).unwrap();
+    let r_v = Value::unit_sum(3, 6).unwrap();
+    let r_w = builder.add_load_value(r_v.clone());
+    let tag = Tag::new(1, vec![type_row![], r_v.get_type().into()]);
```
acl-cqc, line 145: The `1` here should be `TailLoop::BREAK_TAG` or...something....

hugr-passes/src/dataflow/total_context.rs
```diff
@@ -0,0 +1,54 @@
...
+use ascent::lattice::BoundedLattice;
+use hugr_core::{ops::OpTrait, Hugr, HugrView, IncomingPort, Node, OutgoingPort, PortIndex};
+
+use super::partial_value::{PartialValue, Sum};
+use super::{BaseValue, DFContext};
+
+/// A simpler interface like [DFContext] but where the context only cares about
+/// values that are completely known (as `V`s), i.e. not `Bottom`, `Top`, or
+/// Sums of potentially multiple variants.
+pub trait TotalContext<V>: Clone + Eq + Hash + std::ops::Deref<Target = Hugr> {
```
acl-cqc, line 12: This may belong in the follow-up PR (even tho it's fairly general, so I think lives in this module, it's only used there)...

hugr-passes/src/dataflow/partial_value.rs
```diff
@@ -0,0 +1,747 @@
...
+    /// a sum).
+    ///
+    /// The signature is this way to optimize query/inspection (is-it-a-sum),
+    /// at the cost of requiring more cloning during actual conversion
+    /// (inside the lazy Iterator, or for the error case, as Self remains)
+    ///
+    /// [interpret_leaf_op]: super::DFContext::interpret_leaf_op
+    /// [Sum]: TypeEnum::Sum
+    /// [Tag]: hugr_core::ops::Tag
+    fn as_sum(&self) -> Option<(usize, impl Iterator<Item = Self> + '_)> {
```
acl-cqc, line 29: It'd be possible to use `supports_tag` and `variant_values` here, but that's more flexible than we need (this only needs to cope with concrete Sums, which have a single tag); and I think this single-method interface is much better. I tried to change `supports_tag`+`variant_values` into a single method returning an iterator but couldn't figure out anything (able to deal with multiple tags) that got past the borrow checker....

hugr-passes/src/dataflow/test.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    let tlb = builder
+        .tail_loop_builder([], [], vec![r_v.get_type()].into())
+        .unwrap();
+    let tail_loop = tlb.finish_with_outputs(tagged.out_wire(0), []).unwrap();
+    let [tl_o] = tail_loop.outputs_arr();
+    let hugr = builder.finish_hugr(&EMPTY_REG).unwrap();
+
+    let mut machine = Machine::default();
+    machine.propolutate_out_wires([(r_w, PartialValue::new_variant(3, []))]);
```
cqc-alec, line 156: "propolutate"?

hugr-passes/Cargo.toml
```diff
@@ -14,6 +14,10 @@ categories = ["compilers"]
 
 [dependencies]
 hugr-core = { path = "../hugr-core", version = "0.10.0" }
+portgraph = { workspace = true }
+# This ascent commit has a fix for unsoundness in release/tag 0.6.0:
+ascent = {git = "https://github.com/s-arash/ascent", rev="9805d02cb830b6e66abcd4d48836a14cd98366f3"}
```
aborgna-q, line 19: Merging this will prevent us from publishing to crates.

Ascent 0.7.0 is now out, so please fix before merging.

hugr-passes/src/dataflow/test.rs
```diff
@@ -0,0 +1,310 @@
...
+    assert_eq!(
+        tlb.loop_signature().unwrap().signature(),
+        Signature::new_endo(type_row![BOOL_T, BOOL_T])
+    );
+    let [in_w1, in_w2] = tlb.input_wires_arr();
+    let tail_loop = tlb.finish_with_outputs(in_w1, [in_w2, in_w1]).unwrap();
+
+    let hugr = builder.finish_hugr(&EMPTY_REG).unwrap();
+    // TODO once we can do conditionals put these wires inside `just_outputs` and
+    // we should be able to propagate their values...ALAN wtf? loop control type IS bool ATM
```
acl-cqc, line 231: @doug-q please excuse the "wtf" but can you explain what you were planning here?? (They are currently in `inputs_outputs`, do you mean put them in both `just_inputs` and `just_outputs` ?)

hugr-passes/src/dataflow/test.rs
```diff
@@ -0,0 +1,310 @@
...
+    let [o_w1, o_w2] = tail_loop.outputs_arr();
+
+    let mut machine = Machine::default();
+    machine.propolutate_out_wires([(true_w, pv_true()), (false_w, pv_false())]);
+    machine.run(TestContext(Arc::new(&hugr)));
+
+    let true_or_false = PartialValue::new_variant(0, []).join(PartialValue::new_variant(1, []));
+    // TODO these should be the propagated values for now they will be join(true,false) - ALAN wtf?
```
acl-cqc, line 239: Similarly - I guess this might automagically work if we dup'd `inputs_outputs` into `just_inputs` and `just_outputs` (two variables for two cases rather than one variable for both) - is that what you meant?

hugr-passes/src/dataflow/machine.rs
```diff
@@ -0,0 +1,131 @@
...
+impl<V: AbstractValue, C: DFContext<V>> Default for Machine<V, C> {
+    fn default() -> Self {
+        Self(Default::default(), None)
+    }
+}
+
+impl<PV: AbstractValue, C: DFContext<PV>> Machine<PV, C> {
+    /// Provide initial values for some wires.
+    /// (For example, if some properties of the Hugr's inputs are known.)
+    pub fn propolutate_out_wires(&mut self, wires: impl IntoIterator<Item = (Wire, PV)>) {
```
acl-cqc, line 27: This should be renamed - I guess maybe `populate` or `prepopulate` or just (`set`|`add`)(`_initial`|`_wire_value`)

hugr-passes/Cargo.toml
acl-cqc: Thank you! :-) and done

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,324 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum IO {
+    Input,
+    Output,
+}
+
+/// Clients of the dataflow framework (particular analyses, such as constant folding)
+/// must implement this trait (including providing an appropriate domain type `PV`).
+pub trait DFContext<PV: AbstractValue>: Clone + Eq + Hash + std::ops::Deref<Target = Hugr> {
```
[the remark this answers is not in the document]
acl-cqc: `Deref<Hugr>` is semantically equivalent to your `fn hugr(&self) -> &Hugr` but syntactically much neater - you can do just `*x` rather than `x.hugr()` and you can directly call `x.foo()` rather than `x.hugr().foo()`.

So I quite like it, but I haven't understood the objection yet. (IIUC the "proliferation of generic parameters" applies only if we return an arbitrary `HugrView`? So we could return `impl HugrView`, would that avoid the proliferation? Then, what, this change would allow running the dataflow analysis on a sub-hugr (via a RegionView) with certainty that the analysis wasn't touching anything outside that? (I guess that could be good...)

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,324 @@
...
+    out_wire_value(c, cond, OutgoingPort::from(o_p.index()), v) <--
+      case_node(c, cond, _, case),
+      io_node(c, case, o, IO::Output),
+      in_wire_value(c, o, o_p, v);
+
+    lattice case_reachable(C, Node, Node, bool);
+    case_reachable(c, cond, case, reachable) <-- case_node(c,cond,i,case),
+        in_wire_value(c, cond, IncomingPort::from(0), v),
+        let reachable = v.supports_tag(*i);
+
```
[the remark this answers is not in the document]
acl-cqc: Just added CFG, tests in progress.

`Call` raises big questions - do you use one set of lattice variables for the whole function body (shared between callsites), or a new set for each callsite (with the same relations between variables within each call)? The latter ("context-sensitive" analysis) is more powerful+accurate: variables will be smaller/more precise, you'll figure out more. But it takes a lot of memory.

`CallIndirect` raises....all those questions, *plus* those of working out the callee, and what to do when there are multiple callees. FWIW, I am proposing [removal of CallIndirect](#1569).

How we can be extensible over these is a big question. It feels like any decision is gonna be hardcoded into the datalog, perhaps (if we try hard) with an enum for context-sensitivity level (accuracy vs. cost/memory-usage)....

So, yeah, do you see a better answer than a TODO ?

hugr-passes/src/dataflow/test.rs
acl-cqc: https://github.com/CQCL/hugr/pull/1476#discussion_r1793169977

hugr-passes/src/dataflow/test.rs
acl-cqc: Well, I kept thinking that I could see what you were aiming for here, but it kept being harder than I hoped - the inputs to the second iteration are joined with the inputs to the first, so that tends to get you something uninformative.

But I kinda got there in the end - the loop input needs to be a *sum* with >0 (Partial)values in each variant so you can put different data in each. (And variant0 leads to doing another iteration with variant1, whereas variant1 leads to an exit.) Have added this as a separate test, it is quite cool that it does work.

hugr-passes/src/dataflow/datalog.rs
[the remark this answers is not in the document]
acl-cqc: Hmmmm. Just possibly this would enable us to interpret leaf-op `Call`s by firing up a separate analysis, finishing that, and returning results into the parent analysis. So that might allow that as a way to resolve the Call question. (But you could always just make a new Hugr rather than using a view - the cost of copying nodes seems small compared to running the analysis.) And I don't think this generalizes to also beiing able to do a context-*insensitive* analysis (where all leaf-`Call`s share variables), so probably isn't a good technique overall.

zrho:
Is there ever a `T: AbstractValue` that is not a `PartialValue<V>` for some `V: BaseValue`? Since an `AbstractValue` must implement `BoundedLattice`, there must be a top and a bottom representable in `T`. Via the `AbstractValue::new_variant` you can construct a `T` that represents a particular variant of a sum type. Then via `BoundedLattice::join` you can create `T`s that include multiple variants. So this already looks a lot like `PartialValue<V>`.

zrho:
Huh, is there any particular reason why PartialValue<V> does not have a V: Lattice bound? If you start with a V that only has equality, you can still make it into a bounded lattice. You’ll end up with PartialValue::Top and PartialValue::Value(V::top()) but that isn’t too weird.

acl-cqc:
So prior to 7f2a91a5fc5bc26143f5e82543c172e10ebea90d I was using PartialValue in place of AbstractValue everywhere (with the *underlying* type called AbstractValue i.e.  equivalent to current `BaseValue`) - I could revert that, by adding inside `trait BaseValue`:
```
   fn try_join(self, other: Self) -> Option<Self> {
      (self==other).then_some(self)
   }
```
(note trait-default impls); the caller/framework then turns `None` into `PartialValue::Top`. (And similarly `try_meet` with the caller/framework turning `None` into `PartialValue::Bottom`.) Alternatively these could even return `PartialValue<Self>` rather than `Option<Self>`.

That might be better; thoughts?

(I'm also wondering if we can get rid of `as_sum` but will have to look+think carefully at the not-yet-PR'd constant-folding code, and other not-yet-realized analyses, so maybe best to leave that in.)

zrho:
A suggestion that might streamline the types and traits, and make it clearer how the ordering and the sums interact:

Remove `AbstractValue` and use `PartialValue<V>` everywhere. Remove `as_sum` and instead generate a `PartialValue::Sum` for sum values. In that case the `V` in `PartialValue<V>` would stand for all of the base types that are not sums. This then allows to remove `PVEnum` and let `PartialValue` be the enum. Require `V` for `PartialValue<V>` to be a `BoundedLattice` . In case you start out with some values that only are `PartialEq`, you can make them into a `BoundedLattice` using the type:

```rust
enum FreeBoundedLattice<V> {
    Top,
    Value(V),
    Bottom
}
```

As mentioned above, that way `PartialValue::Top` and `PartialValue::Value(V::top())` are different, just like `PartialValue::Bottom` and `PartialValue::Value(V::bottom())`. That does make sense though: `PartialValue::Value(V::top())` then is any value which is not a sum, in contrast to `PartialValue::Top`which is any value (including sums). That distinction might not be necessary, but it isn't harmful.

acl-cqc:
Ok, I've done some fairly major refactoring here. Firstly removing `AbstractValue` and using `PartialValue` everywhere. I went back to calling the thing inside `PartialValue` an `AbstractValue` but maybe that should still be called `BaseValue`. Secondly dropping `as_sum` as suggested - that this means we can lose `PVEnum` is a big plus that I had missed :-).

I've gone with `try_join` and `try_meet` rather than requiring `BoundedLattice` as this seemed a lighter touch, perhaps I could be persuaded but I think BoundedLattice will be a bit annoying for clients and OTT.

Secondly, I realized that previous behaviour of `interpret_leaf_op` was unsound - the default was, if we don't understand the leaf, return `Bottom` but we must return `Top` (we can't deduce anything about the result, but the op still happens, so we have to assume the worst, i.e. assume nothing, i.e. the value could be anything). I tried to do this by allowing `out_wire_value_proto` to *override* the result of `interpret_leaf_op` but couldn't get that past the borrow checker inside the `ascent!` (see fd717be5a2f0f7d9ffcf7976316fd8476057cd5f), so have gone one step further and moved construction of `PartialValue`(`::PartialSum`)s from hugr constants *into* the framework. This has the benefit that tests no longer need to use `propolutate` for boolean constants which had been a PITA :-) :-)

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,324 @@
...
+fn value_inputs(h: &impl HugrView, n: Node) -> impl Iterator<Item = IncomingPort> + '_ {
+    h.in_value_types(n).map(|x| x.0)
+}
+
+fn value_outputs(h: &impl HugrView, n: Node) -> impl Iterator<Item = OutgoingPort> + '_ {
+    h.out_value_types(n).map(|x| x.0)
+}
+
+// Wrap a (known-length) row of values into a lattice. Perhaps could be part of partial_value.rs?
```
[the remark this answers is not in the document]
acl-cqc: I think this is an internal class (wrapping Vec into a lattice) and so rightly hidden, however, so currently I'm minded to keep it here - it's fairly short

hugr-passes/src/dataflow.rs
```diff
@@ -0,0 +1,91 @@
...
+        _ins: &[PartialValue<V>],
+        _outs: &mut [PartialValue<V>],
+    ) {
+    }
+
+    /// Produces an abstract value from a constant. The default impl
+    /// traverses the constant [Value] to its leaves ([Value::Extension] and [Value::Function]),
+    /// converts these using [Self::value_from_custom_const] and [Self::value_from_const_hugr],
+    /// and builds nested [PartialValue::new_variant] to represent the structure.
+    fn value_from_const(&self, n: Node, cst: &Value) -> PartialValue<V> {
```
acl-cqc, line 41: Alternatives include (a) moving `value_from_const` out of the trait and fixing it; I don't have a concrete reason why anyone would need to override it but it seems "reasonable" that one might; or (b) moving `value_from_custom_const` and `value_from_const_hugr` into a separate trait, and then implementing `value_from_const` for any type providing those two - this would be the more proper way of doing what I've done here but feels a bit overkill TBH.

hugr-passes/src/dataflow/test.rs
```diff
@@ -0,0 +1,498 @@
...
+    assert_eq!(machine.case_reachable(&hugr, cond.node()), None);
+}
+
+// Tuple of
+//   1. Hugr being a function on bools: (x, y) => (x XOR y, x AND y)
+//   2. Input node of entry block
+// Result readable from root node outputs
+// Inputs should be placed onto out-wires of the Node (2.)
+#[fixture]
+fn xor_and_cfg() -> (Hugr, Node) {
```
acl-cqc, line 374: I am realizing this is getting to be a seriously complex CFG. Originally I had the goal that we should test on a CFG that does something that can't easily be done in a Conditional or TailLoop. Making this also cover the need for reachability on CFG basic blocks has made it (slightly) more complex still, so could move that into a separate test (e.g. on a simple diamond CFG), although the improvement here would be slight. Or I could just make this test a CFG that's equivalent to a conditional, say...

hugr-passes/src/dataflow/test.rs
acl-cqc: Now gone

hugr-passes/src/dataflow/machine.rs
acl-cqc: I've now got `prepopulate_out_wire` but also added `run` taking *input wires* as this is much easier to use.

hugr-passes/src/dataflow/datalog.rs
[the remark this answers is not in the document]
acl-cqc: Context-insensitive `Call` added, and TODO for CallIndirect. I'm gonna mark as resolved, but we could raise an issue for a follow-up PR if we have a good plan for context-sensitive interprocedural analysis.

hugr-passes/src/dataflow/machine.rs
```diff
@@ -0,0 +1,160 @@
...
+    /// given initial values for some of the root node inputs.
+    /// (Note that `in_values` will not be useful for `Case` or `DFB`-rooted Hugrs,
+    /// but should handle other containers.)
+    /// The context passed in allows interpretation of leaf operations.
+    ///
+    /// # Panics
+    ///
+    /// If this Machine has been run already.
+    ///
+    pub fn run(
```
acl-cqc, line 45: Perhaps `run` should consume the Machine and return the out_wires, perhaps even in  a tuple with the `AscentProgram` (which could then be discarded, or inspected, but not run via Machine - perhaps manually if one wanted) ?

acl-cqc: (done)

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum IO {
+    Input,
+    Output,
+}
+
+ascent::ascent! {
+    pub(super) struct AscentProgram<V: AbstractValue, C: DFContext<V>>;
+    relation context(C);
```
zrho, line 32: There are a lot of copies of `C`, for every element of every relation, and then again in the indexes. Even when `C` is just an `Arc` that presumably adds to be quite substantial. This could be addressed by using `ascent_run` in a function where the context is in scope. We could then also avoid the hacky `Hash` implementations on contexts.

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+        // Handle basics here. I guess (given the current interface) we could allow
+        // DFContext to handle these but at the least we'd want these impls to be
+        // easily available for reuse.
+        op if op.cast::<MakeTuple>().is_some() => Some(ValueRow::from_iter([PV::new_variant(
+            0,
+            ins.iter().cloned(),
+        )])),
+        op if op.cast::<UnpackTuple>().is_some() => {
+            let elem_tys = op.cast::<UnpackTuple>().unwrap().0;
+            let [tup] = ins.iter().collect::<Vec<_>>().try_into().unwrap();
```
zrho, line 223: This could use `Itertools::exactly_one` to avoid allocating a vector.

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+    // Outputs of each reachable block propagated to successor block or (if exit block) then CFG itself
+    out_wire_value(c, dest, OutgoingPort::from(out_p), v) <--
+        dfb_block(c, cfg, pred),
+        bb_reachable(c, cfg, pred),
+        let df_block = c.get_optype(*pred).as_dataflow_block().unwrap(),
+        for (succ_n, succ) in c.output_neighbours(*pred).enumerate(),
+        io_node(c, pred, out_n, IO::Output),
+        _cfg_succ_dest(c, cfg, succ, dest),
+        node_in_value_row(c, out_n, out_in_row),
+        if let Some(fields) = out_in_row.unpack_first(succ_n, df_block.sum_rows.get(succ_n).unwrap().len()),
```
zrho, line 186: This grabs the row in input values for the node and only looks at the first. It could get the first value directly by using `in_wire_value`. There's some other places where this is done as well.

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+    relation context(C);
+
+    relation node(C, Node);
+    relation in_wire(C, Node, IncomingPort);
+    relation out_wire(C, Node, OutgoingPort);
+    relation parent_of_node(C, Node, Node);
+    relation io_node(C, Node, Node, IO);
+    lattice out_wire_value(C, Node, OutgoingPort, PV<V>);
+    lattice in_wire_value(C, Node, IncomingPort, PV<V>);
+    lattice node_in_value_row(C, Node, ValueRow<V>);
```
zrho, line 41: For an input port connected directly to an output port, this stores three copies of the value, which ascent then needs to keep in sync with lattice operations. We could be more economical like this: let a "link" refer to a group of ports that are all connected. We can then store the association of node and port to link, and the value of the link. That would reduce the number of times that values are copied.

hugr-passes/src/dataflow/datalog.rs
acl-cqc: Yup. My bad here - I never looked at the top-level doc for ascent to see `ascent_run`. Thanks :-), that is about 100* better. (Lead to moving a few things around, and separating DFContext from HugrView.)

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+
+type PV<V> = PartialValue<V>;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum IO {
+    Input,
+    Output,
+}
+
+ascent::ascent! {
```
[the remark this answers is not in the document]
acl-cqc: Yes, I wondered if I should do that, then was lazy. Thanks for keeping me honest :). I've kept these terse in format but I hope the meaning is clear, shout if you think it's worth spending more space.

hugr-passes/src/dataflow/datalog.rs
acl-cqc: Done, ta

hugr-passes/src/dataflow/datalog.rs
acl-cqc: I think this is worth noting as a future optimization but not doing yet / until benchmarks demand, as I think the extra level of indirection will reduce clarity. But to clarify (as I haven't written the comment yet), something like...
```
relation link_values(Link, PV<V>);
relation out_wire_link(Node, OutgoingPort, Link);
relation in_wire_link(Node, OutgoingPort, Link);
relation node_in_value_row(Node, Vec<Link>);
```
the last three are fairly trivial. But then, each (say) `out_wire_value(n, OutgoingPort::from(p.index()), v) <-- .....` becomes
```
link_values(lnk, v) <-- .....,
   out_wire_link(n, OutgoingPort::from(p.index()), lnk)
```
?
That didn't seem too bad, hmmm. I can add a comment to that effect, or make a follow-up PR that I'd keep separate (in case we ever need to revert for debugging, say) if you think that's worth it?

hugr-passes/src/dataflow/datalog.rs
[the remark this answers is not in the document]
acl-cqc: Now that we're using `ascent_run` I've separated DFContext from HugrView altogether so I think this is addressed.

hugr-passes/src/dataflow/datalog.rs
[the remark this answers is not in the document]
acl-cqc: Moved into `value_row.rs` to make more space in `datalog.rs` for Machine to come back....(which then didn't take as much space as I expected because I left AnalysisResults behind in `results.rs`. Never mind.)

So, done.

hugr-passes/src/dataflow/datalog.rs
acl-cqc: Ok, so I had a bit more of a go. Defining a `Link` as a `(Node, OutgoingPort)` (it'd be better to use a Wire but there), I defined:
```
relation node_in_value_preds(Node, Vec<(Node, OutgoingPort)>);
```
which could be filled straightforwardly from the edges of the graph (an element being a node, and then the source of each of said node's inputs). The best I could come up with for a standard-control-flow "untuple the first element and append the rest" was then....
```
             for (succ_n, succ) in hugr.output_neighbours(*pred).enumerate(),
             output_child(pred, out_n),
             _cfg_succ_dest(cfg, succ, dest),
-            node_in_value_row(out_n, out_in_row),
-            if let Some(fields) = out_in_row.unpack_first(succ_n, df_block.sum_rows.get(succ_n).unwrap().len()),
-            for (out_p, v) in fields.enumerate();
+            node_in_value_preds(out_n, out_in_row),
+            for (idx, (out_node, out_port)) in out_in_row.iter().enumerate(),
+            out_wire_value(out_node, out_port, v1),
+            let first_tuple_len = df_block.sum_rows.get(succ_n).unwrap().len(),
+            for (out_p, v) in if idx==0 {
+                v1.variant_values(succ_n, first_tuple_len).unwrap_or_default()
+                .into_iter().enumerate().collect()
+            } else {
+                vec![(idx + first_tuple_len, v1.clone())]
+            };
```
which....unless there's some cunning way to make datalog procedures that I'm missing, would be needed about 5 times. (Bad, but not terrible). However I got even more stuck on the call to `propagate_leaf_op` where it's not sufficient to define in-values to the leaf op one IncomingPort at a time, we need to reassemble them into a Vector to pass the whole lot together to `propagate_leaf_op`. AFAICS this would require defining an ascent custom aggregator, and at that point I'm thinking - well we could, but no, let's not do this now, and let's hope we never have to....

hugr-passes/src/dataflow/partial_value.rs
```diff
@@ -0,0 +1,685 @@
...
+    ///
+    /// The default checks equality between `self` and `other` and returns `self` if
+    /// the two are identical, otherwise `None`.
+    fn try_meet(self, other: Self) -> Option<Self> {
+        (self == other).then_some(self)
+    }
+}
+
+/// Represents a sum with a single/known tag, abstracted over the representation of the elements.
+/// (Identical to [Sum](hugr_core::ops::constant::Sum) except for the type abstraction.)
```
acl-cqc, line 35: Commit 7772dfc04b6b99580f9b11a962c98cd3f174ef04 combines these two/dedups, but at the cost of having to manually implement serde Serialize/Deserialize, so see what you think

hugr-passes/src/dataflow/datalog.rs
zrho: There's much simpler and less cursed ways to do this; but I agree, this PR doesn't need it.

hugr-passes/src/dataflow/datalog.rs
zrho: I see! I think it's clear enough now.

acl-cqc:
I had to recombine DFContext with the Hugr to get round some Rust mutable/immutable reference ownership issues in #1603; rather than requiring `Deref<Target = Hugr>` it seem better to require `AsRef<Hugr>` (sufficient to get you the HugrView methods) or indeed just require `HugrView` (as implementing AsRef gets you that anyway). However I could be persuaded to do something like
```
trait DFContext<V> {
  type View : HugrView;
  fn view() -> &Self::View
}
```
with `Machine::run<C: DFContext>` producing an `AnalysisResults<C::View`. @doug-q ?

hugr-passes/src/dataflow/partial_value.rs
```diff
@@ -0,0 +1,718 @@
...
+
+    /// Turns this instance into a [Sum] of some "concrete" value type `V2`,
+    /// *if* this PartialSum has exactly one possible tag.
+    ///
+    /// # Errors
+    ///
+    /// If this PartialSum had multiple possible tags; or if `typ` was not a [TypeEnum::Sum]
+    /// supporting the single possible tag with the correct number of elements and no row variables;
+    /// or if converting a child element failed via [PartialValue::try_into_value].
+    pub fn try_into_value<VE, SE, V2: TryFrom<V, Error = VE> + TryFrom<Sum<V2>, Error = SE>>(
```
acl-cqc, line 153: Yes, this is highly parametric. The first case will be the "easy" one, where `V2` is `hugr_core::ops::Value`, in which case `VE = Infallible` and `SE = ConstTypeError`. But I think it's important to keep this open for other analysis domains.

It might be more legible if written `<V2, VE, SE> ..... where V2: TryFrom<V, Error=VE> + TryFrom<Sum<V2>, Error=SE>>`, happy to do that if anyone shouts.

The bigger issue may be nomenclature - this relates to `ExtractValueError` so perhaps should be (`try_`)`extract_value`? And also to `Machine::try_read_wire_value` - so, `try_extract_wire_value`?? However I quite like the word "concrete" so could perhaps use that naming to tie the three together - "try_into_concrete", "try_read_wire_concrete", and, erm, "ConcretizationError" (hmmm, not so good....). Thoughts?

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,348 @@
...
+/// derived-Default requires the context to be Defaultable, which is unnecessary
+impl<V: AbstractValue> Default for Machine<V> {
+    fn default() -> Self {
+        Self(Default::default())
+    }
+}
+
+impl<V: AbstractValue> Machine<V> {
+    /// Provide initial values for some wires.
+    // Likely for test purposes only - should we make non-pub or #[cfg(test)] ?
```
acl-cqc, line 30: Not sure quite what to do here. @doug-q may have opinions?

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,348 @@
...
+                }
+            }
+        });
+        for (i, b) in have_inputs.into_iter().enumerate() {
+            if !b {
+                self.0
+                    .push((root, IncomingPort::from(i), PartialValue::Top));
+            }
+        }
+        // Note/TODO, if analysis is running on a subregion then we should do similar
```
acl-cqc, line 69: I'll raise an issue for this. It's not clear how to identify nonlocal edges into a subregion, or to differentiate between nonlocal edges whose source we can't see, and ports with no edge at all, where perhaps "Bottom" is appropriate (it really can't happen, that bit of Hugr can't run....).

hugr-passes/src/dataflow/partial_value.rs
croyzor: > It might be more legible if written `<V2, VE, SE> ..... where V2: TryFrom<V, Error=VE> + TryFrom<Sum<V2>, Error=SE>>`, happy to do that if anyone shouts.

I think you should!

hugr-passes/src/dataflow.rs
```diff
@@ -0,0 +1,108 @@
...
+    ///
+    /// [MakeTuple]: hugr_core::extension::prelude::MakeTuple
+    /// [UnpackTuple]: hugr_core::extension::prelude::UnpackTuple
+    fn interpret_leaf_op(
+        &self,
+        _node: Node,
+        _e: &ExtensionOp,
+        _ins: &[PartialValue<V>],
+        _outs: &mut [PartialValue<V>],
+    ) {
```
croyzor, line 40: You don't need to provide default implementations for trait methods, you can just write `fn f(...);` instead of `fn f(...) { }`

acl-cqc: But this impl is totally ok as a default - it means, I know nothing about any leaf op

hugr-passes/src/dataflow/partial_value.rs
acl-cqc: Done - yes good to standardize the order of the args with `try_read_wire_value`.

acl-cqc:
Note we could separate the DFContext from HugrView - which would be significantly neater - if we did #1636 first, but we need that because `ValidationLevel::run_validated_pass` only gives us a `&mut impl HugrMut`.

hugr-passes/src/dataflow/datalog.rs
acl-cqc: We think this potentially useful (albeit not very powerful, as any prepopulated value will be `join`ed with anything computed). A version that "breaks" the wire (replacing any value computed, without `join`) would be more effective, also dangerous....

hugr-passes/src/dataflow/partial_value.rs
acl-cqc: Any thoughts on naming?

hugr-passes/src/dataflow/test.rs
acl-cqc: Decided `1 == true` was obvious enough, dropping

hugr-passes/src/dataflow/test.rs
acl-cqc: Decided it was obvious enough that 0 == false, so dropping

hugr-passes/src/dataflow/results.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// to be reachable. (Returns `None` if argument is not a child of a CFG.)
+    ///
+    /// [CFG]: hugr_core::ops::CFG
+    /// [DataflowBlock]: hugr_core::ops::DataflowBlock
+    /// [ExitBlock]: hugr_core::ops::ExitBlock
+    pub fn bb_reachable(&self, bb: Node) -> Option<bool> {
+        let cfg = self.hugr().get_parent(bb)?; // Not really required...??
+        self.hugr().get_optype(cfg).as_cfg()?;
+        let t = self.hugr().get_optype(bb);
+        if !t.is_dataflow_block() && !t.is_exit_block() {
```
croyzor, line 72: I think this would be slightly more clear with an application de Morgan's law 🙂

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+          case_node(cond, _i, case),
+          case_reachable(cond, case),
+          output_child(case, o),
+          in_wire_value(o, o_p, v);
+
+        // In `Conditional` <Node>, child `Case` <Node> is reachable given our knowledge of predicate:
+        relation case_reachable(Node, Node);
+        case_reachable(cond, case) <-- case_node(cond, i, case),
+            in_wire_value(cond, IncomingPort::from(0), v),
+            if v.supports_tag(*i);
```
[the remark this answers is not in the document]
acl-cqc: `Top` supports any tag, `Bottom` supports none

hugr-passes/src/dataflow.rs
```diff
@@ -0,0 +1,110 @@
...
+    ///
+    /// [FuncDefn]: hugr_core::ops::FuncDefn
+    /// [FuncDecl]: hugr_core::ops::FuncDecl
+    /// [LoadFunction]: hugr_core::ops::LoadFunction
+    fn value_from_function(&self, _node: Node, _type_args: &[TypeArg]) -> Option<V> {
+        None
+    }
+}
+
+fn traverse_value<V>(
```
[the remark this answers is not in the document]
acl-cqc: It's got an extra param vs the version on `ConstLoader` but TBH I am leaning towards removing `ConstLoader::value_from_const` (keeping just the "leaf" methods `value_from_opaque`, `value_from_const_hugr` and `value_from_function`) - assuming we want correct treatment of sums for conditionals, MakeTuple, etc. etc., every valid impl has to do this....

hugr-passes/src/dataflow.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// to be appropriate in most cases) traverses [Sum](Value::Sum) constants
+    /// to their leaves ([Value::Extension] and [Value::Function]),
+    /// converts these using [Self::value_from_opaque] and [Self::value_from_const_hugr],
+    /// and builds nested [PartialValue::new_variant] to represent the structure.
+    fn value_from_const(&self, n: Node, cst: &Value) -> PartialValue<V> {
+        traverse_value(self, n, &mut Vec::new(), cst)
+    }
+
+    /// Produces an abstract value from an [OpaqueValue], if possible.
+    /// The default just returns `None`, which will be interpreted as [PartialValue::Top].
```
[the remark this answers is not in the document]
acl-cqc: I could make it `PartialValue` but then that allows returning lots of things that make no sense at all, like `Bottom` ("can't happen", what, yes it can) or of course a `Sum` - when we know it isn't one...

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+
+        // outputs of case nodes propagate to outputs of conditional *if* case reachable
+        out_wire_value(cond, OutgoingPort::from(o_p.index()), v) <--
+          case_node(cond, _i, case),
+          case_reachable(cond, case),
+          output_child(case, o),
+          in_wire_value(o, o_p, v);
+
+        // In `Conditional` <Node>, child `Case` <Node> is reachable given our knowledge of predicate:
+        relation case_reachable(Node, Node);
```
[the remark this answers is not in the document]
acl-cqc: Yes, sorting out all dependencies, even (mutually-)recursive ones, is one of the great features of datalog!

hugr-passes/src/dataflow/results.rs
acl-cqc: Done with `then_some`, how's that?

hugr-passes/src/dataflow.rs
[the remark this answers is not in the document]
acl-cqc: Yeah, gone in circles here, but after introducing ConstLocation, I then wanted `impl Into<ConstLocation>` which you can't do on a trait. So I've made it an external method, and we can add an extra trait method (defaulting to the external) as a non-breaking change if we want...

hugr-passes/src/dataflow.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn value_from_opaque(&self, _node: Node, _fields: &[usize], _val: &OpaqueValue) -> Option<V> {
```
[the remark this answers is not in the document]
acl-cqc: Switched to `ConstLocation` in the hope that is clearer than `node`/`fields`. LMK if still unclear!

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+        OpType::ExtensionOp(e) => {
+            // Interpret op using DFContext
+            let init = if ins.iter().contains(&PartialValue::Bottom) {
+                // So far we think one or more inputs can't happen.
+                // So, don't pollute outputs with Top, and wait for better knowledge of inputs.
+                PartialValue::Bottom
+            } else {
+                // If we can't figure out anything about the outputs, assume nothing (they still happen!)
+                PartialValue::Top
+            };
```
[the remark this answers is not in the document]
acl-cqc: Yeah, coverage comes in #1603 .

`interpret_leaf_op` can do anything, it gets an `&mut [PartialValue<V>]` so can overwrite, it does not have to use join.

Note that returning `Bottom` erroneously is unsafe, i.e. can lead the analysis to conclude things that are not true. Returning `Top` is conservative.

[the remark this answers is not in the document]
acl-cqc: The analysis could be better than `contains(PartialValue::Bottom)`, i.e. we could apply the `Bottom` case more often, but at least this way is conservative.

(e.g. `Sum(1, [Bottom, foo, bar])` also "can't happen", so must result from not-yet-computed data that'll turn up later in the analysis - that is, for a PartialSum, we can discount any variant at least one of whose value's (recursively) "contains bottom", and then the PartialSum classed as containing-bottom if it has no variants left)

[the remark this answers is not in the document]
acl-cqc: Oooh, but, as you suggest [here](https://github.com/CQCL/hugr/pull/1664/files#r1846407383), maybe we can simplify.

So there are some leaf ops that can figure out properties of their outputs even when *some* of the inputs are completely unknown, and I had thought that that meant we couldn't do as you suggest, but actually....maybe we can; we'll call `interpret_leaf_op` again later when we have non-bottom inputs. In which case we end up
1. `interpret_leaf_op` never sees any `Bottom` inputs
2. initial outputs passed to `interpret_leaf_op` are always `PartialValue::Top`
3. if any input is `Bottom`, all outputs are set to `Bottom` (without even calling `interpret_leaf_op`)

Does that actually work?!?
This is partly about being clear about when Bottom is used - i.e.  only as part of initialization, as a means of bootstrapping cyclic dependencies (maybe non-cyclic depending on how "clever" ascent tries to be)

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+            in_wire_value(cond, IncomingPort::from(0), v),
+            if v.supports_tag(*i);
+
+        // CFG --------------------
+        relation cfg_node(Node); // <Node> is a `CFG`
+        cfg_node(n) <-- node(n), if ctx.get_optype(*n).is_cfg();
+
+        // In `CFG` <Node>, basic block <Node> is reachable given our knowledge of predicates:
+        relation bb_reachable(Node, Node);
+        bb_reachable(cfg, entry) <-- cfg_node(cfg), if let Some(entry) = ctx.children(*cfg).next();
```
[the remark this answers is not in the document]
acl-cqc: Yes, fair. I think the same applies to conditional (if any non-predicate input is bottom, no cases are reachable; if the elements of any variant contain bottom, that case is unreachable), Call, DataflowBlock/ExitBlock (in the same way as conditional, i.e. gate every CF edge), and depending on semantics perhaps also DFG (i.e. if a nested DFG is a scheduling barrier)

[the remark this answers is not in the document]
acl-cqc: Done by augmenting `ValueRow::unpack_first` (now `ValueRow::unpack_first_no_bottom`) and cases for CFG+DFG

hugr-passes/src/dataflow/datalog.rs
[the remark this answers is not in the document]
acl-cqc: Ok, `interpret_leaf_op` no longer called if any input is Bottom. (I considered a default impl that checks row-contains-bottom but I think that if (a) Bottom can only occur during bootstrap/initialisation, and (b) the op doesn't execute unless it gets a valid input row, this is ok. The case of Conditionals being simplifiable because we know their predicate even if not all their inputs, never hits interpret_leaf_op, so I think this is OK.)

I also considered changing it to return `Vec<PartialValue<V>>` rather than take a mutable reference, but this does mean the callee becomes responsible for allocating a vector of the correct length; there are potentially many callees in the future, and only one caller, which can do that work. If you feel strongly that returning `Vec<PartialValue<V>>` would be better, then I'm willing to change.

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+        out_wire_value(inp, OutgoingPort::from(p.index()), v) <--
+            func_call(call, func),
+            input_child(func, inp),
+            in_wire_value(call, p, v);
+
+        out_wire_value(call, OutgoingPort::from(p.index()), v) <--
+            func_call(call, func),
+            output_child(func, outp),
+            in_wire_value(outp, p, v);
+    };
```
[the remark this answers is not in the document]
acl-cqc: Commented that this can be done via `prepopulate_out_wire` for external functions. Whilst very basic, I think we should not do too much until we have settled how Hugr's are used as "libraries" (and perhaps also linked).

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+    pub fn prepopulate_wire(&mut self, h: &impl HugrView, wire: Wire, value: PartialValue<V>) {
+        self.0.extend(
+            h.linked_inputs(wire.node(), wire.source())
+                .map(|(n, inp)| (n, inp, value.clone())),
+        );
+    }
+
+    /// Run the analysis (iterate until a lattice fixpoint is reached),
+    /// given initial values for some of the root node inputs.
+    /// (Note that `in_values` will not be useful for `Case` or `DFB`-rooted Hugrs,
```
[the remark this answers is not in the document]
acl-cqc: Hopefully this has been sorted now - in fact previously passing `in_values` to a `Case` node would just break, but now they are those provided from the `Input` node within the `Case`.

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -0,0 +1,346 @@
...
+/// derived-Default requires the context to be Defaultable, which is unnecessary
+impl<V: AbstractValue> Default for Machine<V> {
+    fn default() -> Self {
+        Self(Default::default())
+    }
+}
+
+impl<V: AbstractValue> Machine<V> {
+    /// Provide initial values for some wires.
+    // Likely for test purposes only - should we make non-pub or #[cfg(test)] ?
```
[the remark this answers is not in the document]
acl-cqc: Went through a few stages here but in the end I've kept this `pub` as a catch-all mechanism for anything we've missed, with example (instructions, doc'd) to use for "externally-callable" functions in libraries

hugr-passes/src/dataflow/partial_value.rs
```diff
@@ -0,0 +1,720 @@
...
+                true
+            }
+            (_, Self::Bottom) => false,
+            (Self::Bottom, other) => {
+                *self = other;
+                true
+            }
+            (Self::Value(h1), Self::Value(h2)) => match h1.clone().try_join(h2) {
+                Some(h3) => {
+                    let ch = h3 != *h1;
```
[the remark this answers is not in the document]
acl-cqc: Well, first I changed `AbstractValue::try_join` to return `(Self, bool)` i.e. whether it's changed, thus avoiding the comparison. Rather than adding the debug_assert, though, I managed to avoid the `clone` (leaving nothing to compare against) by cunning use of `std::mem::swap`.

A large part of my brain here is saying "premature over-optimization" but...

hugr-passes/src/dataflow/partial_value.rs
```diff
@@ -0,0 +1,720 @@
...
+            (_, other @ Self::Top) => {
+                *self = other;
+                true
+            }
+            (_, Self::Bottom) => false,
+            (Self::Bottom, other) => {
+                *self = other;
+                true
+            }
+            (Self::Value(h1), Self::Value(h2)) => match h1.clone().try_join(h2) {
```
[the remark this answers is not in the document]
acl-cqc: Oh, dang, I thought this was a comment on somewhere else, I have now done this. I can undo if you'd rather but it avoids that nasty `unreachable!` too....

hugr-passes/src/dataflow/results.rs
```diff
@@ -0,0 +1,134 @@
...
+    ///
+    /// # Panics
+    ///
+    /// If a [Type](hugr_core::types::Type) for the specified wire could not be extracted from the Hugr
+    pub fn try_read_wire_value<V2, VE, SE>(
+        &self,
+        w: Wire,
+    ) -> Result<V2, Option<ExtractValueError<V, VE, SE>>>
+    where
+        V2: TryFrom<V, Error = VE> + TryFrom<Sum<V2>, Error = SE>,
```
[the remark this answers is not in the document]
acl-cqc: That took a surprising amount of fiddling but yes, done :)

hugr-passes/src/dataflow/partial_value.rs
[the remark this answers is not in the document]
acl-cqc: Nah, this enabled yet more refactoring here, it avoids all the nasty awkward bits we had before with just one `swap`, I think this is both clearest and shortest yet as well as more efficient :)

hugr-passes/src/dataflow/datalog.rs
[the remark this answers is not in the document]
acl-cqc: Ah, crossed-in-post with your lattice suggestion.
Yes, I've been thinking for awhile that a proper hierarchical notion of reachability would be good.
For starters, this could generalize `bb_reachable` and `case_reachable`.
Your idea of gating `in_wire_value` on reachability is quite cool (so if one input is bottom, they all are), although might make flattening conditionals more difficult....and we can't make `in_wire_value` depend on `node_in_value_row` in circular fashion like that, so we'd need

`in_wire_value_pre` and `node_in_value_row_pre` *ignoring* reachability,
PLUS `node_in_value_row` that copies `node_in_value_row` if none of the elements are bottom (and the parent/case/bb are reachable)....I guess `in_wire_value` could read from `node_in_value_row(n)[port]`.

For this PR or later, hmmm.

hugr-passes/src/dataflow/datalog.rs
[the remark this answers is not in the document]
acl-cqc: And added `propopulate_df_inputs` as a utility. Ok to resolve?

croyzor (approved):
Beautiful work 🚀

PR #1480, opened 2024-08-29
fix(hugr-py): record node children correctly when deserializing

### LINKED ISSUE #1479: bug(py): deserialization fails to record node children correctly.
Failing test:

```python
def test_children():
    mod = Module()
    mod.declare_function("foo", tys.PolyFuncType([], tys.FunctionType.empty()))

    h = mod.hugr
    assert len(h.children(h.root)) == 1

    h2 = Hugr.load_json(h.to_json())

    assert len(h2.children(h2.root)) == 1

```

ss2165:
Closes #1479


aborgna-q (approved):
Good catch

PR #1481, opened 2024-08-29
feat: `Option` / `Result` helpers

### LINKED ISSUE #1473: Add `Option`-like type?
It would be useful to export a standard `Option`/`Optional` type definition from the prelude that can be used by fallible operations. (And include a const value definition).

This can currently be defined as a sum `[T] + []`, but having it as a common definition would ensure we have an homogeneous interface.

aborgna-q:
Adds helpers for defining fallible types, so we ensure we always use the same definition. This change does _not_ modify the serialisation format, it's only helpers on top of `Sum`.

On the rust side:

- Adds prelude definitions for `option` and `result` types (special cases of `Sum`s),
including constant values for `some`/`none`, `ok`/`err`.

On the python side:

- Adds an `Option` and `None_` subtype of `tys.Sum`, and `Some`/`None`/`Ok`/`Err` value definitions (also subtypes of `val.Sum`).
  These implement nicer `__repr__`, but the classes get lost on a serialisation roundtrip. I'm not sure if we want to auto-detect the special cases during deserialisation?
  
The names used are rusticisms. Feel free to bikeshed them.

Closes #1473

### DIFF
--- a/hugr-core/src/extension/prelude.rs
+++ b/hugr-core/src/extension/prelude.rs
@@ -248,13 +248,13 @@ pub const ERROR_TYPE_NAME: TypeName = TypeName::new_inline("error");
 
 /// Return a Sum type with the first variant as the given type and the second an Error.
 pub fn sum_with_error(ty: impl Into<TypeRowRV>) -> SumType {
-    SumType::new([ty.into(), ERROR_TYPE.into()])
+    result_type(ty, ERROR_TYPE)
 }
 
 /// An optional type, i.e. a Sum type with the first variant as the given type and the second as an empty tuple.
 #[inline]
 pub fn option_type(ty: impl Into<TypeRowRV>) -> SumType {
-    result_type(ty, Type::UNIT)
+    result_type(ty, TypeRow::new())
 }
 
 /// A result type, i.e. a two-element Sum type where the first variant
@@ -277,7 +277,7 @@ pub fn const_some(value: Value) -> Value {
 ///
 /// See [option_type].
 pub fn const_some_tuple(values: impl IntoIterator<Item = Value>) -> Value {
-    const_ok_tuple(values, Type::UNIT)
+    const_ok_tuple(values, TypeRow::new())
 }
 
 /// A constant optional value with no value.
@@ -882,6 +882,8 @@ impl MakeRegisteredOp for Lift {
 
 #[cfg(test)]
 mod test {
+    use crate::std_extensions::arithmetic::float_ops::FLOAT_OPS_REGISTRY;
+    use crate::std_extensions::arithmetic::float_types::{ConstF64, FLOAT64_TYPE};
     use crate::{
         builder::{endo_sig, inout_sig, DFGBuilder, Dataflow, DataflowHugr},
         utils::test_quantum_extension::cx_gate,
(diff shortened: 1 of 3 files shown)


ss2165:
Auto detection on round trip was left as a todo/future consideration because it also has implications how how we treat implementer of `AsExtOp`. Said operations would also have to register themselves with deserialization. Probably worth making an issue about this to discuss.

hugr-core/src/extension/prelude.rs
```diff
@@ -255,8 +247,92 @@ pub const ERROR_TYPE: Type = Type::new_extension(ERROR_CUSTOM_TYPE);
...
+/// An optional type, i.e. a Sum type with the first variant as the given type and the second as an empty tuple.
+#[inline]
+pub fn option_type(ty: impl Into<TypeRowRV>) -> SumType {
+    result_type(ty, TypeRow::new())
+}
+
+/// A result type, i.e. a two-element Sum type where the first variant
+/// represents the "Ok" value, and the second is the "Error" value.
+#[inline]
+pub fn result_type(ty_ok: impl Into<TypeRowRV>, ty_err: impl Into<TypeRowRV>) -> SumType {
```
ss2165, line 263: it seems more natural to call this `either` rather than result, result things could wrap this when convenient?

hugr-core/src/extension/prelude.rs
```diff
@@ -255,8 +247,92 @@ pub const ERROR_TYPE: Type = Type::new_extension(ERROR_CUSTOM_TYPE);
 pub const ERROR_TYPE_NAME: TypeName = TypeName::new_inline("error");
 
 /// Return a Sum type with the first variant as the given type and the second an Error.
-pub fn sum_with_error(ty: Type) -> SumType {
-    SumType::new([ty, ERROR_TYPE])
+pub fn sum_with_error(ty: impl Into<TypeRowRV>) -> SumType {
```
ss2165, line 250: would be good to use this in the existing code base, e.g. int ops return types (and possibly tests for values).

hugr-py/src/hugr/tys.py
```diff
@@ -303,6 +305,41 @@ def __repr__(self) -> str:
...
+    first variant is a tuple of elements representing the successful state, the
+    second is a tuple of elements representing failure.
+    """
+
+    def __init__(self, ok: Iterable[Type], err: Iterable[Type]):
+        self.variant_rows = [list(ok), list(err)]
+
+    def __repr__(self) -> str:
+        ok = self.variant_rows[0]
+        err = self.variant_rows[1]
```
ss2165, line 337:
```suggestion
        ok, err  = self.variant_rows
```

hugr-py/src/hugr/tys.py
```diff
@@ -303,6 +305,41 @@ def __repr__(self) -> str:
...
+    """Optional tuple of elements.
+
+    Instances of this type correspond to :class:`Sum` with two variants.
+    The first variant is the tuple of elements, the second is empty.
+    """
+
+    def __init__(self, *tys: Type):
+        self.variant_rows = [list(tys), []]
+
+    def __repr__(self) -> str:
```
ss2165, line 319: please add __str__ versions, to go with #1482

hugr-py/src/hugr/val.py
```diff
@@ -149,6 +151,114 @@ def __repr__(self) -> str:
...
+        val_list = list(vals)
+        super().__init__(
+            tag=0, typ=tys.Result([v.type_() for v in val_list], err_typ), vals=val_list
+        )
+
+    def __repr__(self) -> str:
+        vals_str = self.vals[0] if len(self.vals) == 1 else tuple(self.vals)
+        err = self.typ.variant_rows[1]
+        err_str = err[0] if len(err) == 1 else tuple(err)
+        return f"Ok({vals_str}, {err_str})"
```
ss2165, line 229: A __str__ impl that doesn't print the error type would be nice

hugr-py/src/hugr/val.py
```diff
@@ -149,6 +151,114 @@ def __repr__(self) -> str:
...
+    vals: list[Value]
+
+    def __init__(self, ok_typ: Iterable[tys.Type], vals: Iterable[Value]):
+        val_list = list(vals)
+        super().__init__(
+            tag=1, typ=tys.Result(ok_typ, [v.type_() for v in val_list]), vals=val_list
+        )
+
+    def __repr__(self) -> str:
+        ok = self.typ.variant_rows[1]
```
ss2165, line 256: I think this should be `self.type.variant_rows[0]`? 
Suggest:
1. Using different types for ok and err in your doctest to catch this
2.:
```suggestion
        ok, _ = self.typ.variant_rows
```

hugr-py/src/hugr/val.py
```diff
@@ -149,6 +151,114 @@ def __repr__(self) -> str:
...
+
+    def __init__(self, vals: Iterable[Value], err_typ: Iterable[tys.Type]):
+        val_list = list(vals)
+        super().__init__(
+            tag=0, typ=tys.Result([v.type_() for v in val_list], err_typ), vals=val_list
+        )
+
+    def __repr__(self) -> str:
+        vals_str = self.vals[0] if len(self.vals) == 1 else tuple(self.vals)
+        err = self.typ.variant_rows[1]
```
ss2165, line 227:
```suggestion
        _, err = self.typ.variant_rows
```

ss2165 (changes requested):

hugr-core/src/extension/prelude.rs
aborgna-q: This method was already there, and it _is_ used in multiple parts of the code.
I just made the signature sightly more general, to put it in line with the new helper functions.

ss2165: ah woops sorry

PR #1482, opened 2024-08-29
feat(hugr-py)!: pretty printing for ops and types

### LINKED ISSUE #1468: Implement `__repr__` for all `Type` and `Op` subclasses
Otherwise, the new dot visualisation from #1462 is completely unreadable

### LINKED ISSUE #1470: Dot renderer panics
Trying to visualise the following hugr with the new dot renderer from #1462
```
{"version":"live","nodes":[{"parent":0,"op":"Module"},{"parent":0,"op":"FuncDecl","name":"foo","signature":{"params":[],"body":{"t":"G","input":[],"output":[{"t":"Opaque","extension":"arithmetic.int.types","id":"int","args":[{"tya":"BoundedNat","n":6}],"bound":"C"}],"extension_reqs":[]}}}],"edges":[],"metadata":[],"encoder":"hugr-py v0.7.0"}
```
raises an exception
```
graphviz.backend.execute.CalledProcessError: Command '[PosixPath('dot'), '-Kdot', '-Tsvg', '-O', 'test']' returned non-zero exit status 1. [stderr: b"Error: not well-formed (invalid token) in line 9 \n... FuncDecl(name='foo', signature=PolyFuncType(params=[], body=FunctionType([], [ExtType(type_def=TypeDef(name='int', description='Variable-width integer.', params=[BoundedNatParam(upper_bound=7)], bound=ExplicitBound(bound=<TypeBound.Copyable: 'C'> ...\nin label of node 1\n"]
```

Probably a character escaping issue. Also see #1468

ss2165:
Closes #1468 
Closes #1470

Previous testing wasn't covering label formatting because it requires actually rendering the dot in to an image. Have enabled that, but put it behind an env var because it significantly slows down tests.

BREAKING CHANGE: rename `Custom.name` to `Custom.op_name` and `Func(Defn/Decl).name` to `f_name` to allow for new `name` method

### DIFF
--- a/hugr-py/src/hugr/ops.py
+++ b/hugr-py/src/hugr/ops.py
@@ -66,6 +66,10 @@ class Op(Protocol):
     def _invalid_port(self, port: InPort | OutPort) -> InvalidPort:
         return InvalidPort(port, self)
 
+    def name(self) -> str:
+        """Name of the operation."""
+        return str(self)
+
 
 def _sig_port_type(sig: tys.FunctionType, port: InPort | OutPort) -> tys.Type:
     if port.direction == Direction.INCOMING:
@@ -177,6 +181,9 @@ class Input(DataflowOp):
     def __call__(self) -> Command:
         return super().__call__()
 
+    def name(self) -> str:
+        return "Input"
+
 
 @dataclass()
 class Output(DataflowOp, _PartialOp):
@@ -199,6 +206,9 @@ class Output(DataflowOp, _PartialOp):
 
     def _set_in_types(self, types: tys.TypeRow) -> None:
         self._types = types
+
+    def name(self) -> str:
+        return "Output"
 
 
 @runtime_checkable
@@ -286,12 +296,15 @@ class AsExtOp(DataflowOp, Protocol):
     def num_out(self) -> int:
         return len(self.outer_signature().output)
 
+    def name(self) -> str:
+        return f"{self.ext_op._op_def.qualified_name()}"
+
 
 @dataclass(frozen=True, eq=False)
 class Custom(DataflowOp):
     """Serializable version of non-core dataflow operation defined in an extension."""
 
-    name: str
+    op_name: str
     signature: tys.FunctionType = field(default_factory=tys.FunctionType.empty)
     description: str = ""
     extension: tys.ExtensionId = ""
@@ -301,7 +314,7 @@ class Custom(DataflowOp):
         return sops.ExtensionOp(
             parent=parent.idx,
             extension=self.extension,
-            name=self.name,
+            name=self.op_name,
             signature=self.signature._to_serial(),
             description=self.description,
             args=ser_it(self.args),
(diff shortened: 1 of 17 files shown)


hugr-py/tests/conftest.py
```diff
@@ -164,6 +165,10 @@ def validate(
     if snap is not None and isinstance(h, Hugr):
         dot = h.render_dot()
         assert snap == dot.source
+        if os.environ.get("HUGR_RENDER_DOT"):
+            with NamedTemporaryFile(mode="w", delete=True) as f:
+                # test rendering succeeds
+                dot.render(f.name)
```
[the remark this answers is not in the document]
ss2165: much simpler, thanks (though it is similarly slow)

hugr-py/src/hugr/ops.py
```diff
@@ -286,12 +296,15 @@ def _to_serial(self, parent: Node) -> sops.ExtensionOp:
     def num_out(self) -> int:
         return len(self.outer_signature().output)
 
+    def name(self) -> str:
+        return f"{self.ext_op._op_def.qualified_name()}"
```
mark-koch, line 300:
```suggestion
        return self.ext_op._op_def.qualified_name()
```

hugr-py/src/hugr/tys.py
```diff
@@ -180,6 +186,9 @@ class StringArg(TypeArg):
     def _to_serial(self) -> stys.StringArg:
         return stys.StringArg(arg=self.value)
 
+    def __str__(self) -> str:
+        return self.value
```
mark-koch, line 190:
```suggestion
        return f'{"self.value"}'
```

hugr-py/src/hugr/tys.py
```diff
@@ -446,6 +483,10 @@ def resolve(self, registry: ext.ExtensionRegistry) -> PolyFuncType:
             body=self.body.resolve(registry),
         )
 
+    def __str__(self) -> str:
+        # ∀[a]. [list<a>] -> [Bool]
+        return f"∀{self.params}. {self.body!s})"
```
mark-koch, line 488: This would look more like `∀[TypeTypeParam(TypeBound.Any)].[list<Variable(0)>] -> [Bool]`

We probably want more concise `__str__` implementations for `TypeParam`s. Also maybe we should print sth like `$0` instead of `Variable(0)` (same for `VariableArg`)?

hugr-py/src/hugr/tys.py
```diff
@@ -478,6 +519,9 @@ def _to_serial(self) -> stys.Opaque:
             bound=self.type_bound(),
         )
 
+    def __str__(self) -> str:
+        return f"{self.type_def.name}<{', '.join(str(arg) for arg in self.args)}>"
+
```
mark-koch, line 524: `<>` should be left out if `self.args` is empty

hugr-py/src/hugr/tys.py
```diff
@@ -513,6 +557,9 @@ def resolve(self, registry: ext.ExtensionRegistry) -> Type:
 
         return ExtType(type_def, self.args)
 
+    def __str__(self) -> str:
+        return f"{self.id}<{', '.join(str(arg) for arg in self.args)}>"
```
mark-koch, line 561: Same here

mark-koch (commented):
Looks good, the only thing I'm not sure about are generic function types (see comment below)

mark-koch (approved):
Love the tests!

PR #1484, opened 2024-08-29
fix: Sum value equality. Add unit tests

aborgna-q:
This was supposed to be part of #1481, but pushed it to the branch that depended on it instead 🤦 

- Adds the string/repr unit tests suggested by https://github.com/CQCL/hugr/pull/1481#pullrequestreview-2268976031

- Tests—and fixes—equality comparation between Sum values.


ss2165 (approved):
LGTM

PR #1489, opened 2024-08-30
refactor!: Make Either::Right the "success" case

### LINKED ISSUE #1487: Either::Left should be the failing state
Bool uses left for False, right for True.
We should be consistent for `either` and `option`, using the right variant for the success case.

aborgna-q:
Similarly for options, which now have type `[] + [elems...]`.
Adds `const_ok`/`_fail` aliases to `const_right`/`_left`.

Closes #1487.

Modifies the old folding tests that hard-coded the sum tags.

BREAKING CHANGE: Binary sums representing fallible values now use tag `1` for the successful variant


hugr-core/src/extension/prelude.rs
```diff
@@ -397,6 +431,14 @@ impl ConstError {
             message: message.to_string(),
         }
     }
+
+    /// Returns an "either" value with a failure variant.
+    ///
+    /// args:
+    ///     ty_ok: The type of the success variant.
+    pub fn as_either(self, ty_ok: impl Into<TypeRowRV>) -> Value {
```
ss2165, line 439: nice

hugr-py/src/hugr/val.py
```diff
@@ -193,10 +193,10 @@ class None_(Sum):
     """
 
     def __init__(self, *types: tys.Type):
-        super().__init__(tag=1, typ=tys.Option(*types), vals=[])
+        super().__init__(tag=0, typ=tys.Option(*types), vals=[])
 
     def __repr__(self) -> str:
-        return f"None({', '.join(map(repr, self.typ.variant_rows[0]))})"
+        return f"None({', '.join(map(repr, self.typ.variant_rows[1]))})"
```
ss2165, line 199: can use `comma_sep_repr` from utils

ss2165 (approved):

PR #1499, opened 2024-09-02
feat: `is_ext` utility for checking equality against extension ops

ss2165:
Open to bikeshedding on name of `is_ext`

left in an `assert_eq` that just uses cast directly as a comparison (this may not be worthwhile)


ss2165:
Closing as I've convinced myself it's not useful.

PR #1501, opened 2024-09-02
feat(hugr-cli)!: allow registry specification in `run_dump`

ss2165:
To allow downstream cli tools to dump their own extensions.

BREAKING CHANGE: `ExtArgs::run_dump` now takes the registry to dump as an argument.


ss2165 (commented):
Putting on the non-breaking milestone as it only breaks the cli library.

aborgna-q (approved):

PR #1518, opened 2024-09-03
fix: Update collections extension

aborgna-q:
Broken in #1513, it's failing on main

Should we make this a required check?


mark-koch (approved):
> Should we make this a required check?

Yes I think so, this is easy to forget

Issue #1433, opened 2024-08-15
HUGR Model and binary serialisation format.


ss2165:
Goals:
1. Stop serialising core HUGR structures directly and go via a new Model structure. See #1326. 
2. Switch to a binary serialisation format with efficient table-encoding.
Unblocks:
4. Text format in/out from model.
5. Stable-ish serialisation.
6. Resolve conceptual issues with "value". See #1425.

Tasks:

- [x] #1434
- [x] Merge draft of model in to main with feature flag #1542 
- [x] #1453
- [x] Add constants to model
- [x] Add constraint to capture type bounds
- [x] Add extension sets with variables to model
- [x] Add rows with variables to model
- [ ] Extensions expressed in model structure.
- [ ] Add entry block op to core?
- [ ] Finalise binary format and complete implementation from model.
- [ ] #1861
- [ ] #1553 
- [x] #1554
- [x] #1667

zrho:
# `Ref` Types

The model has three `Ref` types: `GlobalRef`, `LocalRef`, and `LinkRef`. These types are pointers to other objects in the model and come in two variants: each type of `Ref` has a named variant and a more compact variant that directly refers to the thing being referenced by its id.

## `GlobalRef`

`GlobalRef`s point to global definitions, things that are defined by a node: this includes functions, aliases, custom operations and custom types. Where a `GlobalRef` is used, there is the ability to pass values to the parameters of the global definition.

Functions and aliases already can be defined or declared by nodes with the corresponding operations. We can therefore refer to them either by their name or by the id of that node. We don't have these operations for custom types and custom operations yet, and so they need to be referred to by their name. As part of the declarative extensions project, we can add the missing operations to declare custom types and custom operations. We can then refer to them simply by the id of the corresponding node. Moreover we can attach metadata to these nodes, including a name and description, to generate documentation or code, where desired.

- [ ] Declarative extensions: declare custom operations in the model.
- [ ] Declarative extensions: declare custom types in the model.

## `LocalRef`

`LocalRef`s point to variables that are bound by a node's parameters. For instance, when a node defines a function that is polymorphic in some type, this type can be referenced by a `LocalRef`. Currently, the indexed variant of `LocalRef` only points at the index of the parameter in the list of parameters of the node. I've come to see that this is not yet ideal, as term deduplication would identify two variables as the same even though they reference parameters of different nodes and potentially even different types. This can be easily fixed by adding the id of the node to the `LocalRef`.

[...]

PR #1542, opened 2024-10-02
feat: Draft for `hugr-model` with export, import, parsing and pretty printing

zrho:
This PR defines a draft of the model format, and the following parts that operate on the format:

 - **Import**: A module in the model format can be imported as a core hugr module.
 - **Export**: A core hugr module can be exported into the model format.
 - **Parsing**: A model hugr can be parsed from an s-expression based text format.
 - **Pretty printing:**: A model hugr can be printed into the s-expression format, with readable indentation.

This PR is very big, and yet it does not present a completed feature. The model and core format differ in several aspects. Where they differ, this is mostly intentional: The model is kept flexible enough to incorporate features that we know that we want in the future but are currently hard or impossible to express in the core. Import and export perform conversions where possible, but do not yet cover everything.

To prevent this PR from becoming even bigger and to allow for core and model to converge incrementally from both directions, I suggest the following plan: We merge this PR into main with the model related code being feature gated and considered experimental. We then perform tweaks to model and core in smaller PRs where necessary until they are sufficiently close.


hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+        /// The type of the items in the list.
+        ///
+        /// `item_type : static`
+        item_type: TermId,
+    },
+
+    /// A literal string.
+    ///
+    /// `"STRING" : str`
+    Str(&'a str),
```
ss2165, line 474: request: add "bytes"

zrho: See #1545.

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+        /// The name of the custom operation.
+        name: GlobalRef<'a>,
+    },
+    /// Custom operation.
+    ///
+    /// The implicit parameters of the operation are included.
+    CustomFull {
+        /// The name of the custom operation.
+        name: GlobalRef<'a>,
+    },
```
ss2165, line 223: why both?

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+/// A region in the hugr.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Region<'a> {
+    /// The kind of the region. See [`RegionKind`] for details.
+    pub kind: RegionKind,
+    /// The source ports of the region.
+    pub sources: &'a [Port<'a>],
+    /// The target ports of the region.
+    pub targets: &'a [Port<'a>],
+    /// The nodes in the region. The order of the nodes is not significant.
```
ss2165, line 275: not clear what the ports of a region are. In dataflow, are they the ports belonging to Input/Output nodes?

ss2165: is this why Input/Output aren't Operation variants?

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+
+/// A function declaration.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct FuncDecl<'a> {
+    /// The name of the function to be declared.
+    pub name: &'a str,
+    /// The static parameters of the function.
+    pub params: &'a [Param<'a>],
+    /// The type of the function.
+    pub func: TermId,
```
ss2165, line 309:
```suggestion
    pub f_type: TermId,
```
or some name more specific than "func"?

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+    }
+}
+
+/// A reference to a link.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum LinkRef<'a> {
+    /// Reference to the link by its id.
+    Id(LinkId),
+    /// Reference to the link by its name.
+    Named(&'a str),
```
ss2165, line 378: when do "named links" pop up?

ss2165: I think I've answered my question, if you're using them to connect things in text I guess? and then presumably we can resolve everything to ids later optionally.

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+
+    /// A local variable.
+    Var(LocalRef<'a>),
+
+    /// A symbolic function application.
+    ///
+    /// `(GLOBAL ARG-0 ... ARG-n)`
+    Apply {
+        // TODO: Should the name be replaced with the id of the node that defines
+        // the function to be applied? This could be a type, alias or function.
```
ss2165, line 419: seems like this todo has been done by using a ref, should `name` field be renamed?

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+    /// There is a reference to a region that does not exist.
+    #[error("region not found: {0:?}")]
+    RegionNotFound(RegionId),
+    /// There is a local reference that does not resolve.
+    #[error("local variable invalid: {0:?}")]
+    InvalidLocal(String),
+    /// There is a global reference that does not resolve to a node
+    /// that defines a global variable.
+    #[error("global variable invalid: {0:?}")]
+    InvalidGlobal(String),
```
ss2165, line 604: should these be taking the relevant `LocalRef` and `GlobalRef` rather than String?

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+    /// The model contains an operation in a place where it is not allowed.
+    #[error("unexpected operation on node: {0:?}")]
+    UnexpectedOperation(NodeId),
```
ss2165, line 607: worth reporting the unexpected operation too?

ss2165 (commented):
first review of mod.rs

hugr-model/src/v0/mod.rs
zrho: In principle yes, and I did that originally. However it would introduce a lifetime parameter to the error, which might not be what we want.

hugr-model/src/v0/mod.rs
zrho: Yes. Currently the resolving is done by the `import` and is global, but should be relegated to a resolver step. Having the links globally scoped is very annoying when writing text since suddenly nodes from different functions want to connect to each other. We might fight inspiration in MLIR's https://mlir.llvm.org/docs/LangRef/#value-scoping.

hugr-model/src/v0/mod.rs
zrho: One only provides the explicit parameters, the other provides all parameters. Ultimately we want the one where all parameters are given, but that requires knowing the declaration of whatever thing the `GlobalRef` references, since we otherwise can't tell how many implicit parameters there are and where they are in the parameter list. A type inference step can convert the partial variant to the full variant since it has that information.

Also in the text format we can express both the full and the partial variant. Most of the time when you write these things manually you wouldn't want to include all implicit parameters, even if just as a wildcard. But rarely you might want to give them explicitly. This is inspired by how Lean does implicit parameters: https://docs.lean-lang.org/lean4/doc/implicit.html

The same happens in `Term::Apply` and `Term::ApplyFull`.

Instead of having these as different variants, we could have one variant with a flag. Matter of taste, I'd be happy with either.

hugr-model/src/v0/mod.rs
zrho: Yes, good catch. Do you have an idea for a good name? In my draft for the capnp serialization I've used `global`, but I'm not entirely sure about that.

ss2165: `func`, `term`?

hugr-model/src/v0/mod.rs
ss2165: Thanks for the explanation - I'm happy with having different variants, would be nice to have a brief version of the above explanation in the docstrings

hugr-model/src/v0/text/parse.rs
```diff
@@ -0,0 +1,716 @@
...
+
+/// A parsed HUGR module.
+///
+/// This consists of the module itself, together with additional information that was
+/// extracted from the text format.
+#[derive(Debug, Clone)]
+pub struct ParsedModule<'a> {
+    /// The parsed module.
+    pub module: Module<'a>,
+    // TODO: Spans
```
cqc-alec, line 34: TODO here?

hugr-model/src/v0/text/parse.rs
```diff
@@ -0,0 +1,716 @@
...
+                }
+            }
+
+            Rule::term_list_type => {
+                let item_type = self.parse_term(inner.next().unwrap())?;
+                Term::ListType { item_type }
+            }
+
+            Rule::term_str => {
+                // TODO: Escaping?
```
cqc-alec, line 164: Also here.

hugr-core/tests/model.rs
```diff
@@ -0,0 +1,57 @@
+#[cfg(feature = "model")]
```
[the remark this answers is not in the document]
zrho: Hm, I just tried this but both the LSP and `cargo test` appear to ignore it. Perhaps I'm doing something wrong.

hugr-model/src/v0/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+    /// The output ports of the node.
+    pub outputs: &'a [Port<'a>],
+    /// The parameters of the node.
+    pub params: &'a [TermId],
+    /// The regions of the node.
+    pub regions: &'a [RegionId],
+    /// The meta information attached to the node.
+    pub meta: &'a [MetaItem<'a>],
+    /// The type of the node.
+    pub r#type: TermId,
```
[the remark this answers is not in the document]
zrho: Yes. `signature` is perhaps the better name for consistency.

aborgna-q (commented):
Some general comments, I'll do mod/export next.

Other tidbits:
- [x] Add `hugr-model` to `.github/change-filters.yml`

hugr-core/tests/model.rs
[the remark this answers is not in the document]
zrho: Fixed it, it's `[[test]]` instead of `[[tests]]`.

cqc-alec (commented):
(Looked at `parse.rs` and `print.rs`.) LGTM, I just highlighted some TODO comments that you may or may not want to address here (if not we should make issues).

hugr-model/src/v0/mod.rs
zrho: `signature` would work well I think.

hugr-model/src/v0/text/print.rs
```diff
@@ -0,0 +1,649 @@
...
+                    this.print_term(*tail)?;
+                }
+                Ok(())
+            }),
+            Term::ListType { item_type } => self.print_parens(|this| {
+                this.print_text("list");
+                this.print_term(*item_type)
+            }),
+            Term::Str(str) => {
+                // TODO: escape
```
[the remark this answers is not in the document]
zrho: Will be done: https://github.com/CQCL/hugr/issues/1549

hugr-model/src/v0/text/parse.rs
zrho: Also postponed to https://github.com/CQCL/hugr/issues/1549.

hugr-model/src/v0/text/parse.rs
zrho: Source spans will be useful, but aren't essential at the moment. So https://github.com/CQCL/hugr/issues/1550

ss2165 (commented):
import first pass, second half coming soon

hugr-core/src/import.rs
```diff
@@ -0,0 +1,1214 @@
...
+                    Some((decl.name, NamedItem::FuncDecl(node_id)))
+                }
+                model::Operation::DeclareFunc { decl } => {
+                    Some((decl.name, NamedItem::FuncDefn(node_id)))
+                }
+                _ => None,
+            };
+
+            if let Some((name, item)) = item {
+                // TODO: Deal with duplicates
```
ss2165, line 1207: todo?

ss2165 (commented):
import round 2

hugr-core/src/export.rs
```diff
@@ -0,0 +1,726 @@
...
+        for child in self.hugr.children(self.hugr.root()) {
+            children.push(self.export_node(child));
+        }
+
+        let root = self.module.insert_region(model::Region {
+            kind: model::RegionKind::DataFlow,
+            sources: &[],
+            targets: &[],
+            children: children.into_bump_slice(),
+            meta: &[],
```
aborgna-q, line 68: Either implement or add a TODO+issue for encoding the metadata.

hugr-core/src/import.rs
```diff
@@ -0,0 +1,1214 @@
...
+        let optype = OpType::Conditional(Conditional {
+            sum_rows,
+            other_inputs,
+            outputs,
+            extension_delta: ExtensionSet::new(),
+        });
+
+        let node = self.make_node(node_id, optype, parent)?;
+
+        for region in node_data.regions {
```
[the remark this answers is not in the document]
zrho: In the model, everything order sensitive has been addressed by regions:
 - Input and Output are replaced by the data flow region's sources and targets
 - The first block and the Exit block are replaced by the control flow region's sources and targets
 - Instead of case nodes, the conditional node has multiple regions, and regions are ordered.

For now, the import and export convert to the order sensitive encoding of the core which is unchanged. How to proceed with that relates to https://github.com/CQCL/hugr/issues/1546 and is probably out of scope of this PR.

hugr-core/src/export.rs
```diff
@@ -0,0 +1,726 @@
...
+
+        Self {
+            hugr,
+            module,
+            bump,
+            links: IndexSet::new(),
+        }
+    }
+
+    pub fn export_root(&mut self) {
```
aborgna-q, line 53: Why is the logic for the root module of the Hugr different from the node's?

This seems to assume the hugr root is a Module.
As far as I understand, passing a DFG-rotted hugr here will fail when calling `self.export_node(child)` on the input / output nodes.

Some root Optypes like `Case` and `ExitBlock` should be detected early and rejected here.

hugr-core/src/import.rs
```diff
@@ -0,0 +1,1214 @@
...
+            model::Term::Wildcard => Err(error_uninferred!("wildcard")),
+
+            model::Term::Var(var) => {
+                let mut es = ExtensionSet::new();
+                let (index, _) = self.resolve_local_ref(var)?;
+                es.insert_type_var(index);
+                Ok(es)
+            }
+
+            model::Term::ExtSet { extensions, rest } => {
```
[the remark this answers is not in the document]
zrho: It's for extension sets where not all extensions are known, so the rest can be a variable. It's like the tail of a list.

There's a mismatch here with the core: The core at the moment allows variables in rows (becoming lists in the model since that's the name of these things in type theory. That also opens up using "row" to refer to actual rows if we want to add them) and in extension sets. But having multiple variables in there leads to a weird type theory in which unification is ambiguous. When we want to union multiple unknown extension sets or lists, we can do that with constraints.

hugr-core/src/import.rs
```diff
@@ -0,0 +1,1214 @@
...
+        &mut self,
+        term_id: model::TermId,
+    ) -> Result<TypeBase<RV>, ImportError> {
+        match self.get_term(term_id)? {
+            model::Term::Wildcard => Err(error_uninferred!("wildcard")),
+            model::Term::Apply { .. } => {
+                Err(error_uninferred!("application with implicit parameters"))
+            }
+
+            model::Term::ApplyFull { name, args } => {
```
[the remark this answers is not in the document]
zrho: I've added some docs by nows. Does it become clearer?

hugr-core/src/export.rs
```diff
@@ -0,0 +1,726 @@
+//! Exporting HUGR graphs to their `hugr-model` representation.
```
aborgna-q, line 1: Given the differences between the core rust representation and the hugr model, it would be nice to list what's different here. (node regions, unsupported optypes, ...)

hugr-core/src/import.rs
```diff
@@ -0,0 +1,1214 @@
...
+        let inputs = self.import_type_row::<RV>(*inputs)?;
+        let outputs = self.import_type_row::<RV>(*outputs)?;
+        // TODO: extensions
+        Ok(FuncTypeBase::new(inputs, outputs))
+    }
+
+    fn import_closed_list(
+        &mut self,
+        mut term_id: model::TermId,
+    ) -> Result<Vec<model::TermId>, ImportError> {
```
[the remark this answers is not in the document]
zrho: Not a nit, every allocation is precious! I wanted to make this an iter at first, but unfortunately that doesn't fly with the borrow checker: in the places where it is iterated over, we call methods that potentially mutate the context. So `import_closed_list` would hold a borrow during the iteration where we need a mutable borrow. If it shows up in the profiler, we could see if we can avoid it, but so far it's not the bottleneck.

hugr-core/src/export.rs
```diff
@@ -0,0 +1,726 @@
...
+            }
+
+            OpType::Conditional(_) => {
+                regions = self.export_conditional_regions(node);
+                model::Operation::Conditional
+            }
+
+            // Opaque/extension operations should in the future support having multiple different
+            // regions of potentially different kinds. At the moment, we check if the node has any
+            // children, in which case we create a dataflow region with those children.
```
aborgna-q, line 315: // We assume that the regions will always have dataflow kind.

hugr-core/src/export.rs
```diff
@@ -0,0 +1,726 @@
...
+                model_ports.push(model_port);
+            }
+        }
+
+        model_ports.into_bump_slice()
+    }
+
+    pub fn make_port(&mut self, node: Node, port: impl Into<Port>) -> Option<model::Port<'a>> {
+        let port: Port = port.into();
+        let op_type = self.hugr.get_optype(node);
```
aborgna-q, line 103: nit: This can be retrieved once in `make_ports` and re-used

hugr-core/src/import.rs
```diff
@@ -0,0 +1,1214 @@
...
+            model::Term::StaticType => Err(error_unsupported!("`static` as `TypeArg`")),
+            model::Term::ControlType => Err(error_unsupported!("`ctrl` as `TypeArg`")),
+
+            model::Term::FuncType { .. }
+            | model::Term::Adt { .. }
+            | model::Term::Control { .. } => Err(model::ModelError::TypeError(term_id).into()),
+        }
+    }
+
+    fn import_extension_set(
```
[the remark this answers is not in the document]
zrho: It is and I found a bug. Good catch.

hugr-core/src/export.rs
```diff
@@ -0,0 +1,726 @@
...
+    /// Mapping from ports to link indices.
+    /// This only includes the minimum port among groups of linked ports.
+    links: IndexSet<(Node, Port)>,
+    bump: &'a Bump,
+}
+
+impl<'a> Context<'a> {
+    pub fn new(hugr: &'a Hugr, bump: &'a Bump) -> Self {
+        let mut module = model::Module::default();
+        module.nodes.reserve(hugr.node_count());
```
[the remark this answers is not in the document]
zrho: Nodes, terms and regions have different capacities, so that'd be a `with_capacity` with quite a few parameters (all of the same type, don't mix them up!).

hugr-core/src/export.rs
zrho: TODO: https://github.com/CQCL/hugr/issues/1553

hugr-core/src/export.rs
zrho: For now

hugr-core/src/export.rs
zrho: `make_port` is used once individually as well. Reusing it would just avoid some bounds checks, `get_optype` luckily does not copy everything.

hugr-core/src/export.rs
zrho: I'm in the process of writing these things up as comments to the issue https://github.com/CQCL/hugr/issues/1433

hugr-core/src/export.rs
zrho: The core root module node is exported not as a node but as a region. I was not aware that there can be `Hugr`s whose roots are not `Module`s...

hugr-core/src/export.rs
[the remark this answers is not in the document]
aborgna-q: Then maybe a `reserve_nodes` method instead, that can be documented instead of relying on knowing how the Model works internally.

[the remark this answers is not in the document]
zrho: There isn't really an internally. It's plain old data. Everything is pub.

hugr-core/src/export.rs
```diff
@@ -0,0 +1,726 @@
...
+        Self {
+            hugr,
+            module,
+            bump,
+            links: IndexSet::new(),
+        }
+    }
+
+    pub fn export_root(&mut self) {
+        let r#type = self.module.insert_term(model::Term::Wildcard);
```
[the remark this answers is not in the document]
zrho: I've changed this where the term is `signature` instead.

hugr-core/src/export.rs
zrho: I propose we postpone this: https://github.com/CQCL/hugr/issues/1554

zrho:
Given that the code is deliberately declared `unstable` and after looking over what the tests cover, I am currently fine with the extent of test coverage. This will increase as more features are built around it. So if you agree do not let that be an obstacle to approving.

cqc-alec:
Happy to approve the parts I looked at.

hugr-core/src/import.rs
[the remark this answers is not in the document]
ss2165: where should I be looking?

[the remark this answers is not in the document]
zrho: At the doc strings of `Term::Apply`, `Term::ApplyFull`, `Operation::Custom` and `Operation::CustomFull`.

ss2165:
Happy to approve mod and import

hugr-core/src/export.rs
zrho: I've added the ones that I am aware of.

aborgna-q:
Same here

aborgna-q (approved):

PR #1547, opened 2024-10-03
fix!: Make list length op give back the list

### LINKED ISSUE #1543: List `length` op must also return the list
Otherwise, we can't really use it in a linear context

mark-koch:
Fixes #1543.

BREAKING CHANGE: The `length` op in the std `collections` extensions now also returns the list.

### DIFF
--- a/specification/std_extensions/collections.json
+++ b/specification/std_extensions/collections.json
@@ -190,6 +190,22 @@
           ],
           "output": [
             {
+              "t": "Opaque",
+              "extension": "collections",
+              "id": "List",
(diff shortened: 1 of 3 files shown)


specification/std_extensions/collections.json
```diff
@@ -189,6 +189,22 @@
             }
           ],
           "output": [
+            {
```
cqc-alec, line 192: Seems wrong that we should have this file duplicated in source control, but that's another issue.

cqc-alec (approved):

Issue #1551, opened 2024-10-03
There is a lot of cloning


zrho:
I have benchmarked the roundtrip of a circuit via the path `core -> model -> capnp -> model -> core`. I was surprised about how slow it was; way slower than the JSON export. So I profiled to see where all the time goes. This is the export:
<img width="1684" alt="Screenshot 2024-10-03 at 18 09 41" src="https://github.com/user-attachments/assets/da1f99c3-a85d-484e-96d3-0ed951367ef8">
Most of the time is spent in `OpType::port_kind`, which takes so long since it copies a lot of stuff:
```rust
pub fn port_kind(&self, port: impl Into<Port>) -> Option<EdgeKind> {
    let signature = self.dataflow_signature().unwrap_or_default();      // <-- this deep copies he signature
    let port: Port = port.into();
    let dir = port.direction();
    let port_count = signature.port_count(dir);

    // Dataflow ports
    if port.index() < port_count {
        return signature.port_type(port).cloned().map(EdgeKind::Value);    // <-- this deep copies the relevant type again
    }

    // ....
}
```

Looking at the import, it's similar:
<img width="1688" alt="Screenshot 2024-10-03 at 18 05 46" src="https://github.com/user-attachments/assets/98e3519a-0250-41fe-ad2c-a4f841323971">

The first bump is mostly due to splitting strings into the extension and the name part. The remaining two bumps come from `OpType::value_port_count`, which spends most of its time copying data.
```rust
pub fn value_port_count(&self, dir: portgraph::Direction) -> usize {
     self.dataflow_signature() // <-- copy all of the signature just to count ports
         .map(|sig| sig.port_count(dir))
         .unwrap_or(0)
}
```

The `hugr-core` data structures are designed in a way that you don't have to worry about ownership, which is not a bad thing. But there is a lot of deep copying, allocating and freeing going on, even just to look at something. There is a massive performance improvement to be had by being more careful here, not just for import or export, but for everything the compiler does that ever looks at types.

The most immediate and pragmatic approach that comes to my mind is to slap a lot of `Arc`s on as many types as possible, ideally in a way that minimally affects the API. (There are other approaches but they are too radical for now). What do you think?

PR #1557, opened 2024-10-07
feat: Binary serialisation format for hugr-model based on capnproto.

zrho:
A binary serialisation format for `hugr-model` based on capnproto.

- Export from `hugr-core` to `hugr-model` now deduplicates terms on the fly.
- Removed types on ports and instead used the signature. This allows for more deduplication and also provides a natural place for extension set annotations which was missing before.
- Added the `NodeId` of the node that defines a parameter to the indexed variant of a `LocalRef` so that deduplication does not accidentally merge two local variables that refer to different parameters.

Serializing the same 2000 2-qubit gate + 1000 1-qubit gate circuit by going from `core` via the `model` takes 5.5682 ms per roundtrip on Apple M3. Meanwhile the JSON roundtrip baseline is 10.760 ms. A significant portion of the time is spent on needless copies of signatures since `export` and `import` have to interact with the `hugr-core` API (#1551), so there is still a lot of potential to make this faster yet.


zrho:
The filesize for the capnp serialization is also better than the JSON baseline: the 2000 2-qubit gate + 1000 1-qubit gate circuit serializes to 295kb with binary serialization in contrast to 695K for JSON. This is not just due to being a binary format, but also due to term deduplication. Note that the JSON serialization also does a form of this by serializing the `prelude.qubit` type simply as `"Q"`. But whereas this only works for qubits for the JSON serializer, the term deduplication performed by the export to the model works for all types.

The file size can be driven down further and the speed increased once we have custom operation declarations: Then nodes that use a custom operation don't have to repeat the operation's full name every time, but can simply refer to the declaring node by index.

This form of deduplication is not only useful to decrease file size; that could similarly be achieved by just compressing the output. When reading a deduplicated file back in, we can use the sharing to also get an efficient in-memory representation, and to avoid redundantly validating the same term over and over. At this point, we do not make good use of this yet since the `hugr-core` data structures do not allow any sharing.

hugr-model/capnp/hugr-v0.capnp
```diff
@@ -0,0 +1,175 @@
...
+    outputs @2 :List(LinkRef);
+    params @3 :List(UInt32);
+    regions @4 :List(UInt32);
+    meta @5 :List(MetaItem);
+    signature @6 :UInt32;
+}
+
+struct Operation {
+    union {
+        invalid @0 :Void;
```
croyzor, line 22: What's the `invalid` for?

zrho: See the docs on Term::Invalid. It serves two purposes: - being able to represent modules whose nodes have non-contiguous indices without needing to compact
- a placeholder operation for when you have partially constructed a module. This was useful to break a cycle in export after adding the defining node index in LocalRef.

hugr-model/src/v0/mod.rs
```diff
@@ -323,15 +347,15 @@ pub struct Region<'a> {
...
     /// The target ports of the region.
-    pub targets: &'a [Port<'a>],
+    pub targets: &'a [LinkRef<'a>],
     /// The nodes in the region. The order of the nodes is not significant.
     pub children: &'a [NodeId],
     /// The metadata attached to the region.
     pub meta: &'a [MetaItem<'a>],
     /// The signature of the region.
-    pub signature: TermId,
+    pub signature: Option<TermId>,
```
ss2165, line 358: what does None mean? (say in docstring)

hugr-core/src/export.rs
```diff
@@ -1,18 +1,23 @@
...
+        TypeBase, TypeEnum,
     },
-    Direction, Hugr, HugrView, IncomingPort, Node, Port, PortIndex,
+    Direction, Hugr, HugrView, IncomingPort, Node, Port,
 };
-use bumpalo::{collections::Vec as BumpVec, Bump};
+use bumpalo::{collections::String as BumpString, collections::Vec as BumpVec, Bump};
+use fxhash::FxHashMap;
 use hugr_model::v0::{self as model};
 use indexmap::IndexSet;
```
ss2165, line 17: is this still used?

hugr-core/src/export.rs
```diff
@@ -344,20 +327,101 @@ impl<'a> Context<'a> {
...
             regions,
             meta: &[], // TODO: Export metadata
             signature,
+        };
+
+        node_id
+    }
+
+    /// Export the signature of a `DataflowBlock`. Here we can't use `OpType::dataflow_signature`
+    /// like for the other nodes since the ports are control flow ports.
```
ss2165, line 372: does this do what you need: https://github.com/CQCL/hugr/blob/a725a29a13648ffa262b53e1ce8dd605711c0540/hugr-core/src/ops/controlflow.rs#L165

hugr-core/src/export.rs
```diff
@@ -398,8 +463,20 @@ impl<'a> Context<'a> {
             region_children.push(self.export_node(child));
         }
 
-        // TODO: We can determine the type of the region
-        let signature = self.module.insert_term(model::Term::Wildcard);
+        let signature = {
+            let inputs = self.export_type_row(&input_op.types);
+            let outputs = self.export_type_row(&output_op.types);
+            // TODO: What about the extensions?
```
ss2165, line 469: extension inference puts the inferred extensions on the parent node signature?

ss2165 (commented):
looks good, minor queries

hugr-model/capnp/hugr-v0.capnp
```diff
@@ -0,0 +1,175 @@
...
+}
+
+struct Node {
+    operation @0 :Operation;
+    inputs @1 :List(LinkRef);
+    outputs @2 :List(LinkRef);
+    params @3 :List(UInt32);
+    regions @4 :List(UInt32);
+    meta @5 :List(MetaItem);
+    signature @6 :UInt32;
```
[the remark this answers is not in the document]
zrho: As far as I can tell, capnproto does not have aliases, unfortunately. Introducing a new a struct would also introduce an indirection via a pointer so affect filesize and performance.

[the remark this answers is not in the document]
ss2165: see "using" here, I used it in my PR for this purpose

hugr-core/src/export.rs
zrho: Not for here: this constructs the outer signature of a dataflow block, not the inner. We construct the inner signature from the input and output nodes in the dfg region export.

hugr-core/src/export.rs
zrho: All of them are used in that file, yes. Maybe I misunderstood?

hugr-core/src/export.rs
zrho: Do they always agree? What about an operation that runs a subgraph somewhere else which has different capabilities? In that case the extensions of the dfg region would be different from the extensions of the parent node.

hugr-core/src/export.rs
ss2165: sorry badly phased. I see you replaced one use of `IndexSet` with `FxIndexSet`, do we still need the `indexmap` dependency. These model related PRs are pulling in a lot of new dependencies, which I am a bit wary of.

hugr-core/src/export.rs
ss2165: Currently they always agree, the semantics of "elsewhere" for extensions aren't really well defined (yet).

hugr/Cargo.toml
```diff
@@ -34,7 +36,11 @@ rstest = { workspace = true }
 lazy_static = { workspace = true }
 criterion = { workspace = true, features = ["html_reports"] }
 serde_json = { workspace = true }
+bumpalo = { workspace = true, features = ["collections"] }
 
 [[bench]]
 name = "bench_main"
 harness = false
+
+[profile.bench]
```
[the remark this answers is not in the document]
zrho: That includes debug symbols into the benchmark binaries, which is very useful for profiling. I could take it out and put it in locally when I need it for flamegraphs, or leave it in.

hugr-core/src/export.rs
zrho: `FxIndexSet` is a type alias: `type FxIndexSet<T> = IndexSet<T, fxhash::FxBuildHasher>;`

hugr/Cargo.toml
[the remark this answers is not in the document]
zrho: It isn't debug mode (as opposed to release mode), but whether to include debug information (like function names) in the compiled binary. It might have a small impact on performance due to locality since the binary is consequently larger.

hugr-core/src/export.rs
ss2165: 🤦

hugr-core/Cargo.toml
```diff
@@ -20,6 +20,7 @@ workspace = true
 extension_inference = []
 declarative = ["serde_yaml"]
 model_unstable = ["hugr-model"]
+default = ["model_unstable"]
```
ss2165, line 23: I don't think we intend to make this default?

zrho: Ah good catch, that was just so that LSP would work for the file while I updated it.

hugr-core/src/export.rs
zrho: This should be addressed now.

ss2165 (approved):
🚀

PR #1572, opened 2024-10-11
chore(ci): Bundle crates under the same version_group

aborgna-q:
This should solve the diamond problem we had where `hugr-core` get's a breaking release but `hugr-passes` only publishes a minor, so cargo automatically updates the latter and we get diverging `hugr-core` versions on the dependency tree (so the public APIs break).

`version_group` defines a grouping of versions, where packages **that had changes** will all be bumped to the same version.
https://release-plz.ieni.dev/docs/config#the-version_group-field

In the previous example, a mayor bump to `hugr-core` will cause a mayor release for both `hugr-passes` and `hugr`.
On the other hand, a change to `hugr-passes` will **not** trigger a bump on `hugr-core`.

An alternative (non-automated) solution to this problem can be done via the `semver-trick` \[0,1\], but that's a one-off manual solution more useful to avoid problems for external libraries.

- \[0\]: https://doc.rust-lang.org/cargo/reference/resolver.html#version-incompatibility-hazards
- \[1\]: https://github.com/dtolnay/semver-trick


zrho (approved):
I think that makes sense.

PR #1573, opened 2024-10-11
feat!: define wrappers around package that point to internals

### LINKED ISSUE #1561: py: Add a "Pointed package" protocol
Guppy can compile functions into hugr packages, which have a function identified as an entry point.
This is similar to tket2 circuits, which point to a region in a hugr.

We should unify this notion in a Protocol in `hugr-py`, so both libraries can interact without depending on each other.
```python
class Proto(Protocol):
    def package(self) -> Package: ...
    def hugr_index(self) -> int: ...
    def root(self) -> Node: ...
```

ss2165:
Closes #1561

So far just top level things, can add inner ones (cfg, blocks, etc.) later.

BREAKING CHANGE: `Package` moved to new `hugr.package` module

### DIFF
--- a/hugr-py/src/hugr/package.py
+++ b/hugr-py/src/hugr/package.py
@@ -1,13 +1,17 @@
 """HUGR package and pointed package interfaces."""
 
+from __future__ import annotations
+
 from dataclasses import dataclass, field
-from typing import Generic, TypeVar, cast
+from typing import TYPE_CHECKING, Generic, TypeVar, cast
 
 import hugr._serialization.extension as ext_s
-from hugr.ext import Extension
-from hugr.hugr.base import Hugr
-from hugr.hugr.node_port import Node
 from hugr.ops import FuncDecl, FuncDefn, Op
+
+if TYPE_CHECKING:
+    from hugr.ext import Extension
+    from hugr.hugr.base import Hugr
+    from hugr.hugr.node_port import Node
 
 __all__ = [
     "Package",
@@ -20,7 +24,7 @@ __all__ = [
 ]
 
 
-@dataclass
+@dataclass(frozen=True)
 class Package:
     """A package of HUGR modules and extensions.
 
@@ -43,17 +47,24 @@ class Package:
         return self._to_serial().model_dump_json()
 
 
-@dataclass
+@dataclass(frozen=True)
 class PackagePointer:
     """Classes that point to packages and their inner contents."""
 
+    #: Package pointed to.
     package: Package
 
 
-@dataclass
+@dataclass(frozen=True)
 class ModulePointer(PackagePointer):
-    """Pointer to a module in a package."""
+    """Pointer to a module in a package.
 
+    Args:
+        package: Package pointed to.
+        module_index: Index of the module in the package.
+    """
+
+    #: Index of the module in the package.
     module_index: int
 
     @property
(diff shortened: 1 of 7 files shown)


hugr-py/src/hugr/package.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def module(self) -> Hugr:
+        """Hugr definition of the module."""
+        return self.package.modules[self.module_index]
+
+    def to_executable_package(self) -> "ExecutablePackage":
+        """Create an executable package from a module containing a main function.
+
+        Raises:
+            StopIteration: If the module does not contain a main function.
```
aborgna-q, line 71: Rethrow a more appropriate exception, with a custom message.
Either a custom one, or `ValueError`

hugr-py/src/hugr/package.py
```diff
@@ -0,0 +1,130 @@
...
+class ModulePointer(PackagePointer):
+    """Pointer to a module in a package."""
+
+    module_index: int
+
+    def module(self) -> Hugr:
+        """Hugr definition of the module."""
+        return self.package.modules[self.module_index]
+
+    def to_executable_package(self) -> "ExecutablePackage":
```
aborgna-q, line 67:
```python
from __future__ import annotations
```
and remove the quotes

hugr-py/src/hugr/package.py
```diff
@@ -0,0 +1,130 @@
...
+
+@dataclass
+class PackagePointer:
+    """Classes that point to packages and their inner contents."""
+
+    package: Package
+
+    def get_package(self) -> Package:
+        """Get the package pointed to."""
+        return self.package
```
mark-koch, line 54: What is the purpose of this method?

ss2165: good point!

ss2165: (vestigial from original issue which proposed protocols)

hugr-py/src/hugr/package.py
```diff
@@ -0,0 +1,130 @@
...
+
+@dataclass
+class ModulePointer(PackagePointer):
+    """Pointer to a module in a package."""
+
+    module_index: int
+
+    def module(self) -> Hugr:
+        """Hugr definition of the module."""
+        return self.package.modules[self.module_index]
```
mark-koch, line 65: This could be a property (same for all other methods in this PR)

hugr-py/src/hugr/package.py
```diff
@@ -0,0 +1,130 @@
...
+class FuncDefnPointer(NodePointer[FuncDefn]):
+    """Pointer to a function definition in a module."""
+
+    def func_defn(self) -> FuncDefn:
+        """Function definition."""
+        return self.node_op()
+
+
+@dataclass
+class ExecutablePackage(FuncDefnPointer):
```
aborgna-q, line 127:
```suggestion
class ExecutablePackage(FuncDefnPointer):
        """PackagePointer with a defined entrypoint node."""
        
```

hugr-py/src/hugr/package.py
aborgna-q, line 1: It'd be useful to add 
```py
    Args:
        package: ...
        module: ...
        ...
```
sections to the dataclass docs.
Python doesn't make too clear what attributes are getting inherited in the constructor.

aborgna-q (approved):
Nice.

The followup question will now be how do we serialize these :)

PR #1587, opened 2024-10-16
feat: Add `Package` definition on `hugr-core`

### LINKED ISSUE #1530: Move `Package` out of `hugr-cli`
`Package` is a core type that should be exported from the core (at least until `hugr-model` is created).
The current state of things requires tket2 to depend on `hugr-cli` for the sole reason of using that type.

aborgna-q:
Moves the definition of packages out of `hugr-cli`. (We leave a re-export there to avoid breaking the API, it should be removed on the next breaking release).

Adds some helper functions to load and validate the package.

Closes #1530 

drive-by: Replace `hugr-core` dependency on `hugr-cli` by `hugr`. There's no need to access the internals there.

### DIFF
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -92,6 +92,9 @@ impl ExtensionRegistry {
     /// If extension IDs match, the extension with the higher version is kept.
     /// If versions match, the original extension is kept.
     /// Returns a reference to the registered extension if successful.
+    ///
+    /// Avoids cloning the extension unless required. For a reference version see
+    /// [`ExtensionRegistry::register_updated_ref`].
     pub fn register_updated(
         &mut self,
         extension: Extension,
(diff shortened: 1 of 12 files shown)


hugr-core/src/extension.rs
```diff
@@ -107,6 +110,30 @@ impl ExtensionRegistry {
...
+    pub fn register_updated_ref(
+        &mut self,
+        extension: &Extension,
+    ) -> Result<&Extension, ExtensionRegistryError> {
+        match self.0.entry(extension.name().clone()) {
+            btree_map::Entry::Occupied(mut prev) => {
+                if prev.get().version() < extension.version() {
+                    *prev.get_mut() = extension.clone();
+                }
+                Ok(prev.into_mut())
```
ss2165, line 131: doesn't appear to be tested

ss2165 (approved):
📦

aborgna-q:
I added some `to_json` methods too, to keep things symmetric.

Note that we are not enforcing that packages must only contain module-rooted HUGRs.
I'll make a separate PR with that, as we need to fix `hugr-cli` once that starts getting checked.

Issue #1388, opened 2024-08-01
Update spec regarding "Package" behaviour


ss2165:
follow up from #1358

Specification needs to be updated to describe what a package is and how it can be well-formed (valid)

aborgna-q:
Regarding the packed extensions, we should define that `prelude` and the std extensions do not need to be included.

The std extensions are [listed on the spec](https://github.com/CQCL/hugr/blob/main/specification/hugr.md#standard-library). (A bit late for this, but it'd have been nice if the extension names started with `std.`)

PR #1589, opened 2024-10-17
feat: Ensure packages always have modules at the root

aborgna-q:
Followup to #1587.

Ensures that rust `Package`s only contain module-rooted hugrs. Returns errors at construction times if the condition is not met.
This required fixing `hugr-cli`, as it should be able to load both packages and arbitrary hugrs.
Added `Package::from_hugr{,s}` methods that try to wrap the hugrs if possible. We'll need these for tket2.

Packages are not on the spec yet, their description should include this restriction. See #1388.

This PR does only modify the (unpublished) API introduced in #1587, so I'm not marking it as breaking.

### DIFF
--- a/hugr-cli/src/lib.rs
+++ b/hugr-cli/src/lib.rs
@@ -101,7 +101,7 @@ impl PackageOrHugr {
         reg: &mut ExtensionRegistry,
     ) -> Result<(), PackageValidationError> {
         match self {
-            PackageOrHugr::Package(pkg) => pkg.validate(reg),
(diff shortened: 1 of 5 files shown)


aborgna-q:
#1587 made some breaking changes. I'll try to fix them in a following PR, and replace them with `#[deprecated]`

aborgna-q:
I renamed the `update` method. Note that this is a breaking release, but https://github.com/CQCL/hugr/pull/1591 will take care of deprecating things.

PR #1590, opened 2024-10-17
docs: Add help docs about `semver-checks`

aborgna-q:
Also makes both `release-plz` and `release-please` mark their PRs with `release` labels.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -66,6 +66,23 @@ cargo +nightly miri test
 
 Run `just` to see all available commands.
 
+### 💥 API-breaking changes
+
+Any breaking change in the public Rust APIs will cause the next release to be a
+major version bump. You can check the next release version [draft release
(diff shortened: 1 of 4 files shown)


DEVELOPMENT.md
```diff
@@ -66,6 +66,23 @@ cargo +nightly miri test
...
 Run `just` to see all available commands.
 
+### 💥 API-breaking changes
+
+Any breaking change in the public Rust APIs will cause the next release to be a
+major version bump. You can check the next release version [draft release
+PR](https://github.com/CQCL/hugr/pulls?q=is%3Aopen+is%3Apr+label%3Arelease) on
+github.
+
+If all the changes before your PR were non-breaking, you can use `cargo
```
croyzor, line 76: Why would i need to do this if all the changes were non breaking?

croyzor (approved):

DEVELOPMENT.md
aborgna-q: Updated the docs to compare against the current `main` instead.
Also added a comment about CI, assuming #1598 gets merged.

PR #1591, opened 2024-10-17
chore: Unbreak release

aborgna-q:
Replaces the breaking changes introduced in #1587 with non-breaking deprecations.
We should be able to remove these on the next breaking release.


ss2165 (approved):
🫠

PR #1593, opened 2024-10-18
feat: Add filter_edge_kind to PortIterator

lmondada:
I need to filter for `EdgeKind::Value(_)` (and possibly `EdgeKind::Const(_)`) so thought of making a more general filter available.


aborgna-q (approved):
👍

PR #1597, opened 2024-10-18
feat: make errors more readable with Display impls

ss2165:
Particularly useful for cli validation errors.

- Blanket Display for OpType that uses name, can be made more specific in future.
- user derive_more::Display in more places


cqc-alec (approved):
Thanks!

PR #1603, opened 2024-10-21
WIP use dataflow analysis for constant folding

acl-cqc:
Remnants of #1157 after the "framework" of #1476 is pulled out. I'm no longer sure there's really very much here left and even `ValueHandle` may want to change significantly now we call `PartialValue::new_variant` from outside rather than having `as_sum` and the framework handles `LoadConstant`.

Main TODO is to transform the graph! Specifically
- [ ] Cut wires where we have constants
- [ ] Remove then-unused nodes
- [ ] Some kind of framework/control for ops that are actually stateful i.e. we can't just remove them
- [ ] When Conditional node's predicate is a constant, turn into DFG
- [ ] Can we remove unreachable cases from Conditionals with >1 reachable case remaining? 
    * only one case => DFG, but the tricky part here is traversing backwards to find Tag nodes and removing them or updating their types; or else unwrap/panic (ugh!)
- [ ] For TailLoops which never iterate, also turn into DFG, similarly
- [ ] And, CFGs! (Perhaps that can wait)


hugr-passes/src/const_fold.rs
```diff
@@ -1,220 +1,270 @@
-//! Constant folding routines.
+#![warn(missing_docs)]
+//! An (example) use of the [super::dataflow](dataflow-analysis framework)
```
croyzor, line 2: Is this not also now our canonical constant folder?

hugr-passes/src/const_fold.rs
```diff
@@ -1,220 +1,287 @@
...
+            .into_iter()
+            .flat_map(|n| hugr.node_inputs(n).map(move |ip| (n, ip)))
+            .filter(|(n, ip)| {
+                *n != hugr.root()
+                    && matches!(hugr.get_optype(*n).port_kind(*ip), Some(EdgeKind::Value(_)))
+            })
+            // Note we COULD filter out (avoid breaking) wires from other nodes that we are keeping.
+            // This would insert fewer constants, but potentially expose less parallelism.
+            .filter_map(|(n, ip)| {
+                let (src, outp) = hugr.single_linked_output(n, ip).unwrap();
```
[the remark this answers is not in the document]
acl-cqc: Perhaps this would be needed if you run on a subgraph-view that has an incoming nonlocal edge? I'd like to have a testcase/bugreport though

hugr-passes/src/const_fold.rs
```diff
@@ -1,220 +1,287 @@
...
+            .zip(ins.iter())
+            .filter_map(|((i, ty), pv)| {
+                Some((IncomingPort::from(i), pv.clone().try_into_value(ty).ok()?))
+            })
+            .collect::<Vec<_>>();
+        for (p, v) in op.constant_fold(&known_ins).unwrap_or_default() {
+            outs[p.index()] =
+                partial_from_const(self, ConstLocation::Field(p.index(), &node.into()), &v);
+        }
+    }
```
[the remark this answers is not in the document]
acl-cqc: Believe this is handled by changes to the framework such that `interpret_leaf_op` is not called if any inputs are Bottom

hugr-passes/src/const_fold.rs
```diff
@@ -1,221 +1,294 @@
...
+        let h = results.hugr();
+        q.push_back(h.root());
+        while let Some(n) = q.pop_front() {
+            if !needed.insert(n) {
+                continue;
+            };
+
+            if h.get_optype(n).is_cfg() {
+                for bb in h.children(n) {
+                    //if results.bb_reachable(bb).unwrap() { // no, we'd need to patch up predicates
```
croyzor, line 164: Remove this comment (this is run before we create results anyway?)

croyzor: but i don't really understand the argument here -- wouldn't the commented code be better?

hugr-passes/src/const_fold.rs
```diff
@@ -1,221 +1,294 @@
...
+        let sig = op.signature();
+        let known_ins = sig
+            .input_types()
+            .iter()
+            .enumerate()
+            .zip(ins.iter())
+            .filter_map(|((i, ty), pv)| {
+                pv.clone()
+                    .try_into_concrete(ty)
+                    .map_or(None, |v| Some((IncomingPort::from(i), v)))
```
croyzor, line 283:
```suggestion
                    .map(|v| (IncomingPort::from(i), v))
```

PR #1605, opened 2024-10-22
feat: Operation and constructor declarations in `hugr-model`

zrho:
This PR adds the ability to declare custom operations and constructors (so static types, runtime types, constraints, etc.) to `hugr-model`. In the case of operations this is used for deduplication when exporting.


hugr-core/src/export.rs
```diff
@@ -392,6 +406,58 @@ impl<'a> Context<'a> {
         node_id
     }
 
+    /// Export an `OpDef` as an operation declaration.
+    ///
+    /// Operations that allow a declarative form are exported as a reference to
```
ss2165, line 411: and those that don't allow this?

hugr-core/src/export.rs
```diff
@@ -392,6 +406,58 @@ impl<'a> Context<'a> {
...
+                    outputs: &[],
+                    params: &[],
+                    regions: &[],
+                    meta: &[], // TODO: Metadata
+                    signature: None,
+                }))
+            }
+        };
+
+        let decl = self.with_local_scope(node, |this| {
```
ss2165, line 444: is this going to overwrite the polyfunctype even if the node is already occupied?

hugr-model/tests/fixtures/model-decl-exts.edn
```diff
@@ -0,0 +1,13 @@
...
+(declare-ctr array.Array
+  (param ?t type)
+  (param ?n nat)
+  type
+  (meta docs.description "Fixed size array."))
+
+(declare-operation array.Init
+  (param ?t type)
+  (param ?n nat)
+  (fn [?t] [(array.Array ?t ?n)] (ext array))
```
ss2165, line 12: nice! do we want to think about some syntax for grouping declarations under one extension?

```
(ext array
    (declare-ctr Array
         ...
```

hugr-core/src/export.rs
zrho: For now we can fall back to just referencing the operation by its name as before. For each of the `SignatureFunc` variants:

- `PolyFuncType` - This is the only case where we export a declarative op specification at the moment.
- `CustomValidator` - This has a `PolyFuncType` and a custom validator. We can't serialize a custom validator, and so this currently can not be expressed as an op declaration. The idea of the constraint system is to give the custom validator a name so it can be referred to. Once we use constraints for validation, we can handle this case. While we could export by dropping the custom validator, that would not be an accurate description of the operation's type.
- `MissingValidateFunc` - This is like `CustomValidator` except that we do not have the validator code. This case would also be covered by the constraint system, when the operation's type is constrained but we do not have the code that solves the constraint.
- `CustomFunc` - Here we have even less information than for `CustomValidator`; the function that computes the `PolyFuncTypeRV` again can not be serialized. It appears as if the static parameter list must still be known, in which case constraints would allow us to merge this with the `CustomValidator` case.
- `MissingComputeFunc` - Here we know literally nothing about the type of the operation, including what its static parameters are. When we add the `import` mode in addition to `declare` and `define` (i.e. like a declaration except that we only know the name and nothing about its type), we could handle this.

hugr-core/src/export.rs
zrho: I don't understand what you are referring to; can you rephrase that?

hugr-model/tests/fixtures/model-decl-exts.edn
zrho: I thought about it but don't see that much would be gained by it. We'd then have to deal with merging these groups when there's two or more for the same extension in a module. This would come up especially when dealing with linking. This isn't hard but a bit of extra complexity.

The text syntax currently also uses qualified names to avoid collisions with future keywords: `array.Array` is never a keyword since keywords are not qualified. But `array` or `Array` alone might become a keyword in the future. For local variables and for link names we solve this via sigils (`?t` and `%x`). If we'd allow the names for globals to occur unqualified, we'd either need to sigil them or accept potential future collisions.

There might be some minor space savings by not having to repeat the name of the extension for every operation and constructor that it defines (and these names do get quite long: `arithmetic.float.types`). However, I'm not sure if that's really worth it.

hugr-core/src/export.rs
ss2165: have you looked at how these are handled in the current serialization? We use a boolean flag to indicate that some code accompanies this time (either to compute it or to validate it), which allows us to transmit the critical information if we have it while still tracking the original intent. I understand you want to capture all of this in constraints, but it is worth thinking about whether something can be added now that captures this. This would allow this system to replace the existing extension serialization much quicker.

hugr-core/src/export.rs
ss2165: IIUC this function looks up whether an opdef with the same name already exists in the map. On line 455 it appears the `operation` field of that node is overwritten by the declaration whether or not it was already occupied by a previous declaration. Is this intended behaviour?

hugr-model/tests/fixtures/model-decl-exts.edn
ss2165: ok, let's at least aim for some convention for grouping one extension together. If not it becomes quite unwieldy when declaring multiple in one module.

hugr-core/src/export.rs
zrho: If the name already exists in the map, the node for that name is returned immediately. If it does not exist, we create a placeholder entry first and then overwrite that placeholder entry on line 455. No previous declaration will be overwritten.

hugr-core/src/export.rs
zrho: I don't see an obvious good solution to doing this with `SignatureFunc` in the current form without implementing some features that we're then likely to remove again (and there's nothing more permanent than a temporary solution). So I propose to postpone the other cases to a future PR.

hugr-model/tests/fixtures/model-decl-exts.edn
zrho: When you write extension declarations by hand, it's likely that you'd use a module per extension. When generating documentation or code, the tool can do the grouping automatically. What situation do you have in mind where some more explicit grouping would help?

ss2165: "module per extension" as the convention is good enough for me

zrho:
CI indicates a breaking change, but the change is confined to the model which is considered unstable.

ss2165 (approved):

PR #1624, opened 2024-10-31
feat: Emulate `TypeBound`s on parameters via constraints.

zrho:
This PR translates some `TypeBound`s in `hugr-core` to the `nonlinear` constraint in `hugr-model`. This translation only occurs on parameters that take a runtime type directly.

As a driveby change before the model stabilises, this PR also moves constraints out of the parameter lists into their own list. In its previous form this could have led to confusions about which parameter a local variable index refers to when a constraint is situated between two parameters in the list. We also remove constraints from aliases for now.

Closes https://github.com/CQCL/hugr/issues/1637.

### DIFF
--- a/hugr-core/src/export.rs
+++ b/hugr-core/src/export.rs
@@ -693,7 +693,11 @@ impl<'a> Context<'a> {
         for (i, param) in t.params().iter().enumerate() {
             let name = self.bump.alloc_str(&i.to_string());
             let r#type = self.export_type_param(param, Some(model::LocalRef::Index(scope, i as _)));
-            let param = model::Param::Implicit { name, r#type };
+            let param = model::Param {
+                name,
+                r#type,
+                sort: model::ParamSort::Implicit,
+            };
             params.push(param)
         }
 
--- a/hugr-core/src/export.rs
+++ b/hugr-core/src/export.rs
@@ -269,7 +269,6 @@ impl<'a> Context<'a> {
                 let decl = this.bump.alloc(model::AliasDecl {
                     name: &alias.name,
                     params: &[],
-                    constraints: &[],
                     r#type,
                 });
                 model::Operation::DeclareAlias { decl }
@@ -282,7 +281,6 @@ impl<'a> Context<'a> {
                 let decl = this.bump.alloc(model::AliasDecl {
                     name: &alias.name,
                     params: &[],
-                    constraints: &[],
                     r#type,
                 });
                 model::Operation::DefineAlias { decl, value }
--- a/hugr-core/src/export.rs
+++ b/hugr-core/src/export.rs
@@ -7,7 +7,7 @@ use crate::{
         type_param::{TypeArgVariable, TypeParam},
         type_row::TypeRowBase,
         CustomType, FuncTypeBase, MaybeRV, PolyFuncTypeBase, RowVariable, SumType, TypeArg,
-        TypeBase, TypeEnum,
+        TypeBase, TypeBound, TypeEnum,
     },
     Direction, Hugr, HugrView, IncomingPort, Node, Port,
 };
@@ -46,6 +46,8 @@ struct Context<'a> {
     term_map: FxHashMap<model::Term<'a>, model::TermId>,
     /// The current scope for local variables.
     local_scope: Option<model::NodeId>,
+    /// Constraints to be added to the local scope.
+    local_constraints: Vec<model::TermId>,
     /// Mapping from extension operations to their declarations.
     decl_operations: FxHashMap<(ExtensionId, OpName), model::NodeId>,
 }
@@ -63,6 +65,7 @@ impl<'a> Context<'a> {
             term_map: FxHashMap::default(),
             local_scope: None,
             decl_operations: FxHashMap::default(),
+            local_constraints: Vec::new(),
         }
     }
 
@@ -173,9 +176,11 @@ impl<'a> Context<'a> {
     }
 
     fn with_local_scope<T>(&mut self, node: model::NodeId, f: impl FnOnce(&mut Self) -> T) -> T {
-        let old_scope = self.local_scope.replace(node);
+        let prev_local_scope = self.local_scope.replace(node);
+        let prev_local_constraints = std::mem::take(&mut self.local_constraints);
         let result = f(self);
-        self.local_scope = old_scope;
+        self.local_scope = prev_local_scope;
+        self.local_constraints = prev_local_constraints;
         result
     }
 
@@ -676,13 +681,22 @@ impl<'a> Context<'a> {
         t: &PolyFuncTypeBase<RV>,
     ) -> (&'a [model::Param<'a>], model::TermId) {
         let mut params = BumpVec::with_capacity_in(t.params().len(), self.bump);
+        let scope = self
+            .local_scope
+            .expect("exporting poly func type outside of local scope");
 
         for (i, param) in t.params().iter().enumerate() {
             let name = self.bump.alloc_str(&i.to_string());
-            let r#type = self.export_type_param(param);
+            let r#type = self.export_type_param(param, Some(model::LocalRef::Index(scope, i as _)));
             let param = model::Param::Implicit { name, r#type };
             params.push(param)
         }
+
+        params.extend(
+            self.local_constraints
+                .drain(..)
+                .map(|constraint| model::Param::Constraint { constraint }),
+        );
 
         let body = self.export_func_type(t.body());
 
@@ -794,20 +808,35 @@ impl<'a> Context<'a> {
         self.make_term(model::Term::List { items, tail: None })
     }
 
-    pub fn export_type_param(&mut self, t: &TypeParam) -> model::TermId {
+    pub fn export_type_param(
+        &mut self,
+        t: &TypeParam,
+        var: Option<model::LocalRef<'static>>,
+    ) -> model::TermId {
         match t {
             // This ignores the type bound for now.
-            TypeParam::Type { .. } => self.make_term(model::Term::Type),
+            TypeParam::Type { b } => {
+                if let (Some(var), TypeBound::Copyable) = (var, b) {
+                    let term = self.make_term(model::Term::Var(var));
+                    let copy = self.make_term(model::Term::CopyConstraint { term });
+                    let discard = self.make_term(model::Term::DiscardConstraint { term });
+                    self.local_constraints.extend([copy, discard]);
+                }
+
+                self.make_term(model::Term::Type)
+            }
             // This ignores the type bound for now.
             TypeParam::BoundedNat { .. } => self.make_term(model::Term::NatType),
             TypeParam::String => self.make_term(model::Term::StrType),
             TypeParam::List { param } => {
-                let item_type = self.export_type_param(param);
+                let item_type = self.export_type_param(param, None);
                 self.make_term(model::Term::ListType { item_type })
             }
             TypeParam::Tuple { params } => {
                 let items = self.bump.alloc_slice_fill_iter(
-                    params.iter().map(|param| self.export_type_param(param)),
+                    params
+                        .iter()
+                        .map(|param| self.export_type_param(param, None)),
                 );
                 let types = self.make_term(model::Term::List { items, tail: None });
                 self.make_term(model::Term::ApplyFull {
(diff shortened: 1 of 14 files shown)


hugr-core/src/import.rs
```diff
@@ -938,7 +956,7 @@ impl<'a> Context<'a> {
             model::Term::FuncType { .. } => Err(error_unsupported!("`(fn ...)` as `TypeParam`")),
 
             model::Term::ListType { item_type } => {
-                let param = Box::new(self.import_type_param(*item_type)?);
+                let param = Box::new(self.import_type_param(*item_type, TypeBound::Any)?);
```
ss2165, line 959: Not sure why this is any (also not covered by tests), is this a todo waiting for a list constraint?

hugr-model/tests/fixtures/model-constraints.edn
```diff
@@ -0,0 +1,15 @@
+(hugr 0)
+
+(declare-func array.replicate
+  (forall ?t type)
+  (forall ?n nat)
+  (where (copy ?t))
+  (where (discard ?t))
```
ss2165, line 7: this looks like it could get quite verbose. Will we need to specify things that are `copy` but not `discard`?  Even if we do, I feel that ought to be the more verbose specification. Ideally I guess I want to write:
1. Nothing for things that are copyable and discardable
2. "linear" or "not copy" for things that are such
3. "affine" when I can't copy but can discard.

Is this just a handy syntax change or are we running in to subtyping problems again?

hugr-core/src/export.rs
```diff
@@ -46,6 +46,8 @@ struct Context<'a> {
     term_map: FxHashMap<model::Term<'a>, model::TermId>,
     /// The current scope for local variables.
     local_scope: Option<model::NodeId>,
+    /// Constraints to be added to the local scope.
```
acl-cqc, line 49: Is `local scope` == `[Self::local_scope]`? Do we actually plan to modify the local scope by adding these? Or are these "Constraints to be added to those from the local scope" (i.e., extra constraints from somewhere else)? "Constraints in addition to those from the local scope" would be clearer still that we don't intend to mutate some list.

hugr-model/src/v0/mod.rs
```diff
@@ -397,6 +397,8 @@ pub struct FuncDecl<'a> {
     pub name: &'a str,
     /// The static parameters of the function.
     pub params: &'a [Param<'a>],
+    /// The constraints on the static parameters.
+    pub constraints: &'a [TermId],
```
acl-cqc, line 401: Good point (in the PR description) about separating these out! :)

hugr-model/src/v0/mod.rs
```diff
@@ -408,6 +410,8 @@ pub struct AliasDecl<'a> {
     pub name: &'a str,
     /// The static parameters of the alias.
     pub params: &'a [Param<'a>],
+    /// The constraints on the static parameters.
+    pub constraints: &'a [TermId],
```
acl-cqc, line 414: So the alias also introduces more constraints. I'd say that might make it slightly more than an alias; I can't introduce constraints just anywhere (without an alias), can I?

And when we have Terms in hugr-core, Alias's are likely to disappear after import (as we can get sharing via TermId, right?) - are we sure we want/need this?

Or maybe the idea is that the constraints here are copied(+substituted) from/according-to the RHS of the alias (they are the constraints that must apply anyway, rather than the Alias being able to specify new ones to introduce). If so, let's be clear in the comment....

hugr-model/src/v0/mod.rs
```diff
@@ -662,40 +670,42 @@ pub enum Term<'a> {
...
 }
 
 /// A parameter to a function or alias.
 ///
 /// Parameter names must be unique within a parameter list.
 /// Implicit and explicit parameters share a namespace.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub enum Param<'a> {
-    /// An implicit parameter that should be inferred, unless a full application form is used
+pub struct Param<'a> {
```
acl-cqc, line 692: +1 to this commoning up, amen.

I am a long way from convinced that we need implicit/explicit, but that's another issue and definitely not for this PR!

hugr-model/src/v0/mod.rs
```diff
@@ -662,40 +670,42 @@ pub enum Term<'a> {
     ///
     /// `ctrl : static`
     ControlType,
+
+    /// Constraint that requires a runtime type to be copyable.
+    CopyConstraint {
```
acl-cqc, line 675: I would not object to a general `Term::Constraint { name: str, args: Vec<TermId> }` with the name looked up in *data* that is static for now (i.e. just the one/two constraints and none user-defined). It does kinda say that there will be more constraints coming, but with the data being static there is no possibility of adding them.

hugr-model/src/v0/mod.rs
acl-cqc: (nice that Constraints are now separate from Parameters here too! 👍  )

hugr-model/src/v0/text/print.rs
```diff
@@ -178,9 +170,8 @@ impl<'p, 'a: 'p> PrintContext<'p, 'a> {
                     this.print_text(decl.name);
                 });
 
-                for param in decl.params {
-                    this.print_param(*param)?;
-                }
+                this.print_params(decl.params)?;
```
acl-cqc, line 173: If we are sure we want constraints in Alias too (I argue against elsewhere), then it is true that everywhere you introduce params you also introduce constraints; maybe a `struct ParamsAndConstraints<'a>` (maybe `ConstrainedParams`?) would be useful, as it'd allow combining all these `print_params(...); print_constraints(...)`

hugr-model/src/v0/text/print.rs
```diff
@@ -460,25 +446,37 @@ impl<'p, 'a: 'p> PrintContext<'p, 'a> {
         }
     }
 
+    fn print_params(&mut self, params: &'a [Param<'a>]) -> PrintResult<()> {
+        for param in params {
```
acl-cqc, line 450: nit: can you just `params.iter().try_for_each(self.print_param)` (maybe `.copied()` or something) ?

hugr-model/tests/fixtures/model-constraints.edn
acl-cqc: Yeah, I feel that the "sales blurb" in the description that "The translation in this PR requires the two to be used together since TypeBound does not have the facility to express one without the other. Once we move to constraints in hugr-core, that restriction would be lifted." misses the point. We *could* have had separate copy/discard/affine/etc. in TypeBound, but [we discussed this](https://github.com/CQCL/hugr/discussions/1298) and did not decide that we wanted it. So I see no reason to do so now.

Indeed, if we change our minds are decide we want to add this in the future, there is still an easy mechanism (right?) - we just add two new constraints (say `copy` and `discard`) alongside the old `copyable` and then define that `copyable` holds if both `copy` and `discard` hold (and that each of the latter/two hold if `copyable` does), right? (so ok we need a cyclicity check in the constraint solver.) The problem would just be better naming, we can't really have both `copy` and `copyable`....

So (the more I think about it, the more) I think we should stick with one combined constraint rather than two for now.

hugr-model/tests/fixtures/model-constraints.edn
```diff
@@ -0,0 +1,15 @@
...
+  [?t] [(@ array.Array ?t ?n)]
+  (ext))
+
+(declare-func tuple.copy
+  (forall ?t type)
+  (where (copy ?t))
+  (where (discard ?t))
+  [(@ foo.tuple ?t)] [(@ foo.tuple ?t) (@ foo.tuple ?t)] (ext))
```
acl-cqc, line 15: What is `foo.tuple` ??

acl-cqc (commented):
Generally I would say this looks pretty good, but I vote for combining copy+discard, also a couple of other ideas (/confusions)...

hugr-model/src/v0/mod.rs
zrho: There is a possibility and moreover a mechanism for adding them: Your `Term::Constraint { name: str, args: Vec<TermId> }` is `Term::Apply`/`Term::ApplyFull` for a constructor that declares a constraint.

There are good arguments for keeping `Term` as minimal as possible by defining anything that does not need special handling as a custom constructor in a standard extension. In particular it'd remove a bunch of boilerplate code. There's also arguments for having anything that is structurally important as builtin; for example `adt`s since the control flow mechanism depends on it, or copyability constraints since they emerge just from the connectivity pattern of ports. I could go either way.

hugr-model/src/v0/text/print.rs
zrho: I agree that we'd probably want to avoid constraints on aliases for now.

hugr-core/src/import.rs
zrho: The item type is nested within the list and since we only constrain parameters directly here, we can not express that the items of the list should be copyable. Therefore the `TypeBound::Any`.

hugr-core/src/export.rs
zrho: When exporting a `PolyFuncType`, we gather constraints in `local_constraints`. `PolyFuncType` itself can not express constraints, but if one of the parameters takes a copyable type, `local_constraints` is where we record the corresponding constraint for that parameter.

hugr-model/tests/fixtures/model-constraints.edn
ss2165: `nonlinear` ..?

zrho: I'm fine with `nonlinear`. @acl-cqc ?

hugr-model/tests/fixtures/model-constraints.edn
zrho: It's some arbitrary custom constructor to trigger a particular case in import/export for more test coverage. `tuple` is for "historical" reasons since the example was different at first... Can rename it into something more nondescript.

hugr-model/tests/fixtures/model-constraints.edn
zrho: The two constraints are joined into a `nonlinear` constraint now.

hugr-model/src/v0/mod.rs
zrho: Constraints on aliases have been removed.

acl-cqc (commented):
Code all looks good to me, thanks @zrho . I would "Approve" all the Rust, but I have many suggestions/asks for comment improvements, please bear in mind that (I think) the majority of people who may have to work with this have probably less type-system-fu than I do....

hugr-core/src/import.rs
```diff
@@ -1285,3 +1302,18 @@ impl<'a> Names<'a> {
...
 }
+
+#[derive(Debug, Clone, Copy)]
+struct LocalVar {
+    r#type: model::TermId,
+    bound: TypeBound,
+}
+
+impl LocalVar {
+    pub fn new(r#type: model::TermId) -> Self {
```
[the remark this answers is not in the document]
zrho: `LocalVar::new` creates the variable in its default state, after which more information is collected. In particular, we call `LocalVar::new` also for variables that we later figure out to be nonlinear.

hugr-core/src/export.rs
```diff
@@ -44,10 +44,21 @@ struct Context<'a> {
     bump: &'a Bump,
     /// Stores the terms that we have already seen to avoid duplicates.
     term_map: FxHashMap<model::Term<'a>, model::TermId>,
+
     /// The current scope for local variables.
+    ///
+    /// This is set to the id of the smallest enclosing node that defines a polymorphic type.
+    /// We use this when exporting local variables in terms.
```
acl-cqc, line 51: Slightly nervous of "we use this when" comments because the far-away thing on the RHS can change and forget to update the comment...

hugr-core/src/export.rs
```diff
@@ -44,10 +44,21 @@ struct Context<'a> {
...
     local_scope: Option<model::NodeId>,
+
     /// Constraints to be added to the local scope.
+    ///
+    /// When exporting a node that defines a polymorphic type, we use this field
```
acl-cqc, line 56: However, this is helpful, thanks :). (The "Currently...." bit is probably overkill but happy to leave it)

acl-cqc (approved):
Thanks @zrho :-). Clean code, nice :)

PR #1626, opened 2024-11-04
feat: Add TailLoop::BREAK_TAG and CONTINUE_TAG

acl-cqc:
It doesn't seem obvious to me which way round these should be, so add constants, and use for the "definitive" `make_break` and `make_continue`.

No change to python as this doesn't even have `make_break` or `make_continue` ATM.

### DIFF
--- a/hugr-core/src/builder/build_traits.rs
+++ b/hugr-core/src/builder/build_traits.rs
@@ -623,7 +623,11 @@ pub trait Dataflow: Container {
         tail_loop: ops::TailLoop,
         values: impl IntoIterator<Item = Wire>,
     ) -> Result<Wire, BuildError> {
-        self.make_sum(TailLoop::CONTINUE_TAG, [tail_loop.just_inputs, tail_loop.just_outputs], values)
+        self.make_sum(
+            TailLoop::CONTINUE_TAG,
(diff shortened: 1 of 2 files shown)


mark-koch (approved):
Looks good to me, AFAIK the Python builder doesn't even have `make_continue` and `make_break`.

Should we use these `CONTINUE_TAG` and `BREAK_TAG` constants also during validation?

acl-cqc:
> Should we use these CONTINUE_TAG and BREAK_TAG constants also during validation?

Yes, I guess so - but I'm looking for validation code that uses constant 0 / 1 for tags/tail-loops and not finding anything. If you can point me where you're looking then I'd be happy to fix?

acl-cqc:
As per discussion on Slack - the other change we could make would be at https://github.com/CQCL/hugr/blob/c987d2f775e310878c8d16874629db7cc8535178/hugr-core/src/ops/controlflow.rs#L44 where we could instead do something like
```
let mut sum_rows = [type_row![], type_row![]];
sum_rows[BREAK_TAG] = self.just_outputs.clone();
sum_rows[CONTINUE_TAG] = self.just_inputs.clone();
```
but we deem this "not worth it".

PR #1628, opened 2024-11-04
feat!: Allow static inputs to extension operations

### LINKED ISSUE #1594: Allow custom ops with static constant input edges
### Discussed in https://github.com/CQCL/hugr/discussions/1342

<sup>Originally posted by **ss2165** July 24, 2024</sup>
This follows discussions mainly in #1308 

A constant static input to an op would would be some semantically significant data, that cannot be lost like metadata. This could be thought of as "literal properties" or "intrinsic parameters". 

If further we can elevate values to constants at compile time (constexpr evaluation) we can do powerful things with ops that need some structured, complicated, compile-time known data.

An open question is whether this data is available to binary compute_signature functions.</div>

ss2165:
Extension operations can now have some number of static inputs that come between their dataflow input ports and the state order input port.

This can be used for compile-time-known inputs to operations. 
I have attempted to keep the serialization backwards compatible.

OpDefs can either declare their static inputs as a row, or it can be calculated like the rest of the signature.

Closes #1594

BREAKING CHANGE: `OpType, OpTrait, DataflowOpTrait` methods which assumed one static input now support many. E.g. `static_input_port` -> `static_input_ports`. Returning a `Vec` rather than an `Option`. Operation definitions support static input type declarations, so a new type `OpDefSignature` has been introduced for this to hold the `PolyFuncTypeRV` and the static inputs. The cached signature of an ExtensionOp is also a new type `ExtOpSignature` that holds the dataflow `Signature` and the static inputs. `OpTag::StaticInput` removed because it doesn't actually narrow things down any more.

### DIFF
--- a/hugr-core/src/extension/op_def.rs
+++ b/hugr-core/src/extension/op_def.rs
@@ -274,6 +274,14 @@ pub struct ExtOpSignature {
 }
 
 impl ExtOpSignature {
+    /// Create a new [ExtOpSignature] from a [Signature] and a [TypeRow].
+    pub fn new(func_type: Signature, static_inputs: impl Into<TypeRow>) -> Self {
+        Self {
+            func_type,
+            static_inputs: static_inputs.into(),
+        }
+    }
+
     /// Returns the function type of the signature.
     pub fn func_type(&self) -> &Signature {
         &self.func_type
@@ -549,10 +557,10 @@ pub(super) mod test {
 
     use itertools::Itertools;
 
-    use super::SignatureFromArgs;
+    use super::{ExtOpSignature, SignatureFromArgs};
     use crate::builder::{endo_sig, DFGBuilder, Dataflow, DataflowHugr};
     use crate::extension::op_def::{CustomValidator, LowerFunc, OpDef, SignatureFunc};
-    use crate::extension::prelude::USIZE_T;
+    use crate::extension::prelude::{QB_T, USIZE_T};
     use crate::extension::{ExtensionRegistry, ExtensionSet, PRELUDE};
     use crate::extension::{SignatureError, EMPTY_REG, PRELUDE_REGISTRY};
     use crate::ops::OpName;
@@ -837,6 +845,36 @@ pub(super) mod test {
             def.compute_signature(&args, &PRELUDE_REGISTRY),
             Ok(exp_fun_ty.into())
         );
+        Ok(())
+    }
+
+    pub(crate) fn static_input_op() -> (Extension, OpName) {
+        let mut e = Extension::new_test(EXT_ID);
+        let def = e
+            .add_op(
+                "StaticInOp".into(),
+                "".into(),
+                OpDefSignature::new(vec![], Signature::new(vec![], vec![QB_T]))
+                    .with_static_inputs(USIZE_T),
+            )
+            .unwrap();
+        let op_name = def.name().clone();
+        (e, op_name)
+    }
+
+    #[test]
+    fn test_static_input_op() -> Result<(), Box<dyn std::error::Error>> {
+        let (e, op_name) = static_input_op();
+        let reg = ExtensionRegistry::try_new([PRELUDE.to_owned(), e])?;
+        let e = reg.get(&EXT_ID).unwrap();
+
+        let ext_op = e.instantiate_extension_op(&op_name, vec![], &reg).unwrap();
+        let sig = ext_op.ext_op_signature();
+        let expected = ExtOpSignature::new(
+            Signature::new(vec![], vec![QB_T]).with_extension_delta(e.name().clone()),
+            USIZE_T,
+        );
+        assert_eq!(sig, &expected);
         Ok(())
     }
 
--- a/hugr-core/src/extension/op_def.rs
+++ b/hugr-core/src/extension/op_def.rs
@@ -4,6 +4,9 @@ use std::collections::HashMap;
 use std::fmt::{Debug, Formatter};
 use std::sync::Arc;
 
+#[cfg(test)]
+use {crate::proptest::RecursionDepth, ::proptest::prelude::*, proptest_derive::Arbitrary};
+
 use super::{
     ConstFold, ConstFoldResult, Extension, ExtensionBuildError, ExtensionId, ExtensionRegistry,
     ExtensionSet, SignatureError,
@@ -11,7 +14,7 @@ use super::{
 
 use crate::ops::{OpName, OpNameRef};
 use crate::types::type_param::{check_type_args, TypeArg, TypeParam};
-use crate::types::{FuncValueType, OpDefSignature, PolyFuncType, Signature};
+use crate::types::{FuncValueType, OpDefSignature, PolyFuncType, Signature, TypeRow};
 use crate::Hugr;
 mod serialize_signature_func;
 
@@ -224,7 +227,7 @@ impl SignatureFunc {
         def: &OpDef,
         args: &[TypeArg],
         exts: &ExtensionRegistry,
-    ) -> Result<Signature, SignatureError> {
+    ) -> Result<ExtOpSignature, SignatureError> {
         let temp: OpDefSignature; // to keep alive
         let (pf, args) = match &self {
             SignatureFunc::CustomValidator(custom) => {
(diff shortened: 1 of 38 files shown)


hugr-core/src/hugr/views.rs
```diff
@@ -416,11 +416,13 @@ pub trait HugrView: HugrInternals {
             .finish()
     }
 
-    /// If a node has a static input, return the source node.
-    fn static_source(&self, node: Node) -> Option<Node> {
-        self.linked_outputs(node, self.get_optype(node).static_input_port()?)
-            .next()
-            .map(|(n, _)| n)
+    /// If a node has static inputs, return the source nodes.
+    fn static_sources(&self, node: Node) -> Vec<Node> {
```
aborgna-q, line 420: Can't we return an `impl Iterator<Item = Node> + &'_` ?

hugr-core/src/import.rs
```diff
@@ -251,7 +252,12 @@ impl<'a> Context<'a> {
             let src = self.nodes[&src_id];
             let dst = self.nodes[&dst_id];
             let src_port = self.hugr.get_optype(src).static_output_port().unwrap();
-            let dst_port = self.hugr.get_optype(dst).static_input_port().unwrap();
+            let dst_port = *self
+                .hugr
+                .get_optype(dst)
+                .static_input_ports()
+                .first()
+                .unwrap();
```
aborgna-q, line 260:
```suggestion
                .exactly_one()
                .unwrap();
```

hugr-core/src/ops.rs
```diff
@@ -157,10 +157,10 @@ impl OpType {
     /// given direction after any dataflow ports and before any
     /// [`OpType::other_port_kind`] ports.
     #[inline]
-    pub fn static_port_kind(&self, dir: Direction) -> Option<EdgeKind> {
+    pub fn static_port_kind(&self, dir: Direction) -> Vec<EdgeKind> {
```
aborgna-q, line 160: Rename to `_kinds` ?
Also update the docs.

hugr-core/src/ops.rs
```diff
@@ -225,28 +224,36 @@ impl OpType {
...
     #[inline]
     pub fn static_output_port(&self) -> Option<OutgoingPort> {
-        self.static_port(Direction::Outgoing)
-            .map(|p| p.as_outgoing().unwrap())
+        if let [p] = self.static_ports(Direction::Outgoing).as_slice() {
+            Some(p.as_outgoing().unwrap())
+        } else {
+            None
+        }
+        // .map(|p| p.as_outgoing().unwrap()).collect()
```
aborgna-q, line 256:
```suggestion
        match self.static_ports(Direction::Outgoing).as_slice() {
            [p] => Some(p.as_outgoing().unwrap()),
            _ => None,
        }
```

hugr-core/src/ops.rs
aborgna-q, line 1: Many of the `OpType` methods look like they could return an `impl Iterator` instead of collecting a vec.
Are there any lifetime problems with that?

hugr-core/src/ops/custom.rs
```diff
@@ -108,24 +111,35 @@ impl ExtensionOp {
             name: self.def.name().clone(),
             description: self.def.description().into(),
             args: self.args.clone(),
-            signature: self.signature.clone(),
+            signature: self.signature.func_type.clone(),
+            static_inputs: self.signature.static_inputs.clone(),
         }
     }
+
+    /// Returns the [`ExtOpSignature`] of this [`ExtensionOp`].
```
aborgna-q, line 119: nit: We can already see the types.
```suggestion
    /// Return the instantiated signature of this [`ExtensionOp`].
```

hugr-py/src/hugr/build/dfg.py
```diff
@@ -195,12 +195,19 @@ def add_op(
...
         return replace(new_n, _num_out_ports=op.num_out)
 
-    def add(self, com: ops.Command, *, metadata: dict[str, Any] | None = None) -> Node:
+    def add(
+        self,
+        com: ops.Command,
+        *,
+        static_in: Iterable[Wire] | None = None,
+        metadata: dict[str, Any] | None = None,
+    ) -> Node:
```
aborgna-q, line 204: Add the `static_in` argument to `add_op` too (and maybe to `_wire_up`), and use it here.

hugr-py/src/hugr/build/tracked_dfg.py
```diff
@@ -124,7 +124,13 @@ def tracked_wire(self, index: int) -> Wire:
             raise IndexError(msg)
         return tracked
 
-    def add(self, com: Command, *, metadata: dict[str, Any] | None = None) -> Node:
+    def add(
+        self,
+        com: Command,
+        *,
+        static_in: Iterable[Wire] | None = None,
```
aborgna-q, line 131: The parameter is not used?

aborgna-q (commented):
Long review, but the PR looks nice.

You mentioned that some `Vec` return types cannot be the rewritten as `impl Iterator`s, but some of them surely can?

zrho:
I'm wondering about the purpose of static inputs. So these are statically known runtime values, which are statically known because they are produced by a const node? We already have a mechanism to provide statically known information to an operation via `TypeArg`s; this isn't a hack but appears to be the most natural way to do this anyway. In that light I'd advocate for removing static inputs entirely instead of allowing multiple of them. Am I missing something?

acl-cqc:
....and specifically, `TypeArg` does not allow to contain static functions.

Now, IIUC there are proposals to allow a `TypeArg:Node`, which if it referenced a funcdefn, would allow a "loadconst"-like op to produce a function value (the TypeArg replacing the Const, or indeed, you could say it was a "loadfunction"-like op) - and then removing the ability to have Hugr's nested inside consts (==> equivalent to saying static function values can *only* be produced by LoadFunction). I haven't thought through monomorphic-vs-polymorphic, typeapply, etc. yet tho

acl-cqc:
> In that light I'd advocate for removing static inputs entirely instead of allowing multiple of them

In some sense, this could well be where we are headed - but we certainly wouldn't want to remove them until we have finalized+implemented https://github.com/CQCL/hugr/discussions/1425! So in the meantime, aren't "static inputs" just the hugr-core way of representing (with nodes and edges) the noderef-like "terms" proposed in that discussion?

zrho:
> So in the meantime, aren't "static inputs" just the hugr-core way of representing (with nodes and edges) the noderef-like "terms" proposed in that discussion?

I don't think so. This PR introduces an additional parameter list to operations, so that operations can have three different types of inputs:
 - static values (`TypeArg` in core, `Term` in model)
 - runtime values, but statically known
 - runtime values
And this does not appear to be an implementation detail, but part of the specification of an operation.

acl-cqc:
> operations can have three different types of inputs:
>
> * static values (TypeArg in core, Term in model)
> * runtime values, but statically known
> * runtime values

I thought the idea of #1425 was that the first two of these would be Terms in the model, which would include all "constant values" (probably excluding nested Hugrs, instead allowing references to function nodes). That some of these Terms would (at least initially) be turned into nodes with edges and others would be left as TypeArgs is an implementation detail of hugr-core and not part of the model.

zrho:
> That some of these Terms would (at least initially) be turned into nodes with edges and others would be left as TypeArgs is an implementation detail of hugr-core and not part of the model.

Yes, in the standup where we gave the thumbs up for this I understood it in that way. But that appears not compatible with how this PR does it since it makes static inputs into their own part of the signature. It therefore can't be a `hugr-core` implementation detail.

hugr-core/src/extension/op_def.rs
```diff
@@ -141,11 +144,11 @@ impl CustomValidator {
 /// The ways in which an OpDef may compute the Signature of each operation node.
 pub enum SignatureFunc {
     /// An explicit polymorphic function type.
-    PolyFuncType(PolyFuncTypeRV),
+    PolyFuncType(OpDefSignature),
     /// A polymorphic function type (like [Self::PolyFuncType] but also with a custom binary for validating type arguments.
```
acl-cqc, line 148:
```suggestion
    /// A polymorphic function type (like [Self::PolyFuncType]) but also with a custom binary for validating type arguments.
```

hugr-core/src/types/poly_func.rs
```diff
@@ -47,9 +54,93 @@ pub type PolyFuncType = PolyFuncTypeBase<NoRV>;
 /// The polymorphic type of an [OpDef], whose number of input and outputs
 /// may vary according to how [RowVariable]s therein are instantiated.
 ///
+/// It may also have a row of static inputs.
+///
 /// [OpDef]: crate::extension::OpDef
-pub type PolyFuncTypeRV = PolyFuncTypeBase<RowVariable>;
+#[derive(Clone, PartialEq, Debug, Default, Eq, Hash, derive_more::Display)]
+#[cfg_attr(test, derive(Arbitrary), proptest(params = "RecursionDepth"))]
+#[display("{}{}{}", self.display_params(), self.static_inputs, self.body())]
```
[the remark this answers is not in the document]
acl-cqc: I wonder if `OpDefSignature` should now move into `hugr_core::extension`

acl-cqc:
> > That some of these Terms would (at least initially) be turned into nodes with edges and others would be left as TypeArgs is an implementation detail of hugr-core and not part of the model.
> 
> But that appears not compatible with how this PR does it since it makes static inputs into their own part of the signature. It therefore can't be a `hugr-core` implementation detail.

I'm not sure I understand here - PR leaves the signature of a *node* unchanged, as a `pub type Signature = FuncTypeBase<NoRV>`. An Op*Def* now has a type_row of static inputs as well, i.e. an `OpDefSignature` rather than a `PolyFuncTypeBase<RowVariable>` - is that the problem, i.e. in the serialization==model-representation of extension definitions ?

ss2165:
>  PR leaves the signature of a node unchanged,
Don't think this is true it may not be stored in the dataflow signature part but nodes still have to report their static signature, like using `ExtOpSignature` for extension ops.

acl-cqc:
> > Don't think this is true it may not be stored in the dataflow signature part but nodes still have to report their static signature, like using `ExtOpSignature` for extension ops.

Ah I see. Yes if that gets written out in the model then that doesn't sound like what we want. That info will eventually go in the model in the Term (equiv to TypeArg) part, IIUC.

hugr-core/src/extension/op_def.rs
```diff
@@ -244,11 +247,72 @@ impl SignatureFunc {
...
+        let func_type = res.try_into()?;
+
+        Ok(ExtOpSignature {
+            func_type,
+            static_inputs,
+        })
+    }
+}
+
+/// Instantiated [OpDef] signature.
```
acl-cqc, line 264: Suggest this goes in `hugr_core::ops::custom`

ss2165:
Closing as stale, subsumed by work in #1433

PR #1633, opened 2024-11-05
feat: Add array `repeat` and `scan` ops

### LINKED ISSUE #1627: Add op to initialise an array using a function
Required for https://github.com/CQCL/guppylang/pull/616.

Signature `forall n, T. (() -> T) -> array<n, T>`

mark-koch:
Closes #1627


hugr-core/src/extension/prelude/array.rs
```diff
@@ -40,6 +40,7 @@ use crate::types::type_param::TypeParam;
 #[non_exhaustive]
 pub enum ArrayOpDef {
     new_array,
+    new_uninitialized_array,
     get,
```
aborgna-q, line 44: Move this to the end or add an out-of-order discriminant to avoid the breaking change.
```suggestion
    new_uninitialized_array=100,
    get=2,
```

hugr-core/src/extension/prelude/array.rs
```diff
@@ -179,6 +184,10 @@ impl MakeOpDef for ArrayOpDef {
     fn description(&self) -> String {
         match self {
             ArrayOpDef::new_array => "Create a new array from elements",
+            ArrayOpDef::new_uninitialized_array => {
+                "Create a new array with uninitialised elements. \
+                Reading such an unitialised element is undefined behaviour."
```
cqc-alec, line 189: Do we really want to have undefined behaviour in hugr? Perhaps it should panic instead (though that would require storing a "poison" value I suppose).

mark-koch: I think guaranteeing a panic would be too much overhead. An alternative would be to pass a default value to the op and renaming it to `array_repeat` or something like that.

For linear arrays, this default would need to be provided via a function that is called for each element

cqc-alec: Yeah that could work. May add some runtime overhead (constructing values that are then replaced), but at least some of that could be eliminated by compiler optimizations.

hugr-core/src/extension/prelude/array.rs
```diff
@@ -312,6 +331,192 @@ pub fn new_array_op(element_ty: Type, size: u64) -> ExtensionOp {
...
+    }
+
+    fn extension(&self) -> ExtensionId {
+        PRELUDE_ID
+    }
+
+    fn description(&self) -> String {
+        "A combination of map and foldl. Applies a function to each element \
+        of the array with an accumulator that is passed trough from left to \
+        right. Returns the resulting array and the final state of the \
```
cqc-alec, line 411:
```suggestion
        of the array with an accumulator that is passed through from start to \
        finish. Returns the resulting array and the final state of the \
```

hugr-core/src/extension/prelude/array.rs
```diff
@@ -312,6 +331,192 @@ pub fn new_array_op(element_ty: Type, size: u64) -> ExtensionOp {
...
+        accumulator."
+            .into()
+    }
+
+    /// Add an operation implemented as an [MakeOpDef], which can provide the data
```
cqc-alec, line 416:
```suggestion
    /// Add an operation implemented as a [MakeOpDef], which can provide the data
```

hugr-core/src/extension/prelude/array.rs
```diff
@@ -312,6 +331,192 @@ pub fn new_array_op(element_ty: Type, size: u64) -> ExtensionOp {
...
+    /// required to define an [OpDef], to an extension.
+    //
+    // This method is re-defined here since we need to pass the array type def while
+    //  computing the signature, to avoid recursive loops initializing the extension.
```
cqc-alec, line 420:
```suggestion
    // computing the signature, to avoid recursive loops initializing the extension.
```

hugr-core/src/extension/prelude/array.rs
```diff
@@ -312,6 +331,192 @@ pub fn new_array_op(element_ty: Type, size: u64) -> ExtensionOp {
...
+
+/// Definition of the array scan op.
+#[derive(Clone, Debug, PartialEq)]
+pub struct ArrayScan {
+    /// The element type of the input array.
+    src_ty: Type,
+    /// The target element type of the output array.
+    tgt_ty: Type,
+    /// The accumulator types.
+    acc_tys: Vec<Type>,
```
cqc-alec, line 442: It's not clear to me why this is an array of types and not just a type -- isn't the accumulator a single type?

mark-koch: I used a row variable for the accumulator. Setting the accumulator to the empty row recovers the signature of `map`

cqc-alec (approved):

Issue #1613, opened 2024-10-24
feat(core): Put extension declarations in top level of `Hugr`


ss2165:
To allow validation of individual HUGRs with any root, without resorting to packages, which only currently allow modules. This would then mean packages would only need to contain modules (no extensions) and can be used mostly for linking. This also aligns well with the model work so should be done as a follow up to that. See #1433.

PR #1647, opened 2024-11-12
feat!: Share `Extension`s under `Arc`s

aborgna-q:
Extensions are defined once and shared throughout `ExtensionRegistr`ies, `Package`s, and soon `Hugr`s (#1613).
This _write-once then share around_ is a good usecase for `Arc`s, specially since the definitions are mostly read and rarely cloned.

This is a requisite for #1613, to avoid cloning all extensions for each new hugr.

BREAKING CHANGE: `ExtensionRegistry` and `Package` now wrap `Extension`s in `Arc`s.

### DIFF
--- a/hugr-core/src/extension/op_def.rs
+++ b/hugr-core/src/extension/op_def.rs
@@ -617,7 +617,7 @@ pub(super) mod test {
         assert_eq!(def.misc.len(), 1);
 
         let reg =
-            ExtensionRegistry::try_new([PRELUDE.to_owned(), EXTENSION.to_owned(), e]).unwrap();
+            ExtensionRegistry::try_new([PRELUDE.clone(), EXTENSION.clone(), e.into()]).unwrap();
         let e = reg.get(&EXT_ID).unwrap();
 
         let list_usize =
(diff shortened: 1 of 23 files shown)


.github/workflows/python-wheels.yml
```diff
@@ -74,4 +74,4 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           verbose: true
-          skip-existing: true
\ No newline at end of file
+          skip-existing: true
```
aborgna-q, line 77: Drive-by from the pre-commit checks.

hugr-core/src/extension/op_def.rs
```diff
@@ -611,7 +611,7 @@ pub(super) mod test {
         assert_eq!(def.misc.len(), 1);
 
         let reg =
-            ExtensionRegistry::try_new([PRELUDE.to_owned(), EXTENSION.to_owned(), e]).unwrap();
+            ExtensionRegistry::try_new([PRELUDE.clone(), EXTENSION.clone(), e.into()]).unwrap();
```
aborgna-q, line 614: Note that for all of these `to_owned` was still valid.
I just think that `clone`-ing the `Arc` is more explicit now, that we can use that.

mark-koch (approved):

Issue #1652, opened 2024-11-13
`hugr_passes::lower` is slow on graphs with large Dataflow regions because of convexity checking


doug-q:
Running `lower_ops` on a graph with a large dataflow regions instantiates a `SiblingSubgraph` per op to be replaced. Each instantiation does a convexity check, which is slow.

I suggest we add `SiblingSubgraph::from_node` which is sufficient in this case, and which does not need convexity checking because a graph consisting of a single node is always convex.

PR #1654, opened 2024-11-13
perf: Faster singleton SiblingSubgraph construction

aborgna-q:
Wraps the topoConvexChecker used by `SiblingSubgraph` in a `OnceCell` so we can delay initialization until it's needed.

This lets us avoid traversing the graph (and computing a topological order) when creating sibling subgraphs with zero or one nodes.

This partially helps with #1652.

See benchmark results:
```
$ critcmp before after -f '.*subgraph'
group                      after                                  before
-----                      -----                                  ------
singleton_subgraph/10      1.00      3.0±0.08µs        ? ?/sec    2.71      8.1±0.18µs        ? ?/sec
singleton_subgraph/100     1.00      4.8±0.07µs        ? ?/sec    9.81     47.4±1.16µs        ? ?/sec
singleton_subgraph/1000    1.00     23.2±1.86µs        ? ?/sec    18.94  439.3±17.04µs        ? ?/sec
multinode_subgraph/10      1.01     17.9±0.25µs        ? ?/sec    1.00     17.7±0.29µs        ? ?/sec
multinode_subgraph/100     1.01    170.0±3.72µs        ? ?/sec    1.00    168.5±3.04µs        ? ?/sec
multinode_subgraph/1000    1.02      2.4±0.02ms        ? ?/sec    1.00      2.3±0.04ms        ? ?/sec
```
`singleton_subgraph` creates a single-node subgraph in a region with around `k * 3` nodes.
`multinode_subgraph` creates a subgraph with 1/3rd of the nodes for the same region.

This PR is quite noisy since it's adding those two new benchmarks, and tidying up the files in the process.

drive-by: Add `bench = false` for all the targets in the workspace. Otherwise, the auto-added test harness threw errors when passing criterion flags to `cargo bench`.


aborgna-q:
Uh, looks like I introduced a breaking change -.-
(`TopoConvexChecker` is no longer `Sync` now that it uses an `OnceCell` internally...)

PR #1657, opened 2024-11-14
feat: Lists and extension sets with splicing

### LINKED ISSUE #1609: Extension sets and lists with variables in `hugr-model`
`hugr-model` at this point only supports extension sets and lists that have at most one variable, which needs to be at the end in the case of lists. `hugr-core` allows any number of variables to occur in extension sets and rows. I was initially skeptical that this might make the type theory quite a bit more complicated. I now think that this additional complexity is manageable enough so that I don't have problem with making `hugr-model` support multiple variables in extension sets and rows. That would fix one of the remaining mismatches.

zrho:
Closes #1609.


hugr-core/Cargo.toml
```diff
@@ -20,6 +20,7 @@ workspace = true
 extension_inference = []
 declarative = ["serde_yaml"]
 model_unstable = ["hugr-model"]
+default = ["model_unstable"]
```
acl-cqc, line 23: At the point where we are making "unstable" the default I have to wonder if we still need a flag for the base/stable model ?

zrho: This isn’t intentional, good for spotting it. I got to figure out how to tell the lsp to work without needing to make the feature default, because I keep forgetting to take it out…

hugr-core/src/export.rs
```diff
@@ -767,31 +745,47 @@ impl<'a> Context<'a> {
...
-        items.extend(t.iter().map(|row| self.export_type(row)));
+    pub fn export_type_row<RV: MaybeRV>(
+        &mut self,
+        row: &TypeRowBase<RV>,
+        tail: Option<model::TermId>,
+    ) -> model::TermId {
+        let mut items = BumpVec::with_capacity_in(row.len() + tail.is_some() as usize, self.bump);
+
+        for t in row.iter() {
+            match t.as_type_enum() {
```
acl-cqc, line 773: Consider `items.push(match t.as_type_enum() { .... })` or even `items.push(if let TypeEnum::RowVar(var)=t.as_type_enum() {...} else {...})`

acl-cqc: A shame there is no `from_sized_iter_in` so you could create the BumpVec and fill it in one go!

hugr-core/src/export.rs
```diff
@@ -767,31 +745,47 @@ impl<'a> Context<'a> {
...
+                TypeEnum::RowVar(var) => {
+                    items.push(model::ListItem::Splice(self.export_row_var(var.as_rv())));
+                }
+                _ => {
+                    items.push(model::ListItem::Item(self.export_type(t)));
+                }
+            }
+        }
+
+        if let Some(tail) = tail {
```
acl-cqc, line 783: Consider renaming `export_type_row` to `export_type_row_with_tail` and then adding `export_type_row` that just adds the `None`

hugr-core/src/export.rs
```diff
@@ -822,54 +818,25 @@ impl<'a> Context<'a> {
...
-                // However, it appears that so far we never actually use extension sets with
-                // multiple variables, except for extension sets that are generated through
-                // property testing.
-                if rest.is_some() {
-                    // TODO: We won't need this anymore once we have a core representation
-                    // that ensures that extension sets have at most one variable.
-                    panic!("Extension set with multiple variables")
-                }
+    pub fn export_ext_set(&mut self, ext_set: &ExtensionSet) -> model::TermId {
+        let capacity = ext_set.iter().size_hint().0;
```
acl-cqc, line 822: Hmmm, so `ExtensionSet::iter()` just returns an `impl Iterator`. I wasn't familiar with this `size_hint` but...are we sure `BumpVec::from_iter_in` doesn't use it?

hugr-core/src/export.rs
```diff
@@ -822,54 +818,25 @@ impl<'a> Context<'a> {
...
 
-                let node = self.local_scope.expect("local variable out of scope");
-                rest = Some(
-                    self.module
-                        .insert_term(model::Term::Var(model::LocalRef::Index(node, index as _))),
-                );
-            } else {
-                extensions.push(self.bump.alloc_str(ext) as &str);
+        for ext in ext_set.iter() {
+            match ext.parse::<u16>() {
```
acl-cqc, line 826: This hack from `hugr-core` is probably worth noting here (sthg like "Variables in extension sets are represented as numbers as these are not valid as Extension names")

hugr-core/src/import.rs
```diff
@@ -1022,32 +1022,38 @@ impl<'a> Context<'a> {
...
+                    for item in *items {
+                        match item {
+                            model::ExtSetItem::Extension(ext) => {
+                                let ext_ident = IdentList::new(*ext).map_err(|_| {
+                                    model::ModelError::MalformedName(ext.to_smolstr())
+                                })?;
+                                es.insert(&ext_ident);
+                            }
+                            model::ExtSetItem::Splice(term_id) => {
+                                stack.push(*term_id);
```
acl-cqc, line 1047: Might be worth a comment that the final order doesn't matter (i.e. because it's an Extension_Set_)

hugr-core/src/import.rs
```diff
@@ -1138,62 +1144,86 @@ impl<'a> Context<'a> {
...
             match self.get_term(term_id)? {
-                model::Term::Var(_) => return Err(error_unsupported!("open lists")),
-                model::Term::List { items, tail } => {
-                    list_items.extend(items.iter());
-
-                    match tail {
-                        Some(tail) => term_id = *tail,
-                        None => break,
+                model::Term::List { items } => {
+                    types.reserve(items.len());
```
acl-cqc, line 1163: Is this worth it? It's a conservative overestimate and we'll be doing it repeatedly

hugr-model/src/v0/mod.rs
```diff
@@ -563,19 +563,12 @@ pub enum Term<'a> {
         r#type: TermId,
     },
 
-    /// A list, with an optional tail.
-    ///
-    /// - `[ITEM-0 ... ITEM-n] : (list T)` where `T : static`, `ITEM-i : T`.
-    /// - `[ITEM-0 ... ITEM-n . TAIL] : (list item-type)` where `T : static`, `ITEM-i : T`, `TAIL : (list T)`.
+    /// A list.
```
acl-cqc, line 566:
```suggestion
    /// A list. May include a mixture of single elements and also other list(variable)s to splice in
```
Or similar

hugr-model/src/v0/mod.rs
```diff
@@ -609,14 +602,9 @@ pub enum Term<'a> {
     NatType,
 
     /// Extension set.
-    ///
-    /// - `(ext EXT-0 ... EXT-n) : ext-set`
-    /// - `(ext EXT-0 ... EXT-n . REST) : ext-set` where `REST : ext-set`.
     ExtSet {
         /// The items in the extension set.
```
acl-cqc, line 606:
```suggestion
        /// The items in the extension set, in arbitrary order.
```

hugr-model/src/v0/mod.rs
```diff
@@ -664,6 +652,24 @@ pub enum Term<'a> {
     ControlType,
 }
 
+/// An item in a list term.
```
acl-cqc, line 655: Given one of the enum variants here is called `Item` I suspect we should find a different term for the enum itself. `Element`, `Member`, ... or maybe rename the variant to `ListItem::Element` - that might be best?

hugr-model/src/v0/text/parse.rs
```diff
@@ -170,21 +171,23 @@ impl<'a> ParseContext<'a> {
             }
 
             Rule::term_ext_set => {
-                let mut extensions = Vec::new();
-                let mut rest = None;
+                let mut items = BumpVec::with_capacity_in(inner.len(), self.bump);
```
acl-cqc, line 174: Another place where I wonder if `BumpVec::from_iter_in` would do the job?

hugr-model/src/v0/mod.rs
```diff
@@ -664,6 +652,24 @@ pub enum Term<'a> {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ListItem {
+    /// A single item.
+    Item(TermId),
+    /// A list to be spliced into the parent list.
+    Splice(TermId),
```
acl-cqc, line 661: I guess the model-validity-checker that tests that this TermId refers to a thing of appropriate List type, whereas the TermId in a `ListItem::Item` refers to something of appropriate element type, does not exist yet??

hugr-core/src/export.rs
acl-cqc: Does `self.bump.alloc_slice_fill_iter` do it?

hugr-model/src/v0/mod.rs
zrho: That's correct; this would be done by typechecking.

zrho:
Yeah, I was also wondering about functional style lists. This is one of a few attempts, which turned out to be the most natural. I also tried an option where there was `Term::ListEmpty`, `Term::ListCons` and `Term::ListConcat`. However the code to handle these turned out quite tedious.

hugr-core/src/export.rs
zrho: There's a few places where `Bump::alloc_slice_fill_iter` would in principle work. I've kept the `BumpVec` at the moment since I am leaving open the option that export may be fallible and return a `Result`, in which case it'd be easier to change the `BumpVec` option.

Also `Bump::alloc_slice_fill_iter` takes an `ExactSizeIterator` which we do not have in some cases.

acl-cqc: > Also Bump::alloc_slice_fill_iter takes an ExactSizeIterator which we do not have in some cases.

Ah ok - I wondered whether there was a problem something like that, and then when I saw that every Iterator has a `size_hint` I thought maybe it'd be ok anyway (i.e. sizing with capacity on a best-effort basis). If `alloc_slice_fill_iter` of a non-ExactSizeIterator doesn't compile...then never mind, BumpVec it'll have to be.

hugr-core/src/export.rs
acl-cqc: It fact it does - `from_iter_in` goes to `extend` [here](https://github.com/fitzgen/bumpalo/blob/8392c63d956962011a127bfd2f98c64aa36ff24a/src/collections/vec.rs#L2149-L2159)

zrho: `size_hint` gives a lower and an optional upper bound for the contents of an iterator. Extension sets do not report their size with the current API, and I didn't want to do a `hugr-core` API change in this PR.

`BumpVec::from_iter_in` very likely uses the `size_hint`. Same reason here for `BumpVec::push` as in other places: this makes it a bit easier to refactor for now, when everything is still quite in flux.

ss2165:
Looks good to me!

hugr-model/src/v0/mod.rs
acl-cqc: Yes, I thought of `ListComponent` and then indeed ListPart :-), yours are shorter==better

zrho: There's `ListPart`s and `ExtSetPart`s now.

hugr-model/src/v0/text/parse.rs
zrho: `BumpVec::from_iter_in` does not support propagating errors via `Result`s.

hugr-model/src/v0/mod.rs
```diff
@@ -676,6 +664,24 @@ pub enum Term<'a> {
...
+pub enum ListPart {
+    /// A single item.
+    Item(TermId),
+    /// A list to be spliced into the parent list.
+    Splice(TermId),
+}
+
+/// A part of an extension set term.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ExtSetPart<'a> {
```
acl-cqc, line 678: One might say that a part of a set is a subset... ;-)

acl-cqc (approved):
Looks good @zrho, thanks for changes!

PR #1664, opened 2024-11-18
DRAFT: feat: Prototype dataflow analysis for static circuit extraction

doug-q:
Based on #1603.

Includes an rough inliner and call graph.

See the new snapshots for current capabilities.

### DIFF
--- a/hugr-passes/src/static_circuit.rs
+++ b/hugr-passes/src/static_circuit.rs
@@ -114,7 +114,9 @@ impl TryFrom<Sum<QbHistory>> for QbHistory {
 impl AbstractValue for QbHistory {
     fn try_join(self, other: Self) -> Option<Self> {
         eprintln!("join: {:?} ^ {:?} ", &self, &other);
-        let r = if self.0 == other.0 && self.1.len() == other.1.len() {
+        
+        // eprintln!("join:  => {:?}", &r);
+        if self.0 == other.0 && self.1.len() == other.1.len() {
             Some(QbHistory(
                 self.0,
                 zip_eq(self.1, other.1)
@@ -123,10 +125,7 @@ impl AbstractValue for QbHistory {
             ))
         } else {
             None
-        };
-        panic!("dougrulz");
-        eprintln!("join:  => {:?}", &r);
-        r
+        }
     }
 
     fn try_meet(self, other: Self) -> Option<Self> {
@@ -192,9 +191,11 @@ impl<'a, H: HugrView> DFContext<QbHistory> for StaticCircuitContext<'a, H> {
         }
 
         if ins.iter().any(|x| x == &PartialValue::Bottom) {
+            eprintln!("interpret_leaf_op: {node}: At least one bottom input");
             return;
         }
 
+        eprintln!("interpret_leaf_op: {node}: No bottom inputs, continuing");
         // assert_eq!(ins.len(), self.hugr.num_inputs(node));
         // assert_eq!(outs.len(), self.hugr.num_outputs(node));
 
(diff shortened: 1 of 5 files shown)


hugr-passes/src/static_circuit.rs
```diff
@@ -0,0 +1,398 @@
...
+        ins: &[PartialValue<QbHistory>],
+        outs: &mut [PartialValue<QbHistory>],
+    ) {
+        self.assert_invariants();
+        let extension = e.def().extension();
+        if !self.extensions.contains(extension) {
+            return;
+        }
+
+        if ins.iter().any(|x| x == &PartialValue::Bottom) {
```
acl-cqc, line 194: I think this demonstrates exactly why I put in the test [you comment on here](https://github.com/CQCL/hugr/pull/1476/#discussion_r1843963953) !!

hugr-passes/src/static_circuit.rs
```diff
@@ -0,0 +1,398 @@
...
+            let mb_in_qbs = qb_ins
+                .iter()
+                .map(|&i| match &ins[i.index()] {
+                    PartialValue::Value(qbh) => Some((i, qbh)),
+                    _ => None,
+                })
+                .collect::<Option<HashMap<_, _>>>();
+            let gating_qbs = mb_in_qbs
+                .iter()
+                .flat_map(|hash_map| hash_map.iter().map(|(_, qbh)| qbh.qb_index()))
```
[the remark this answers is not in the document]
acl-cqc: So the "gating_nodes" of a new node is the ordered list of input gates, right? That could go in a comment on `Gate::gating_nodes`.

I'm *slightly* concerned that `Node` (rather than `(Node, Port)`) is sufficiently accurate - the wire coming from one or another port sounds like an important difference - I think the counterargument is that if a different port were used, *because of linearity* other places would see different `Node` instead, so we might be OK.

hugr-passes/src/static_circuit.rs
```diff
@@ -0,0 +1,398 @@
...
+                let mb_join_val = (|| {
+                    let in_p = qb_out_to_in.get(&OutgoingPort::from(i))?;
+                    let hash_map = mb_in_qbs.as_ref()?;
+                    let qbh: &QbHistory = hash_map.get(in_p)?;
+                    let mut qbh = qbh.clone();
+                    qbh.push(node, e.name(), gating_qbs.clone());
+                    Some(qbh.into())
+                })();
+
+                v.join_mut(mb_join_val.unwrap_or(PartialValue::Top));
```
acl-cqc, line 247: Short comment - if  the corresponding in-port has a QbHistory, push this gate (+ ordered list of its inputs) onto that QbHistory...

But, if this in-port is classical, join with Top. Doesn't this hit `3.py` where you have two different classical values?

hugr-passes/src/static_circuit.rs
```diff
@@ -0,0 +1,398 @@
...
+            nodes: BTreeSet::from_iter([node]),
+            gate: gate.into(),
+            gating_nodes: BTreeSet::from_iter(gating_nodes),
+            hash: 0,
+        };
+        g.calc_hash();
+        g
+    }
+    pub fn join(mut self, other: Self) -> Option<Self> {
+        if self.gate != other.gate || self.gating_nodes != other.gating_nodes {
```
acl-cqc, line 66: So we require `gating_nodes` to be exactly equal?

Doesn't this rule out cases like a conditional where each Case contains "the same" circuit, as the Node IDs will be different? You'd need some kind of isomorphism....

acl-cqc (commented):
Ok... so my overall thought is this is not really a dataflow analysis. What you want is the part of dataflow analysis that does "join" at the exit of every conditional, and indeed on every control-flow merge; and possibly for TailLoops; but the things you want to join are *circuits*, not any kind of value that's per-qubit. (Because that join/merge logic currently only exists in dataflow analysis, "give a man a hammer and everything looks like a nail", but the other bits of the dataflow framework are really not helping here, and it'd be better to do something else - another datalog, probably, with just the bit you need. If you really want analysis like "only these cases are reachable" we should determine that using DF and then simplify the Hugr with this knowledge.)

Also, you need to do some much-more-complex factoring of classical values so that you can catch "quantum-part-of-circuit same, classical part different" which you clearly want to do. And there is the isomorphism concern....

So, overall, afraid I still think that the best approach is to say flat circuits (no nested control flow) are statically scheduled; and then we transform into that. That lets us deploy a wide range of different optimizations to get there, *perhaps* including this or something like it, but with flexibility to choose an optimization according to what works on a particular circuit. (I slightly favour a rewriting approach but don't mean to rule this approach out.)

hugr-passes/src/static_circuit.rs
```diff
@@ -0,0 +1,398 @@
...
+            assert_eq!(0, qb_outs.len());
+            for out in outs.iter_mut() {
+                out.join_mut(PartialValue::Top);
+            }
+        } else {
+            assert_eq!(qb_outs.len(), qb_ins.len(), "{e:?}");
+            let mb_in_qbs = qb_ins
+                .iter()
+                .map(|&i| match &ins[i.index()] {
+                    PartialValue::Value(qbh) => Some((i, qbh)),
```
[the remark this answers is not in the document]
acl-cqc: Ok, I think that makes sense

acl-cqc:
> What is your criteria for being a dataflow analysis?

An abstract value per wire (perhaps only a subset of wires, i.e. only those of type Qubit). Here you want an abstract value *per circuit* (per sibling subgraph).

I think the analysis framework you want is:
* client provides a type of values `V`. Yes, these will need `join`.
* client provides a function `interpret_dfg(parent: Node, child_results: Map<Node, V>) -> V` which is called with a map whose keys are the container nodes in the dfg.
   * For "nested DFG" nodes, the value in the map is the result of calling `interpret_dfg`, so we could rely on `interpret_dfg` to recursively call itself, but I think this just makes the rules more complicated
   * The *framework* handles Conditional nodes by recursing on each child Case, and then `joining` the resulting V's
   * The framework handles CFG nodes similarly by merging at CF joins and copying at control-flow splits. (There is some value in doing Tag/Value/Sum analysis on wires here but I think that could be done separately as a first step)
   * The framework handles TailLoop's...somehow (I mean, any quantum op in a tailloop will report "no static circuit" unless it's a never-iterates loop, right?)

The use-case here then instantiates V == Circuit == `Hugr<RootNode=DFG>`, and `interpret_dfg` inlines the map (replacing each container node keyed in the Map with the DFG value from the Map)

I think this is necessary to handle qalloc/qfree which are not represented in the outputs (where you suggest "this might require Dense Analysis")

acl-cqc:
Ok, so qalloc/qfree could be represented in the containing region's outputs mentioning (in their `gating_nodes`) 2-qubit ops with the ancillae. But then you've got only a nodeindex, not a description of the op (and the history of the ancilla), so if you had a conditional which in each arm did qalloc, ..., qfree with (case 1) the same, or (case 2) different ops, to combine the ancilla with the conditional's input-and-output qubit, we still need to do some graph traversal (not here) to identify or distinguish the two arms?

Is recording all classical values/inputs as `PartialValue::Top` sufficient here? How does one proceed in that case? Is the goal here just to identify cases which could be statically scheduled if we then did a separate transformation pass?

hugr-passes/src/static_circuit.rs
acl-cqc: I think gating_nodes gets the node-id of the gate added, not the node id of the qalloc (/function input)?

hugr-passes/src/static_circuit.rs
acl-cqc: How does one proceed in that case (given a TOP)? Is the goal here just to identify cases which could be statically scheduled if we then did a separate transformation pass?

hugr-passes/src/static_circuit.rs
acl-cqc: Ah I see - yes much clearer now you've refactored `interpret_leaf_op` a bit

hugr-passes/src/static_circuit.rs
acl-cqc: Yes, "then wire up the angles" is the "separate transformation" i.e. where you add extra edges, Output node ports, etc. to get the angles from wherever they were computed.

acl-cqc:
Yeah ok so I think the things I would disagree with are not actually in this PR - they are the architecture, what analysis/transformation goes where in the pipeline, what the intermediates exchanged look like.

That you *can* use dataflow analysis to extract (some) static circuits, I do not dispute :-) :-). Whether it's the best tool for the job - well that depends what else is in the toolbox, and right now the toolbox doesn't have a lot else in it and dataflow analysis is the most powerful tool in there.

> We are surely in agreement that we should do transformations on the hugr first to make the analysis work better.

Generally I would follow the school of thought that analysis enables transformation, rather than the other way around...

PR #1677, opened 2024-11-22
feat: export llvm test utilities under llvm-test feature

### LINKED ISSUE #1673: hugr-llvm: Export testing utilities
Once https://github.com/CQCL/hugr/pull/1601 is merged:

hugr-llvm has some useful testing utilities in `src/test.hs` and `src/emit/test.rs`. In moving the lowering for the rotation extension to tket2 in https://github.com/CQCL/tket2/pull/697 we removed some tests because they could no longer access those testing utilities.

Let's export these tools from `hugr-llvm` so that downstream crates can benefit from them. Being test code, it's not as polished as we would want, so a bit of tidying/renaming/documenting will be required. We can guard these tools with a feature if there is a benefit in doing so(i.e. making some dependencies optional).

We could also put the code in a new crate (say `hugr-llvm-test`) and add that crate as a dev-dependency to `hugr-llvm`. This would be a circular dependency (hugr-llvm -> hugr-llvm-test -> hugr-llvm), but this is allowed as one of the edges is a dev-dependency edge. It's not clear to me what the benefit of this would be.

ss2165:
Closes #1673


ss2165:
I believe all the necessary things are documented, I did check docs builds locally

PR #1678, opened 2024-11-25
feat!: impl HugrView for any &(mut) to a HugrView

### LINKED ISSUE #1636: Reference to HugrView is not a HugrView
All(?) the methods in HugrView take `&self`, so to use it, `h: &impl HugrView` is all you need - easy for functions. However, for structs/storage, you have to decide between `MyStruct<H:HugrView>{ h: H .... }`, or `MyStruct<'a, H:HugrView> { h: &'a H ... }`, which is awkward and inflexible.

In practice this will probably mean replacing the current `impl<T:AsRef<Hugr>> HugrView for T` with *two* impls:
```
impl HugrView for Hugr { ... }
impl<T: HugrView> HugrView for &T
```
(possibly three, for `&mut T`, if we extend this to HugrMut, which we may want to too, although perhaps less important). Unfortunately this means more code/duplication, but the current situation is awkward for (increasing numbers of) clients....

acl-cqc:
closes #1636 .

Ideally we'd like a blanket implementation over all things that Deref to anything that's a HugrView, but AFAICS this isn't possible in Rust ATM, although might become so in the future given constrained HRTBs (higher-ranked type bounds). So, implement manually for `&T`, `&mut T`, `Rc<T>`, `Arc<T>`, `Box<T>`, `Cow<...,T>` and also `RootChecked`, using a macro in views/impls.rs.

BREAKING CHANGE: types which implement AsRef<Hugr> - both library ones such as Rc<Hugr> and custom ones - no longer get a blanket impl of HugrView. Workaround by manually calling `as_ref()` and using the `&Hugr` yourself.

### DIFF
--- a/hugr-core/src/hugr/views.rs
+++ b/hugr-core/src/hugr/views.rs
@@ -595,51 +595,61 @@ impl HugrView for Hugr {
     }
 }
 
-impl<T: HugrView> HugrView for &T {
-    delegate! {
+macro_rules! hugr_view_methods {
+    () => {
+        delegate! {
             to (**self) {
 
-        #[inline]
-        fn contains_node(&self, node: Node) -> bool;
+                #[inline]
+                fn contains_node(&self, node: Node) -> bool;
 
-        #[inline]
-        fn node_count(&self) -> usize;
+                #[inline]
+                fn node_count(&self) -> usize;
 
-        #[inline]
-        fn edge_count(&self) -> usize;
+                #[inline]
+                fn edge_count(&self) -> usize;
 
-        #[inline]
-        fn nodes(&self) -> impl Iterator<Item = Node> + Clone;
+                #[inline]
+                fn nodes(&self) -> impl Iterator<Item = Node> + Clone;
 
-        #[inline]
-        fn node_ports(&self, node: Node, dir: Direction) -> impl Iterator<Item = Port> + Clone;
+                #[inline]
+                fn node_ports(&self, node: Node, dir: Direction) -> impl Iterator<Item = Port> + Clone;
 
-        #[inline]
-        fn all_node_ports(&self, node: Node) -> impl Iterator<Item = Port> + Clone;
+                #[inline]
+                fn all_node_ports(&self, node: Node) -> impl Iterator<Item = Port> + Clone;
 
-        #[inline]
-        fn linked_ports(
-            &self,
-            node: Node,
-            port: impl Into<Port>,
-        ) -> impl Iterator<Item = (Node, Port)> + Clone;
+                #[inline]
+                fn linked_ports(
+                    &self,
+                    node: Node,
+                    port: impl Into<Port>,
+                ) -> impl Iterator<Item = (Node, Port)> + Clone;
 
-        #[inline]
-        fn node_connections(&self, node: Node, other: Node) -> impl Iterator<Item = [Port; 2]> + Clone;
+                #[inline]
+                fn node_connections(&self, node: Node, other: Node) -> impl Iterator<Item = [Port; 2]> + Clone;
 
-        #[inline]
-        fn num_ports(&self, node: Node, dir: Direction) -> usize;
+                #[inline]
+                fn num_ports(&self, node: Node, dir: Direction) -> usize;
 
-        #[inline]
-        fn children(&self, node: Node) -> impl DoubleEndedIterator<Item = Node> + Clone;
+                #[inline]
+                fn children(&self, node: Node) -> impl DoubleEndedIterator<Item = Node> + Clone;
 
-        #[inline]
-        fn neighbours(&self, node: Node, dir: Direction) -> impl Iterator<Item = Node> + Clone;
+                #[inline]
+                fn neighbours(&self, node: Node, dir: Direction) -> impl Iterator<Item = Node> + Clone;
 
-        #[inline]
-        fn all_neighbours(&self, node: Node) -> impl Iterator<Item = Node> + Clone;
+                #[inline]
+                fn all_neighbours(&self, node: Node) -> impl Iterator<Item = Node> + Clone;
+            }
+        }
     }
-    }
+}
+
+impl<T: HugrView> HugrView for &T {
+    hugr_view_methods! {}
+}
+
+impl<T: HugrView> HugrView for &mut T {
+    hugr_view_methods! {}
 }
 
 /// Trait implementing methods on port iterators.
--- a/hugr-core/src/hugr/views.rs
+++ b/hugr-core/src/hugr/views.rs
@@ -596,9 +596,9 @@ impl HugrView for Hugr {
 }
 
 macro_rules! hugr_view_methods {
-    () => {
+    ($arg:ident, $e:expr) => {
         delegate! {
-            to (**self) {
+            to ({let $arg=self; $e}) {
 
                 #[inline]
                 fn contains_node(&self, node: Node) -> bool;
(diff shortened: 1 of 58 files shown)


hugr-core/src/hugr/views.rs
```diff
@@ -557,57 +552,106 @@ impl<T: AsRef<Hugr>> HugrView for T {
...
     }
 }
 
+impl<T: HugrView> HugrView for &T {
+    hugr_view_methods! {}
+}
+
+impl<T: HugrView> HugrView for &mut T {
+    hugr_view_methods! {}
+}
```
aborgna-q, line 653: - There's no need to add `inline`s. `derive_more` takes care of that already.
- We normally single-line the methods when using `derive!`
- With the above, I don't think the macro simplifies the code much.
  Writing it explicitly avoids the extra layer of indirection.

```suggestion
impl<T: HugrView> HugrView for &T {
    delegate! {
        to (**self) {
            fn contains_node(&self, node: Node) -> bool;
            fn node_count(&self) -> usize;
            fn edge_count(&self) -> usize;
            fn nodes(&self) -> impl Iterator<Item = Node> + Clone;
            fn node_ports(&self, node: Node, dir: Direction) -> impl Iterator<Item = Port> + Clone;
            fn all_node_ports(&self, node: Node) -> impl Iterator<Item = Port> + Clone;
            fn linked_ports(&self, node: Node, port: impl Into<Port>) -> impl Iterator<Item = (Node, Port)> + Clone;
            fn node_connections(&self, node: Node, other: Node) -> impl Iterator<Item = [Port; 2]> + Clone;
            fn num_ports(&self, node: Node, dir: Direction) -> usize;
            fn children(&self, node: Node) -> impl DoubleEndedIterator<Item = Node> + Clone;
            fn neighbours(&self, node: Node, dir: Direction) -> impl Iterator<Item = Node> + Clone;
            fn all_neighbours(&self, node: Node) -> impl Iterator<Item = Node> + Clone;
        }
    }
}

impl<T: HugrView> HugrView for &mut T {
    delegate! {
        to (**self) {
            fn contains_node(&self, node: Node) -> bool;
            fn node_count(&self) -> usize;
            fn edge_count(&self) -> usize;
            fn nodes(&self) -> impl Iterator<Item = Node> + Clone;
            fn node_ports(&self, node: Node, dir: Direction) -> impl Iterator<Item = Port> + Clone;
            fn all_node_ports(&self, node: Node) -> impl Iterator<Item = Port> + Clone;
            fn linked_ports(&self, node: Node, port: impl Into<Port>) -> impl Iterator<Item = (Node, Port)> + Clone;
            fn node_connections(&self, node: Node, other: Node) -> impl Iterator<Item = [Port; 2]> + Clone;
            fn num_ports(&self, node: Node, dir: Direction) -> usize;
            fn children(&self, node: Node) -> impl DoubleEndedIterator<Item = Node> + Clone;
            fn neighbours(&self, node: Node, dir: Direction) -> impl Iterator<Item = Node> + Clone;
            fn all_neighbours(&self, node: Node) -> impl Iterator<Item = Node> + Clone;
        }
    }
}
```

hugr-core/src/hugr/views/root_checked.rs
```diff
@@ -45,6 +48,62 @@ impl<Root> RootChecked<&mut Hugr, Root> {
...
+            #[inline]
+            fn children(&self, node: Node) -> impl DoubleEndedIterator<Item = Node> + Clone;
+
+            #[inline]
+            fn neighbours(&self, node: Node, dir: Direction) -> impl Iterator<Item = Node> + Clone;
+
+            #[inline]
+            fn all_neighbours(&self, node: Node) -> impl Iterator<Item = Node> + Clone;
+        }
+    }
```
aborgna-q, line 104:
```suggestion
    delegate! {
        to self.as_ref() {
            fn contains_node(&self, node: Node) -> bool;
            fn node_count(&self) -> usize;
            fn edge_count(&self) -> usize;
            fn nodes(&self) -> impl Iterator<Item = Node> + Clone;
            fn node_ports(&self, node: Node, dir: Direction) -> impl Iterator<Item = Port> + Clone;
            fn all_node_ports(&self, node: Node) -> impl Iterator<Item = Port> + Clone;
            fn linked_ports(&self, node: Node, port: impl Into<Port>) -> impl Iterator<Item = (Node, Port)> + Clone;
            fn node_connections(&self, node: Node, other: Node) -> impl Iterator<Item = [Port; 2]> + Clone;
            fn num_ports(&self, node: Node, dir: Direction) -> usize;
            fn children(&self, node: Node) -> impl DoubleEndedIterator<Item = Node> + Clone;
            fn neighbours(&self, node: Node, dir: Direction) -> impl Iterator<Item = Node> + Clone;
            fn all_neighbours(&self, node: Node) -> impl Iterator<Item = Node> + Clone;
        }
    }
```

aborgna-q:
Any `impl<any kind of bound> HugrView for T` is a blanket implementation.
Unless you are in nightly, blanket implementations are **unique** so they prevent any other impl from being written (since an external definition would be able to satisfy the overlapping constraints otherwise).

I don't think the delegate block is annoying enough to merit much more workarounds. We can add the handful implementations to `Rc`, `Cow`, ... in addition to the ones here (in a submodule!) and be done with it.

hugr-core/src/hugr/views.rs
acl-cqc: > * There's no need to add `inline`s. `derive_more` takes care of that already.

Sounds good, but how does `derive_more` get involved here? I took the `[inline]`s from the old `impl<T: AsRef<Hugr>> HugrView for T {`...

I've now managed to common up all three of the delegations with the same macro...

aborgna-q: Sorry, _s/derive_more/delegate/g_ -.-'

acl-cqc: Ah....right. (That makes sense!)

Ok, so if `delegate` automatically adds `[inline]`....then (a) that shortens the macro :), and (b) yes it does reduce the utility of the macro a bit. OTOH there are now three instances of it, so...we can drop it....but do you really want to? :)

Note the macro is local to `hugr::views`, which I think works quite well, but we'd have to put `Rc` / `Cow` / `Arc` versions thereof in `views::refs` or something - not a bad place I guess. If we do those as well it'd be 5-6 instances of the macro.... (but they are easy to add, even in this PR)

acl-cqc: Done!

hugr-core/src/hugr/views/root_checked.rs
acl-cqc: Ok to resolve?

acl-cqc:
> Any `impl<any kind of bound> HugrView for T` is a blanket implementation. Unless you are in nightly, blanket implementations are **unique** so they prevent any other impl from being written (since an external definition would be able to satisfy the overlapping constraints otherwise).
> 
> I don't think the delegate block is annoying enough to merit much more workarounds. We can add the handful implementations to `Rc`, `Cow`, ... in addition to the ones here (in a submodule!) and be done with it.

I've added `Rc` and `Arc` but couldn't get `Cow` to work. I would love the blanket impl for Deref, so tried in commit 743aac7e56870cce1a75b498d72632f7bdb74e90, but couldn't figure out the lifetime issues. (Seems like the lifetime I need to use to tell the borrow checker that everything is ok...can't be named?!)

acl-cqc:
> I've been looking at how this could improve some of the `hugr: &impl HugrView` parameters in `hugr-llvm`.
> 
> I think the following is representative of the code this change is supposed to enable.....can you see how to make it work?

You had me worried for a moment - but then I found that I had checked this did what I want in the dataflow PR - I've now added a test/demo, does that help?

acl-cqc:
Ok I find I am hating this myself but it still seems like the only option that can be made to work :-(. @aborgna-q can you have another look please?

aborgna-q (approved):

PR #1684, opened 2024-11-25
fix: unused import from #1683

ss2165:
fix #1683


aborgna-q (approved):
I guess clippy is not checking these on CI

ss2165:
It should be?? The clippy command used in CI failed for me locally

PR #1717, opened 2024-11-26
feat: Lower array repeat and scan ops

### LINKED ISSUE #1681: Lower array `scan` and `repeat` ops

mark-koch:
Closes #1681


hugr-llvm/src/extension/prelude/array.rs
```diff
@@ -12,29 +15,26 @@ use hugr_core::{
...
 
-/// Helper function to allocate an array on the stack and pass a pointer to it
-/// to a closure.
+/// Helper function to allocate an array on the stack.
 ///
-/// The pointer forwarded to the closure is a pointer to the first element of
-/// the array. I.e. it is of type `array.get_element_type().ptr_type()` not
-/// `array.ptr_type()`
-fn with_array_alloca<'c, T, E: From<BuilderError>>(
+/// Returns two pointers: The first one
```
croyzor, line 33: This comment looks unfinished

hugr-llvm/src/extension/prelude/array.rs
```diff
@@ -779,4 +938,191 @@ mod test {
...
         assert_eq!(expected, exec_ctx.exec_hugr_u64(hugr, "main"));
     }
+
+    #[rstest]
+    #[case(5, 42, 0)]
+    #[case(5, 42, 1)]
+    #[case(5, 42, 2)]
+    #[case(5, 42, 3)]
+    #[case(5, 42, 4)]
+    fn exec_repeat(
```
croyzor, line 948: This isn't a very satisfying test -- it'd be nicer to have something where the function inspects the inputs and puts a different value in each location

hugr-llvm/src/extension/prelude/array.rs
```diff
@@ -779,4 +938,191 @@ mod test {
...
+
+    #[rstest]
+    #[case(10, 1)]
+    #[case(10, 2)]
+    #[case(0, 1)]
+    fn exec_scan_map(mut exec_ctx: TestContext, #[case] size: u64, #[case] inc: u64) {
+        // We build a HUGR that:
+        // - Creates an array [1, 2, 3, ..., size]
+        // - Maps a function that increments each element by `inc`
+        // - Returns the sum of the
```
croyzor, line 1002: the what?? 😛

hugr-llvm/src/extension/prelude/array.rs
```diff
@@ -779,4 +938,191 @@ mod test {
...
+                            pop_res,
+                        )
+                        .unwrap();
+                    arr = new_arr;
+                    r = builder.add_iadd(6, r, elem).unwrap();
+                }
+                builder.finish_with_outputs([r]).unwrap()
+            });
+        exec_ctx.add_extensions(|cge| cge.add_default_prelude_extensions().add_int_extensions());
+        let expected: u64 = (inc..size + inc).sum();
```
croyzor, line 1065: I don't think this should work for the third case? We're summing `map (n+) (0..0)`, which should be 0, but this would have `expected = 1`

hugr-llvm/src/extension/prelude/array.rs
mark-koch: That's not possible with repeat, all values in the array must have the same value

croyzor: Ah I see, I was thinking repeat was haskell's `Data.List.iterate`

croyzor (approved):
Approved once you address my comment on the `map` test

hugr-llvm/src/extension/prelude/array.rs
mark-koch: `inc..0 + inc == inc..inc` which is an empty iterator, so shouldn't it be `expected = 0`?

croyzor: Yes you're right, somehow I read it as `(inc..size).sum() = inc`

PR #1718, opened 2024-11-26
fix: hierarchical simple replacement using insert_hugr

### LINKED ISSUE #1715: bug(core): simple replace does not insert children when replacement is nested
Only inserts immediate children of root in the replacement.

Demonstrated in test in this commit: https://github.com/CQCL/hugr/commit/c19132054f27865a11f9ae2534ad53985f2df02e

ss2165:
Closes #1715


aborgna-q (approved):
Nice! Looks cleaner now

PR #1719, opened 2024-11-26
feat!: OpDefs and TypeDefs keep a reference to their extension

aborgna-q:
This change was extracted from the work towards #1613.
Now `OpDef`s and `TypeDef`s keep a `Weak` reference to their extension's `Arc`.

This way we will be able to automatically set the extension requirements when adding operations, so we can get rid of `update_validate` and the explicit registries when building hugrs.

To implement this, the building interface for `Extension`s is sightly modified.
Once an `Arc` is built it cannot be modified without doing internal mutation.
But we need the `Arc`'s weak reference to define the ops and types.
Thankfully, we can use `Arc::new_cyclic` which provides us with a `Weak` ref at build time so we are able to define things as needed.

This is wrapped in a new `Extension::new_arc` method, so the user doesn't need to think about that.

BREAKING CHANGE: Renamed `OpDef::extension` and `TypeDef::extension` to `extension_id`. `extension` now returns weak references to the `Extension` defining them.
BREAKING CHANGE: `Extension::with_reqs` moved to `set_reqs`, which takes `&mut self` instead of `self`.
BREAKING CHANGE: `Extension::add_type` and `Extension::add_op` now take an extra parameter. See docs for example usage.
BREAKING CHANGE: `ExtensionRegistry::register_updated` and `register_updated_ref` are no longer fallible.

### DIFF
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -104,10 +104,7 @@ impl ExtensionRegistry {
     ///
     /// Takes an Arc to the extension. To avoid cloning Arcs unless necessary, see
     /// [`ExtensionRegistry::register_updated_ref`].
-    pub fn register_updated(
-        &mut self,
-        extension: impl Into<Arc<Extension>>,
-    ) -> Result<(), ExtensionRegistryError> {
+    pub fn register_updated(&mut self, extension: impl Into<Arc<Extension>>) {
         let extension = extension.into();
         match self.0.entry(extension.name().clone()) {
             btree_map::Entry::Occupied(mut prev) => {
@@ -119,7 +116,6 @@ impl ExtensionRegistry {
                 ve.insert(extension);
             }
         }
-        Ok(())
     }
 
     /// Registers a new extension to the registry, keeping most up to date if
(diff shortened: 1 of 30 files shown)


hugr-core/src/extension.rs
```diff
@@ -372,14 +410,63 @@ impl Extension {
...
+            }
+        });
+        match error {
+            Some(e) => Err(e),
+            None => Ok(ext),
         }
     }
 
+    /// Extend the requirements of this extension with another set of extensions.
+    pub fn set_reqs(&mut self, extension_reqs: impl Into<ExtensionSet>) {
```
mark-koch, line 465: nit: "set" doesn't sound like extending the existing reqs to me. Maybe something like `add_reqs`?

mark-koch (approved):
Nice!

I think you missed the breaking change that `ExtensionRegistry::register_updated` and `ExtensionRegistry::register_updated_ref` are no longer fallible?

hugr-core/src/extension.rs
aborgna-q: Renamed it to `add_requirements`

Issue #1323, opened 2024-07-17
Unwrap panic on `SimpleReplace`


aborgna-q:
While trying to replace the two nodes in the middle of this diagram
```mermaid
graph LR
    subgraph 0 ["(0) DFG"]
        direction LR
        1["(1) Input"]
        2["(2) Output"]
        3["(3) MakeTuple"]
        4["(4) UnpackTuple"]
        1--"0:0<br>[]+[]"-->3
        1--"1:1<br>[]+[]"-->3
        3--"0:0<br>[[]+[], []+[]]"-->4
        3--"0:2<br>[[]+[], []+[]]"-->2
        4--"0:0<br>[]+[]"-->2
        4--"1:1<br>[]+[]"-->2
    end
```
with this graph
```mermaid
graph LR
    subgraph 0 ["(0) DFG"]
        direction LR
        1["(1) Input"]
        2["(2) Output"]
        3["(3) MakeTuple"]
        1--"0:0<br>[]+[]"-->3
        1--"0:0<br>[]+[]"-->2
        1--"1:1<br>[]+[]"-->3
        1--"1:1<br>[]+[]"-->2
        3--"0:2<br>[[]+[], []+[]]"-->2
    end
```

I got an `unwrap` panic on [SimpleReplacement::apply:158](https://github.com/CQCL/hugr/blob/4e0ad64c0a53155bb3b9936d86fe8e3a2bfb9e3d/hugr-core/src/hugr/rewrite/simple_replace.rs#L158)

I wrote a pure-hugr failing test in
https://github.com/CQCL/hugr/blob/4e0ad64c0a53155bb3b9936d86fe8e3a2bfb9e3d/hugr-core/src/hugr/rewrite/simple_replace.rs#L720-L721

PR #1724, opened 2024-11-27
refactor: avoid hugr clone in simple replace

ss2165:
Sercond commit is optional, see PR comment on comment

### DIFF
--- a/hugr-core/src/hugr/rewrite/simple_replace.rs
+++ b/hugr-core/src/hugr/rewrite/simple_replace.rs
@@ -1,13 +1,13 @@
 //! Implementation of the `SimpleReplace` operation.
 
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 
 use crate::hugr::hugrmut::InsertionResult;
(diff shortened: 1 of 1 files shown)


hugr-core/src/hugr/rewrite/simple_replace.rs
```diff
@@ -161,23 +195,17 @@ impl Rewrite for SimpleReplacement {
                 //
                 // Otherwise, we might disconnect other wires in `rem_inp_node`
                 // that are needed for the following iterations.
-                connect.insert((
+                h.connect(
```
ss2165, line 198: the comment above is also no longer true - but it is not clear to me how any disconnects might happen, and all tests (including those involving direct wires from input to output in the replacement) work

aborgna-q: This was meant to fix #1323. See #1324 / `test_half_nots`.
I'm not sure why that it's not a problem anymore, I'd have to look at the logic here for a bit.

aborgna-q (approved):

PR #1729, opened 2024-11-29
feat: add op for creating new/empty list

acl-cqc:
Constants cannot refer to type variables, so you can't have a constant for an empty `List<T>`...
* Add `ListOp::new` to handle this
* Constant-fold it to an empty ListValue *if* the element type has no type_vars
* (Most of the PR) add new functions (kept `pub(crate)` for now) for whether a Type, TypeArg or ExtensionSet contains variables.

I'm a bit, ugh, do we have to do this; when we move to hugr-model, we'll have a better way, so we'll end up redoing this again. Hence I've not written any tests yet, but will do if we think we want this....?

### DIFF
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -681,6 +681,10 @@ impl ExtensionSet {
             },
         }))
     }
+
+    pub(crate) fn contains_vars(&self) -> bool {
+        self.0.iter().any(|id| as_typevar(id).is_some())
+    }
 }
 
 impl From<ExtensionId> for ExtensionSet {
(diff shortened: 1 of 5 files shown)


mark-koch (commented):
I think it would be worth doing this, but I don't feel strongly. Also, I'm not super familiar with the discussion around hugr model, so a second opinion might be good.

In case we add the op, we should also implement the llvm lowering

acl-cqc:
Sorry @mark-koch for dropping the ball here, I presume we have found other ways. Shout if this wants resurrecting!

PR #1730, opened 2024-11-29
add HugrView impls for Box and Cow

doug-q:


acl-cqc:
I think this is probably not useful for @tatiana-s to review, given it's into my PR #1678. Moreover, review of that PR could probably do with seeing the full amount of duplication involved, hence I've merged in there...

PR #1739, opened 2024-12-03
feat!: `used_extensions` calls for both ops and signatures

aborgna-q:
Adds methods
- `OpType::used_extensions(&self) -> Result<ExtensionRegistry, _>`
- `Signature::used_extensions(&self) -> Result<ExtensionRegistry, _>`
and tests these along with the code merged in #1735.

Moves the code from #1735 into `resolution::types_mut`, and adds a (quite-similar) non-mutable version in `::types` that only collects the extensions without modifying the `OpType`. 

Fixes the resolution not exploring types inside a `CustomType` type arguments.

drive-by: Implement `Display`, `Serialize`, and `::new` for `ExtensionRegistry`.
drive-by: `ExtensionSet` should take ids by value when inserting.
drive-by: Fix `Hugr::resolve_extension_defs` not scanning all the ops.

These changes were extracted from #1738.

BREAKING CHANGE: Removed `ExtensionRegistry::try_new`. Use `new` instead, and call `ExtensionRegistry::validate` to validate.
BREAKING CHANGE: `ExtensionSet::insert` and `singleton` take extension ids by value instead of cloning internally.

### DIFF
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -705,7 +705,7 @@ impl ExtensionSet {
     }
 
     /// Adds a extension to the set.
-    pub fn insert(&mut self, extension: &ExtensionId) {
+    pub fn insert(&mut self, extension: ExtensionId) {
         self.0.insert(extension.clone());
     }
 
@@ -733,7 +733,7 @@ impl ExtensionSet {
     }
 
     /// Create a extension set with a single element.
-    pub fn singleton(extension: &ExtensionId) -> Self {
+    pub fn singleton(extension: ExtensionId) -> Self {
         let mut set = Self::new();
         set.insert(extension);
         set
@@ -797,7 +797,25 @@ impl ExtensionSet {
 
 impl From<ExtensionId> for ExtensionSet {
     fn from(id: ExtensionId) -> Self {
-        Self::singleton(&id)
+        Self::singleton(id)
+    }
+}
+
+impl IntoIterator for ExtensionSet {
+    type Item = ExtensionId;
+    type IntoIter = std::collections::btree_set::IntoIter<ExtensionId>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+impl<'a> IntoIterator for &'a ExtensionSet {
+    type Item = &'a ExtensionId;
+    type IntoIter = std::collections::btree_set::Iter<'a, ExtensionId>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.iter()
     }
 }
 
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -52,8 +52,9 @@ pub struct ExtensionRegistry {
     /// A flag indicating whether the current set of extensions has been
     /// validated.
     ///
-    /// This is used to avoid re-validating the extensions every time they are
-    /// used, and is set to `false` whenever a new extension is added.
+    /// This is used to avoid re-validating the extensions every time the
+    /// registry is validated, and is set to `false` whenever a new extension is
+    /// added.
     valid: AtomicBool,
 }
 
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -3,13 +3,16 @@
 //! TODO: YAML declaration and parsing. This should be similar to a plugin
 //! system (outside the `types` module), which also parses nested [`OpDef`]s.
 
+use itertools::Itertools;
 pub use semver::Version;
+use serde::{Deserialize, Deserializer, Serialize};
 use std::collections::btree_map;
 use std::collections::{BTreeMap, BTreeSet};
-use std::fmt::{Debug, Display, Formatter};
+use std::fmt::Debug;
 use std::mem;
 use std::sync::{Arc, Weak};
 
+use derive_more::Display;
 use thiserror::Error;
 
 use crate::hugr::IdentList;
@@ -40,10 +43,22 @@ pub use type_def::{TypeDef, TypeDefBound};
 pub mod declarative;
 
 /// Extension Registries store extensions to be looked up e.g. during validation.
-#[derive(Clone, Debug, Default, PartialEq)]
+#[derive(Clone, Debug, Display, Default, PartialEq)]
+#[display("ExtensionRegistry[{}]", _0.keys().join(", "))]
 pub struct ExtensionRegistry(BTreeMap<ExtensionId, Arc<Extension>>);
 
 impl ExtensionRegistry {
+    /// Create a new empty extension registry.
+    ///
+    /// For a version that checks the validity of the extensions, see [`ExtensionRegistry::try_new`].
+    pub fn new(extensions: impl IntoIterator<Item = Arc<Extension>>) -> Self {
+        let mut res = Self::default();
+        for ext in extensions.into_iter() {
+            res.register_updated(ext);
+        }
+        res
+    }
+
     /// Gets the Extension with the given name
     pub fn get(&self, name: &str) -> Option<&Arc<Extension>> {
         self.0.get(name)
@@ -55,14 +70,12 @@ impl ExtensionRegistry {
     }
 
     /// Makes a new [ExtensionRegistry], validating all the extensions in it.
+    ///
+    /// For an unvalidated version, see [`ExtensionRegistry::new`].
     pub fn try_new(
         value: impl IntoIterator<Item = Arc<Extension>>,
     ) -> Result<Self, ExtensionRegistryError> {
-        let mut res = ExtensionRegistry(BTreeMap::new());
-
-        for ext in value.into_iter() {
-            res.register(ext)?;
-        }
+        let res = ExtensionRegistry::new(value);
 
         // Note this potentially asks extensions to validate themselves against other extensions that
         // may *not* be valid themselves yet. It'd be better to order these respecting dependencies,
@@ -205,6 +218,28 @@ impl Extend<Arc<Extension>> for ExtensionRegistry {
     }
 }
 
+// Encode/decode ExtensionRegistry as a list of extensions.
+// We can get the map key from the extension itself.
+impl<'de> Deserialize<'de> for ExtensionRegistry {
+    fn deserialize<D>(deserializer: D) -> Result<ExtensionRegistry, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let extensions: Vec<Arc<Extension>> = Vec::deserialize(deserializer)?;
+        Ok(ExtensionRegistry::new(extensions))
+    }
+}
+
+impl Serialize for ExtensionRegistry {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let extensions: Vec<Arc<Extension>> = self.0.values().cloned().collect();
+        extensions.serialize(serializer)
+    }
+}
+
 /// An Extension Registry containing no extensions.
 pub const EMPTY_REG: ExtensionRegistry = ExtensionRegistry(BTreeMap::new());
 
@@ -226,7 +261,7 @@ pub enum SignatureError {
     #[error("Invalid type arguments for operation")]
     InvalidTypeArgs,
     /// The Extension Registry did not contain an Extension referenced by the Signature
-    #[error("Extension '{missing}' not found. Available extensions: {}",
+    #[error("Extension '{missing}' is not part of the declared HUGR extensions [{}]",
         available.iter().map(|e| e.to_string()).collect::<Vec<_>>().join(", ")
     )]
     ExtensionNotFound {
@@ -614,7 +649,10 @@ pub enum ExtensionBuildError {
 }
 
 /// A set of extensions identified by their unique [`ExtensionId`].
-#[derive(Clone, Debug, Default, Hash, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[derive(
+    Clone, Debug, Display, Default, Hash, PartialEq, Eq, serde::Serialize, serde::Deserialize,
+)]
+#[display("[{}]", _0.iter().join(", "))]
 pub struct ExtensionSet(BTreeSet<ExtensionId>);
 
 /// A special ExtensionId which indicates that the delta of a non-Function
@@ -735,12 +773,6 @@ fn as_typevar(e: &ExtensionId) -> Option<usize> {
     match e.chars().next() {
         Some(c) if c.is_ascii_digit() => Some(str::parse(e).unwrap()),
         _ => None,
-    }
-}
-
-impl Display for ExtensionSet {
-    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
-        f.debug_list().entries(self.0.iter()).finish()
     }
 }
 
--- a/hugr-core/src/extension.rs
+++ b/hugr-core/src/extension.rs
@@ -10,6 +10,7 @@ use std::collections::btree_map;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::Debug;
 use std::mem;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Weak};
 
 use derive_more::Display;
@@ -43,14 +44,36 @@ pub use type_def::{TypeDef, TypeDefBound};
 pub mod declarative;
 
 /// Extension Registries store extensions to be looked up e.g. during validation.
-#[derive(Clone, Debug, Display, Default, PartialEq)]
-#[display("ExtensionRegistry[{}]", _0.keys().join(", "))]
-pub struct ExtensionRegistry(BTreeMap<ExtensionId, Arc<Extension>>);
+#[derive(Debug, Display, Default)]
+#[display("ExtensionRegistry[{}]", exts.keys().join(", "))]
+pub struct ExtensionRegistry {
+    /// The extensions in the registry.
+    exts: BTreeMap<ExtensionId, Arc<Extension>>,
+    /// A flag indicating whether the current set of extensions has been
+    /// validated.
+    ///
+    /// This is used to avoid re-validating the extensions every time they are
+    /// used, and is set to `false` whenever a new extension is added.
+    valid: AtomicBool,
+}
+
+impl PartialEq for ExtensionRegistry {
+    fn eq(&self, other: &Self) -> bool {
+        self.exts == other.exts
+    }
+}
+
+impl Clone for ExtensionRegistry {
+    fn clone(&self) -> Self {
+        Self {
+            exts: self.exts.clone(),
+            valid: self.valid.load(Ordering::Relaxed).into(),
+        }
+    }
+}
 
 impl ExtensionRegistry {
     /// Create a new empty extension registry.
-    ///
-    /// For a version that checks the validity of the extensions, see [`ExtensionRegistry::try_new`].
     pub fn new(extensions: impl IntoIterator<Item = Arc<Extension>>) -> Self {
         let mut res = Self::default();
         for ext in extensions.into_iter() {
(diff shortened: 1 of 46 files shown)


aborgna-q:
Coverage is quite bad. I'll add some extra test today.

hugr-core/src/extension/resolution/types_mut.rs
```diff
@@ -0,0 +1,212 @@
...
+            update_type_row_exts(node, &mut db.other_outputs, extensions, used_extensions)?;
+            for row in db.sum_rows.iter_mut() {
+                update_type_row_exts(node, row, extensions, used_extensions)?;
+            }
+        }
+        OpType::ExitBlock(e) => {
+            update_type_row_exts(node, &mut e.cfg_outputs, extensions, used_extensions)?;
+        }
+        OpType::TailLoop(tl) => {
+            update_type_row_exts(node, &mut tl.just_inputs, extensions, used_extensions)?;
```
acl-cqc, line 93: Hmmm, so the trick with just scanning the signature doesn't work here. I'm tempted to suggest a method on OpType `fn all_types_mut(&mut self) -> Iterator<Item = &mut Type>` but not sure that'd be an improvement (it's probably much like this, in contrast to using Signature that exists already)

hugr-core/src/extension/resolution/types_mut.rs
```diff
@@ -0,0 +1,212 @@
...
+        TypeEnum::Function(f) => {
+            update_type_row_exts(node, &mut f.input, extensions, used_extensions)?;
+            update_type_row_exts(node, &mut f.output, extensions, used_extensions)?;
+        }
+        TypeEnum::Sum(SumType::General { rows }) => {
+            for row in rows.iter_mut() {
+                update_type_row_exts(node, row, extensions, used_extensions)?;
+            }
+        }
+        _ => {}
```
acl-cqc, line 188: Ditto maybe list them

hugr-core/src/hugr.rs
```diff
@@ -213,7 +213,7 @@ impl Hugr {
         //
         // This is not something we want to expose it the API, so we manually
         // iterate instead of writing it as a method.
-        for n in 0..self.node_count() {
+        for n in 0..self.graph.node_capacity() {
```
acl-cqc, line 216: This looks like it belongs in a different PR

hugr-core/src/hugr/rewrite/simple_replace.rs
```diff
@@ -793,7 +793,7 @@ pub(in crate::hugr::rewrite) mod test {
         };
         rewrite.apply(&mut hugr).unwrap_or_else(|e| panic!("{e}"));
 
-        assert_eq!(hugr.update_validate(&PRELUDE_REGISTRY), Ok(()));
+        assert_eq!(hugr.update_validate(&test_quantum_extension::REG), Ok(()));
```
acl-cqc, line 796: Whereas this might or might not be related, I'm really not sure. If it is required, sorry for noise!

hugr-core/src/types/signature.rs
```diff
@@ -118,6 +120,28 @@ impl<RV: MaybeRV> FuncTypeBase<RV> {
...
+    /// Returns a registry with the concrete extensions used by this signature.
+    ///
+    /// Note that extension type parameters are not included, as they have not
+    /// been instantiated yet.
+    ///
+    /// This method only returns extensions actually used by the types in the
+    /// signature. The extension deltas added via [`Self::with_extension_delta`]
+    /// refer to _runtime_ extensions, which may not be in all places that
+    /// manipulate a HUGR.
+    pub fn used_extensions(&self) -> Result<ExtensionRegistry, ExtensionCollectionError> {
```
acl-cqc, line 133: So the other way to do this - is to put the `collect_blah_exts` methods as impls of a single method in a trait, and then you can add a trait-default method (or a method taking an `impl ExtensionCollectionTrait`) that does this...that'd get you the two `used_extensions` methods that you have plus a load more with only having to write this once. (I wasn't going to suggest the trait approach until I saw this...)

hugr-core/src/extension/resolution/types.rs
```diff
@@ -1,174 +1,214 @@
...
+/// the missing extension names.
 ///
-/// This is a helper function used right after deserializing a Hugr.
-pub fn update_op_types_extensions(
-    node: Node,
-    op: &mut OpType,
-    extensions: &ExtensionRegistry,
-    used_extensions: &mut ExtensionRegistry,
-) -> Result<(), ExtensionResolutionError> {
+/// Use [`collect_op_types_extensions`] instead to update the weak references and
```
acl-cqc, line 25: Think this should point to `update_` something, rather than (AFAICS) the method this doc comment is on?

hugr-core/src/extension/resolution/test.rs
```diff
@@ -0,0 +1,292 @@
...
+    func.finish_with_outputs(vec![]).unwrap();
+    let mut hugr = module
+        .finish_hugr(&build_extensions)
+        .unwrap_or_else(|e| panic!("{e}"));
+
+    // Check that the read-only methods collect the same extensions.
+    let mut collected_exts = ExtensionRegistry::default();
+    for node in hugr.nodes() {
+        let op = hugr.get_optype(node);
+        collected_exts.extend(collect_op_extensions(Some(node), op).unwrap());
```
acl-cqc, line 224: If you changed OpType::used_extensions to take the `Option<Node>` param as extra, which seems reasonable, then you could call that here and remove all direct calls to `collect_op_`(`type_`)`extensions`. Then hide those (or inline into the single caller `OpType::used_extensions`).

hugr-core/src/types/signature.rs
acl-cqc: This doesn't work if you change `OpType::used_extensions` to take the extra `Option<Node>` parameter I suggested elsewhere, though. So feel free to drop.

hugr-core/src/extension/resolution/types.rs
```diff
@@ -1,174 +1,214 @@
...
+/// ensure they point to valid extensions.
+///
+/// # Attributes
+///
+/// - `node`: The node where the operation is located, if available.
+///   This is used to provide context in the error message.
+/// - `op`: The operation to collect the extensions from.
+pub fn collect_op_types_extensions(
```
acl-cqc, line 33: Should this be `pub`? The nomenclature is a bit confusing here - it feels equivalent to `Signature::used_extensions` but actually it's only one part of `OpType::used_extensions`. However the other `collect` things take two `&mut`s. Should one really use this directly?

hugr-core/src/extension/resolution/types_mut.rs
```diff
@@ -0,0 +1,212 @@
+//! Resolve weak links inside `CustomType`s in an optype's signature, while
+//! collecting all used extensions.
+//!
+//! For a non-mutating option see [`super::collect_op_types_extensions`].
```
acl-cqc, line 4: I reckon the pub function here should be `OpType::used_extensions`.

hugr-core/src/extension.rs
```diff
@@ -40,10 +43,22 @@ pub use type_def::{TypeDef, TypeDefBound};
...
-#[derive(Clone, Debug, Default, PartialEq)]
+#[derive(Clone, Debug, Display, Default, PartialEq)]
+#[display("ExtensionRegistry[{}]", _0.keys().join(", "))]
 pub struct ExtensionRegistry(BTreeMap<ExtensionId, Arc<Extension>>);
 
 impl ExtensionRegistry {
+    /// Create a new empty extension registry.
+    ///
+    /// For a version that checks the validity of the extensions, see [`ExtensionRegistry::try_new`].
+    pub fn new(extensions: impl IntoIterator<Item = Arc<Extension>>) -> Self {
```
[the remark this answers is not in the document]
aborgna-q: I added `validate(&self)` and an internal `AtomicBool` that caches the validation status.

hugr-core/src/extension.rs
```diff
@@ -205,6 +218,28 @@ impl Extend<Arc<Extension>> for ExtensionRegistry {
...
 
+// Encode/decode ExtensionRegistry as a list of extensions.
+// We can get the map key from the extension itself.
+impl<'de> Deserialize<'de> for ExtensionRegistry {
+    fn deserialize<D>(deserializer: D) -> Result<ExtensionRegistry, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let extensions: Vec<Arc<Extension>> = Vec::deserialize(deserializer)?;
+        Ok(ExtensionRegistry::new(extensions))
```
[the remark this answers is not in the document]
aborgna-q: It's probably too restrictive to validate here. We may want to add extra extensions before `validate`ing.
This way it follows the same `create -> validate` pattern as `new()`,

hugr-core/src/extension/resolution/types.rs
```diff
@@ -1,174 +1,214 @@
...
+    node: Option<Node>,
+    op: &OpType,
+) -> Result<ExtensionRegistry, ExtensionCollectionError> {
+    let mut used = ExtensionRegistry::default();
+    let mut missing = HashSet::<ExtensionId>::new();
```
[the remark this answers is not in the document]
aborgna-q: Done. I also changed `ExtensionSet::insert` and `::singleton` to take ids by value, to avoid unnecessary cloning.

hugr-core/src/extension/resolution/types.rs
```diff
@@ -1,174 +1,214 @@
...
+            for variant in t.variants.iter() {
+                collect_type_row_exts(variant, &mut used, &mut missing)
             }
         }
         OpType::DataflowBlock(db) => {
-            update_type_row_exts(node, &mut db.inputs, extensions, used_extensions)?;
-            update_type_row_exts(node, &mut db.other_outputs, extensions, used_extensions)?;
-            for row in db.sum_rows.iter_mut() {
-                update_type_row_exts(node, row, extensions, used_extensions)?;
+            collect_type_row_exts(&db.inputs, &mut used, &mut missing);
```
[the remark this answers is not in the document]
aborgna-q: Yes, but that requires allocating a `Signature`, cloning all row elements, computing the concrete input/output rows, and immediately destructuring them to run this same logic.

hugr-core/src/extension/resolution/types.rs
```diff
@@ -1,174 +1,214 @@
...
-        OpType::DFG(dfg) => {
-            update_signature_exts(node, &mut dfg.signature, extensions, used_extensions)?
-        }
-        OpType::OpaqueOp(op) => {
-            update_signature_exts(node, op.signature_mut(), extensions, used_extensions)?
+            collect_signature_exts(lf.func_sig.body(), &mut used, &mut missing);
+            collect_signature_exts(&lf.signature, &mut used, &mut missing);
         }
+        OpType::DFG(dfg) => collect_signature_exts(&dfg.signature, &mut used, &mut missing),
+        OpType::OpaqueOp(op) => collect_signature_exts(&op.signature(), &mut used, &mut missing),
```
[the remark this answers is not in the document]
aborgna-q: Right. I added it both here and on `ExtensionOp`, including the mutable versions.

hugr-core/src/extension/resolution/types.rs
```diff
@@ -1,174 +1,214 @@
...
         }
         OpType::Case(case) => {
-            update_signature_exts(node, &mut case.signature, extensions, used_extensions)?;
+            collect_signature_exts(&case.signature, &mut used, &mut missing);
         }
         // Ignore optypes that do not store a signature.
         _ => {}
+    };
+
+    if missing.is_empty() {
```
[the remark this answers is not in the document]
aborgna-q: It's a bit less readable at-a-glance imho, but sure.

hugr-core/src/extension/resolution/types.rs
```diff
@@ -1,174 +1,214 @@
...
-    node: Node,
-    row: &mut TypeRowBase<RV>,
-    extensions: &ExtensionRegistry,
+/// # Attributes
+///
+/// - `row`: The type row to collect the extensions from.
+/// - `used_extensions`: A The registry where to store the used extensions.
+/// - `missing_extensions`: A set of `ExtensionId`s of which the
+///   `Weak<Extension>` pointer has been invalidated.
+fn collect_type_row_exts<RV: MaybeRV>(
```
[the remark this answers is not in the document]
aborgna-q: I think I prefer the explicitness here, it's not too much noise.

hugr-core/src/extension/resolution/types.rs
```diff
@@ -1,174 +1,214 @@
...
+                    // The extension has been dropped.
+                    // Register it in the missing set.
+                    missing_extensions.insert(custom.extension().clone());
+                }
+            }
         }
         TypeEnum::Function(f) => {
-            update_type_row_exts(node, &mut f.input, extensions, used_extensions)?;
-            update_type_row_exts(node, &mut f.output, extensions, used_extensions)?;
+            collect_type_row_exts(&f.input, used_extensions, missing_extensions);
```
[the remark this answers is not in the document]
aborgna-q: Yes, but again here we can avoid unnecessary allocs.

hugr-core/src/extension/resolution/types.rs
```diff
@@ -1,174 +1,214 @@
...
-        OpType::Const(_c) => {
-            // TODO: Is it OK to assume that `Value::get_type` returns a well-resolved value?
+            collect_signature_exts(&ext.signature(), &mut used, &mut missing)
         }
-        OpType::Input(inp) => {
-            update_type_row_exts(node, &mut inp.types, extensions, used_extensions)?
-        }
-        OpType::Output(out) => {
-            update_type_row_exts(node, &mut out.types, extensions, used_extensions)?
+        OpType::FuncDefn(f) => collect_signature_exts(f.signature.body(), &mut used, &mut missing),
```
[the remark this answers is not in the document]
aborgna-q: Those are just the kind of params, aren't they? There's no extension names to extract from there.

hugr-core/src/extension/resolution/types_mut.rs
aborgna-q: That's too specific to what we are doing here. This is an internal method that fiddles with the internal, `all_types_mut` wouldn't have much use outside of it.

hugr-core/src/hugr.rs
aborgna-q: It's actually a bugfix for the used_extensions resolution. `node_count` doesn't cover all the indices 🤦 

I'll add the drive-by comment

hugr-core/src/hugr/rewrite/simple_replace.rs
aborgna-q: Tests fail without this. Because using only `Prelude` is actually invalid... (same with all the other test fixes).

hugr-core/src/types/signature.rs
aborgna-q: I opted for keeping all the logic in a single module, as it's short and quite similar between all the ops.

hugr-core/src/extension/resolution/test.rs
aborgna-q: I was calling the resolve methods explicitly, but it's exactly the same as `op.used_exts`.
I just changed it to that.

hugr-core/src/extension/resolution/types.rs
aborgna-q: It's hidden inside the private `types` module. But I'll add the `(crate)` for clarity.

hugr-core/src/extension/resolution/types_mut.rs
aborgna-q: It's a private module. (Otherwise the docs check would complain)

hugr-core/src/extension/resolution/types.rs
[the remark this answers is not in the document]
acl-cqc: Ah yes right you are, I was thinking of the old days before we removed that lot for `TypeParam::String`. Jolly good.

hugr-core/src/extension/resolution/ops.rs
```diff
@@ -1,12 +1,50 @@
...
+/// [`ExtensionOp`]s store a [`Weak`] reference to their extension, which can be
+/// invalidated if the original `Arc<Extension>` is dropped. On such cases, we
+/// return an error with the missing extension names.
+///
+/// # Attributes
+///
+/// - `node`: The node where the operation is located, if available. This is
+///   used to provide context in the error message.
+/// - `op`: The operation to collect the extensions from.
+pub(crate) fn collect_op_extensions(
```
acl-cqc, line 28: super-super-nit: `collect_op_extension` or maybe `identify_op_extension`(?) since collecting kinda means more than one

hugr-core/src/extension/resolution/types.rs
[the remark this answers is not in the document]
acl-cqc: Hmmmm, even when we have `Cow<Signature>`? I mean, we are gonna be constructing that thing a few times, clearly it's gonna help to cache it in the op. OK don't add that for this PR, but if you're optimizing that anyway and could use that here...

hugr-core/src/hugr.rs
acl-cqc: Yes please do! The comment above does not explain (to me) why we are using `node_capacity`

hugr-core/src/std_extensions.rs
```diff
@@ -21,8 +21,10 @@ pub fn std_reg() -> ExtensionRegistry {
         collections::EXTENSION.to_owned(),
         logic::EXTENSION.to_owned(),
         ptr::EXTENSION.to_owned(),
-    ])
-    .unwrap()
+    ]);
+    reg.validate()
```
acl-cqc, line 25: Many examples here and elsewhere, I wonder if it's worth adding to ExtensionRegistry
```
   fn validated(mut self) -> Result<Self, ....> {
      self.validate()?;
      self
   }
```

acl-cqc (approved):
Generally - great :). I think all of these comments are now small and local, so looks good and thanks @aborgna-q , let's get it in :)

I think there is a bit of an issue that ideally one would like `Hugr::validate` to also validate the registry (if not cached in that `AtomicBool`), but that requires passing an `&mut ExtensionRegistry` which is gonna require changing everywhere (and annoyingly). Otherwise you could add methods to validate immutably (i.e. without updating the cached bool), I guess....

hugr-core/src/extension/resolution/types_mut.rs
acl-cqc: Ah ok, yes I missed that. Ok I'll hold some fire there then ;)

hugr-core/src/extension/resolution/types.rs
```diff
@@ -1,174 +1,214 @@
...
+            collect_type_row_exts(&f.output, used_extensions, missing_extensions);
         }
         TypeEnum::Sum(SumType::General { rows }) => {
-            for row in rows.iter_mut() {
-                update_type_row_exts(node, row, extensions, used_extensions)?;
+            for row in rows.iter() {
+                collect_type_row_exts(row, used_extensions, missing_extensions);
+            }
+        }
+        _ => {}
```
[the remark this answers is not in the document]
aborgna-q: Ah, I added the variants to the `OpType`s up top instead.
Fixed

hugr-core/src/extension.rs
```diff
@@ -141,31 +181,36 @@ impl ExtensionRegistry {
...
     /// Returns an iterator over the extensions ids in the registry.
     pub fn ids(&self) -> impl Iterator<Item = &ExtensionId> {
-        self.0.keys()
+        self.exts.keys()
     }
 
     /// Delete an extension from the registry and return it if it was present.
     pub fn remove_extension(&mut self, name: &ExtensionId) -> Option<Arc<Extension>> {
-        self.0.remove(name)
+        // Clear the valid flag so that the registry is re-validated.
```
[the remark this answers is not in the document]
aborgna-q: `remove` returns the `Option<Arc<Ext>>`, so moving it requires declaring local vars..

hugr-core/src/extension.rs
```diff
@@ -205,8 +250,33 @@ impl Extend<Arc<Extension>> for ExtensionRegistry {
...
+        S: serde::Serializer,
+    {
+        let extensions: Vec<Arc<Extension>> = self.exts.values().cloned().collect();
+        extensions.serialize(serializer)
+    }
+}
+
 /// An Extension Registry containing no extensions.
-pub const EMPTY_REG: ExtensionRegistry = ExtensionRegistry(BTreeMap::new());
+pub static EMPTY_REG: ExtensionRegistry = ExtensionRegistry {
```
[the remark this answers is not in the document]
aborgna-q: ExtensionRegistry now has internal mutability, so it **should not** be used as a const.
(it does compile, but using it just raises linting errors everywhere, with good reason)

hugr-core/src/std_extensions.rs
aborgna-q: We'll call `ExtRegistry::validate` on each `Hugr::validate` as soon as #1738 gets merged, so this shouldn't be needed.

PR #1745, opened 2024-12-09
feat: Print []+[] as Bool and [] as Unit in user-facing messages

### LINKED ISSUE #1736: Print `[] + []` as `bool` in user-facing messages
In `hugr-py`, `UnitSum` already has a custom `__repr__` for [`bool` and `unit`](https://github.com/CQCL/hugr/blob/88d8b2500986c68e72eb636c06ea3e9b0c746a97/hugr-py/src/hugr/tys.py#L325-L330).
This however is not implemented on the rust side, so validation errors can throw errors like:

```
RuntimeError: Error validating HUGR:
An operation FuncDefn contains invalid children:
The Input node of a DataflowParent has a signature of [qubit, []+[]],
which differs from the expected type row [qubit]. In parent Node(1), child Node(2)
```

See the [`Display` implementation](https://github.com/CQCL/hugr/blob/88d8b2500986c68e72eb636c06ea3e9b0c746a97/hugr-core/src/types.rs#L151) of `SumType`.

aborgna-q:
Closes #1736


hugr-core/src/types.rs
```diff
@@ -148,10 +148,16 @@ impl std::fmt::Display for SumType {
         }
 
         match self {
+            SumType::Unit { size: 1 } => write!(f, "Unit"),
+            SumType::Unit { size: 2 } => write!(f, "Bool"),
             SumType::Unit { size } => {
                 display_list_with_separator(repeat_n("[]", *size as usize), f, "+")
```
ss2165, line 154: I think this should use something other than square brackets since they are also used for type parameters/extensions

ss2165 (approved):

hugr-core/src/types.rs
aborgna-q: Type rows always get printed as `[t1, t2, ...]` since they are a _list_ of types. Should we change that?

ss2165: that makes sense, but a sum holding a empty rows shouldn't just be `[]+[]+...` right? Maybe just put in parens? Or something like `([]|[]|..)` ?

aborgna-q: Currently the sum of three rows where a single one is empty gets printed as
`[T]+[]+[T]`
if we empty them we get `Sum::Unit{3}`, displayed as
`[]+[]+[]`
adding parenthesis anywhere would be inconsistent with the non-empty case and it would look noisy when printed (keep in mind this is a `Display` impl, so being easy to read is more important that not being ambiguous).

Issue #1688, opened 2024-11-25
Implement emission for `TailLoop`


hugrbot:
Originally raised by @doug-q

PR #1749, opened 2024-12-10
feat(llvm): Tail loop emission

croyzor:
Work in progress. Will address #1688

### DIFF
--- a/hugr-llvm/src/emit/func.rs
+++ b/hugr-llvm/src/emit/func.rs
@@ -252,7 +252,7 @@ impl<'c, 'a, H: HugrView> EmitFuncContext<'c, 'a, H> {
         Ok(r)
     }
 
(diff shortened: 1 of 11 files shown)


croyzor:
I think the problem with this code (the reason it's failing the simple test anyway) is that it's not strict enough about when we should be expecting the data to be packed in a sum type or not.

* Entry to the whole thing should expect `#I, #X`
* Entry to the loop should also expect `#I, #X`
* The loop body should return `Sum(#I + #O), #X`
* The postprocessing step should then unpack the data and call either the "continue" branch or "break" branch. Hence it expects the `Sum(#I + #O), #X` emitted by the loop body
* Hence, the "continue" branch should expect `#I, #X`
* And the "break" branch should expect `#O, #X`
* The exit from the whole tail loop should be `#O, #X`
* I.e. the signature of the tail loop is `#I, #X -> #O, #X`

(Using terminology from the spec)

PR #1754, opened 2024-12-10
feat: Scoping rules and utilities for symbols, links and variables

zrho:
This PR introduces scoping for symbols, links and variables. It comes with utilities that can be used to resolve names appropriately. Moreover the model data structures are changed so that they always use direct references by indices instead of names in order to streamline the serialisation format.

### DIFF
--- a/hugr-core/src/import.rs
+++ b/hugr-core/src/import.rs
@@ -22,15 +22,12 @@ use crate::{
     Direction, Hugr, HugrView, Node, Port,
 };
 use fxhash::FxHashMap;
-use hugr_model::v0::{self as model, GlobalRef};
-use indexmap::IndexMap;
+use hugr_model::v0::{self as model};
 use itertools::Either;
 use smol_str::{SmolStr, ToSmolStr};
 use thiserror::Error;
 
 const TERM_JSON: &str = "prelude.json";
-
-type FxIndexMap<K, V> = IndexMap<K, V, fxhash::FxBuildHasher>;
 
 /// Error during import.
 #[derive(Debug, Clone, Error)]
@@ -76,22 +73,18 @@ pub fn import_hugr(
     module: &model::Module,
     extensions: &ExtensionRegistry,
 ) -> Result<Hugr, ImportError> {
-    let names = Names::new(module)?;
-
     // TODO: Module should know about the number of edges, so that we can use a vector here.
     // For now we use a hashmap, which will be slower.
-    let edge_ports = FxHashMap::default();
-
     let mut ctx = Context {
         module,
-        names,
         hugr: Hugr::new(OpType::Module(Module {})),
-        link_ports: edge_ports,
+        link_ports: FxHashMap::default(),
         static_edges: Vec::new(),
         extensions,
         nodes: FxHashMap::default(),
-        local_variables: IndexMap::default(),
+        local_vars: FxHashMap::default(),
         custom_name_cache: FxHashMap::default(),
+        region_scope: model::RegionId::default(),
     };
 
     ctx.import_root()?;
@@ -105,31 +98,28 @@ struct Context<'a> {
     /// The module being imported.
     module: &'a model::Module<'a>,
 
-    names: Names<'a>,
-
     /// The HUGR graph being constructed.
     hugr: Hugr,
 
     /// The ports that are part of each link. This is used to connect the ports at the end of the
     /// import process.
-    link_ports: FxHashMap<model::LinkRef<'a>, Vec<(Node, Port)>>,
+    link_ports: FxHashMap<(model::RegionId, model::LinkIndex), Vec<(Node, Port)>>,
 
     /// Pairs of nodes that should be connected by a static edge.
     /// These are collected during the import process and connected at the end.
     static_edges: Vec<(model::NodeId, model::NodeId)>,
 
-    // /// The `(Node, Port)` pairs for each `PortId` in the module.
-    // imported_ports: Vec<Option<(Node, Port)>>,
     /// The ambient extension registry to use for importing.
     extensions: &'a ExtensionRegistry,
 
     /// A map from `NodeId` to the imported `Node`.
     nodes: FxHashMap<model::NodeId, Node>,
 
-    /// The local variables that are currently in scope.
-    local_variables: FxIndexMap<&'a str, LocalVar>,
+    local_vars: FxHashMap<model::VarId, LocalVar>,
 
     custom_name_cache: FxHashMap<&'a str, (ExtensionId, SmolStr)>,
+
+    region_scope: model::RegionId,
 }
 
 impl<'a> Context<'a> {
@@ -166,25 +156,6 @@ impl<'a> Context<'a> {
             .ok_or_else(|| model::ModelError::RegionNotFound(region_id).into())
     }
 
-    /// Looks up a [`LocalRef`] within the current scope.
-    fn resolve_local_ref(
-        &self,
-        local_ref: &model::LocalRef,
-    ) -> Result<(usize, LocalVar), ImportError> {
-        let term = match local_ref {
-            model::LocalRef::Index(_, index) => self
-                .local_variables
-                .get_index(*index as usize)
-                .map(|(_, v)| (*index as usize, *v)),
-            model::LocalRef::Named(name) => self
-                .local_variables
-                .get_full(name)
-                .map(|(index, _, v)| (index, *v)),
-        };
-
-        term.ok_or_else(|| model::ModelError::InvalidLocal(local_ref.to_string()).into())
-    }
-
     fn make_node(
         &mut self,
         node_id: model::NodeId,
@@ -209,13 +180,16 @@ impl<'a> Context<'a> {
     }
 
     /// Associate links with the ports of the given node in the given direction.
-    fn record_links(&mut self, node: Node, direction: Direction, links: &'a [model::LinkRef<'a>]) {
+    fn record_links(&mut self, node: Node, direction: Direction, links: &'a [model::LinkIndex]) {
         let optype = self.hugr.get_optype(node);
         // NOTE: `OpType::port_count` copies the signature, which significantly slows down the import.
         debug_assert!(links.len() <= optype.port_count(direction));
 
         for (link, port) in links.iter().zip(self.hugr.node_ports(node, direction)) {
-            self.link_ports.entry(*link).or_default().push((node, port));
+            self.link_ports
+                .entry((self.region_scope, *link))
+                .or_default()
+                .push((node, port));
         }
     }
 
@@ -242,8 +216,9 @@ impl<'a> Context<'a> {
 
             if inputs.is_empty() || outputs.is_empty() {
                 return Err(error_unsupported!(
-                    "link {} is missing either an input or an output port",
-                    link_id
+                    "link {}#{} is missing either an input or an output port",
+                    link_id.0,
+                    link_id.1
                 ));
             }
 
@@ -279,60 +254,13 @@ impl<'a> Context<'a> {
         Ok(())
     }
 
-    fn with_local_socpe<T>(
-        &mut self,
-        f: impl FnOnce(&mut Self) -> Result<T, ImportError>,
-    ) -> Result<T, ImportError> {
-        let previous = std::mem::take(&mut self.local_variables);
-        let result = f(self);
-        self.local_variables = previous;
-        result
-    }
-
-    fn resolve_global_ref(
-        &self,
-        global_ref: &model::GlobalRef,
-    ) -> Result<model::NodeId, ImportError> {
-        match global_ref {
-            model::GlobalRef::Direct(node_id) => Ok(*node_id),
-            model::GlobalRef::Named(name) => {
-                let item = self
-                    .names
-                    .items
-                    .get(name)
-                    .ok_or_else(|| model::ModelError::InvalidGlobal(global_ref.to_string()))?;
-
-                match item {
-                    NamedItem::FuncDecl(node) => Ok(*node),
-                    NamedItem::FuncDefn(node) => Ok(*node),
-                    NamedItem::CtrDecl(node) => Ok(*node),
-                    NamedItem::OperationDecl(node) => Ok(*node),
-                }
-            }
-        }
-    }
-
-    fn get_global_name(&self, global_ref: model::GlobalRef<'a>) -> Result<&'a str, ImportError> {
-        match global_ref {
-            model::GlobalRef::Direct(node_id) => {
-                let node_data = self.get_node(node_id)?;
-
-                let name = match node_data.operation {
-                    model::Operation::DefineFunc { decl } => decl.name,
-                    model::Operation::DeclareFunc { decl } => decl.name,
-                    model::Operation::DefineAlias { decl, .. } => decl.name,
-                    model::Operation::DeclareAlias { decl } => decl.name,
-                    model::Operation::DeclareConstructor { decl } => decl.name,
-                    model::Operation::DeclareOperation { decl } => decl.name,
-                    _ => {
-                        return Err(model::ModelError::InvalidGlobal(global_ref.to_string()).into());
-                    }
-                };
-
-                Ok(name)
-            }
-            model::GlobalRef::Named(name) => Ok(name),
-        }
+    fn get_symbol_name(&self, node_id: model::NodeId) -> Result<&'a str, ImportError> {
+        let node_data = self.get_node(node_id)?;
+        let name = node_data
+            .operation
+            .symbol()
+            .ok_or(model::ModelError::InvalidSymbol(node_id))?;
+        Ok(name)
     }
 
     fn get_func_signature(
@@ -345,11 +273,12 @@ impl<'a> Context<'a> {
             _ => return Err(model::ModelError::UnexpectedOperation(func_node).into()),
         };
 
-        self.import_poly_func_type(*decl, |_, signature| Ok(signature))
+        self.import_poly_func_type(func_node, *decl, |_, signature| Ok(signature))
     }
 
     /// Import the root region of the module.
     fn import_root(&mut self) -> Result<(), ImportError> {
+        self.region_scope = self.module.root;
         let region_data = self.get_region(self.module.root)?;
 
         for node in region_data.children {
@@ -400,7 +329,7 @@ impl<'a> Context<'a> {
             }
 
             model::Operation::DefineFunc { decl } => {
-                self.import_poly_func_type(*decl, |ctx, signature| {
+                self.import_poly_func_type(node_id, *decl, |ctx, signature| {
                     let optype = OpType::FuncDefn(FuncDefn {
                         name: decl.name.to_string(),
                         signature,
@@ -419,7 +348,7 @@ impl<'a> Context<'a> {
             }
 
             model::Operation::DeclareFunc { decl } => {
-                self.import_poly_func_type(*decl, |ctx, signature| {
+                self.import_poly_func_type(node_id, *decl, |ctx, signature| {
                     let optype = OpType::FuncDecl(FuncDecl {
                         name: decl.name.to_string(),
                         signature,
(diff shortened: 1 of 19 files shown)


hugr-model/src/v0/text/parse.rs
```diff
@@ -627,6 +751,7 @@ impl<'a> ParseContext<'a> {
 
         for pair in filter_rule(pairs, Rule::param) {
             let param = pair.into_inner().next().unwrap();
+            let param_span = param.as_span();
```
ss2165, line 754: might as well only call this in the error closure?

hugr-core/tests/snapshots/model__roundtrip_call.snap
```diff
@@ -4,6 +4,10 @@ expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-call
 ---
 (hugr 0)
 
+(import prelude.json)
+
+(import arithmetic.int.types.int)
```
ss2165, line 9: i was hoping the import would make the subsequent uses more compact - is that not yet the case?

ss2165 (commented):
Looks much nicer than before thanks

hugr-model/src/v0/mod.rs
```diff
@@ -377,12 +410,34 @@ pub struct Region<'a> {
...
+pub struct RegionScope {
+    /// The number of links in the scope.
+    pub links: u32,
+    /// The number of ports in the scope.
+    pub ports: u32,
+}
+
+/// The link scope of a region.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+pub enum LinkScope {
```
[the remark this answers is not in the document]
zrho: Almost; `LinkScope` here has additional properties beyond whether the scope is closed. So there'd need to be another.

hugr-core/src/import.rs
```diff
@@ -865,6 +805,8 @@ impl<'a> Context<'a> {
             self.record_links(exit, Direction::Incoming, region_data.targets);
         }
 
+        self.region_scope = prev_region;
```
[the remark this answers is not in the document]
zrho: For this it would be possible. However it wouldn't be possible for stuff like `SymbolTable` due to borrow checker limitations. Using `enter` and `exit` for `SymbolTable` and the other scoping utilities is also not entirely ideal (a Python-like `with` would be cool), but it showed me that perhaps complexity-wise avoiding the closure isn't a terrible tradeoff.

hugr-model/src/v0/text/parse.rs
zrho: It'll probably be optimised to be that, but I agree that it might be nicer to be explicit about it and put it in the closure.

hugr-core/tests/snapshots/model__roundtrip_call.snap
zrho: At this point it makes subsequent uses more compact in the binary format, since every named reference is replaced with a reference by node index referring to the import node. It doesn't yet allow to "import as". That'd definitely be useful since `arithmetic.int.types.int` is a bit unwieldy. Since that feature can be added independently on top, I propose to do it in a later PR.

hugr-core/src/import.rs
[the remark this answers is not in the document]
ss2165: finite number of places to forget to exit I guess

hugr-core/tests/snapshots/model__roundtrip_call.snap
ss2165: sounds good yep 👍

hugr-model/src/v0/text/parse.rs
zrho: Ah, I remember now why I did that. The subsequent code calls `.into_inner` on `param`, which consumes `param`. Therefore `param` is no longer available when the error is constructed. So either we'd have to clone `param` to get the error span from it later or do it like it is now. The clone would have to `Rc` ref count increases, so it's probably no different performance wise. Also `pest` is rather slow anyway and parsing isn't on the hot path (currently).

hugr-model/capnp/hugr-v0.capnp
```diff
@@ -97,13 +101,22 @@ struct Operation {
...
+    sources @1 :List(LinkIndex);
+    targets @2 :List(LinkIndex);
     children @3 :List(NodeId);
     meta @4 :List(MetaItem);
     signature @5 :OptionalTermId;
+    scope @6 :RegionScope;
+}
+
+struct RegionScope {
+    links @0 :UInt32;
```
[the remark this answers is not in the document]
acl-cqc: Perhaps you could just cross-link to `RegionScope` in `mod.rs`

hugr-model/src/v0/scope/link.rs
```diff
@@ -0,0 +1,113 @@
...
+        if inserted {
+            scope.link_count += 1;
+        }
+
+        scope.port_count += 1;
+        LinkIndex::new(map_index - scope.link_stack)
+    }
+
+    /// Reset the link table to an empty state while preserving allocated memory.
+    pub fn clear(&mut self) {
```
acl-cqc, line 88: Is this worth having over just `x = LinkTable::default()` ?

hugr-model/src/v0/scope/link.rs
```diff
@@ -0,0 +1,113 @@
...
+    links: FxIndexSet<(RegionId, K)>,
+    scopes: Vec<LinkScope>,
+}
+
+impl<K> LinkTable<K>
+where
+    K: Copy + Eq + Hash,
+{
+    /// Create a new empty link table.
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
acl-cqc: (Indeed, you implement `Default` on top of this, but I think derive will do it, and then I don't see much need for `new()`. Unless we particularly want a `const fn new()` perhaps, but clearly we don't *yet*!)

hugr-core/src/import.rs
```diff
@@ -279,60 +254,13 @@ impl<'a> Context<'a> {
...
-                    _ => {
-                        return Err(model::ModelError::InvalidGlobal(global_ref.to_string()).into());
-                    }
-                };
-
-                Ok(name)
-            }
-            model::GlobalRef::Named(name) => Ok(name),
-        }
+    fn get_global_name(&self, node_id: model::NodeId) -> Result<&'a str, ImportError> {
```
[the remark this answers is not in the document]
zrho: It doesn't. Fixed.

hugr-model/src/v0/scope/symbol.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl Default for SymbolTable<'_> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+struct Binding {
```
acl-cqc, line 173: If you put these `struct`s up after the `struct SymbolTable` but before `impl SymbolTable` I think it'd make this much easier to read, in that you'd see things in a order useful for comprehension

hugr-model/src/v0/scope/symbol.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// The symbol table always points to the currently visible binding for a
+    /// symbol. Therefore this index is only valid if this binding is not shadowed.
+    /// In particular, we detect shadowing by checking if the entry in the symbol
+    /// table at this index does indeed point to this binding.
+    symbol_index: SymbolIndex,
+}
+
+#[derive(Debug, Clone, Copy)]
+struct Scope {
+    /// The length of the `bindings` stack when this scope was entered.
```
acl-cqc, line 191: TBH I think this comment is the key here - you could just put this on `SymbolTable::scopes` and then elide `struct Scope` altogether...

hugr-model/src/v0/scope/vars.rs
```diff
@@ -0,0 +1,155 @@
...
+/// let mut vars = VarTable::new();
+/// vars.enter(NodeId(0));
+/// vars.insert("foo").unwrap();
+/// assert_eq!(vars.resolve("foo").unwrap(), VarId(NodeId(0), 0));
+/// assert!(!vars.is_visible(VarId(NodeId(0), 1)));
+/// vars.insert("bar").unwrap();
+/// assert!(vars.is_visible(VarId(NodeId(0), 1)));
+/// assert_eq!(vars.resolve("bar").unwrap(), VarId(NodeId(0), 1));
+/// vars.enter(NodeId(1));
+/// assert!(vars.resolve("foo").is_err());
```
acl-cqc, line 33: Again, nice example :) - this makes clear the difference vs. bindings

hugr-model/src/v0/scope/vars.rs
```diff
@@ -0,0 +1,155 @@
...
+/// ```
+#[derive(Debug, Clone)]
+pub struct VarTable<'a> {
+    vars: FxIndexSet<(NodeId, &'a str)>,
+    scopes: Vec<VarScope>,
+}
+
+impl<'a> VarTable<'a> {
+    /// Create a new empty variable table.
+    pub fn new() -> Self {
```
acl-cqc, line 47: derive-Default again

hugr-model/src/v0/scope/vars.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Returns an error if the variable is not defined in the current scope.
+    ///
+    /// # Panics
+    ///
+    /// Panics if there are no open scopes.
+    pub fn resolve(&self, name: &'a str) -> Result<VarId, UnknownVarError<'a>> {
+        let scope = self.scopes.last().unwrap();
+        let (set_index, _) = self
+            .vars
+            .get_full(&(scope.node, name))
```
acl-cqc, line 86: Does `get_full` return the pair? Can you do `.get` and then not have to discard the second element (as in `let (set_index, _) = ...`)?

hugr-model/src/v0/mod.rs
```diff
@@ -132,29 +133,39 @@ macro_rules! define_index {
...
     #[derive(Debug, derive_more::Display, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
     pub struct TermId(pub u32);
 }
 
+/// The id of a link consisting of its region and the link index.
+#[derive(Debug, derive_more::Display, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[display("{_0}#{_1}")]
+pub struct LinkId(pub RegionId, pub LinkIndex);
+
+/// The id of a variable consisting of its node and the variable index.
```
acl-cqc, line 164: Might be overkill but consider ultra-explicitness "The id of a variable being the node declaring the variable and the index within the variables declared by that node"

hugr-model/src/v0/text/parse.rs
```diff
@@ -50,33 +52,59 @@ pub fn parse<'a>(input: &'a str, bump: &'a Bump) -> Result<ParsedModule<'a>, Par
...
 
     fn parse_module(&mut self, pair: Pair<'a, Rule>) -> ParseResult<()> {
         debug_assert_eq!(pair.as_rule(), Rule::module);
         let mut inner = pair.into_inner();
+
+        self.module.root = self.module.insert_region(Region::default());
+        self.symbols.enter(self.module.root);
+        self.links.enter(self.module.root);
+
+        // TODO: What scope does the metadata live in?
```
acl-cqc, line 81: I'd guess that metadata attached to a node lives just outside the scope of that node itself (e.g. cannot refer to variables defined by that node, or stuff inside it), but can refer to anything outside it. That's a bit of a guess, though, happy to leave this as TODO.

hugr-model/src/v0/text/parse.rs
```diff
@@ -50,33 +52,59 @@ pub fn parse<'a>(input: &'a str, bump: &'a Bump) -> Result<ParsedModule<'a>, Par
...
         let meta = self.parse_meta(&mut inner)?;
+        let explicit_children = self.parse_nodes(&mut inner)?;
 
-        let children = self.parse_nodes(&mut inner)?;
+        let mut children = BumpVec::with_capacity_in(
```
acl-cqc, line 85: Can't this be something like `from_slice(explicit_children.into_iter().chain(self.implicit_imports.drain().....))`?

hugr-model/src/v0/text/parse.rs
```diff
@@ -495,25 +587,34 @@ impl<'a> ParseContext<'a> {
...
-
-        Ok(node_id)
+        Ok(node)
     }
 
-    fn parse_regions(&mut self, pairs: &mut Pairs<'a, Rule>) -> ParseResult<&'a [RegionId]> {
+    fn parse_regions(
+        &mut self,
+        pairs: &mut Pairs<'a, Rule>,
+        closed: bool,
```
acl-cqc, line 596: This might be worth defining an `enum RegionType {Open, Closed}`

hugr-core/tests/snapshots/model__roundtrip_constraints.snap
```diff
@@ -4,6 +4,8 @@ expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-cons
 ---
 (hugr 0)
 
+(import prelude.Array)
```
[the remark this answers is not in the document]
zrho: Agree. Before we actually stabilise on this, there needs to be a pass through to get consistent names.

hugr-model/capnp/hugr-v0.capnp
```diff
@@ -97,13 +101,22 @@ struct Operation {
...
+    ports @1 :UInt32;
 }
 
+# Either `0` for an open scope, or the number of links in the closed scope incremented by `1`.
```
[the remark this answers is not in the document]
zrho: Links don't have a global index space, but every closed region defines an index space. There can't be non-local edges across closed regions. This is a first step to simplify upcoming work on consts and nested hugrs, so that ideally closed regions can be loaded as `Hugr`s themselves.

hugr-model/src/v0/binary/read.rs
```diff
@@ -245,22 +211,35 @@ fn read_region<'a>(
...
 
-    let sources = read_list!(bump, reader, get_sources, read_link_ref);
-    let targets = read_list!(bump, reader, get_targets, read_link_ref);
+    let sources = read_scalar_list!(bump, reader, get_sources, model::LinkIndex);
+    let targets = read_scalar_list!(bump, reader, get_targets, model::LinkIndex);
     let children = read_scalar_list!(bump, reader, get_children, model::NodeId);
     let meta = read_list!(bump, reader, get_meta, read_meta_item);
     let signature = reader.get_signature().checked_sub(1).map(model::TermId);
 
+    let scope = if reader.has_scope() {
```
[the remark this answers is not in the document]
zrho: The `if` checks if the field is set to see if the region is open (field absent) or closed (field present). I'm not entirely sure if `get_scope` on an absent scope would be an error or return a default value, but either would not be desirable: An error would be undesirable since we want to distinguish between an open region and a corrupted file. A default value would make every region closed.

hugr-model/src/v0/binary/write.rs
```diff
@@ -31,8 +31,8 @@ fn write_module(mut builder: hugr_capnp::module::Builder, module: &model::Module
 
 fn write_node(mut builder: hugr_capnp::node::Builder, node: &model::Node) {
     write_operation(builder.reborrow().init_operation(), &node.operation);
-    write_list!(builder, init_inputs, write_link_ref, node.inputs);
-    write_list!(builder, init_outputs, write_link_ref, node.outputs);
+    let _ = builder.set_inputs(model::LinkIndex::unwrap_slice(node.inputs));
```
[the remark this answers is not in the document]
zrho: I think that the reason for the API to be able to return an error is as follows: you can pass a capnp reader object to a setter. The reader can point to invalid data, in which case the setter attempts to read that invalid data and must forward the error.

hugr-model/src/v0/mod.rs
```diff
@@ -132,29 +133,39 @@ macro_rules! define_index {
...
 
 define_index! {
-    /// Index of a node in a hugr graph.
+    /// Id of a node in a hugr graph.
     #[derive(Debug, derive_more::Display, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
     pub struct NodeId(pub u32);
 }
 
 define_index! {
     /// Index of a link in a hugr graph.
```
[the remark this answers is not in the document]
zrho: The idea was that an `Id` is globally unique while an index (like `VarIndex` as well) is valid within a scope. See `LinkId`, which combines `LinkIndex` with the `RegionId` of the closed region in which the link lives.

hugr-model/src/v0/mod.rs
```diff
@@ -290,17 +301,17 @@ pub enum Operation<'a> {
     /// becomes known by resolving the reference, the node can be transformed into a [`Operation::CustomFull`]
     /// by inferring terms for the implicit parameters or at least filling them in with a wildcard term.
     Custom {
-        /// The name of the custom operation.
-        operation: GlobalRef<'a>,
+        /// The symbol of the custom operation.
```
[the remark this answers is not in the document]
zrho: Yes, it's the operation declaration or the import.

It's intentional that this is not returned by `Operation::symbol`. `Operation::symbol` returns the name of the symbol that is *introduced* (i.e. imported, declared or defined) by the node. `Operation::Custom` and `Operation::CustomFull` do not introduce any symbols but merely refer to one.

hugr-model/src/v0/mod.rs
```diff
@@ -290,17 +301,17 @@ pub enum Operation<'a> {
...
     },
     /// Custom operation with full parameters.
     ///
     /// The node's parameters correspond to both the explicit and implicit parameters of the custom operation.
     /// Since this can be tedious to write, the [`Operation::Custom`] variant can be used to indicate that
     /// the implicit parameters should be inferred.
     CustomFull {
-        /// The name of the custom operation.
-        operation: GlobalRef<'a>,
+        /// The symbol of the custom operation.
```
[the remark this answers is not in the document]
zrho: See above.

hugr-model/src/v0/text/parse.rs
zrho: Yes, this is the indended scoping for metadata and it is implemented as such for almost all nodes. However this is the root node: everything, _including imports_, lives within the root node. So the metadata on the root node can not refer to anything...

hugr-model/src/v0/text/parse.rs
zrho: Afaik `chain` does not implement `ExactSizeIterator`...

hugr-model/src/v0/scope/link.rs
```diff
@@ -0,0 +1,113 @@
...
+/// let mut links = LinkTable::new();
+/// links.enter(RegionId(0));
+/// let foo_0 = links.use_link("foo");
+/// let bar_0 = links.use_link("bar");
+/// assert_eq!(foo_0, links.use_link("foo"));
+/// assert_eq!(bar_0, links.use_link("bar"));
+/// let (num_links, num_ports) = links.exit();
+/// assert_eq!(num_links, 2);
+/// assert_eq!(num_ports, 4);
+/// ```
```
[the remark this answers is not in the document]
zrho: Different links that are defined in different (closed) regions can have the same index. So the `assert_ne` would fail! A link is only uniquely identified by its index together with its closed region.

hugr-model/src/v0/scope/symbol.rs
```diff
@@ -0,0 +1,208 @@
...
+
+/// Symbol binding table that keeps track of symbol resolution and scoping.
+///
+/// Nodes may introduce a symbol so that other parts of the IR can refer to the
+/// node. Symbols have an associated name and are scoped via regions. A symbol
+/// can shadow another symbol with the same name from an outer region, but
+/// within any single region each symbol name must be unique.
+///
+/// When a symbol is referred to directly by the id of the node, the symbol must
+/// be in scope at the point of reference as if the reference was by name. This
```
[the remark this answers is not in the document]
zrho: The part "as if the reference was by name" is supposed to clarify this: the symbol is not allowed to be shadowed. If it was shadowed, it wouldn't be in scope when referenced by name.

hugr-model/src/v0/scope/symbol.rs
```diff
@@ -0,0 +1,208 @@
...
+    ///
+    /// Panics if there is no current scope.
+    pub fn insert(&mut self, name: &'a str, node: NodeId) -> Result<(), DuplicateSymbolError> {
+        let scope_depth = self.scopes.len() as u16 - 1;
+        let (symbol_index, shadowed) = self.symbols.insert_full(name, self.bindings.len());
+
+        if let Some(shadowed) = shadowed {
+            let (shadowed_node, shadowed_binding) = self.bindings.get_index(shadowed).unwrap();
+            if shadowed_binding.scope_depth == scope_depth {
+                self.symbols.insert(name, shadowed);
```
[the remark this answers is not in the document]
zrho: If you used `get` and `insert`, the happy path would have two hash lookups. This way the happy path has just one and the error path has two. We could use the `entry` API to get down to one lookup in both paths; I did this initially and the code was quite unwieldy. Decided ultimately to go with this since we're stopping at an error anyway. I've only reinserted the symbol in the error path since APIs that leave inconconsistent state on error cause me nightmares.

hugr-model/src/v0/scope/vars.rs
zrho: This code was written before I knew of `get_index_of`. I replaced it with `get_index_of` now.

PR #1762, opened 2024-12-11
feat: Add `PartialEq` impls for `FuncTypeBase` and `Cow<FuncTypeBase>`

doug-q:
Some existing out-of-tree tests are broken by the `Cow` changes. Seems better to add these than fix the callsites.


aborgna-q (approved):
👍

PR #1763, opened 2024-12-11
feat: Add `LoadNat` operation to enable loading generic `BoundedNat`s into runtime values

### LINKED ISSUE #1629: OpType to load BoundedNat / TypeArg into runtime value
Functions that are generic over bounded nats might want to use their static arguments as runtime values. For example, given an array of generic length `n`, we might want to write a loop that runs `n` times and does some indexing into the array. I think it would be worth to give up some parametricity to allow that.

I'd suggest to add a new op (to the prelude extension?) that loads these static arguments into a `usize` (?) runtime value. This op would be parametrised over the de Bruijn index of the argument it should load. During monomorphisation, we could then easily lower them to concrete constant values.

We could also consider extending this to `Tuple` or even `List` params?

tatiana-s:
Closes #1629


hugr-core/src/types/signature.rs
cqc-alec, line 1: Don't think these changes belong in this PR?

hugr-core/src/extension/prelude/generic.rs
```diff
@@ -0,0 +1,204 @@
...
+    #[test]
+    fn test_load_nat() {
+        let mut b = DFGBuilder::new(inout_sig(type_row![], vec![usize_t()])).unwrap();
+
+        let arg = TypeArg::BoundedNat { n: 4 };
+        let op = LoadNat::new(arg);
+
+        let out = b.add_dataflow_op(op, []).unwrap();
+
+        b.finish_hugr_with_outputs(out.outputs()).unwrap();
```
cqc-alec, line 184: Can we add some kind of assertion about the result, such as that is contains the expected operations?

cqc-alec:
Looks good but there seem to be some merge conflicts.

cqc-alec (approved):

Issue #1497, opened 2024-09-02
Move Array ops to a dedicated extension


doug-q:
I'm not sure we've got the new array ops naming quite right. The operation names (mostly) do not mention array  This would be fine if they were in an array extension, but in prelude we now have `prelude.get`, `prelude.set` . These are very general names for ops that only work on a specific type. I suggest creating an `array` , or `prelude.array` extension.

aborgna-q:
Also, add a value declaration in the extension

PR #1764, opened 2024-12-11
feat!: Rename `collections` extension to `collections.list`

aborgna-q:
In preparation to moving arrays out of the prelude (#1497), this PR renames the `collections` extension to `collections.list`. Arrays will then be called `collections.array`.

I renamed the related module and struct definitions as well.

BREAKING CHANGE: `collections` extension renamed to `collections.list`


ss2165 (approved):
diff is not kind

aborgna-q:
No... moved files + struct renames breaks the diff move detection :/

Issue #1766, opened 2024-12-11
SiblingSubgraph::try_from_nodes should not demand connected outputs in subgraph


ss2165:
Currently `try_from_nodes` expects the nodes in the subgraph to have singly connected outputs, this is too restrictive especially since we use it heavily for constructing rewrites. Rewrites on subrgaphs with discarded or multi-used outputs should be allowed.

PR #1769, opened 2024-12-11
fix: allow disconnected outputs in SiblingSubgraph::from_node

### LINKED ISSUE #1746: bug: `SiblingSubgraph::create_simple_replacement` compares signatures incorrectly
The function compares the signatures of the subgraph and replacement for equality:

https://github.com/CQCL/hugr/blob/4c1c6ee4c7d657c4bdb6b37c2237ae3f06b8d0be/hugr-core/src/hugr/views/sibling_subgraph.rs#L384-L385

However if the replacement had any disconnected outputs or multi-connected outputs this would fail because the current signature is calculated by iterating of the subgraph edges. 

Suggestion that this check is disabled as a temporary workaround and replaced with a correct signature computation for the current graph as a fix.

ss2165:
Uses `from_node` in lowering code to take advantage of this

Closes #1746

### DIFF
--- a/hugr-core/src/hugr/views/sibling_subgraph.rs
+++ b/hugr-core/src/hugr/views/sibling_subgraph.rs
@@ -785,9 +785,11 @@ mod tests {
     use cool_asserts::assert_matches;
 
     use crate::builder::inout_sig;
+    use crate::hugr::Rewrite;
     use crate::ops::Const;
     use crate::std_extensions::arithmetic::float_types::{self, ConstF64};
     use crate::std_extensions::logic::{self, LogicOp};
+    use crate::type_row;
     use crate::utils::test_quantum_extension::{self, cx_gate, rz_f64};
     use crate::{
         builder::{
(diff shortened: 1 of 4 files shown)


hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -277,9 +277,6 @@ impl SiblingSubgraph {
     ///
     /// The subgraph signature will be given by signature of the node.
     pub fn from_node(node: Node, hugr: &impl HugrView) -> Self {
-        // TODO once https://github.com/CQCL/portgraph/issues/155
```
ss2165, line 280: this comment is no longer true since the two methods have different functionality

If we can extend the functionality of `try_from_nodes` that would be great: #1766

aborgna-q:
Do you mind rebasing on `main`?

aborgna-q (approved):

PR #1770, opened 2024-12-11
feat!: Move arrays from prelude into new extension

### LINKED ISSUE #1497: Move Array ops to a dedicated extension

aborgna-q:
Moves the prelude array to a new `collections.array` extension.
This does not yet include the array value requested in https://github.com/CQCL/hugr/issues/1497#issuecomment-2527487618, I'll make a separate PR for that.

Closes #1497

BREAKING CHANGE: Array type and operations have been moved out of `prelude` and into a new `collections.array` extension.
BREAKING CHANGE: (py) `list_type` method replaced with `List` class.
BREAKING CHANGE: Removed `Array` type variant from the serialization format.


aborgna-q:
I may be going crazy, but this line in the failing `clippy` check does not exist?

hugr-py/src/hugr/std/collections/array.py
```diff
@@ -0,0 +1,34 @@
...
+
+EXTENSION = _load_extension("collections.array")
+
+
+@dataclass(eq=False)
+class Array(tys.ExtType):
+    """Fixed `size` array of `ty` elements."""
+
+    ty: tys.Type = field(default_factory=lambda: tys.Unit)
+    size: int = field(default=0)
```
ss2165, line 18: i think this stops a type arg being used here?

hugr-py/src/hugr/std/collections/list.py
```diff
@@ -12,10 +12,22 @@
 EXTENSION = _load_extension("collections.list")
 
 
-def list_type(ty: tys.Type) -> tys.ExtType:
-    """Returns a list type with a fixed element type."""
-    arg = tys.TypeTypeArg(ty)
-    return EXTENSION.types["List"].instantiate([arg])
+@dataclass(eq=False)
+class List(tys.ExtType):
```
ss2165, line 16: needs a test

hugr-py/src/hugr/std/collections/array.py
aborgna-q: Ah, yes. The parameter should be a `TypeArg` instead, and we cannot store a concrete type.

I'll add some tests

hugr-llvm/src/emit/test.rs
```diff
@@ -733,7 +731,7 @@ mod test_fns {
     ) {
         exec_ctx.add_extensions(add_int_extensions);
         assert_eq!(
-            input * 1 << (iters + 1),
+            input << (iters + 1),
```
ss2165: these should be in main now

ss2165 (approved):
:shipit:

Issue #1774, opened 2024-12-12
Instantiation fails for `ArrayOp`s with non-prelude element type


aborgna-q:
Failing test case, similar to `array_op::tests::test_get`.
```rust
    #[test]
    /// Initialize an array operation where the element type is not from the prelude.
    fn test_non_prelude_op() {
        let size = 2;
        let element_ty = float64_type();
        let op = ArrayOpDef::get.to_concrete(element_ty.clone(), size);

        let optype: OpType = op.into();

        let sig = optype.dataflow_signature().unwrap();

        assert_eq!(
            sig.io(),
            (
                &vec![array_type(size, element_ty.clone()), usize_t()].into(),
                &vec![option_type(element_ty.clone()).into()].into()
            )
        );
    }
```
This should get fixed if we remove the extension registries from validation.

PR #1779, opened 2024-12-12
fix: Resolve types in `Value`s and custom consts

### LINKED ISSUE #1742: `OpaqueValue` should keep a `Weak<Extension>` pointer
Similarly to what we did with `ExtensionOp`s and `CustomType`s, `OpaqueValue`/`CustomConst` should keep a weak pointer to their extension and provide a `used_extensions` call.

aborgna-q:
Collects and resolves extensions in the types stored inside a `Value`. This includes the specified type of a `Sum` and other cached types.

Since we expect `CustomConst`'s `get_type` method to always return signatures computed by a binary definition.

For some reason there's a random test on `hugr-py` that starts failing when we enable this:
```
=================================== FAILURES ===================================
______________________________ test_higher_order _______________________________
Error parsing package: Error resolving opaque operation: Error in signature of operation 'prelude.Noop' in Node(3): Type arguments of node did not match params declared by definition: Wrong number of type arguments: 0 vs expected 1 declared type parameters
```

The fix for #1774 that I'm submitting immediately after this PR fixes it so I'm skipping the test in this PR 🤷 
I'll open an issue to investigate further after we make the release

### DIFF
--- a/hugr-core/src/extension/resolution/types.rs
+++ b/hugr-core/src/extension/resolution/types.rs
@@ -44,9 +44,7 @@ pub(crate) fn collect_op_types_extensions(
         }
         OpType::FuncDefn(f) => collect_signature_exts(f.signature.body(), &mut used, &mut missing),
         OpType::FuncDecl(f) => collect_signature_exts(f.signature.body(), &mut used, &mut missing),
-        OpType::Const(c) => {
-            collect_value_exts(&c.value, &mut used, &mut missing);
-        }
+        OpType::Const(c) => collect_value_exts(&c.value, &mut used, &mut missing),
         OpType::Input(inp) => collect_type_row_exts(&inp.types, &mut used, &mut missing),
         OpType::Output(out) => collect_type_row_exts(&out.types, &mut used, &mut missing),
         OpType::Call(c) => {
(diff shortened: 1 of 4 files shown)


hugr-core/src/extension/resolution/types.rs
```diff
@@ -45,8 +45,7 @@ pub(crate) fn collect_op_types_extensions(
         OpType::FuncDefn(f) => collect_signature_exts(f.signature.body(), &mut used, &mut missing),
         OpType::FuncDecl(f) => collect_signature_exts(f.signature.body(), &mut used, &mut missing),
         OpType::Const(c) => {
-            let typ = c.get_type();
-            collect_type_exts(&typ, &mut used, &mut missing);
+            collect_value_exts(&c.value, &mut used, &mut missing);
         }
```
ss2165, line 49: nit: looks like the braces are no longer necessary

hugr-core/src/extension/resolution/types.rs
```diff
@@ -221,3 +220,37 @@ fn collect_typearg_exts(
         _ => {}
     }
 }
+
+/// Collect the Extension pointers in the [`CustomType`]s inside a value.
+///
+/// # Attributes
+///
+/// - `arg`: The type argument to collect the extensions from.
```
ss2165, line 228:
```suggestion
/// - `value`: The value to collect the extensions from.
```

ss2165 (approved):
thanks for the nice comments :)

PR #1780, opened 2024-12-12
feat: Update extension pointers in customConsts

### LINKED ISSUE #1742: `OpaqueValue` should keep a `Weak<Extension>` pointer

aborgna-q:
Closes #1742.

Adds a (default defined) `update_extensions` call to `CustomConst` so it can update its internal extension `Weak` pointers after loading a serialized hugr.


aborgna-q:
There are no array values yet, I covered all things that implement `CustomConst` at the moment.
I'll try to add a quick test

PR #1781, opened 2024-12-12
feat!: Use registries of `Weak<Extension>`s when doing resolution 

aborgna-q:
We'll need this to resolve extension references inside an extension that's being created.

drive-by: Make "node" optional in resolution errors.

BREAKING CHANGE: `::extension::resolve` operations now use `WeakExtensionRegistry`es.

### DIFF
--- a/hugr-core/src/extension/resolution.rs
+++ b/hugr-core/src/extension/resolution.rs
@@ -22,6 +22,9 @@
 mod ops;
 mod types;
 mod types_mut;
+mod weak_registry;
+
+pub use weak_registry::WeakExtensionRegistry;
 
 pub(crate) use ops::{collect_op_extension, resolve_op_extensions};
 pub(crate) use types::{collect_op_types_extensions, collect_signature_exts};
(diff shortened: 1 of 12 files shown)


hugr-core/src/extension/resolution/weak_registry.rs
```diff
@@ -0,0 +1,88 @@
...
+use itertools::Itertools;
+
+use derive_more::Display;
+
+use crate::extension::{ExtensionId, ExtensionRegistry};
+use crate::Extension;
+
+/// The equivalent to an [`ExtensionRegistry`] that only contains weak references.
+///
+/// This is used to resolve extensions pointers while
```
ss2165, line 13: while what?

hugr-core/src/extension/resolution/weak_registry.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub fn register(&mut self, id: ExtensionId, ext: impl Into<Weak<Extension>>) -> bool {
+        self.exts.insert(id, ext.into()).is_none()
+    }
+
+    /// Returns an iterator over the weak references in the registry.
+    pub fn iter(&self) -> impl Iterator<Item = &Weak<Extension>> {
+        self.exts.values()
+    }
+
+    /// Returns an iterator over the extensions ids in the registry.
```
ss2165, line 44:
```suggestion
    /// Returns an iterator over the extension ids in the registry.
```

hugr-core/src/extension/resolution.rs
```diff
@@ -111,13 +103,14 @@ pub enum ExtensionResolutionError {
         available_extensions: Vec<ExtensionId>,
     },
     #[display(
-        "Type {ty} in {node} requires extension {missing_extension}, but it could not be found in the extension list used during resolution. The available extensions are: {}",
+        "Type {ty}{} requires extension {missing_extension}, but it could not be found in the extension list used during resolution. The available extensions are: {}",
+        node.map(|n| format!(" in {}", n)).unwrap_or_default(),
```
ss2165, line 107: what's the default behaviour?

hugr-core/src/extension/resolution/types_mut.rs
```diff
@@ -20,13 +20,16 @@ use crate::Node;
 /// When a pointer is replaced, the extension is added to the
 /// `used_extensions` registry.
 ///
+/// Returns
```
ss2165, line 23: returns what?

ss2165 (commented):
How much use remains for a normal registry now?

hugr-core/src/extension/resolution.rs
aborgna-q: If there's a node: "Type Unit in Node(4) requires..."
If there's no node: "Type Unit requires ..."

ss2165: ah yeah sorry misread, thanks

aborgna-q:
> How much use remains for a normal registry now?

That's for the last PR in the chain!
`ExtensionRegistry` is still needed when we need as _someone_ has to have a strong pointer to the `Arc`s, but we'll remove most uses of them.

hugr-core/src/extension/resolution/weak_registry.rs
```diff
@@ -21,6 +21,15 @@ pub struct WeakExtensionRegistry {
 }
 
 impl WeakExtensionRegistry {
+    /// Create a new empty extension registry.
```
ss2165, line 24: doesn't look empty....

aborgna-q: Sorry for the hundred typos 🤦

ss2165 (approved):

PR #1783, opened 2024-12-13
feat!: Resolve extension references inside the extension themselves

aborgna-q:
When deserializing an `ExtensionRegistry`, traverses the extension definitions to update any `Weak<Extension>` so they point to valid extensions (including to extensions in the registry being deserialized!).

This needed a bit of `unsafe` trickery to mimic [`Arc::new_cyclic`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.new_cyclic) but for vectors of extensions.

As with `Package`, I left the `serde::Deserialize` implementation available but now extension registries have a `load_json` method that should be used when possible, as it runs the required pointer updates.

BREAKING CHANGE: Marked `ExtensionBuildError` and `ExtensionRegistryError` as non-exhaustive.

### DIFF
--- a/hugr-core/src/extension/resolution/test.rs
+++ b/hugr-core/src/extension/resolution/test.rs
@@ -15,7 +15,7 @@ use crate::extension::resolution::WeakExtensionRegistry;
 use crate::extension::resolution::{
     resolve_op_extensions, resolve_op_types_extensions, ExtensionCollectionError,
 };
-use crate::extension::{ExtensionId, ExtensionRegistry, ExtensionSet, PRELUDE};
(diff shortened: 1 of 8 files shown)


hugr-core/src/extension/resolution/test.rs
```diff
@@ -86,6 +88,33 @@ fn make_extension(name: &str, op_name: &str) -> (Arc<Extension>, OpType) {
     (ext, op.into())
 }
 
+/// Create a new test extension with a type and an op using that type
+///
+/// Returns an instance of the defined op.
```
ss2165, line 93:
```suggestion
/// Returns the defined extension.
```

ss2165 (approved):

PR #1784, opened 2024-12-13
feat!: Don't require explicit extension registers for validation

aborgna-q:
Now that we keep references to the extensions used around the hugr, we no longer need to pass `ExtensionRegistry`es when calling `validate` on ops, types, etc. This PR removes most uses of the registries.

drive-by: Fix `Noop` not declaring its type argument in `hugr-py`.

BREAKING CHANGE: Removed the extension registry argument from `validate` calls.
BREAKING CHANGE: Removed the extension registry argument from operation instantiation methods.


hugr-core/src/extension/resolution/test.rs
```diff
@@ -153,6 +154,64 @@ fn check_extension_resolution(mut hugr: Hugr) {
     );
 }
 
+/// Build a hugr with all possible op nodes and resolve the extensions.
```
ss2165, line 157: docstring is a lie

ss2165 (approved):
😍

PR #1802, opened 2024-12-17
feat: Cleanup `Display` of types and arguments

aborgna-q:
Updates the `Display` implementation of poly functions, functions, types, type arguments, and type params so its easier to read.

Comparison below. I added the newlines manually.

Before:
```swift
forall Any _0.map(|i|i.to_string()).unwrap_or("-".to_string()).
    [array([
        Variable { v: TypeArgVariable { idx: 1, cached_decl: BoundedNat { bound: UpperBound(None) } } },
        Type { ty: TypeBase(Sum(General { rows: [TypeRowBase { types: [] }, TypeRowBase { types: [TypeBase(Variable(0, Any), Any)] }] }), Any) }
    ])]
 -> [[]][[
    array([
            Variable { v: TypeArgVariable { idx: 1, cached_decl: BoundedNat { bound: UpperBound(None) } } },
            Type { ty: TypeBase(Sum(General { rows: [TypeRowBase { types: [] }, TypeRowBase { types: [TypeBase(Variable(0, Any), Any)] }] }), Any) }
        ]),
        int([BoundedNat { n: 6 }])
    ]]
```
(yes, that is a hardcoded rust code string)

After:
```swift
∀ (var#0 : Type) (var#1 : Nat). [array(var#1, Type([]+[var#0]))]
                             -> [[array(var#1, Type([]+[var#0])), int(6)]]
```

That is, a generic function on a type var and an unbounded natural that takes an array of (nat, option type) and returns a tuple with the array and an int.

It may be better if the tuple display used `()` instead of `[]`, but that can be discussed later.


aborgna-q:
I could also remove the `var` part of the variables, and just call them `#42` (and/or use some other reserved character)

hugr-core/src/types.rs
```diff
@@ -265,24 +265,27 @@ impl<RV: MaybeRV> From<SumType> for TypeBase<RV> {
...
+    /// A function type.
+    #[display("{_0}")]
     Function(Box<FuncValueType>),
-    // Index into TypeParams, and cache of TypeBound (checked in validation)
-    #[allow(missing_docs)]
-    #[display("Variable({_0})")]
+    /// A type variable, defined by an index into a list of type parameters.
+    //
+    // We cache the TypeBound here (checked in validation)
+    #[display("var#{_0}")]
```
mark-koch, line 282: I'd prefer to get rid of the `var` part

hugr-core/src/types/type_param.rs
```diff
@@ -144,29 +154,31 @@ impl From<UpperBound> for TypeParam {
 #[serde(tag = "tya")]
 pub enum TypeArg {
     /// Where the (Type/Op)Def declares that an argument is a [TypeParam::Type]
+    #[display("Type({ty})")]
```
mark-koch, line 157: Or just `{ty}`?

mark-koch (approved):
Very nice, this is a lot more readable!

PR #1809, opened 2024-12-18
feat: add MonomorphizePass and deprecate monomorphize

ss2165:


hugr-passes/src/monomorphize.rs
```diff
@@ -28,26 +29,33 @@ use itertools::Itertools as _;
...
     validate(&h);
 
+    monomorphize_ref(&mut h);
+
+    #[cfg(debug_assertions)]
+    validate(&h);
+    h
+}
+
+fn monomorphize_ref(h: &mut impl HugrMut) {
```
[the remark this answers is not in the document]
ss2165: this is a private function so we can rename to `monomorphize` when we remove the public one, will add comment

hugr-passes/src/lib.rs
```diff
@@ -7,7 +7,14 @@ mod half_node;
 pub mod lower;
 pub mod merge_bbs;
 mod monomorphize;
-pub use monomorphize::{monomorphize, remove_polyfuncs};
+// TODO: Deprecated re-export. Remove on a breaking release.
+#[deprecated(
+    since = "0.14.1",
+    note = "Use `hugr::algorithms::MonomorphizePass` instead."
```
acl-cqc, line 13: Where is `hugr::algorithms` ?

acl-cqc: 😮

Issue #1819, opened 2024-12-18
DFBs need `add_{in,out}put` like the `FunctionBuilder`


qartik:
Similar to https://github.com/CQCL/hugr/issues/1562

PR #1824, opened 2024-12-19
fix: allow add_input/add_output on all DFGs

qartik:
Closes: #1819

### DIFF
--- a/hugr-core/src/builder/dataflow.rs
+++ b/hugr-core/src/builder/dataflow.rs
@@ -79,6 +79,109 @@ impl DFGBuilder<Hugr> {
         let root = base.root();
         DFGBuilder::create_with_io(base, root, signature)
     }
+
+    /// Add a new input to the function being constructed.
+    ///
+    /// Returns the new wire from the input node.
+    pub fn add_input(&mut self, input_type: Type) -> Wire {
+        let [inp_node, _] = self.io();
+
+        // Update the parent's root type
+        let new_optype = self.update_fn_signature(|mut s| {
+            s.input.to_mut().push(input_type);
+            s
+        });
+
+        // Update the inner input node
+        let types = new_optype.signature.body().input.clone();
+        self.hugr_mut()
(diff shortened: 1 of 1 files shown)


hugr-core/src/builder/dataflow.rs
```diff
@@ -231,10 +152,10 @@ impl FunctionBuilder<Hugr> {
...
         // Update the builder metadata
-        self.0.num_out_wires += 1;
+        self.num_out_wires += 1;
     }
 
-    /// Update the function builder's parent signature.
+    /// Update the DFG builder's parent signature.
     ///
     /// Internal function used in [add_input] and [add_output].
     ///
```
qartik, line 161: `update_fn_signature` still needs to be updated or it will panic for anything other than `FuncDefn`.

acl-cqc: Yeah, so you've moved `add_input` and `add_output` from  `FunctionBuilder<Hugr>` into `DFGBuilder<Hugr>` and a proxy for that in `FunctionBuilder<Hugr>`. Instead you could define
```
trait MutableIO : Dataflow {
  fn update_signature(...); // no default impl
  fn add_input(...) { .... } // code you've moved becomes default impl delegating to self.update_signature
  fn add_output(...) { ....} // similarly
}
```
then you `impl MutableIO for FunctionBuilder<Hugr>` and `impl MutableIO for DFGBuilder<Hugr>` and just have to provide the two `update_signature` methods, i.e. then you don't need the `delegate! { to self.0 { fn add_input(...); fn add_output(...); } }`. Does that work??

aborgna-q:
I agree that `MutableIO` is too much of a footgun. It wouldn't hurt to keep it just with the limited `add_input` / `add_output` methods, and keep the rest of the implementation private to each type similar to @doug-q patch.

I made a couple comments in #1866 too.

qartik:
Not needed anymore.

PR #1827, opened 2024-12-20
fix(py): Fix array/list value serialization

### LINKED ISSUE #1826: `ArrayValue` created by hugr-py cannot be resolved by hugr-rs
See this test case for guppy:
```py
def test_array_const(validate, run_int_fn):
    module = GuppyModule("test")
    module.load_all(quantum)

    bools = [True, False]

    @guppy(module)
    def main() -> int:
        bs = py(bools)
        return int(bs[0])

    package = module.compile()
    validate(package)

    run_int_fn(package, expected=1)
```

We get an internal error during hugr-llvm execution because the constant array from `py(bools)` gets deserialized as a `CustomConst` instead of `ArrayValue`.

aborgna-q:
The rust-side serialization definition for custom constants is a bit more complex than what we had written on the python side for array and list values.

This fixes #1826, and was tested by running the issue's example from guppy.

- > Why aren't we testing this locally?

  Constant value deserialization is a "best effort" thing.
  If we can match the serialized thing with one of the hardcoded implementors of `CustomConst` (registered in a global var by `typetag`) then we deserialize into the specific class.
  If we cannot deserialize into any specific class, we use `CustomSerialized` by default.
  
  So even if we build a constant from python and `validate` using the rust checker, the deserialization will succeed and validate correctly.
  
  It'd be nice to fail the deserialization if a constant claims to be of some specific kind but cannot be deserialized into it. I'll open an issue and look a bit into it.

### DIFF
--- a/hugr-py/src/hugr/std/collections/array.py
+++ b/hugr-py/src/hugr/std/collections/array.py
@@ -60,7 +60,7 @@ class ArrayVal(val.ExtensionValue):
     """Constant value for a statically sized array of elements."""
 
     v: list[val.Value]
-    ty: tys.Type
+    ty: Array
 
     def __init__(self, v: list[val.Value], elem_ty: tys.Type) -> None:
         self.v = v
@@ -71,7 +71,11 @@ class ArrayVal(val.ExtensionValue):
         # The value list must be serialized at this point, otherwise the
         # `Extension` value would not be serializable.
         vs = [v._to_serial_root() for v in self.v]
-        return val.Extension(name, typ=self.ty, val=vs, extensions=[EXTENSION.name])
+        element_ty = self.ty.ty._to_serial_root()
+        serial_val = {"values": vs, "typ": element_ty}
+        return val.Extension(
+            name, typ=self.ty, val=serial_val, extensions=[EXTENSION.name]
+        )
 
     def __str__(self) -> str:
         return f"array({comma_sep_str(self.v)})"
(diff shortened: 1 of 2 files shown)


aborgna-q:
Will `hugr-model` redefine constant serialization? I didn't find references to that after a quick search

hugr-py/src/hugr/std/collections/array.py
```diff
@@ -60,7 +60,7 @@ class ArrayVal(val.ExtensionValue):
     """Constant value for a statically sized array of elements."""
 
     v: list[val.Value]
-    ty: tys.Type
+    ty: Array
```
ss2165, line 63: this is non-breaking because `Array` satisfies the Type protocol, correct?

ss2165 (approved):
thanks for this!

model will change constant serialization, to become static "recipes" for constructing runtime values. There is some discussion here: https://github.com/CQCL/hugr/discussions/1425.

ss2165:
have checked this test passes with this: https://github.com/quantinuum-dev/guppy-integration/blob/d8c5dfa6cd876d82dcc694da944690b16edce813/tests/test_simple.py#L147

hugr-py/src/hugr/std/collections/array.py
aborgna-q: It's fine for reading since it's a type refinement.

For _writing_ into the variable it is actually a breaking change. I assumed the dataclass was frozen, but it seems not.

Anyways I'd say mutating the attribute directly is not common (since the type is normally constructed using its custom `__init__`), and putting in a non-array type should be a bug anyways.

Issue #1702, opened 2024-11-25
Emit remaining int operations


hugrbot:
Originally raised by @croyzor:

Having done [some int ops](https://github.com/CQCL/hugr-llvm/issues/20) We're still missing emission for some int ops from the arithmetic extension:

Done in #1835 
- [x] ine,
- [x] imax_u,
- [x] imax_s,
- [x] imin_u,
- [x] imin_s,
- [x] iabs,
- [x] iand,
- [x] ior,
- [x] ixor,
- [x] inot,
- [x] ishl,
- [x] ishr,

Done in #1839
- [x] ipow

Done in #1946
- [x] iwiden_u,
- [x] iwiden_s,

Done in #1955
- [x] inarrow_u,
- [x] inarrow_s,

Done(?)
- [x] itobool,
- [x] ifrombool,

Done in #2025 
- [X] idivmod_u,
- [X] idivmod_s,
- [X] imod_u,
- [X] imod_checked_u,
- [X] imod_checked_s,
- [X] idiv_checked_u,
- [X] idiv_checked_s,
- [X] idivmod_checked_u,
- [X] idivmod_checked_s,

Still todo:
- [ ] irotl,
- [ ] irotr,

These seem hard, require dynamic memory allocation on the target 
- [ ] itostring_u,
- [ ] itostring_s,

ss2165:
priorities: 
`ine, ixor, ipow`

croyzor:
Split remaining ops into [#2103](https://github.com/CQCL/hugr/issues/2103) and [#2102](https://github.com/CQCL/hugr/issues/2102) on the hugr repo

PR #1835, opened 2025-01-03
feat(hugr-llvm): Emit more int ops

croyzor:
More work on #1702. Adds `ine`, `iabs`, `imax_{s,u}`, `imin_{s,u}`, `ishl`, `ishr`, `ixor`, `ior`, `inot`, `iand`

### DIFF
--- a/hugr-llvm/src/extension/int.rs
+++ b/hugr-llvm/src/extension/int.rs
@@ -217,6 +217,12 @@ fn emit_int_op<'c, H: HugrView>(
                 .build_not(arg.into_int_value(), "")?
                 .as_basic_value_enum()])
         }),
+        IntOpDef::iand => emit_custom_binary_op(context, args, |ctx, (lhs, rhs), _| {
+            Ok(vec![ctx
+                .builder()
+                .build_and(lhs.into_int_value(), rhs.into_int_value(), "")?
+                .as_basic_value_enum()])
+        }),
         _ => Err(anyhow!("IntOpEmitter: unimplemented op: {}", op.name())),
     }
 }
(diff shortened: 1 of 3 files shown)


hugr-llvm/src/extension/int.rs
```diff
@@ -104,6 +199,24 @@ fn emit_int_op<'c, H: HugrView>(
         IntOpDef::igt_u => emit_icmp(context, args, inkwell::IntPredicate::UGT),
         IntOpDef::ile_u => emit_icmp(context, args, inkwell::IntPredicate::ULE),
         IntOpDef::ige_u => emit_icmp(context, args, inkwell::IntPredicate::UGE),
+        IntOpDef::ixor => emit_custom_binary_op(context, args, |ctx, (lhs, rhs), _| {
```
ss2165, line 202: seems a shame not to include `iand` here

hugr-llvm/src/extension/int.rs
```diff
@@ -95,7 +95,102 @@ fn emit_int_op<'c, H: HugrView>(
                 .build_int_neg(arg.into_int_value(), "")?
                 .as_basic_value_enum()])
         }),
+        IntOpDef::iabs => emit_custom_unary_op(context, args, |ctx, arg, _| {
```
ss2165, line 98: how bad is `ipow`?

hugr-llvm/src/extension/int.rs
croyzor: True! I've added it now

hugr-llvm/src/extension/int.rs
croyzor: It's more annoying than this - `ipow` isn't implemented in llvm, so we have to turn it into a list of multiplications ourselves. I'm thinking about how to do this in hugr-llvm without too much code duplication - doing it as a lowering hugr pass is also an option...

hugr-llvm/src/extension/int.rs
```diff
@@ -256,4 +378,101 @@ mod test {
...
+        exec_ctx.add_extensions(add_int_extensions);
+        let input = ConstInt::new_s(6, arg).unwrap();
+        let ty = INT_TYPES[6].clone();
+        let hugr = test_int_op_with_results::<1>(op, 6, Some([input]), ty.clone());
+        assert_eq!(exec_ctx.exec_hugr_i64(hugr, "main"), result);
+    }
+
+    #[rstest]
+    // LHS: Most significant bit (2^63), RHS: All the other bits combined
+    #[case::inot("inot", 9223372036854775808, 9223372036854775807)]
```
ss2165, line 465:
```suggestion
    #[case::inot("inot", 9223372036854775808, 9223372036854775807)]
    #[case::inot("inot", 1, 0)]
```

ss2165 (approved):
thanks!

PR #1838, opened 2025-01-06
feat: Constant values in `hugr-model`

zrho:
Import and export constant values via `hugr-model`.

Includes terms corresponding to `Value::Function` and allows exporting; importing these is left for a later PR since sum/extension constants are the most important for now. In the future, we should move from the JSON/typetag based model for extension constants to one based on custom constructors.


flamegraph.svg
ss2165, line 1: should this be checked in?

hugr-core/src/export.rs
```diff
@@ -71,7 +72,10 @@ struct Context<'a> {
     implicit_imports: FxHashMap<&'a str, model::NodeId>,
 
     /// Map from node ids in the [`Hugr`] to the corresponding node ids in the model.
-    node_indices: FxHashMap<Node, model::NodeId>,
+    node_to_id: FxHashMap<Node, model::NodeId>,
+
+    /// Mapping from node ids in the [`Hugr`] to the corresponding model nodes.
+    id_to_node: FxHashMap<model::NodeId, Node>,
```
ss2165, line 78: if these are meant to stay in sync would be better to have a dedicated struct holding both that has responsibility for their correctness.

hugr-core/src/export.rs
```diff
@@ -1029,6 +996,56 @@ impl<'a> Context<'a> {
...
+                        let extensions = self.export_ext_set(&dfg.extension_delta());
+                        self.export_dfg(hugr.root(), extensions, model::ScopeClosure::Closed)
+                    }
+                    _ => panic!("Value::Function root must be a DFG"),
+                };
+
+                self.node_to_id = outer_node_to_id;
+                self.hugr = outer_hugr;
+
+                self.make_term(model::Term::ConstFunc { region })
```
ss2165, line 1029: no coverage? good to add something simple

hugr-core/src/export.rs
```diff
@@ -1029,6 +996,56 @@ impl<'a> Context<'a> {
...
+            }
+
+            Value::Sum(sum) => {
+                let tag = sum.tag as _;
+                let mut values = BumpVec::with_capacity_in(sum.values.len(), self.bump);
+
+                for value in &sum.values {
+                    values.push(model::ListPart::Item(self.export_value(value)));
```
ss2165, line 1037: no test included Sums holding values?

hugr-core/src/import.rs
```diff
@@ -1263,6 +1291,116 @@ impl<'a> Context<'a> {
...
+                let symbol_name = self.get_symbol_name(*symbol)?;
+
+                if symbol_name == TERM_JSON_CONST {
+                    let value = args.get(1).ok_or(model::ModelError::TypeError(term_id))?;
+
+                    let model::Term::Str(json) = self.get_term(*value)? else {
+                        return Err(model::ModelError::TypeError(term_id).into());
+                    };
+
+                    // We attempt to deserialize as the custom const directly.
```
ss2165, line 1319: 👍

hugr-core/src/import.rs
```diff
@@ -1263,6 +1291,116 @@ impl<'a> Context<'a> {
...
+                    // This might fail due to the custom const struct not being included when
+                    // this code was compiled; in that case, we fall back to the serialized form.
+                    let value: Option<Box<dyn CustomConst>> = serde_json::from_str(json).ok();
+
+                    if let Some(value) = value {
+                        let opaque_value = OpaqueValue::from(value);
+                        return Ok(Value::Extension { e: opaque_value });
+                    } else {
+                        let runtime_type =
```
ss2165, line 1328: I think we need test coverage of this case

hugr-core/src/import.rs
```diff
@@ -1263,6 +1291,116 @@ impl<'a> Context<'a> {
...
+                let variant = self.import_closed_list(*variant)?;
+
+                let items = values
+                    .iter()
+                    .zip(variant.iter())
+                    .map(|(value, typ)| self.import_value(*value, *typ))
+                    .collect::<Result<Vec<_>, _>>()?;
+
+                let typ = {
+                    // TODO: Import as a `SumType` directly and avoid the copy.
```
ss2165, line 1392: TODO now or later?

hugr-model/src/v0/binary/read.rs
```diff
@@ -274,9 +277,13 @@ fn read_term<'a>(bump: &'a Bump, reader: hugr_capnp::term::Reader) -> ReadResult
             model::Term::ApplyFull { symbol, args }
         }
 
-        Which::Quote(r#type) => model::Term::Quote {
-            r#type: model::TermId(r#type),
-        },
+        Which::Const(reader) => {
+            let reader = reader?;
+            model::Term::Const {
+                r#type: model::TermId(reader.get_type()),
```
ss2165, line 283: I'm surprised this isn't being covered by the float test...?

hugr-model/src/v0/text/print.rs
```diff
@@ -563,7 +572,7 @@ impl<'p, 'a: 'p> PrintContext<'p, 'a> {
                 outputs,
                 extensions,
             } => self.print_parens(|this| {
-                this.print_text("fn");
+                this.print_text("->");
```
ss2165, line 575: dumb question but should this come between inputs and outputs

ss2165 (commented):
Looks good, biggest question is about `->` placement

hugr-model/src/v0/text/print.rs
zrho: In s-expression based syntax it's common to have the function and operator names in the front, e.g. `(* 2 (+ 1 3))` or `(-> ins outs)`. So it's not a mistake. Overall the syntax needs some work anyway, but it won't be stabilised this month, so we can address syntax.

hugr-model/src/v0/binary/read.rs
zrho: This is the *term* for the type of constants. The float test has a const *node*. The const node has a parameter `(@ prelude.const-json (@ arithmetic.float.types.float64) "{\"c\":\"ConstF64\",\"v\":{\"value\":1.0}}" (ext))` which is of type `(const (@ arithmetic.float.types.float64) (ext))`, but that type does not occur itself in the example.

hugr-core/src/import.rs
zrho: Later. This only affects performance and doesn't need any breaking changes to do right.

flamegraph.svg
zrho: Indeed it shouldn't, removed.

hugr-core/src/export.rs
zrho: Test case added.

hugr-core/src/import.rs
zrho: Test case added.

hugr-core/src/export.rs
zrho: `Value::Function` can not yet be imported, so we can not roundtrip test this. I propose we delay that until we can import functions.

hugr-core/src/export.rs
zrho: Ideally yes. I'm not entirely sure yet if the way that functions are exported is ideal here, so there might be a bit of churn around these things happening. In that light I'm not sure how much it is worth to polish.

hugr-core/Cargo.toml
```diff
@@ -20,6 +20,7 @@ workspace = true
 extension_inference = []
 declarative = ["serde_yaml"]
 model_unstable = ["hugr-model"]
+default = ["model_unstable"]
```
ss2165, line 23: it's back!

ss2165 (changes requested):

hugr-core/Cargo.toml
zrho: Removed

hugr-core/src/export.rs
zrho: TODO added.

ss2165 (approved):
LGTM modulo CI fixes 👍

PR #1839, opened 2025-01-06
feat(hugr-llvm): Emit ipow

croyzor:


hugr-llvm/src/extension/int.rs
```diff
@@ -397,6 +460,9 @@ mod test {
     #[case::iand("iand", 6, 15, 6)]
     #[case::iand("iand", 15, 6, 6)]
     #[case::iand("iand", 15, 15, 15)]
+    #[case::ipow("ipow", 2, 3, 8)]
+    #[case::ipow("ipow", 42, 1, 42)]
+    #[case::ipow("ipow", 42, 0, 1)]
```
mark-koch, line 465: Could you also add some similar tests to `test_exec_signed_bin_op` for negative bases?

mark-koch (approved):
Looks good!

We should update the description of the `ipow` op to specify that the exponent is treated as an unsigned integer. This also matches the current implementation in the constant folder for `ipow`.

PR #1841, opened 2025-01-07
test: Fix failing inot test

croyzor:
The `!1 = 0` test doesn't work because the ints it's run on are more than 1 bit wide. I tried reducing the width but it seems the hugr-llvm conversion isn't faithful, downcasting smaller int types to `u8`, meaning `!1` would be `254`

### DIFF
--- a/hugr-llvm/src/extension/int.rs
+++ b/hugr-llvm/src/extension/int.rs
@@ -387,6 +387,7 @@ mod test {
     #[case::imin("imin_u", 2, 1, 1)]
     #[case::imin("imin_u", 2, 2, 2)]
     #[case::ishl("ishl", 73, 1, 146)]
+    // (2^64 - 1) << 1 = (2^64 - 2)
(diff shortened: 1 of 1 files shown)


hugr-llvm/src/extension/int.rs
```diff
@@ -463,7 +463,7 @@ mod test {
     #[rstest]
     // LHS: Most significant bit (2^63), RHS: All the other bits combined
     #[case::inot("inot", 9223372036854775808, 9223372036854775807)]
-    #[case::inot("inot", 1, 0)]
+    #[case::inot("inot", 18446744073709551615, 0)]
```
ss2165, line 466:
```suggestion
    #[case::inot("inot", !0u64, 0)]
```

All I'm after is something easy to verify by eye really

ss2165 (approved):

hugr-llvm/src/extension/int.rs
croyzor: 😮

PR #1844, opened 2025-01-08
fix: Three bugfixes in model import and export.

zrho:
This PR fixes three bugs in the import/export between `hugr_model` and `hugr_core`.

1. The types of the source and target ports of a CFG region must be wrapped in a `ctrl` type on export.
2. Importing links that have a single input or output but aren't otherwise connected should be valid.
3. Runtime types are valid `TypeArg`s.

### DIFF
--- a/hugr-core/src/import.rs
+++ b/hugr-core/src/import.rs
@@ -216,24 +216,26 @@ impl<'a> Context<'a> {
                 }
             }
 
-            if inputs.is_empty() || outputs.is_empty() {
-                return Err(error_unsupported!(
-                    "link {}#{} is missing either an input or an output port",
-                    link_id.0,
-                    link_id.1
-                ));
-            }
-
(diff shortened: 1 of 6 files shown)


hugr-core/src/import.rs
```diff
@@ -216,24 +216,26 @@ impl<'a> Context<'a> {
...
+            match (inputs.as_slice(), outputs.as_slice()) {
+                ([], []) => {
+                    unreachable!();
+                }
+                (_, [output]) => {
+                    for (node, port) in inputs.iter() {
+                        self.hugr.connect(output.0, output.1, *node, *port);
+                    }
+                }
+                ([input], _) => {
```
ss2165, line 228: not great to have these untested lines especially since they are adjacent to a bug, I think benefits from a dedicated test

zrho: Covered now.

ss2165 (approved):
LGTM

PR #1845, opened 2025-01-08
feat: Bytes literal in hugr-model.

### LINKED ISSUE #1545: `model`: Add terms for bytes
As @ss2165 pointed out, it would be useful to have a bytes literal in the hugr model. Once the initial version of the model is merged, this can be added easily. In particular, this would be a new `Bytes(&'a [u8])` variant and a `BytesType` variant in `Term`.

zrho:
Adds a term for byte literals to hugr-model.

In the text format, byte literals are represented as base64 encoded strings.

Closes #1545.


tatiana-s (approved):
👍

PR #1849, opened 2025-01-08
feat: Improved representation for metadata in `hugr-model`

### LINKED ISSUE #1848: Improved representation for metadata in `hugr-model`
Currently metadata in `hugr-model` is represented as a list of `MetaItem`s defined as follows:
```rust
pub struct MetaItem<'a> {
    pub name: &'a str,
    pub value: TermId,
}
```

This issue proposes an alternative representation. We introduce a new term `meta` which is the type of metadata. Then extensions can define custom constructors that produce terms of type `meta`:
```edn
(declare-ctr core.meta.description (param ?description str) meta)
(declare-ctr compat.meta-json (param ?name str) (param ?json str) meta)
```
Nodes and regions then have a list of `TermId`s that refer to instantiations of these terms. We can use the `compat.meta-json` constructor for compatibility with the existing JSON based metadata representation in `hugr-core`.

The proposed representation has a number of advantages:

- Before, metadata names were simply strings and therefore used their own namespacing mechanism. In the new representation, the type of metadata is identified by a symbol.
- Before, the strings for metadata names were not deduplicated. This is solved in the new representation via the already existing deduplication mechanism for symbols.
- Before, there was no mechanism to associate between the name of metadata and the type expected for its value. This is solved in the new representation as each constructor for metadata can specify the types of parameters and even constraints.
- Nodes and regions have a bunch of lists of `u32`s to refer to inputs, outputs, sources, targets, parameters, regions and children. With the new representation, metadata also is just a list of `u32`s. This opens up optimisations in the serialisation/in-memory format in the future.

zrho:
Closes #1848 by using a term-based representation for metadata items.

### DIFF
--- a/hugr-core/src/export.rs
+++ b/hugr-core/src/export.rs
@@ -18,9 +18,6 @@ use std::fmt::Write;
 
 pub(crate) const OP_FUNC_CALL_INDIRECT: &str = "func.call-indirect";
 const TERM_PARAM_TUPLE: &str = "param.tuple";
-const TERM_JSON: &str = "prelude.json";
-const META_DESCRIPTION: &str = "docs.description";
-const TERM_JSON_CONST: &str = "prelude.const-json";
 
 /// Export a [`Hugr`] graph to its representation in the model.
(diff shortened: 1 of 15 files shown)


croyzor (approved):
LGTM

PR #1855, opened 2025-01-13
refactor(hugr-llvm)!: Optimise the llvm types used to represent hugr sums.

doug-q:
We rework `LLVMSumType` and `LLVMValueType` to have llvm representations that
* elide the tag field when the sum type has only one variant
* elide any "uninformational" fields, i.e. `Type::UNIT`.
* elide the struct wrapper when there is only a single field
* reuse fields from other variants where possible.

In particular, the hugr `bool_t()` is now represented by `i1`.

BREAKING CHANGE: `LLVMSumValue::get_tag_type` renamed to `tag_type`. `LLVMSumType::try_new2` renamed to `LLVMSumType::try_new`. `LLVMSumType::get_variant` removed.

### DIFF
--- a/hugr-llvm/src/sum/layout.rs
+++ b/hugr-llvm/src/sum/layout.rs
@@ -174,7 +174,7 @@ fn layout_variants_impl<T: Ord + Clone + fmt::Debug>(
                 .map(|t| {
                     (!elide(t)).then(|| {
                         t_to_range_map
-                            .get_mut(&t)
+                            .get_mut(t)
                             .and_then(Iterator::next)
                             .expect("We have ensured that there are enough fields of type t")
                     })


zrho:
LGTM but I currently lack context to find anything subtle here.

qartik:
The LLVM output looks a lot more readable now!

hugr-llvm/src/sum/layout.rs
```diff
@@ -0,0 +1,216 @@
...
+}
+
+#[derive(derive_more::Debug, Clone, PartialEq, Eq)]
+/// `BasicTypeEnum` does not `impl` `Ord`. We use this type to put an ordering `BasicTypeEnum`,
+/// first by reverse-size and then by, string representation.
+struct BasicTypeOrd<'c>(
+    BasicTypeEnum<'c>,
+    #[debug(skip)] u64,
+    #[debug(skip)] Cow<'c, String>,
+);
```
mark-koch, line 59: Not a big fan of the string ordering performance wise, but I can't think of a better solution either that doesn't require a lot of work...

hugr-llvm/src/types.rs
```diff
@@ -124,9 +124,9 @@ pub mod test {
     }
 
     #[rstest]
-    #[case(0, SumType::new_unary(0))]
+    // #[case(0, SumType::new_unary(0))]
```
mark-koch, line 127: Should be reenabled if allow empty sums as suggested

hugr-llvm/src/sum.rs
```diff
@@ -60,114 +124,358 @@ impl<'c> LLVMSumType<'c> {
...
+    pub(self) fn num_fields_for_variant(&self, variant: usize) -> usize {
+        self.fields_for_variant(variant).len()
     }
 
-    delegate! {
-        to self.1 {
-            pub(self) fn num_variants(&self) -> usize;
+    /// The LLVM types representing the fields in the `tag` variant of the represented [HugrSumType].
+    /// Panics if `tag` is out of bounds.
+    pub(self) fn fields_for_variant(&self, variant: usize) -> &[BasicTypeEnum<'c>] {
```
mark-koch, line 451: Argument doesn't match the docstring: `tag` vs `variant`

hugr-llvm/src/sum.rs
```diff
@@ -198,52 +506,81 @@ unsafe impl<'c> BasicValue<'c> for LLVMSumValue<'c> {}
...
+        ensure!(tag < self.num_variants(), "Bad tag {tag} in {}", self.1);
+        let results = match self.get_type().0 {
+            LLVMSumTypeEnum::Unit { .. } | LLVMSumTypeEnum::NoFields { .. } => Ok(vec![]),
+            LLVMSumTypeEnum::SingleVariantSingleField { .. } => Ok(vec![self.0]),
+            LLVMSumTypeEnum::SingleVariantMultiField { .. } => {
+                let value = self.0.into_struct_value();
+                (0..value.get_type().count_fields())
+                    .map(|i| anyhow::Ok(builder.build_extract_value(value, i, "")?))
+                    .collect()
+            }
```
mark-koch, line 560: Don't you need to fill in `undefined` for all the elided fields in these cases?

hugr-llvm/src/emit/ops/snapshots/hugr_llvm__emit__ops__cfg__test__diverse_outputs@llvm14.snap
```diff
@@ -1,6 +1,6 @@
 ---
-source: src/emit/ops/cfg.rs
-expression: module.to_string()
+source: hugr-llvm/src/emit/ops/cfg.rs
+expression: mod_str
```
mark-koch, line 3: Are all of these path changes intentional?

hugr-llvm/src/emit/ops/snapshots/hugr_llvm__emit__ops__cfg__test__diverse_outputs@llvm14.snap
```diff
@@ -12,38 +12,33 @@ alloca_block:
...
   br label %2
 
-2:                                                ; preds = %14, %entry_block
-  %"7_0.0" = phi i8 [ %0, %entry_block ], [ %16, %14 ]
-  %"7_1.0" = phi i8 [ %1, %entry_block ], [ %9, %14 ]
-  %3 = insertvalue { i8, i8 } undef, i8 %"7_0.0", 0
+2:                                                ; preds = %11, %entry_block
+  %"7_0.0" = phi i8 [ %0, %entry_block ], [ %12, %11 ]
+  %"7_1.0" = phi i8 [ %1, %entry_block ], [ %7, %11 ]
+  %3 = insertvalue { i8, i8 } poison, i8 %"7_0.0", 0
```
mark-koch, line 18: Is it better to have `undef` or `poison` here?

hugr-llvm/src/sum.rs
```diff
@@ -60,114 +124,358 @@ impl<'c> LLVMSumType<'c> {
...
+    /// Constructs a new [LLVMSumTypeEnum] from a `Vec` of variants.
+    /// Each variant is a `Vec` of LLVM types each corresponding to a field in the sum.
+    pub fn try_new(
+        context: &'c Context,
+        variant_types: Vec<Vec<BasicTypeEnum<'c>>>,
+    ) -> Result<Self> {
+        let result = match variant_types.len() {
+            0 => anyhow::bail!(
+                "LLVMSumType constructed with no variants. Void is not representable in LLVM"
+            ),
```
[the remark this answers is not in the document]
mark-koch: Ok, but do you agree that we need to handle empty sums one way or another?

My feeling is that removing them before going to LLVM would be more effort than allowing them here as a type but not as a value

hugr-llvm/src/sum/layout.rs
```diff
@@ -0,0 +1,234 @@
...
+                .and_modify(|x| *x = count.max(*x))
+                .or_insert(count);
+        }
+        t_counts
+    };
+    let sorted_fields = t_counts
+        .iter()
+        .flat_map(|(t, &count)| itertools::repeat_n(t.clone(), count))
+        .collect_vec();
+    // Map from each t to its first occurrence in sorted_fields
```
mark-koch, line 127:
```suggestion
    // Map from each t to its index range in sorted_fields
```

hugr-llvm/src/sum/layout.rs
```diff
@@ -0,0 +1,234 @@
...
+            sorted_fields.len()
+        );
+        for (t, range) in &t_to_range {
+            assert!(range.clone().all(|i| &sorted_fields[i] == t));
+        }
+    }
+    (sorted_fields, t_to_range)
+}
+
+/// The implemenation of the layout algorithm.
```
mark-koch, line 151:
```suggestion
/// The implementation of the layout algorithm.
```

hugr-llvm/src/sum.rs
```diff
@@ -60,114 +124,365 @@ impl<'c> LLVMSumType<'c> {
...
+                .into_int_type(),
+        }
     }
 
-    fn get_variant_field_index(&self, tag: usize) -> usize {
-        tag + (if self.has_tag_field() { 1 } else { 0 })
+    /// The underlying LLVM type.
+    pub fn value_type(&self) -> BasicTypeEnum<'c> {
+        match self {
+            Self::Void { tag_type, .. } => (*tag_type).into(),
```
mark-koch, line 423: Shouldn't this be an empty struct according to the doc on `LLVMSumTypeEnum::Void`?

hugr-llvm/src/sum.rs
```diff
@@ -280,7 +669,142 @@ impl<'c> LLVMSumValue<'c> {
...
+    #[case(4, 2)]
+    #[case(5, 3)]
+    #[case(8, 3)]
+    #[case(9, 4)]
+    fn tag_width(#[case] num_variants: usize, #[case] expected: u32) {
+        assert_eq!(tag_width_for_num_variants(num_variants), expected);
+    }
+
+    #[rstest]
+    fn sum_types(mut llvm_ctx: TestContext) {
```
mark-koch, line 705: Could you add a test case for the empty sum?

hugr-llvm/src/sum/layout.rs
```diff
@@ -0,0 +1,234 @@
...
+        .scan(0, |offset, (t, count)| {
+            let t_offset = *offset;
+            *offset += count;
+            Some((t, t_offset..*offset))
+        })
+        .collect();
+
+    #[cfg(debug_assertions)]
+    {
+        assert_eq!(t_to_range.first_key_value().unwrap().1.start, 0);
```
mark-koch, line 139: Somehow this `unwrap` seems to fail in all of the tests?

mark-koch (commented):
Thanks Doug, looks good!

However, something seems off in the layout since tests are failing the asserts?

PR #1857, opened 2025-01-13
feat: Special cased array, float and int constants in hugr-model export

zrho:
This PR special cases the import/export of array, float and int constants so that they are represented as terms in `hugr-model`. This representation is a lot more efficient to handle than the encoding as JSON. In the future we should do this for all constants.

### DIFF
--- a/hugr-core/tests/snapshots/model__roundtrip_const.snap
+++ b/hugr-core/tests/snapshots/model__roundtrip_const.snap
@@ -4,9 +4,19 @@ expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-cons
 ---
 (hugr 0)
 
+(import collections.array.array)
+
+(import collections.array.const)
+
 (import compat.const-json)
 
 (import arithmetic.float.types.float64)
+
+(import arithmetic.int.const)
+
+(import arithmetic.int.types.int)
+
+(import arithmetic.float.const-f64)
 
 (define-func example.bools
   [] [(adt [[] []]) (adt [[] []])] (ext)
@@ -19,7 +29,8 @@ expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-cons
 (define-func example.make-pair
   []
   [(adt
-     [[(@ arithmetic.float.types.float64) (@ arithmetic.float.types.float64)]])]
+     [[(@ collections.array.array 5 (@ arithmetic.int.types.int 6))
+       (@ arithmetic.float.types.float64)]])]
   (ext)
   (dfg
     [] [%0]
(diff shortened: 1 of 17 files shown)


hugr-core/tests/snapshots/model__roundtrip_const.snap
```diff
@@ -19,40 +29,41 @@ expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-cons
...
         (->
           []
           [(adt
-             [[(@ arithmetic.float.types.float64)
+             [[(@ collections.array.array 5 (@ arithmetic.int.types.int 6))
                (@ arithmetic.float.types.float64)]])]
           (ext))))))
 
-(define-func example.f64
+(define-func example.f64-json
```
ss2165, line 66: is this still using json?

hugr/Cargo.toml
```diff
@@ -27,12 +27,13 @@ declarative = ["hugr-core/declarative"]
 model_unstable = ["hugr-core/model_unstable", "hugr-model"]
 llvm = ["hugr-llvm/llvm14-0"]
 llvm-test = ["hugr-llvm/llvm14-0", "hugr-llvm/test-utils"]
+default = ["model_unstable"]
```
ss2165, line 30: 😓

hugr-model/src/v0/text/print.rs
```diff
@@ -612,6 +612,15 @@ impl<'p, 'a: 'p> PrintContext<'p, 'a> {
                 self.print_text("meta");
                 Ok(())
             }
+            Term::Float { value } => {
+                // The debug representation of a float always includes a decimal point.
+                self.print_text(format!("{:?}", value.into_inner()));
+                Ok(())
+            }
+            Term::FloatType => {
```
ss2165, line 620: looks like float type isn't being covered, pls add to test

ss2165 (commented):
🚀

hugr/Cargo.toml
aborgna-q: Uhm. I think we should do `unstable XOR default`

hugr-core/tests/snapshots/model__roundtrip_const.snap
zrho: Yes: if you check `hugr-model/tests/fixtures/model-const.edn` you see that it loads two json constants. Only the unknown one is exported again as json. This is to have the json fallback path in the test coverage.

hugr/Cargo.toml
zrho: Fixed

hugr-model/src/v0/text/print.rs
zrho: Should be covered now; can't check because the codecov CI links seem to be broken...

ss2165 (approved):
thanks

PR #1866, opened 2025-01-16
refactor: remove MutableIO

doug-q:


hugr-core/src/builder/dataflow.rs
```diff
@@ -331,6 +352,46 @@ impl<T> HugrBuilder for DFGWrapper<Hugr, T> {
     }
 }
 
+/// TODO perhaps these should be on HugrMut?
```
aborgna-q, line 355: Yes, it should probably go there.
I'd add a single function though, parametric on the direction. (And add specializations that cast the output type if needed)

hugr-core/src/builder/dataflow.rs
```diff
@@ -160,18 +78,123 @@ impl DFGBuilder<Hugr> {
...
+    // wrap [DFGBuilder].
+    //
+    // `new_optype` must:
+    //   - be a of the same OpType as the node under construction.
+    //   - `DataflowParent::inner_signature` should be the same, except that the
+    //     output row should have `output_type` appended.
+    //
+    // Replaces the DFG under construction with new_optype and mutates the
+    // Output node by appending `output_type``.
+    pub(crate) fn add_output_impl(&mut self, new_optype: impl Into<OpType>, output_type: Type) {
```
aborgna-q, line 169: This is the same as `add_input_impl` bar the returned wire, which can be computed in `add_input` directly.

acl-cqc:
This targets @qartik's #1824 right? On its own it looks a little odd and I don't see quite why you're doing it (less parametricity), but maybe with some docs I might. Or perhaps if you retarget this at main and say "closes #1816"? (right?)

qartik:
This PR may not be needed anymore.

PR #1870, opened 2025-01-17
refactor(hugr-model): Reexport `bumpalo` from `hugr-model`

doug-q:


ss2165:
is it possible to export just the relevant types, perhaps using aliases? e.g. `pub type Bump = bumpalo::Bump;`

rather than re-exporting entire dependencies. Changes to these exports would be breaking in future.

ss2165:
fair enough, I'd like to hear @zrho 's opinion on what are the useful set of things to make part of the interface

zrho:
`indexmap` and `fxhash` are implementation details. `bumpalo::Bump` is the only part of `bumpalo` that is exposed in the API, so perhaps that one may be reexported. For the other stuff I don't really see a reason.

zrho (approved):

Issue #1869, opened 2025-01-16
perf: Only check partial convexity for `SimpleReplacement` subgraphs.


aborgna-q:
See discussion https://github.com/CQCL/hugr/discussions/1656 ("_Can we avoid the convexity work altogether?_").

Requires https://github.com/CQCL/portgraph/pull/161 to be published in portgraph.

`SiblingSubgraph` currently requires the subgraph to be convex. This is no longer needed for doing replacements if we can prove that the new node boundary orders are no stricter than the original ones. We should only need to check if the new port pairs in `PortOrdering::missing_pairs` had an external reverse dependency before.

PR #1878, opened 2025-01-20
chore!: Update to `portgraph 0.13` / `petgraph 0.7`

### LINKED ISSUE #1872: chore!: Update to `portgraph 0.13` & `petgraph 0.7`
We moved to RpiTiT on portgraph and changed the signature of subgraphs, so this brings some migration pains.

aborgna-q:
The new portgraph release comes with some perf improvements.
We require this update for #1869.

Closes #1872.

I added a benchmark variation for subgraphs with few nodes as a drive-by.
This got improved by the new portgraph (skipping a full graph traversal), but the runtime is still mostly the `O(n)` full-graph traversal required for the convexity checking when building the subgraph.

BREAKING CHANGE: Bumped public dependency `portgraph` to breaking version `0.13`.

### DIFF
--- a/hugr-core/src/hugr/views/descendants.rs
+++ b/hugr-core/src/hugr/views/descendants.rs
@@ -137,7 +137,7 @@ where
         let hugr = hugr.base_hugr();
         Ok(Self {
             root,
-            graph: RegionGraph::new_region(&hugr.graph, &hugr.hierarchy, root.pg_index()),
+            graph: RegionGraph::new(&hugr.graph, &hugr.hierarchy, root.pg_index()),
             hugr,
             _phantom: std::marker::PhantomData,
         })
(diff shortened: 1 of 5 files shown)


hugr-core/src/hugr/views/descendants.rs
```diff
@@ -247,7 +247,7 @@ pub(super) mod test {
             Some(Signature::new(vec![usize_t()], vec![usize_t()]))
         );
         assert_eq!(inner_region.node_count(), 3);
-        assert_eq!(inner_region.edge_count(), 2);
+        assert_eq!(inner_region.edge_count(), 1);
```
aborgna-q, line 250: This was previously wrong. The `inner` subgraph is just a DFG with one input / output connected directly to each other.

mark-koch (approved):

PR #1888, opened 2025-01-21
fix: determine correct bounds of custom types

### LINKED ISSUE #1876: hugr-model: panic while round-tripping export/import
See failing test case in https://github.com/CQCL/hugr/pull/1875. A json Hugr is committed to the tree there, generated from this guppy program:

```python
from guppylang import guppy, quantum, qubit
from guppylang.std.angles import angle, angles
from guppylang.std.builtins import result
from guppylang.std.quantum import h, measure, rz

guppy.load(quantum)
guppy.load(angles)


@guppy
def rx(q: qubit, x: angle) -> None:
    # Implement Rx via Rz rotation
    h(q)
    rz(q, x)
    h(q)


@guppy
def main() -> None:
    q = qubit()
    rx(q, angle(1.5))
    result("1", measure(q))


if __name__ == "__main__":
    print(guppy.get_module().compile().package.to_json())
```

zrho:
Import now determines the bound of the custom type correctly.

Fixes #1876.

### DIFF
--- a/hugr-core/src/import.rs
+++ b/hugr-core/src/import.rs
@@ -18,7 +18,7 @@ use crate::{
     types::{
         type_param::TypeParam, type_row::TypeRowBase, CustomType, FuncTypeBase, MaybeRV,
         PolyFuncType, PolyFuncTypeBase, RowVariable, Signature, Type, TypeArg, TypeBase, TypeBound,
-        TypeEnum, TypeRow,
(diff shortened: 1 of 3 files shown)


cqc-alec (approved):
Checked that the test in https://github.com/CQCL/hugr/tree/doug/bounds-bug now does not panic. LGTM.

PR #1890, opened 2025-01-22
fix: Exporting converging control flow edges

### LINKED ISSUE #1889: Export does not correctly handle converging control flow edges
When a `Hugr` contains control flow edges from two ports which converge to a single port, the export to the model does not correctly detect the connectivity. The error is caused by the logic in `hugr_core::export::Context::get_link_index` which was written under the misconception that the underlying data structure of `Hugr` as implemented was a hypergraph. I have a fix, but it still needs some polishing.

<details>
<summary>Guppy code generating a module that exports incorrectly</summary>

```python
from guppylang import guppy
from guppylang.std.builtins import result
from guppylang.std.quantum import measure, qubit, discard, h, tdg, cx, t, z

@guppy
def repeat_until_success(q: qubit) -> None:
    attempts = 0
    while True:
        attempts += 1
        
        # Prepare ancillas
        a, b = qubit(), qubit()
        h(a)
        h(b)
        
        tdg(a)
        cx(b, a)
        t(a)
        h(a)
        if measure(a):
            # First part failed, try again
            discard(b)
            continue

        t(q)
        z(q)
        cx(q, b)
        t(b)
        h(b)
        if measure(b):
            # Second part failed, apply correction
            # and try again 
            z(q)
            continue
        
        result("attempts", attempts)
        break

@guppy
def main() -> None:
    q = qubit()
    repeat_until_success(q)
    measure(q)

with open("cfg_hugr.json", "w") as f:
    f.write(guppy.compile_module().package.to_json())
```
</details>

zrho:
Fixes #1889.

### DIFF
--- a/hugr-core/tests/snapshots/model__roundtrip_cfg.snap
+++ b/hugr-core/tests/snapshots/model__roundtrip_cfg.snap
@@ -21,7 +21,7 @@ expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-cfg.
             [%4] [%5]
             (signature (-> [?0] [(adt [[?0]])] (ext)))
             (tag 0 [%4] [%5] (signature (-> [?0] [(adt [[?0]])] (ext))))))
-        (block [%6] [%3 %9]
+        (block [%6] [%3 %6]
           (signature (-> [(ctrl [?0])] [(ctrl [?0]) (ctrl [?0])] (ext)))
           (dfg
             [%7] [%8]
--- a/hugr-core/src/export.rs
+++ b/hugr-core/src/export.rs
@@ -16,8 +16,9 @@ use crate::{
     Direction, Hugr, HugrView, IncomingPort, Node, Port,
 };
 use bumpalo::{collections::String as BumpString, collections::Vec as BumpVec, Bump};
-use fxhash::FxHashMap;
+use fxhash::{FxBuildHasher, FxHashMap};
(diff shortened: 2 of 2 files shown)


hugr-core/tests/snapshots/model__roundtrip_cfg.snap
```diff
@@ -21,7 +21,7 @@ expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-cfg.
             [%4] [%5]
             (signature (-> [?0] [(adt [[?0]])] (ext)))
             (tag 0 [%4] [%5] (signature (-> [?0] [(adt [[?0]])] (ext))))))
-        (block [%6] [%3 %9]
+        (block [%6] [%3 %6]
```
ss2165, line 24: does this change make sense?

ss2165 (changes requested):
Please add a test that fails without this fix.

Constructing the links struct looks like it adds a non-trivial amount of pre-computation. Do you have a measure of performance impact?

hugr-core/tests/snapshots/model__roundtrip_cfg.snap
zrho: Yes. This is a test that should have failed but I overlooked it earlier. It's correct now. So this is the test you asked for.

I measured the performance impact and it was very small.

ss2165 (approved):

PR #1893, opened 2025-01-27
feat: Simplify hugr-model

zrho:
This PR simplifies `hugr_model` by moving every operation and term that can be defined as an extension to a core extension.


hugr-model/capnp/hugr-v0.capnp
aborgna-q, line 1: Do we need to change the unique ID at line 1, given the breaking changes to the protocol?

hugr-core/src/export.rs
```diff
@@ -371,22 +361,29 @@ impl<'a> Context<'a> {
                 let mut args = BumpVec::new_in(self.bump);
                 args.extend(call.type_args.iter().map(|arg| self.export_type_arg(arg)));
                 let args = args.into_bump_slice();
+                let func = self.make_term(model::Term::Apply(symbol, args));
+
+                // PERFORMANCE: Avoid exporting the signature here again.
```
aborgna-q, line 366: Add a TODO tag too, so it's still highlighted/searchable?
```suggestion
                // TODO PERFORMANCE: Avoid exporting the signature here again.
```

aborgna-q (changes requested):
Looking clean!

My main request is taking it out of the default features.

hugr-model/capnp/hugr-v0.capnp
zrho: Given that there is nothing using it yet, we don't really need to change it.

hugr-core/Cargo.toml
```diff
@@ -20,6 +20,7 @@ workspace = true
 extension_inference = []
 declarative = ["serde_yaml"]
 model_unstable = ["hugr-model"]
+default = ["model_unstable"]
```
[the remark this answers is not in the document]
zrho: I just keep forgetting removing this line. I put it in to make the LSP work, but it's not intended to remain there.

aborgna-q (approved):

PR #1895, opened 2025-01-28
feat!: ComposablePass trait allowing sequencing and validation

### LINKED ISSUE #1832: Better Pass Infrastructure
At present we have a bunch of things that call themselves Passes - [ConstantFoldPass](https://github.com/CQCL/hugr/blob/cad74844fcfef9d4bcd5e0c36335c26d88fa791f/hugr-passes/src/const_fold.rs#L31), [MonomorphizePass](https://github.com/CQCL/hugr/blob/cad74844fcfef9d4bcd5e0c36335c26d88fa791f/hugr-passes/src/monomorphize.rs#L275), [RemoveDeadFuncsPass](https://github.com/CQCL/hugr/blob/cad74844fcfef9d4bcd5e0c36335c26d88fa791f/hugr-passes/src/dead_funcs.rs#L60).

Each of these has its own copy of code to allow setting a `ValidationLevel`, and some have other pass-specific configuration. There is no common interface - the idea of a "pass" is purely conceptual. They then run via `ValidationLevel::run_validated_pass` with a callback (leading to quite a complicated signature, and possibly fun with the borrow-checker).

We could improve on this via:
* A `trait Pass` which could be roughly, `fn run(h: &mut impl HugrMut) -> Result<(), Self::Error>` (plus associated type `Error`)
* A `ValidatingPass<P: Pass>(P, ValidationLevel)` that wraps another pass with something like the current [ValidationLevel](https://github.com/CQCL/hugr/blob/cad74844fcfef9d4bcd5e0c36335c26d88fa791f/hugr-passes/src/validation.rs#L16) but *without* ValidationLevel::None. We could have a single `ValidationPassError<E>` that reports either an error in validation (before/after) or an underlying error `E`
* Some kind of "SequencePass" that wraps a bunch of other passes (I think this would have to use `dyn` so maybe just two  arbitrary passes) - and we can add methods to `trait Pass` that combine Passes e.g. `and_then(other: impl Pass) -> impl Pass`. This would combine with the previous to allow validation before/after a number of passes but not at steps inbetween.
* Maybe common config for specifying inputs, reachable functions, anything else - e.g. #1797 / #1752

acl-cqc:
Currently We have several "passes": monomorphization, dead function removal, constant folding. Each has its own code to allow setting a validation level (before and after that pass).

This PR adds the ability chain (sequence) passes;, and to add validation before+after any pass or sequence; and commons up validation code. The top-level `constant_fold_pass` (etc.) functions are left as wrappers that do a single pass with validation only in test.

I've left ConstFoldPass as always including DCE, but an alternative could be to return a sequence of the two - ATM that means a tuple `(ConstFoldPass, DeadCodeElimPass)`.

I also wondered about including a method `add_entry_point` in ComposablePass (e.g. for ConstFoldPass, that means `with_inputs` but no inputs, i.e. all Top). I feel this is not applicable to *all* passes, but near enough. This could be done in a later PR but `add_entry_point` would need a no-op default for that to be a non-breaking change. So if we wouldn't be happy with the no-op default then I could just add it here...

Finally...docs are extremely minimal ATM (this is hugr-passes), I am hoping that most of this is reasonably obvious (it doesn't really do a lot!), but please flag anything you think is particularly in need of a doc comment!

BREAKING CHANGE: quite a lot of calls to current pass routines will break, specific cases include (a) `with_validation_level` should be done by wrapping a ValidatingPass around the receiver; (b) XXXPass::run() requires `use ...ComposablePass` (however, such calls will cease to do any validation).

closes #1832


hugr-passes/src/composable.rs
```diff
@@ -0,0 +1,223 @@
...
+    type Err: Error;
+    fn run(&self, hugr: &mut impl HugrMut) -> Result<(), Self::Err>;
+
+    fn map_err<E2: Error>(self, f: impl Fn(Self::Err) -> E2) -> impl ComposablePass<Err = E2> {
+        ErrMapper::new(self, f)
+    }
+
+    fn sequence(
+        self,
+        other: impl ComposablePass<Err = Self::Err>,
```
acl-cqc, line 21: The other way to do this is
```
fn sequence_before<P2: ComposablePass>(self, other: P2) -> impl ComposablePass<Err = Self::Err>
   where P2::Err : Into<Self::Err>
```
(and similarly `sequence_after`)
which avoids what may be a common-case use of `map_err` (i.e. with `Into`). However, Infallible *doesn't* implement Into<any type>, so it's probably not much of a win. (I could add `sequence_infallible(self, other: ComposablePass<Err=Infallible>)` I guess, but would then want that before/after versions of that too)

cqc-alec (commented):
Looks good! I do think some more docs would be good. Regarding the suggestion to add `add_entry_point` -- does it mean to apply the pass to everything below that point in the hierarchy but not above it? I think that sounds reasonable.

hugr-passes/src/composable.rs
```diff
@@ -0,0 +1,223 @@
...
+/// e.g. by [ValidatingPass]
+pub trait ComposablePass: Sized {
+    type Err: Error;
+    fn run(&self, hugr: &mut impl HugrMut) -> Result<(), Self::Err>;
+
+    fn map_err<E2: Error>(self, f: impl Fn(Self::Err) -> E2) -> impl ComposablePass<Err = E2> {
+        ErrMapper::new(self, f)
+    }
+
+    fn sequence(
```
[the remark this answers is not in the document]
acl-cqc: Renamed to `then` and `then_either`. Docs might now be a bit redundant, I dunno....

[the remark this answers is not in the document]
acl-cqc: This will be much better when Rust `never_type` becomes usable and we can use `!` instead of `Infallible`...

hugr-passes/src/composable.rs
```diff
@@ -0,0 +1,223 @@
...
+        })?;
+        self.0.run(hugr).map_err(ValidatePassError::Underlying)?;
+        self.validation_impl(hugr, |err, pretty_hugr| ValidatePassError::Output {
+            err,
+            pretty_hugr,
+        })
+    }
+}
+
+pub(crate) fn validate_if_test<P: ComposablePass>(
```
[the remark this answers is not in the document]
acl-cqc: Tried a few things as this sounded like a good suggestion but couldn't really get anything to work well. In the end I conclude this *isn't* the principal way to run a pass - it's `pub (crate)` after all - rather, it's a way for us to hack in a bit of extra test coverage.

If it were the principal way, I'd like
(a) to rename `run` to something that makes it obvious to implement it but not to call it
(b) to make the "principal" `run` method *not* add extra (repeated) validation for ValidatingPass....so a trait method with a default that adds validation, and override in ValidatingPass not to....but

Unfortunately the addition of the extra validation wrapper changes the return/Result type. The "result type with extra validation" could be made a trait associated type, but then you can't provide a trait-default implementation. One could make the extra validation be *unwrapped* but then using it on a ValidatingPass turns a reported error into a panic....

So I didn't really see a way to make a nice API out of this, and so felt it best to leave it as internal i.e. `pub (crate)`. If you have better ideas please shout...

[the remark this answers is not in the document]
cqc-alec: Nice.

acl-cqc:
Breaking, so will leave merging until after end March

acl-cqc:
> This interface works alright with global passes that traverse the whole hugr. It's not sufficient however for region-focused ones

Yeah, that's a good point. The interface imagines we have a region-focused HugrMut, but it is unlikely we ever will due to the performance issues.

Note that most global passes can be applied to a region that does not have incoming edges (e.g. as captured by a RegionView), but if there are unfilled inputs (because of edges from outside the region) then applying the pass locally will invalidate the hugr. So global/local isn't quite as much of a hard distinction as that, although, there are some pure optimization (interface-preserving) passes that definitely *can* be run "just anywhere".

Pointers to within a hugr (#1642) may be a solution, and are slightly better (in that you can see the source of the incoming edges), but what do you do if your pass changes the type of an incoming port with an edge from above the root-pointer?

`with_hugr` might be an option. Then different passes could define `with_hugr` (if global) or `with_hugr_subtree(&mut impl HugrMut, Node)` to return some new interface `Foo<T>::run(self) -> T`...whether that Foo is the new ComposablePass or separate/additional I'm not sure (does one sequence Foo's, or what?). Maybe we define `ComposablePass<T>` and some passes implement that with T being a tuple of (some form of Hugr, Node)?

Note that a local pass run on a subtree rooted at node `N` should not invalidate N or anything above it, so there is some benefit to being able to sequence passes *before* specifying the `N`....this needs sketching out in more detail....

aborgna-q:
I'd argue "running a pass on a region" is a separate concept from "only modify this part".
It's focused on the rewrite-finding part of the pass instead.

This depends on each pass semantics, but I can imagine the `untuple` pass adding non-local edges to previously disconnected nodes outside the region, or `inline` removing a global node that's not used anymore (given the appropriate config).

aborgna-q:
The tuple-targets for `ComposablePass` carry the risk of overcomplicating the API.
This is something we expect users to call regularly, so adding arbitrary parameter tupleing may feel unexpected.

acl-cqc:
> This depends on each pass semantics,

Yes, it's starting to sound difficult to find a common interface. And if you want to allow combining global passes with local ones, then there are much bigger concerns about nodes being invalidated - you might have to provide an `Fn(&Hugr)->Node` rather than a `Node`...

acl-cqc:
> The tuple-targets for `ComposablePass` carry the risk of overcomplicating the API. This is something we expect users to call regularly, so adding arbitrary parameter tupleing may feel unexpected.

@aborgna-q do you mean `impl ComposablePass for (A,B)`? I have removed these now in favour of a hidden struct as that let me do some magic with error traits and phantomdata.

hugr-passes/src/composable.rs
```diff
@@ -0,0 +1,361 @@
...
+    fn map_err<E2: Error>(
+        self,
+        f: impl Fn(Self::Error) -> E2,
+    ) -> impl ComposablePass<Result = Self::Result, Error = E2> {
+        ErrMapper::new(self, f)
+    }
+
+    /// Returns a [ComposablePass] that does "`self` then `other`", so long as
+    /// `other::Err` can be combined with ours.
+    fn then<P: ComposablePass, E: ErrorCombiner<Self::Error, P::Error>>(
```
acl-cqc, line 26: There's a possible method `then2` that requires `E: ErrorCombiner<P::Error, Self::Error>` i.e. that would allow when our error-type maps into P's (as opposed to `then` which requires the other way around).

However, both ways also allow returning an `Either` error and `then2` would reverse that too (you probably wouldn't bother using `then2` if you wanted an `Either` but). So then I wonder, should we also reverse the order of *results*...and the doubt as to what's the right API has lead me to avoid providing `then2` entirely. But I could definitely be persuaded....

(Sadly we can't allow `ErrorCombiner` to work for both `<A, B: Into<A>>` *and* `<A: Into<B>, B>` because that would provide two conflicting impls when `A` and `B` are the same type :-( :-( )

cqc-alec (approved):

hugr-passes/src/composable.rs
```diff
@@ -0,0 +1,361 @@
...
+}
+
+/// Trait for combining the error types from two different passes
+/// into a single error.
+pub trait ErrorCombiner<A, B>: Error {
+    fn from_first(a: A) -> Self;
+    fn from_second(b: B) -> Self;
+}
+
+impl<A: Error, B: Into<A>> ErrorCombiner<A, B> for A {
```
aborgna-q, line 59: It'd be nice to have
```rust
impl<E: Error, A: Into<E>, B: Into<E>> ErrorCombiner<A, B> for E
```
instead, which covers the Either impl.
But that breaks the tests below as it doesn't allow us to have `ErrorCombiner<A, Infallible>` -.-

hugr-passes/src/composable.rs
```diff
@@ -0,0 +1,361 @@
...
+}
+
+// Note: in the short term we could wish for two more impls:
+//   impl<E:Error> ErrorCombiner<Infallible, E> for E
+//   impl<E:Error> ErrorCombiner<E, Infallible> for E
+// however, these aren't possible as they conflict with
+//   impl<A, B:Into<A>> ErrorCombiner<A,B> for A
+// when A=E=Infallible, boo :-(.
+// However this will become possible, indeed automatic, when Infallible is replaced
+// by ! (never_type) as (unlike Infallible) ! converts Into anything
```
aborgna-q, line 86: This is covered by `impl<A: Error, B: Error> ErrorCombiner<A, B> for Either<A, B>`, no?
It is required to compose `DeadCodeElimPass` in `test_then` below.

aborgna-q (approved):
Nice!

hugr-passes/src/composable.rs
acl-cqc: FTR...AFAICS Either does *not* implement From from either left or right, e.g. I get two errors here (one on each `.into()`):
```
fn foo<A,B>(a:A, b:B, which: bool) -> Either<A,B> {
    if which {a.into()} else {b.into()}
}
```

Issue #1797, opened 2024-12-17
constant folding: FuncDefns and FuncDecls are not considered needed


doug-q:
In `find_needed_nodes`

On reflection this is not entirely true. We consider all incoming edges of a needed node to come from needed nodes. So a FuncDefn/Decl is needed if it's called by a needed node.

The issue is that "main" (or more generally, public functions) are not considered needed. 

Constant folding already special cases the inputs of "main", we should at least special case that "main" is needed.

acl-cqc:
Is it worth generalizing this to "establish a convention/API for indicating the entry-points to a Hugr" i.e. not just for constant folding but generally? (DCE/DeadFuncElim, etc. etc.)

acl-cqc:
and/or does it suffice to change this line: https://github.com/CQCL/hugr/blob/cad74844fcfef9d4bcd5e0c36335c26d88fa791f/hugr-passes/src/const_fold.rs#L152
to look for `main` if the root is not a DFG/function but a module?

acl-cqc:
This was fixed by #1901

PR #1896, opened 2025-01-29
feat!: remove implicit "main"-function support from dataflow analysis

acl-cqc:
Passes do not have any automatic support for "main" - when the root is a Module,
* Monomorphization applies to all FuncDefn children
* remove_dead_funcs takes an explicit list of entry points
* ConstantFoldPass removes the entire program (moreover, the doc *says* any inputs provided must be for the root node; in fact if the root is a Module, inputs were indeed applied to `main`, but then the `main` function is removed along with all the others). Yes this is [a bug](#1797) but a better solution will be to explicitly specify what function any provided inputs are for (in progress).

Since dataflow analysis is a lower-level utililty supporting these "passes", it seems odd for it to provide this automatic support (that feels like a higher-level feature). Moreover the above analysis suggests it doesn't really break anything that matters, hence I think it can be done first.

I've also tried to reorganise the code a bit better wrt. `prepopulate_df_inputs` as we can also handle `Conditional` and `CFG` nodes by using existing code, hence renaming to just `prepopulate_inputs`

BREAKING CHANGE: dataflow analysis on Module-rooted Hugrs will no longer accept input PartialValues for the main function, these should be passed to `prepopulate_inputs` instead. Also `prepopulate_df_inputs` is gone, again use `prepopulate_inputs`.


hugr-passes/src/dataflow/datalog.rs
```diff
@@ -36,93 +38,97 @@ impl<H: HugrView, V: AbstractValue> Machine<H, V> {
...
-                .filter_map(|(n, p, _)| (n == &root).then_some(*p))
-                .collect();
-            for p in self.0.signature(root).unwrap_or_default().input_ports() {
-                if !got_inputs.contains(&p) {
-                    self.1.push((root, p, PartialValue::Top));
-                }
+            let mut p = in_values.into_iter().peekable();
+            // We must provide some inputs to the root so that they are Top rather than Bottom.
+            // (However, this test will fail for DataflowBlock or Conditional roots, i.e. if no
+            // inputs have been provided they will still see Bottom. We could store the "input"
```
acl-cqc, line 114: This is a very long note to have as a "We could" comment, it'll probably just get out-of-date ;). Perhaps I should just do it now...

acl-cqc:
Hmmm, whilst I still think this is the right thing to do, I am starting to think that I can tackle #1797 first and then this PR may be better as a followup to that.

hugr-passes/src/dataflow/datalog.rs
acl-cqc: But supporting DataflowBlock/Case roots doesn't seem a high priority. Maybe I should just remove the comment??

hugr-passes/src/const_fold.rs
```diff
@@ -133,10 +146,7 @@ impl ConstantFoldPass {
         }
         // Dataflow analysis applies our inputs to the 'main' function if this is a Module, so do the same here
```
aborgna-q, line 147: Outdated comment

hugr-passes/src/const_fold.rs
```diff
@@ -78,18 +86,23 @@ impl ConstantFoldPass {
...
+                    let const_with_dummy_loc = partial_from_const(
+                        &ConstFoldContext(hugr),
+                        ConstLocation::Field(p.index(), &fresh_node.into()),
+                        v,
+                    );
+                    (*p, const_with_dummy_loc)
+                }),
             )
-        });
+            .unwrap(); // TODO return error?
```
aborgna-q, line 102: Yes please.
This does a partial check for the precondition in `with_inputs`, no?

> Specifies a number of external inputs to an entry point of the Hugr.
> For Module-rooted Hugrs, `node` should be a FuncDefn child of the root.
> For non-Module-rooted Hugrs, `node` should be the root of the Hugr.

(here afaik there's no check that `n` is an entrypoint). Could have the check there and leave the unwrap here?

aborgna-q (commented):
LGTM. Next release will be breaking already, so we should be able to merge this now.

hugr-passes/src/const_fold.rs
```diff
@@ -62,13 +64,19 @@ impl ConstantFoldPass {
         self
     }
 
-    /// Specifies any number of external inputs to provide to the Hugr (on root-node
-    /// in-ports). Each supersedes any previous value on the same in-port.
+    /// Specifies a number of external inputs to an entry point of the Hugr.
+    /// For Module-rooted Hugrs, `node` should be a FuncDefn child of the root.
```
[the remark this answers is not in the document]
acl-cqc: Well, there's no way it could be called other than from nodes already in the Hugr, so I don't see the point? You could use `prepopulate_inputs` if you really wanted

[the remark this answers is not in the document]
acl-cqc: So I think the answer is probably the latter - use `prepopulate_inputs`. Unless you really want to do DF analysis to decide whether to add new uses of a currently-unused function, which sounds a bit wild

hugr-passes/src/const_fold.rs
acl-cqc: Good spot, thanks

hugr-passes/src/const_fold.rs
acl-cqc: Very partial - it checks that it's not, say, an AliasDefn or some other useless node that could be a child of Module, but that's about it. I've added complete checks and error variants, or would be happy to leave out the `EntryPointNotChildOfModule` and `EntryPointNotRoot` checks. These checks are not tested, either - @doug perhaps you could pick that up...

acl-cqc: Removed `EntryPountNotChildOfModule` and `EntryPointNotRoot` in c8fce24b with lots of docs changes instead.

aborgna-q (approved):

Issue #1807, opened 2024-12-18
Separate "dead-code-elimination" from constant folding


doug-q:
the constant folding pass includes a "dead-code-elimination" phase at the end. Let's extract this into a separate pass. Constant folding can (perhaps optionally) invoke that pass.

Problems with the existing dead code elimination in constant folding:
* We need a notion of public function for the "roots" of the pass. #1797 
* We follow edges back from roots. 
  * There is a special case for CFG edges, perhaps this should be cleaned up.  
  * This means an order edge from an op to an output node keeps the op alive, but an order edge from an input node to an op does not keep that op alive. This seems wrong.
   * We have no explicit notion of ops having "side effects". However they can be simulated with order edges from that op to the output node for example:
     * a panic op with no neighbours is considered dead. 
     * A call node is dead if it has no out neighbours.
   * a QFree with no order edge to the output is dead by this definition. The graph will end up invalid, with unused linear ports

PR #1897, opened 2025-01-29
feat: separate DeadCodeElimPass, fix constant-folding of Modules

### LINKED ISSUE #1797: constant folding: FuncDefns and FuncDecls are not considered needed

acl-cqc:
* Constant Folding now just breaks wires and inserts Const + LoadConst even for "unnecessary" (DCE-able) ops
* Followup DCE pass removes these
* DCE pass allows explicit specification of entry points for Module-rooted Hugrs
   * constant folding assumes `main` entry-point (in common with dataflow analysis, for now) - I've added a test for #1797 that this fixes
* DCE pass takes an arbitrary callback for `NodeDivergence` aka "purity" - naming suggestions most welcome here

Could use more DCE-specific tests.

Thus, two in one:
closes #1797, #1807

### DIFF
--- a/hugr-passes/src/const_fold.rs
+++ b/hugr-passes/src/const_fold.rs
@@ -3,7 +3,7 @@
 //! An (example) use of the [dataflow analysis framework](super::dataflow).
 
 pub mod value_handle;
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 use thiserror::Error;
 
 use hugr_core::{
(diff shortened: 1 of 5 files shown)


acl-cqc:
Merging with #1896

PR #1907, opened 2025-02-04
feat: Do not require `capnp` to be installed to compile `hugr-model`

### LINKED ISSUE #1667: (model): standalone py/rust packages for capnproto generated code
Separate workflows should generate code from capnproto, so that it is not needed as a system dependency for hugr use. 
It may be convenient to make the generated code standalone-separately versioned packages.

See: https://github.com/cqcl-dev/quantinuumapis/

aborgna-q:
Now we do the same as with the schema and extensions definition; provide a `just` command to update it manually and check on CI that things are up-to-date.

This lets us remove the `build.rs` from `hugr-model`, removing the need for users to have `capnproto` installed to compile our crates.

Closes #1667


zrho (approved):
LGTM

aborgna-q:
I tried to merge this without rebasing #1893 and the merge queue check failed due to not updating the generated artifact :)

PR #1908, opened 2025-02-05
chore: fix semver checks in CI

doug-q:


aborgna-q:
I think semver checks are working already?
https://github.com/CQCL/hugr/actions/runs/13155044605/job/36710195910

Note that workflow calls are executed in a separate runner, so installing things here won't change the runtime.
I had to add an `apt-dependencies` parameter to the workflow to deal with `capnproto` before.
https://github.com/CQCL/hugrverse-actions/blob/97c4297b1a493b5bf6827ba38b6f7d7fa4891348/.github/workflows/rs-semver-checks.yml#L11-L15

aborgna-q:
Uhm. I see it failing here.

https://github.com/CQCL/hugr/actions/runs/13155673562/job/36712137197#step:10:162

Do you know why it was working before? Is this a spurious error?

aborgna-q:
See #1909

PR #1911, opened 2025-02-06
feat: add xor to logic extension

### LINKED ISSUE #1418: Add `Xor` to the logic extension

ss2165:
Closes  #1418

### DIFF
--- a/hugr-llvm/src/extension/logic.rs
+++ b/hugr-llvm/src/extension/logic.rs
@@ -32,38 +32,11 @@ fn emit_logic_op<'c, H: HugrView>(
         inputs.push(bool_val.build_get_tag(builder)?);
     }
     let res = match lot {
-        LogicOp::And => {
-            let mut acc = inputs[0];
-            for inp in inputs.into_iter().skip(1) {
-                acc = builder.build_and(acc, inp, "")?;
-            }
-            acc
-        }
-        LogicOp::Or => {
-            let mut acc = inputs[0];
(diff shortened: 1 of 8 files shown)


hugr-llvm/src/extension/logic.rs
```diff
@@ -57,6 +57,13 @@ fn emit_logic_op<'c, H: HugrView>(
             acc
         }
         LogicOp::Not => builder.build_not(inputs[0], "")?,
+        LogicOp::Xor => {
+            let mut acc = inputs[0];
+            for inp in inputs.into_iter().skip(1) {
+                acc = builder.build_xor(acc, inp, "")?;
+            }
```
aborgna-q, line 64: Not sure why we do it generically instead of just
```rust
builder.build_xor(inputs[0], inputs[1], "")
```
and even for the n-inputs case, it could use a fold instead
```rust
inputs[1..].fold(Ok(inputs[0]), |a, b| builder.build_xor(a, b, ""))
```
(no [`try_reduce`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_reduce) yet :/ )

aborgna-q (approved):
🤖

hugr-llvm/src/extension/logic.rs
ss2165: I think this is vestigial from when the ops were generic, have simplified as a drive-by

PR #1920, opened 2025-02-13
feat: Expand SimpleReplacement API

lmondada:
I would like to reuse the replacement logic in `tket2`, so I've taken the opportunity to expand the API of SimpleReplacement and refactor things a bit.


croyzor (approved):
Nice! 👍

PR #1921, opened 2025-02-18
fix: delegate default impls in HugrView

lmondada:
Without delegating all methods with default impls, you get the following weirdness that drove me crazy this morning

```rust
trait TraitFoo {
  fn something(&self) -> usize {
    return 42;
  }
}

impl<'t, T: TraitFoo> &'t T {}

struct A;

impl TraitFoo for A {
  fn something(&self) -> usize {
    return 0;
  }
}

let a = A();
a.something()     // is 0
(&a).something()  // is 42 :(
```

### DIFF
--- a/hugr-core/src/hugr/views/impls.rs
+++ b/hugr-core/src/hugr/views/impls.rs
@@ -1,31 +1,87 @@
 use std::{borrow::Cow, rc::Rc, sync::Arc};
 
 use delegate::delegate;
+use itertools::Either;
 
-use super::{HugrView, RootChecked};
-use crate::{Direction, Hugr, Node, Port};
+use super::{render::RenderConfig, HugrView, RootChecked};
+use crate::{
+    extension::ExtensionRegistry,
+    hugr::{NodeMetadata, NodeMetadataMap, ValidationError},
+    ops::OpType,
+    types::{PolyFuncType, Signature, Type},
+    Direction, Hugr, IncomingPort, Node, OutgoingPort, Port,
+};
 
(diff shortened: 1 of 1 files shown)


hugr-core/src/hugr/views/impls.rs
```diff
@@ -1,31 +1,87 @@
...
                 fn neighbours(&self, node: Node, dir: Direction) -> impl Iterator<Item = Node> + Clone;
+                fn input_neighbours(&self, node: Node) -> impl Iterator<Item = Node> + Clone;
+                fn output_neighbours(&self, node: Node) -> impl Iterator<Item = Node> + Clone;
                 fn all_neighbours(&self, node: Node) -> impl Iterator<Item = Node> + Clone;
+                fn get_io(&self, node: Node) -> Option<[Node; 2]>;
+                fn inner_function_type(&self) -> Option<Cow<'_, Signature>>;
+                fn poly_func_type(&self) -> Option<PolyFuncType>;
+                // TODO: cannot use delegate: return type of `as_petgraph`
+                // depends on Self
+                // fn as_petgraph(&self) -> PetgraphWrapper<'_, Self>;
```
aborgna-q, line 72: I think it's fine to keep this out.
`PetgraphWrapper` is a thin wrapper over a `HugrView`, so the default impl should be valid for all implementations.

I'd just leave a comment stating that.

aborgna-q (approved):
Ouch, nice catch -.-'

The `HugrInternal` delegates look alright, so hopefully this is the only place where this can happen.

PR #1924, opened 2025-02-21
fix: Serialize index of control-flow ports

### LINKED ISSUE #1923: hugr-py cannot load CFG edges json-serialized from Rust
Rust produces edges something like:
```
"edges":[[[1,null],[4,null]],[[1,null],[3,null]],[[3,null],[2,null]],[[3,null],[4,null]],[[4,null],[2,null]],
```
...(where nodes 1-4 are DataflowBlocks - note the two `[1, null]` outports for successors 0 and 1 of the same block)

Whereas hugr-py produces
```
"edges":[[[1,0],[5,0]],[[1,1],[10,0]],[[5,0],[4,0]],[[10,0],[4,0]]],
```
...(eliding non-block edges)

In practice, both roundtrip, and Rust is able to load hugr-py output, but not vice versa.

acl-cqc:
and remove logic duplicated in `OpType::port_kind`

The main goal here is to enable hugr-py to read in CFGs serialized from Rust, by changing the Rust to output in the same form as the python.

So, to test this...I should write out a file from the Rust test (i.e. comparing against snapshot)...and then load that into a test in python (that checks the CFG still has edges, it won't without this fix). We don't have any code that does this yet AFAICS so this is a draft while that remains TODO.

fixes #1923

### DIFF
--- a/hugr-core/src/hugr/serialize.rs
+++ b/hugr-core/src/hugr/serialize.rs
@@ -8,7 +8,8 @@ use thiserror::Error;
 use crate::core::NodeIndex;
 use crate::hugr::Hugr;
 use crate::ops::OpType;
-use crate::{Node, PortIndex};
+use crate::types::EdgeKind;
+use crate::{Node, Port, PortIndex};
 use portgraph::hierarchy::AttachError;
 use portgraph::{Direction, LinkError, PortView};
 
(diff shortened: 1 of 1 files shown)


acl-cqc:
I think `hugr-model`s arrival means we can drop this....

PR #1932, opened 2025-02-24
chore!: Add associated type Node to HugrView

lmondada:
Yey, this is a fun one 🫠 

This makes `HugrView` general over the `Node` type. Note that `HugrMut` and all operations that mutate hugrs still only act on the node type `Node`.

I suggest once this is merged in that we add a `.git-blame-ignore-revs` file to simplify future git blames:
```
# Commits to be ignored by git blame

# Add Node associated type to HugrView (large mechanical change)
<commit of this PR on main>
```

### DIFF
--- a/hugr-core/src/core.rs
+++ b/hugr-core/src/core.rs
@@ -44,9 +44,14 @@ pub trait PortIndex {
 }
 
 /// A trait for getting the index of a node.
-pub trait NodeIndex {
+pub trait NodeIndex:
+    Copy + Ord + std::fmt::Debug + std::fmt::Display + std::hash::Hash + From<portgraph::NodeIndex>
+{
     /// Returns the index of the node.
     fn index(self) -> usize;
+
+    /// Returns the node as a portgraph `NodeIndex`.
+    fn pg_index(self) -> portgraph::NodeIndex;
 }
 
 /// A port in the incoming direction.
@@ -73,7 +78,7 @@ pub type Direction = portgraph::Direction;
 )]
 /// A DataFlow wire, defined by a Value-kind output port of a node
 // Stores node and offset to output port
-pub struct Wire(Node, OutgoingPort);
+pub struct Wire<N = Node>(N, OutgoingPort);
 
 impl Node {
     /// Returns the node as a portgraph `NodeIndex`.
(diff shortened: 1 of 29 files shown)


lmondada:
@doug-q for `hugr-llvm`, I have restricted all implementations to `Hugr<Node = Node>`. I hit a roadblock trying to implement it for any `Self::Node` in `Namer`, because it assumes a `Node` can be converted to an integer index.

I don't think we need to support more general `Node` types for LLVM emitting, so this should be harmless.

cqc-alec (approved):

hugr-core/src/core.rs
```diff
@@ -49,6 +49,11 @@ pub trait NodeIndex {
     fn index(self) -> usize;
 }
 
+/// A trait for nodes in the Hugr.
+pub trait HugrNode: Copy + Ord + std::fmt::Debug + std::fmt::Display + std::hash::Hash {}
+
+impl<T: Copy + Ord + std::fmt::Debug + std::fmt::Display + std::hash::Hash> HugrNode for T {}
+
```
aborgna-q, line 56: The reason we don't have `to_pg_index` and `to_node` in the `HugrNode` trait instead of `HugrInternals` is that we need context to create one?

aborgna-q (commented):
🚀

hugr-core/src/core.rs
lmondada: Yes, I'm expecting that node types that are made of (node id, hugr id) pairs might map in different ways to NodeIndex (otherwise might as well stick with Node)

hugr-core/src/hugr/internal.rs
```diff
@@ -26,14 +26,23 @@ pub trait HugrInternals {
...
 
     /// Return the root node of this view.
-    fn root_node(&self) -> Node;
+    fn root_node(&self) -> Self::Node;
+
+    /// Convert a node to a portgraph node index.
+    fn to_pg_index(&self, node: Self::Node) -> portgraph::NodeIndex;
+
+    /// Convert a portgraph node index to a node.
+    fn to_node(&self, index: portgraph::NodeIndex) -> Self::Node;
```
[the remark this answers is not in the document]
lmondada: Good point, will do!

PR #1953, opened 2025-03-06
feat: `hugr-model` AST

### LINKED ISSUE #1600: `Display` implementations using the `hugr-model` text format
The `hugr-model` text format has a pretty printing implementation. To ensure consistency we could use that to implement `Display` for operations and types in `hugr-core`. This would require an export from `hugr-core` to `hugr-model` which works for individual types/operations outside of the context of a module.

zrho:
This PR introduces the AST representation of `hugr-model`, which is less efficient but more ergonomic than the tabled variant. It is used for parsing and pretty printing, and can be converted from and into the tabled representation. Individual parts, such as single terms, can be parsed and printed which will help resolve #1600. The AST representation can also become useful for Python bindings.

* Parsing used to be source -> table, but now source -> AST (-> table)
* Binary representation still loads directly into table form
* table `View`s itself as AST; tree form `resolve`s itself into table
* `struct Literal` factored out from `enum table::Term` and shared with `enum syntax::Term`
* Model `Node`s no longer include params, these are now put inside an `Apply` inside the node.
* Alias values therefore no longer in params but made part of `Operation::DefineAlias`.
* `ListPart` and `TuplePart` combined into `SeqPart`


hugr-core/src/export.rs
```diff
@@ -364,16 +363,16 @@ impl<'a> Context<'a> {
...
                 // TODO PERFORMANCE: Avoid exporting the signature here again.
                 let signature = call.signature();
                 let inputs = self.export_type_row(&signature.input);
                 let outputs = self.export_type_row(&signature.output);
                 let ext = self.export_ext_set(&signature.runtime_reqs);
-
-                params = self.bump.alloc_slice_copy(&[inputs, outputs, ext, func]);
-                model::Operation::Custom(self.resolve_symbol(model::CORE_CALL_INDIRECT))
+                let operation =
+                    self.make_term_apply(model::CORE_CALL_INDIRECT, &[inputs, outputs, ext, func]);
```
acl-cqc, line 374: Not new here but it is a little surprising that we are converting an `OpType::Call` (statically-known target) to a `CORE_CALL_INDIRECT`. The latter seems to take the inputs I'd expect for an `OpType::Call` (an Apply of a symbol term for the static bit, plus runtime value inputs/outputs, and extensionset).hugr-core also has `OpType:CallIndirect` (target determined at runtime), which we export into model without the  static `func` param. This sounds like substantial adhoc overloading of CORE_CALL_INDIRECT, to put it mildly??

hugr-core/src/export.rs
```diff
@@ -364,16 +363,16 @@ impl<'a> Context<'a> {
...
+                table::Operation::Custom(operation)
```
acl-cqc, line 375: driveby: it's a bit sad to call this `Custom` when much of `hugr-core` falls into this!

acl-cqc:
* I'm not entirely swayed by calling the module `syntax`. To me that suggests more textual/frontend than is actually the case. `ast` would be better (the *abstract* syntax preventing the confusion with source/text), but then, it's not really even a tree - it has sharing with Arcs. Perhaps even calling it `arc` then??
* A 3.5k line PR needs a changelist. I've started trying to write one in the description but you could contribute ;-)
* And of course, this 3.5k line PR could be broken up into a large and very boring "Rename model to table" and then an "add AST" PR which would be that much smaller....

hugr-core/src/export.rs
acl-cqc: I note CORE_CALL does exist, should this be that?

hugr-core/src/import.rs
```diff
@@ -372,32 +373,26 @@ impl<'a> Context<'a> {
...
                 }
 
-                if name == model::CORE_CALL {
-                    let &[_, _, _, func] = node_data.params else {
-                        return Err(model::ModelError::InvalidOperation(node_id).into());
-                    };
-
-                    let model::Term::Apply(symbol, args) = self.get_term(func)? else {
-                        return Err(model::ModelError::TypeError(func).into());
+                if let Some([_, _, _, func]) = self.match_symbol(operation, model::CORE_CALL)? {
```
acl-cqc, line 393: This doesn't look to line up with export of `OpType::Call`

hugr-core/src/import.rs
```diff
@@ -500,14 +487,16 @@ impl<'a> Context<'a> {
                     return Ok(Some(node));
                 }
 
-                let signature = self.get_node_signature(node_id)?;
-                let args = node_data
-                    .params
+                let table::Term::Apply(node, params) = self.get_term(operation)? else {
+                    return Err(table::ModelError::TypeError(operation).into());
```
acl-cqc, line 491: This looks a bit like "unknown operation (I've tried everything I know)" - is there no better error type?

hugr-core/tests/model.rs
```diff
@@ -1,15 +1,19 @@
 #![allow(missing_docs)]
 
+use std::str::FromStr;
+
 use hugr::std_extensions::std_reg;
 use hugr_core::{export::export_hugr, import::import_hugr};
-use hugr_model::v0 as model;
+use hugr_model::v0::{self as model};
```
acl-cqc, line 7: This seems a very odd one-line change

hugr-model/src/v0/binary/write.rs
```diff
@@ -41,156 +42,152 @@ pub fn write_to_vec(module: &model::Module) -> Vec<u8> {
...
-        model::Operation::DeclareFunc(symbol) => {
+        table::Operation::DeclareFunc(symbol) => {
             let builder = builder.init_func_decl();
             write_symbol(builder, symbol);
         }
 
-        model::Operation::DefineAlias(symbol) => {
-            let builder = builder.init_alias_defn();
-            write_symbol(builder, symbol);
+        table::Operation::DefineAlias(symbol, value) => {
```
acl-cqc, line 79: What's the change here? (Add to PR description?)

hugr-model/src/v0/mod.rs
```diff
@@ -4,43 +4,38 @@
...
-//! specific in-memory representation.
-//!
-//! The tabling is also used for deduplication of terms. In practice, many terms will share
-//! the same subterms, and we can save memory and validation time by storing them only once.
-//! However we allow non-deduplicated terms for cases in which terms carry additional identity
-//! over just their structure. For instance, structurally identical terms could originate
-//! from different locations in a text file and therefore should be treated differently when
-//! locating type errors.
+//! - [Table]: Efficient intermediate data structure to facilitate conversions.
+//! - [Binary]: Binary serialisation format optimised for performance.
```
acl-cqc, line 14:
```suggestion
//! - [Binary]: Binary serialisation format optimised for compactness.
```

hugr-model/src/v0/mod.rs
```diff
@@ -4,43 +4,38 @@
...
+//! - [AST]: Abstract syntax tree that is more ergonomic than the table form.
```
acl-cqc, line 15:
```suggestion
//! - [AST]: Abstract syntax tree using direct Arcs rather than table indices
```

hugr-model/src/v0/binary/mod.rs
```diff
@@ -1,4 +1,12 @@
-//! The HUGR binary representation.
+//! Binary format based on capnproto.
+//!
+//! The binary format is optimised for fast serialization and deserialization of
+//! hugr modules in the [table] representation. It is the preferred format to
+//! communicate hugr graphs between machines. When a hugr module is to be
+//! written or read by humans, the [text] format can be used instead.
```
acl-cqc, line 6: This probably needs updating, particularly that `[text]` is a reference to `syntax`!

hugr-model/src/v0/mod.rs
```diff
@@ -4,43 +4,38 @@
...
+//! - [Text]: Human readable s-expression based text format.
```
acl-cqc, line 16: Hang on. This is a reference to exactly the same thing as the previous `[AST]`

hugr-model/src/v0/scope/symbol.rs
```diff
@@ -4,7 +4,7 @@ use fxhash::FxHasher;
 use indexmap::IndexMap;
 use thiserror::Error;
 
-use crate::v0::{NodeId, RegionId};
+use crate::v0::table::{NodeId, RegionId};
```
acl-cqc, line 7: How much of this is common, how much of this belongs in `table` instead?

hugr-model/src/v0/syntax/mod.rs
```diff
@@ -0,0 +1,264 @@
+//! Abstract syntax tree representation of hugr modules.
+//!
+//! This module defines the abstract syntax tree for hugr graphs. The AST types
+//! implement [`Display`] and [`FromStr`] for pretty printing into and parsing
+//! from the s-expression based hugr text form. This is useful for debugging and
+//! writing hugr modules by hand. For a more performant serialization format, see
+//! [binary] instead.
+//!
+//! The data types in this module logically mirror those of the [tabled]
```
acl-cqc, line 9:
```suggestion
//! The data types in this module logically mirror those of the [table]
```

hugr-model/src/v0/scope/symbol.rs
acl-cqc: Argh there is a `table::Symbol` as well (and a `syntax::Symbol`...)

hugr-model/src/v0/syntax/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub operation: Operation,
+
+    /// The names of the links connected to the input ports of the node.
+    pub inputs: Box<[LinkName]>,
+
+    /// The names of the links connected to the output ports of the node.
+    pub outputs: Box<[LinkName]>,
+
+    /// The regions in the node.
+    pub regions: Box<[Region]>,
```
acl-cqc, line 87: Why `Box<[Region]>` rather than `Vec<Region` (etc.) ?

hugr-model/src/v0/syntax/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+    DeclareOperation(Box<Symbol>),
+    /// Symbol imports.
+    Import(SymbolName),
+}
+
+impl Operation {
+    /// The name of the symbol introduced by this operation, if any.
+    pub fn symbol_name(&self) -> Option<&SymbolName> {
+        match self {
+            Operation::DefineFunc(symbol) => Some(&symbol.name),
```
acl-cqc, line 138: nit: You might use a big `|` here

hugr-model/src/v0/syntax/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// Operation declarations.
+    DeclareOperation(Box<Symbol>),
+    /// Symbol imports.
+    Import(SymbolName),
+}
+
+impl Operation {
+    /// The name of the symbol introduced by this operation, if any.
+    pub fn symbol_name(&self) -> Option<&SymbolName> {
+        match self {
```
acl-cqc, line 137: Consider
```
if let Operation::Import(symbol_name) = self {
   Some(symbol_name)
} else self.symbol().map(|s| &s.name)
```

hugr-model/src/v0/syntax/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// See [`table::Region`] for the table representation.
+///
+/// [`table::Region`]: crate::v0::table::Region
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct Region {
+    /// The kind of the region. See [`RegionKind`] for details.
+    pub kind: RegionKind,
+    /// The names of the links connected to the source ports of the region.
+    pub sources: Box<[LinkName]>,
```
acl-cqc, line 203: Box<[LinkName]> vs Vec<LinkName> as above.

Is it the case that this will be empty if `scope == ScopeClosure::Closed`? In which case is there a possibility of putting the `sources` and `targets` inside the closure enum?

hugr-model/src/v0/syntax/parse.rs
```diff
@@ -0,0 +1,455 @@
...
+
+fn parse_seq_part(pair: Pair<Rule>) -> ParseResult<SeqPart> {
+    debug_assert_eq!(pair.as_rule(), Rule::part);
+    let pair = pair.into_inner().next().unwrap();
+
+    Ok(match pair.as_rule() {
+        Rule::term => SeqPart::Item(parse_term(pair)?),
+        Rule::spliced_term => {
+            let mut pairs = pair.into_inner();
+            let term = parse_term(pairs.next().unwrap())?;
```
acl-cqc, line 118:
```suggestion
            let term = parse_term(pairs.exactly_one().unwrap())?;
```

hugr-model/src/v0/syntax/print.rs
```diff
@@ -0,0 +1,446 @@
...
+            arena,
+            docs: Vec::new(),
+            docs_stack: Vec::new(),
+        }
+    }
+
+    fn finish(self) -> RefDoc<'a> {
+        let sep = self
+            .arena
+            .concat([self.arena.hardline(), self.arena.hardline()]);
```
acl-cqc, line 33: nit: `vec![self.arena.hardline(); 2]` might work

hugr-model/src/v0/syntax/print.rs
```diff
@@ -0,0 +1,446 @@
...
+
+    fn parens_enter(&mut self) {
+        self.delim_open();
+    }
+
+    fn parens_exit(&mut self) {
+        self.delim_close("(", ")", 2);
+    }
+
+    fn brackets_enter(&mut self) {
```
acl-cqc, line 45: maybe `sqbrackets`?

hugr-model/src/v0/syntax/resolve.rs
```diff
@@ -0,0 +1,373 @@
...
+        };
+
+        self.module.regions[self.module.root.index()] = table::Region {
+            kind: RegionKind::Module,
+            sources: &[],
+            targets: &[],
+            children: all_children,
+            meta,
+            signature: None,
+            scope,
```
acl-cqc, line 64: It'd be clearer if you Inlined `scope` here, it's not used elsewhere and this way we can see the outer structure is fixed

hugr-model/src/v0/syntax/resolve.rs
```diff
@@ -0,0 +1,373 @@
...
+        // Allocate ids for all nodes by introducing placeholders into the module.
+        let ids: &[_] = self.bump.alloc_slice_fill_with(nodes.len(), |_| {
+            self.module.insert_node(table::Node::default())
+        });
+
+        // For those nodes that introduce symbols, we then associate the symbol
+        // with the id of the node. This serves as a form of forward declaration
+        // so that the symbol is visible in the current region regardless of the
+        // order of the nodes.
+        for (id, node) in ids.iter().zip(nodes) {
```
acl-cqc, line 120: Would `itertools.zip_eq` work here? Or are we expecting the nodes that introduce symbols to be before the nodes that don't?

acl-cqc: Actually I might just do this is the same loop as where you make `ids` above. Shorter and more obvious that it works.

acl-cqc: I mean, there are many places where I'd like to see `alloc_slice_fill_with` rather than making a thing with capacity and then filling one element at a time, but I don't think this is one

hugr-model/src/v0/syntax/resolve.rs
```diff
@@ -0,0 +1,373 @@
...
+            meta,
+            signature,
+            scope,
+        };
+
+        Ok(region_id)
+    }
+
+    fn resolve_symbol(&mut self, symbol: &'a Symbol) -> BuildResult<&'a table::Symbol<'a>> {
+        let name = symbol.name.as_ref();
```
acl-cqc, line 273: Do you not need to do something to enter a new scope/layer of Vars here?

hugr-model/src/v0/syntax/resolve.rs
```diff
@@ -0,0 +1,373 @@
...
+            .map_err(|_| ResolveError::UnknownVar(var_name.clone()))
+    }
+
+    /// Resolves a symbol name and returns the node that introduces the symbol.
+    ///
+    /// When there is no symbol with this name in scope, we create a new import
+    /// node in the module and record that the symbol has been implicitly
+    /// imported. At the end of the building process, these import nodes are
+    /// inserted into the module's scope.
+    fn resolve_symbol_name(&mut self, symbol_name: &'a SymbolName) -> BuildResult<NodeId> {
```
acl-cqc, line 325: Does this function ever return an error rather than `Ok(someNodeId)` ?

hugr-model/src/v0/table/mod.rs
acl-cqc, line 1: Modulo SeqPart and Literal I think this is not changed, so not diving in

acl-cqc (commented):
Oooof. Got there. General approval on the large questions but a bit of renaming, fixing doclinks etc. would definitely help.

hugr-core/src/export.rs
zrho: I agree that `Custom` isn't really a good name for this. Do you have an idea for a more appropriate name?

hugr-core/src/import.rs
zrho: The error in that line doesn't get triggered when the operation is unknown, but when the term is not a symbol application. But yes, the way errors get handled here isn't ideal yet. I've not put much more effort into this yet since it also overlaps with future validation code.

hugr-core/tests/model.rs
```diff
@@ -1,15 +1,19 @@
...
 
 fn roundtrip(source: &str) -> String {
     let bump = model::bumpalo::Bump::new();
-    let parsed_model = model::text::parse(source, &bump).unwrap();
-    let imported_hugr = import_hugr(&parsed_model.module, &std_reg()).unwrap();
-    let exported_model = export_hugr(&imported_hugr, &bump);
-    model::text::print_to_string(&exported_model, 80).unwrap()
+    let module_ast = model::syntax::Module::from_str(source).unwrap();
```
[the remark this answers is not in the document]
zrho: It probably should, yes. I can try it.

hugr-core/tests/snapshots/model__roundtrip_add.snap
```diff
@@ -21,4 +27,4 @@ expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-add.
         (core.fn
           [arithmetic.int.types.int arithmetic.int.types.int]
           [arithmetic.int.types.int]
-          (ext arithmetic.int))))))
+          (ext))))))
```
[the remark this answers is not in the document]
zrho: Indeed it would break the snapshot comparison. But that's a temporary issue.

hugr-model/src/v0/syntax/mod.rs
zrho: No: there are regions that are closed but have sources and targets. Closed regions just can't have non-local edges passing into them.

hugr-model/src/v0/syntax/resolve.rs
zrho: I'll do the zipeq, but I've played around with different ways to phrase that function and this has turned out to be the cleanest given the need to forward declare the symbols.

hugr-model/src/v0/syntax/resolve.rs
```diff
@@ -0,0 +1,373 @@
...
+        let resolved = self.symbols.resolve(symbol_name.as_ref());
```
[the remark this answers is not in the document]
zrho: I have commented this! `self.symbols.resolve` errors when the symbol can not be found. As per the comment on the function, this leads to an import node being created.

hugr-model/src/v0/syntax/resolve.rs
zrho: It used to, but doesn't any longer. I'll change it to return a `NodeId`.

hugr-core/tests/model.rs
zrho: LSP doing LSP things, will fix it.

hugr-model/src/v0/syntax/print.rs
zrho: But will allocate.

hugr-core/src/import.rs
zrho: Good catch, the export of `OpType::Call` accidentally generated `model::CORE_CALL_INDIRECT` instead of `model::CORE_CALL`.

hugr-core/src/export.rs
zrho: Yes, fixed.

hugr-model/src/v0/binary/write.rs
zrho: Added to PR description.

zrho:
Regarding the naming, I'm also fine with calling it `ast` instead of `syntax`. The sharing in the terms via `Arc`s is meant as a potential optimisation of what conceptually is still tree. Any sharing is there to save space or time, but has no semantic significance and isn't even used at the moment. So I don't see a problem with calling it a tree.

hugr-model/src/v0/syntax/resolve.rs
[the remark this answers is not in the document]
acl-cqc: Ah yes, didn't quite connect the doc comment about implicit import with the resolved being an `Err`. Consider an inline non-doc `// Symbol not found, import instead` but probably ok

hugr-core/src/export.rs
acl-cqc: That's one for standup or hugr channel, I reckon, rather than this PR

hugr-model/src/v0/mod.rs
acl-cqc: Yes:
```suggestion
//! - [AST]: Human readable s-expression based text format.
```
and I think we're good

hugr-model/src/v0/syntax/resolve.rs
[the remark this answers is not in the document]
acl-cqc: I meant, inline `resolved` - it makes it clear you've not doing anything else with the error later

hugr-model/src/v0/syntax/mod.rs
acl-cqc: Given this structure is not used for the performance-critical binary serialization case, I'd argue that you are probably over-optimizing (and micro-optimizing) here. However it makes a certain amount of sense what with the AST not really being designed for mutation. It's probably worth noting your comment about 8 bytes somewhere though...

hugr-model/src/v0/syntax/mod.rs
acl-cqc: In which case ignore what I said :). I like what you've done with removing the ScopeClosure parameter and moving it to `resolve_region`(s), though.

Might be good to add an `That is, there are no incoming nonlocal edges` to the comment on ScopeClosure.

hugr-model/src/v0/syntax/resolve.rs
acl-cqc: Still this one.

hugr-model/src/v0/syntax/resolve.rs
acl-cqc: Ok, your call

acl-cqc (approved):
Approved after renaming `syntax` to `ast` and any others you feel like doing ;-)

PR #1956, opened 2025-03-06
wip: start on bytecasting ops

croyzor:
Start work on bytecast ops for #1937.

1) I didn't implement constant folding, because I thought these ops should only ever come up in the llvm pipeline. My test fails because constant folding isn't implemented and i have no idea why that should be

2) I think there's a schema that needs to be regenerated with the new op signatures before this goes in

### DIFF
--- a/hugr-core/src/std_extensions/arithmetic/conversions/const_fold.rs
+++ b/hugr-core/src/std_extensions/arithmetic/conversions/const_fold.rs
@@ -32,6 +32,8 @@ pub(super) fn set_fold(op: &ConvertOpDef, def: &mut OpDef) {
         itostring_s => def.set_constant_folder(IToStringS),
         itousize => def.set_constant_folder(IToUsize),
         ifromusize => def.set_constant_folder(IFromUsize),
+        bytecast_float_int => panic!("Bytecasting not supported in constant folding"),
+        bytecast_int_float => panic!("Bytecasting not supported in constant folding"),
     }
 }
 
--- a/hugr-llvm/src/extension/conversions.rs
+++ b/hugr-llvm/src/extension/conversions.rs
@@ -213,7 +213,15 @@ fn emit_conversion_op<'c, H: HugrView<Node = Node>>(
                 };
                 Ok(vec![res])
             })
-        }
+        },
+        ConvertOpDef::bytecast_int_float => emit_custom_unary_op(context, args, |ctx, arg, outs| {
+            let [out] = outs.try_into()?;
+            Ok(vec![ctx.builder().build_signed_int_to_float(arg.into_int_value(), out.into_float_type(), "")?.as_basic_value_enum()])
+        }),
+        ConvertOpDef::bytecast_float_int => emit_custom_unary_op(context, args, |ctx, arg, outs| {
+            let [out] = outs.try_into()?;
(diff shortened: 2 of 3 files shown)


hugr-core/src/std_extensions/arithmetic/conversions/const_fold.rs
```diff
@@ -32,6 +32,7 @@ pub(super) fn set_fold(op: &ConvertOpDef, def: &mut OpDef) {
         itostring_s => def.set_constant_folder(IToStringS),
         itousize => def.set_constant_folder(IToUsize),
         ifromusize => def.set_constant_folder(IFromUsize),
+        bytecast_float_int | bytecast_int_float => (), // We don't have constant folders for bytecasting yet
```
acl-cqc, line 35: Doesn't look that hard using `std::mem::transmute` or some kind of cast. Of course, either is "unsafe". Or, just serialize to a byte array? I guess OK to leave for now.

hugr-llvm/src/extension/conversions.rs
```diff
@@ -214,6 +214,21 @@ fn emit_conversion_op<'c, H: HugrView<Node = Node>>(
...
+                Ok(vec![ctx.builder().build_bit_cast(arg, out, "")?])
+            })
+        }
+        ConvertOpDef::bytecast_float_int => {
+            emit_custom_unary_op(context, args, |ctx, arg, outs| {
+                let [out] = outs.try_into()?;
+                Ok(vec![ctx
+                    .builder()
+                    .build_bit_cast(arg, out, "")?
+                    .as_basic_value_enum()])
```
acl-cqc, line 229: it's not obvious how adding a `.as_basic_value_enum()` makes what would otherwise be an int->float bytecast into a float->int one. That's the only difference I can see, but I suspect it doesn't - is this `as_basic_value_enum` necessary?

It'd be nice to have some type annotations or (`debug_`)`asserts` that make the difference obvious.

hugr-llvm/src/extension/conversions.rs
```diff
@@ -619,4 +634,38 @@ mod test {
...
     }
+
+    #[rstest]
+    #[case(42)]
+    #[case(f64::INFINITY.to_bits())]
+    #[case(f64::NEG_INFINITY.to_bits())]
+    #[case(f64::NAN.to_bits())]
+    #[case((-0.0f64).to_bits())]
+    #[case(0.0f64.to_bits())]
+    fn bytecast_roundtrip(mut exec_ctx: TestContext, #[case] i: u64) {
```
acl-cqc, line 645: These are all int -> float -> int, AFAICS, with initial **Rust** float->int in all cases bar the first. I think it'd be good to check the float in the middle is actually what we expect (in some cases if not all).

Or are we saying that the floating-point representation is not specified? (not e.g. IEEE? and if it is IEEE, then there is still the question of endianness. I guess it may depend upon host platform, I'd be happy to have some platform-specific test config....)

acl-cqc: If we really want to just leave it as "we let LLVM do whatever it wants to do and trust that that's ok" then let's at least put a comment to that effect!

acl-cqc (commented):
Generally, nice, neat! I am wondering about the spec of these ops though...

cqc-alec (approved):

PR #1958, opened 2025-03-10
feat!: Generic HUGR serialization with envelopes

### LINKED ISSUE #1874: `Package::to_json_file` is wrong by inspection
It opens the file in read mode then writes to it. No users(or tests), but it's a landmine.

Locally I now have:
```
    /// Write the Package into a json file.
    pub fn to_json_file(&self, path: impl AsRef<Path>) -> Result<(), PackageEncodingError> {
        let file = fs::OpenOptions::new().write(true).truncate(true).create(true).open(path)?;
        let writer = io::BufWriter::new(file);
        self.to_json_writer(writer)
    }
```

ss2165:
Adds a "hugr envelope" serialization format composed by a 10-byte header and some payload.

The header contains a magic string at the start, some flags, and a format identifier.
The payload can then be encoded using either the old package json encoding, or hugr-module binary format. There's also a flag to do zstd compression.

I added new methods to `Package` both on the rust and python API, and deprecated the json-based ones. We should be able to add new serialization options and formats in the future without changing this new API.

`hugr-cli` now expects these envelopes by default. Since we still support hugr jsons for some use cases (e.g. non-module rooted hugrs that specify rewrites) I added a `--hugr-json` flag to read these jsons instead.

drive-by: Fix #1874 

BREAKING CHANGE: hugr-cli now expects hugr envelopes. Use `--hugr-json` to load raw hugrs.

---

Override for python CHANGELOG (to make them non-breaking)
BEGIN_COMMIT_OVERRIDE
feat: Generic HUGR serialization with envelopes
END_COMMIT_OVERRIDE


hugr-core/Cargo.toml
aborgna-q, line 1: Couldn't help myself from cleaning up the messy dependency list -.-

hugr-py/src/hugr/envelope.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+
+    format: EnvelopeFormat = EnvelopeFormat.JSON
+    zstd: int | None = None
+
+    def _make_header(self) -> EnvelopeHeader:
+        return EnvelopeHeader(format=self.format, zstd=self.zstd is not None)
+
+    @staticmethod
+    def TEXT() -> EnvelopeConfig:
```
aborgna-q, line 180: I'd love to make this a `@property`, but it doesn't seem to work with the static decorator.

ss2165: should it be a classvar with a default factory?

hugr-cli/src/hugr_io.rs
```diff
@@ -0,0 +1,85 @@
...
+use std::io::BufReader;
+use std::path::PathBuf;
+
+use crate::CliError;
+
+/// Arguments for reading a HUGR input.
+#[derive(Debug, clap::Args)]
+pub struct HugrInputArgs {
+    /// Input file. Defaults to `-` for stdin.
+    #[arg(value_parser, default_value = "-", help_heading = "Input")]
```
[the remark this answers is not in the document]
aborgna-q: This is formatting of the help output, independent from `group`.
I haven't found a way to avoid repeating it -.-

Current help output:
```
$ hugr validate -h
Validate and visualize a HUGR file

Usage: hugr validate [OPTIONS] [INPUT]

Options:
  -v, --verbose...  Increase logging verbosity
  -q, --quiet...    Decrease logging verbosity
  -h, --help        Print help (see more with '--help')
  -V, --version     Print version

Input:
      --no-std                   Don't use standard extensions when validating hugrs. Prelude is still used.
  -e, --extensions <EXTENSIONS>  Paths to serialised extensions to validate against.
      --hugr-json                Read the input as a HUGR JSON file instead of an envelope
  [INPUT]                    Input file. Defaults to `-` for stdin [default: -]
```

hugr-core/Cargo.toml
```diff
@@ -29,30 +30,35 @@ name = "model"
...
+html-escape = { workspace = true }
+indexmap.workspace = true
+itertools = { workspace = true }
 lazy_static = { workspace = true }
+packed_struct = { workspace = true, features = ["byte_types_64"] }
+paste = { workspace = true }
 petgraph = { workspace = true }
+portgraph = { workspace = true, features = ["serde", "petgraph"] }
+regex = { workspace = true }
+# Rc used here for Extension, but unfortunately we must turn the feature on globally
```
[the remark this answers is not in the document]
aborgna-q: The comment was there before, but yes.
The "unfortunate" part is that instead of getting a warning from serde when deriving serialization of an object containing an `Rc` or `Arc`, things will work but maybe have unexpected behaviour since serde [doesn't deduplicate values](https://serde.rs/feature-flags.html#rc).

It's not super terrible though.

aborgna-q:
I'll add an issue to add zstd to the python side and make it a runtime dependency by default after this.

I agree that it should be the default _sometimes_, although the "text" config will remain uncompressed json for now (until we replace it with hugr-model's text repr ?).

Issue #882, opened 2024-03-15
Linking


ss2165:
Roughly, this means combining two Hugrs into a single Hugr, and in the process replacing static edges from FuncDecls in each with edges from corresponding FuncDefns in the other Hugr. Note this may be done via `hugr-model` (@zrho link?). There are a number of questions still to be resolved....
* Do both input Hugrs have to have a module root? or can one be a non-module (the children of the module can be transferred into any non-module, but potentially "one way" i.e. the Module cannot refer to FuncDefns in the non-module). I don't see how both can be non-module...
* (How does the linker identify) **which** FuncDecls correspond to which FuncDefns?
   * We may wish to consider the ability to pass extra command-line flags to the linker to change this behaviour e.g. aliasing, extra directives, ....
   * The ability to rename functions could be useful to avoid conflicts, too
* **which** FuncDefns/FuncDecls does it consider - all? or only children of the (Module?) root(s)?

One answer to both of the "which" questions might be metadata. Or, function name - if we consider only FuncDefns that are immediate children of the (Module) root, we can require these to be unique.

* Does it make sense to link a Hugr against itself? There are some interpretations where this makes no sense (e.g. if we only consider FuncDecls+Defns that are immediate children of the root, and require them all to have different names), but others where it might (e.g. if we consider FuncDecls further down the tree, we might link these with FuncDefns just beneath the root)

zrho:
The input of linking is a `Package`, i.e. a sequence of `Module`s. Each `Module` has imports, declarations and definitions of symbols in its root region. The goal of linking then is to match up these symbols across the modules by their name. It may then also create a merged module (optionally stripped of symbols that aren't used).

Imports, declarations and definitions are increasingly specific. An import just declares a name but carries no further information. A declaration has the signature of the symbol being defined and knows whether it is a function, an alias, an operation or a constructor. A definition of a function additionally includes the function body, and an alias definition the value of the alias.

Linking should fail when:
- A symbol is declared to be different things (e.g. once as a function and once as a constructor).
- A symbol is defined twice.
- (Tricky) A symbol is declared/defined with different types. Checking this might be slightly tricky.

There are still some open questions:
- Are all symbols in the module roots subject to linking or do we require symbols to be explicitly marked as public in order for them to be visible from other modules?
  - If so, do we add special support for this in the data structures or do we use metadata?
  - Beyond just visibility, do we also include the option to export the symbol by a different name?
[...]

PR #1959, opened 2025-03-10
feat: Python bindings for `hugr-model`.

zrho:
This PR introduces Python bindings to `hugr-model`. On the Python side, we have dataclasses that mirror the model AST. On the Rust side, these classes are converted into the Rust AST structure. Using that, we can generate the text and binary representation.


ss2165:
related #882

hugr-py/src/hugr/tys.py
```diff
@@ -82,6 +91,10 @@ def resolve(self, registry: ext.ExtensionRegistry) -> Type:
         """Resolve types in the type using the given registry."""
         return self
 
+    def to_model(self) -> model.Term | model.Splice:
+        """Convert the type to a model Term."""
+        raise NotImplementedError(self)
```
croyzor, line 96: The absence of implementations for `Type`, `TypeParam`, `TypeArg` seem like significant omissions! Is there a reasoning behind it?

hugr-py/src/hugr/tys.py
```diff
@@ -103,6 +116,10 @@ def _to_serial(self) -> stys.TypeTypeParam:
     def __str__(self) -> str:
         return str(self.bound)
 
+    def to_model(self) -> model.Term:
+        # Note that we drop the bound.
```
croyzor, line 120: Why are we dropping the bounds here and in `BoundedNatParam`?

hugr-py/src/hugr/tys.py
zrho: These classes are protocols. The actual types, type params, etc inherit from the class and implement the method accordingly.

hugr-py/src/hugr/tys.py
zrho: They have no representation in hugr-model. There was a big discussion on this half a year ago, resulting in the realisation that the bounds introduce subtyping and therefore make the type system unnecessarily complex. But without having a stable serialisation format, we were unable to change anything about that, leading to an in part still unresolved chicken egg problem. Bounds for type variables in some situations are emulated with constraints.

hugr-py/src/hugr/tys.py
```diff
@@ -656,6 +759,10 @@ def _to_serial(self) -> stys.Qubit:
     def __repr__(self) -> str:
         return "Qubit"
 
+    def to_model(self) -> model.Term:
+        # TODO: Is this the correct name?
+        return model.Apply("prelude.Qubit", [])
```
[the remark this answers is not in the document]
zrho: Corrected. It's rather odd that qubits live in prelude, but that's something for another time.

hugr-py/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "hugr-py"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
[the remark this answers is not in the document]
zrho: I've made the edition and rust-version track the workspace now. Do you mean to set it explicitly to 2024 and 1.85 in the crate or in the workspace instead?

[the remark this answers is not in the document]
aborgna-q: Not in the workspace, `hugr` is at the base of a big dependency tree.
We could set the version/edition for this single crate since it's not depended on by anyone if we wanted to, but it's simpler to just follow the workspace

PR #1964, opened 2025-03-12
feat: Add collections.static_array extension.

doug-q:


hugr-core/src/std_extensions/collections/static_array.rs
```diff
@@ -0,0 +1,396 @@
...
+            pub fn get_element_type(&self) -> &Type;
+        }
+
+    }
+
+    /// Create a new [CustomConst] for an array of values of type `typ`.
+    /// That all values are of type `typ` is not checked here.
+    pub fn new(name: impl ToString, typ: Type, contents: impl IntoIterator<Item = Value>) -> Self {
+        Self {
+            value: ArrayValue::new(typ, contents),
```
mark-koch, line 63: We should check that `typ` is copyable (and also document that this is a requirement)

hugr-core/src/std_extensions/collections/static_array.rs
```diff
@@ -0,0 +1,396 @@
...
+        })
+    };
+}
+
+fn instantiate_const_static_array_custom_type(
+    def: &TypeDef,
+    element_ty: impl Into<TypeArg>,
+) -> CustomType {
+    def.instantiate([element_ty.into()])
+        .expect("const array parameters are valid")
```
mark-koch, line 136: This `expect` will fail if someone passes a non-copyable type. Maybe a more informative panic message would be good?

hugr-core/src/std_extensions/collections/static_array.rs
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    fn extension(&self) -> ExtensionId {
+        EXTENSION_ID.to_owned()
+    }
+
+    fn description(&self) -> String {
+        match self {
+            Self::get => "TODO",
+            Self::len => "TODO",
```
mark-koch, line 214: Still needs filling in

hugr-core/src/std_extensions/collections/static_array.rs
```diff
@@ -0,0 +1,396 @@
...
+        }
+        .into()
+    }
+
+    // This method is re-defined here since we need to pass the list type def while computing the signature,
```
mark-koch, line 219:
```suggestion
    // This method is re-defined here since we need to pass the static array type def while computing the signature,
```

hugr-core/src/std_extensions/collections/static_array.rs
```diff
@@ -0,0 +1,396 @@
...
+    type Def = StaticArrayOpDef;
+}
+
+impl HasConcrete for StaticArrayOpDef {
+    type Concrete = StaticArrayOp;
+
+    fn instantiate(&self, type_args: &[TypeArg]) -> Result<Self::Concrete, OpLoadError> {
+        match type_args {
+            [arg] => {
+                let elem_ty = arg.as_type().ok_or(SignatureError::TypeArgMismatch(
```
mark-koch, line 277: Also need to check that the type is copyable

hugr-core/src/std_extensions/collections/static_array.rs
```diff
@@ -0,0 +1,396 @@
...
+/// Reported unique name of the extension
+pub const EXTENSION_ID: ExtensionId = ExtensionId::new_static_unchecked("collections.static_array");
+/// Reported unique name of the array type.
+pub const STATIC_ARRAY_TYPENAME: TypeName = TypeName::new_inline("static_array");
+/// Extension version.
+pub const VERSION: semver::Version = semver::Version::new(0, 1, 0);
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq, derive_more::From)]
+/// Statically sized array of values, all of the same type.
+pub struct StaticArrayValue {
```
mark-koch, line 43: Need to check that serde serialisation is space efficient enough for O(100.000) elements

mark-koch (approved):
Looks good! Note that the json version of the extension is missing. This should trigger a CI failure, but probably not in your case since it's a new extension?

Also it would be good to add the extension to hugr-py, but happy for that to be in a different PR

hugr-core/src/std_extensions/collections/static_array.rs
mark-koch: Mhmm `as_type_with_bound`?

mark-koch (approved):
👍

PR #1977, opened 2025-03-14
chore: update to devenv 1.4.1

ss2165:


devenv.nix
```diff
@@ -12,7 +12,7 @@ in
 
     llvmVersion = lib.mkOption {
       type = lib.types.str;
-      default = "14";
+      default = "16";
```
ss2165, line 15: i needed this for compilation now

ss2165: you're right, and it looks like I don't actually need it, confusion

PR #1978, opened 2025-03-14
feat(hugr-llvm): Emit iu_to_s and is_to_u

### LINKED ISSUE #1806: Lower `arithmetic.int_ops.iu_to_s` to LLVM
In principle this should be a noop, but we could check for overflow and panic?

croyzor:
Resolves #1806

### DIFF
--- a/hugr-llvm/src/extension/int.rs
+++ b/hugr-llvm/src/extension/int.rs
@@ -358,7 +358,7 @@ fn emit_int_op<'c, H: HugrView<Node = Node>>(
             let [TypeArg::BoundedNat { n: log_width }] =
                 TryInto::<[TypeArg; 1]>::try_into(args.node.args().to_vec()).unwrap()
             else {
-                panic!("Type argument to iu_to_s wasn't a number");
+                bail!("Type argument to iu_to_s wasn't a number");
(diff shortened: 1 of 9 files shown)


hugr-llvm/src/extension/int.rs
```diff
@@ -353,6 +354,51 @@ fn emit_int_op<'c, H: HugrView<Node = Node>>(
                 Ok(vec![result])
             })
         }
+        IntOpDef::iu_to_s => {
+            let [TypeArg::BoundedNat { n: log_width }] =
+                TryInto::<[TypeArg; 1]>::try_into(args.node.args().to_vec()).unwrap()
+            else {
+                panic!("Type argument to iu_to_s wasn't a number");
```
aborgna-q, line 361: The other branches use `bail!` instead (which returns an error instead of stopping execution)

aborgna-q (commented):
LGTM 👍

hugr-llvm/src/extension/int.rs
croyzor: Good shout!

aborgna-q (approved):

PR #1980, opened 2025-03-14
feat: Add default envelope config to `Package.to_str`/`to_bytes`

aborgna-q:


qartik (approved):

qartik:
This is required to or guppy files that use `hugr.to_str(EnvelopeConfig.TEXT)` require `from hugr.envelope import EnvelopeConfig` adding an unnecessary dependency on hugr-py.

aborgna-q:
This PR adds default params so it's not the case. But yes, we'd need to either require importing hugr or re-export the config class

PR #1989, opened 2025-03-17
feat: LowerTypes pass allows replacing extension types and ops

acl-cqc:
DRAFT because no tests yet!

If you replace a Copyable type with a Linear one, you'll either get a SignatureError (if this results in a type like `ArgMustBeCopyable[Qubit]`) or an invalid Hugr. (Or you might get lucky, if every outgoing port of the type happens to have one successor!)

Added HugrMut::optype_mut, and Type::transform via a new trait TypeTransformer


hugr-core/src/types.rs
```diff
@@ -666,6 +699,53 @@ impl<'a> Substitution<'a> {
...
+    /// Applies a [TypeTransformer] to this instance.
+    ///
+    /// Returns true if any part may have changed, or false for definitely no change.
+    ///
+    /// If an Err occurs, `self` may be left in an inconsistent state (e.g. partially
+    /// transformed).
+    fn transform<T: TypeTransformer>(&mut self, t: &T) -> Result<bool, T::Err>;
+}
+
+impl<E: Transformable> Transformable for [E] {
```
acl-cqc, line 739: This is the *raison d'être* for `Transformable` as it allows very easy iteration over arrays of Type, TypeRV, TypeArg, TypeRow(RV)

hugr-passes/src/lower_types.rs
```diff
@@ -0,0 +1,686 @@
...
+    },
+    #[error(transparent)]
+    #[allow(missing_docs)]
+    ValidationError(#[from] ValidatePassError),
+}
+
+impl LowerTypes {
+    /// Sets the validation level used before and after the pass is run.
+    // Note the self -> Self style is consistent with other passes, but not the other methods here.
+    // TODO change the others? But we are planning to drop validation_level in https://github.com/CQCL/hugr/pull/1895
```
acl-cqc, line 154: Unfortunately this is likely to go in before #1895 as this is needed now but #1895 is a breaking change that will likely come in April

hugr-passes/src/lower_types.rs
```diff
@@ -0,0 +1,686 @@
...
+        self.consts
+            .insert(Either::Right(src_ty.into()), Arc::from(const_fn));
+    }
+
+    /// Configures this instance to check signatures of ops lowered following [Self::lower_op]
+    /// and [Self::lower_parametric_op] are as expected, i.e. match the signatures of the
+    /// original op modulo the required type substitutions. (If signatures are incorrect,
+    /// it is likely that the wires in the Hugr will be invalid, so this gives an early warning
+    /// by instead raising [ChangeTypeError::SignatureMismatch].)
+    pub fn check_signatures(&mut self, check_sig: bool) {
```
acl-cqc, line 243: I'm not really sure about this checking stuff, hence no tests. I could drop it altogether, or convert to a panic (as it was before a0ac6d6f7c03f79ac1c44c6346fb8d3117053de7). Thoughts?

hugr-passes/src/lower_types.rs
```diff
@@ -0,0 +1,686 @@
...
+
+    fn run_no_validate(&self, hugr: &mut impl HugrMut) -> Result<bool, ChangeTypeError> {
+        let mut changed = false;
+        for n in hugr.nodes().collect::<Vec<_>>() {
+            let maybe_check_sig = if self.check_sig {
+                Some(
+                    if let Some(old_sig) = hugr.get_optype(n).dataflow_signature() {
+                        let old_sig = old_sig.into_owned();
+                        let mut expected_sig = old_sig.clone();
+                        expected_sig.transform(self)?;
```
acl-cqc, line 261: Hence can't use `self.check_sig.then_some()`

hugr-passes/src/lower_types.rs
```diff
@@ -0,0 +1,686 @@
...
+            .unwrap()
+            .outputs_arr();
+        let entry = entry.finish_with_outputs(tagged, []).unwrap();
+        cfg.branch(&entry, 0, &cfg.exit_block()).unwrap();
+        let cfg = cfg.finish_sub_container().unwrap();
+        fb.finish_with_outputs(cfg.outputs()).unwrap();
+        let mut h = mb.finish_hugr().unwrap();
+
+        assert!(lower_types(&ext).run(&mut h).unwrap());
+        // We do not update the extension delta
```
acl-cqc, line 602: If we need to update the delta, probably by rerunning extension "inference" to recompute union over children, we can do this, but would definitely prefer to leave for another PR unless it's urgent

hugr-passes/src/lower_types.rs
acl-cqc: Opinions welcome, if I'm gonna change the others I'll do it in this PR

hugr-passes/src/lower_types.rs
```diff
@@ -0,0 +1,686 @@
...
+    }
+
+    /// Configures this instance to change occurrences of a parametrized type `src`
+    /// via a callback that builds the replacement type given the [TypeArg]s.
+    /// Note that the TypeArgs will already have been lowered (e.g. they may not
+    /// fit the bounds of the original type).
+    pub fn lower_parametric_type(
+        &mut self,
+        src: &TypeDef,
+        dest_fn: Box<dyn Fn(&[TypeArg]) -> Type>, // TODO should we return Option<Type> ?
```
acl-cqc, line 179: Thoughts? Even returning `Type` doesn't actually force replacing all instances of the parametric type as the Fn could return the same type again (with the same or different typeargs)

hugr-passes/src/lower_types.rs
```diff
@@ -0,0 +1,686 @@
...
+    /// Configures this instance to change occurrences of a parametrized op `src`
+    /// via a callback that builds the replacement type given the [TypeArg]s.
+    /// Note that the TypeArgs will already have been lowered (e.g. they may not
+    /// fit the bounds of the original op).
+    ///
+    /// If the Callback returns None, the new typeargs will be applied to the original op.
+    pub fn lower_parametric_op(
+        &mut self,
+        src: &OpDef,
+        dest_fn: Box<dyn Fn(&[TypeArg]) -> Option<OpReplacement>>, // TODO or just OpReplacement?
```
acl-cqc, line 208: I'm leaning towards keeping the Option here

hugr-passes/src/lower_types.rs
```diff
@@ -0,0 +1,686 @@
...
+    }
+
+    /// Configures this instance to change occurrences consts of type `src_ty`, using
+    /// a callback given the value of the constant (of that type).
+    /// Note that if `src_ty` is an instance of a *parametrized* [TypeDef], this
+    /// takes precedence over [Self::lower_consts_parametric] where the `src_ty`s overlap.
+    pub fn lower_consts(
+        &mut self,
+        src_ty: &CustomType,
+        const_fn: Box<dyn Fn(&OpaqueValue) -> Option<Value>>, // TODO should we return Value??
```
acl-cqc, line 220: Leaning towards keeping the Option. Oh, I should doc that None means "no change"

hugr-passes/src/lower_types.rs
```diff
@@ -0,0 +1,686 @@
...
+                        }
+                        ch
+                    }
+                },
+            ),
+
+            OpType::OpaqueOp(_) => panic!("OpaqueOp should not be in a Hugr"),
+
+            OpType::AliasDecl(_) | OpType::Module(_) => Ok(false),
+            _ => todo!(),
```
acl-cqc, line 376: This is the catch-all because OpType is `#[non_exhaustive]`, I believe I've handled all current OpTypes and there is no more to do right now

acl-cqc:
Converting to draft, rerolling the die...

acl-cqc:
I'm wondering about combining `lower_type` with `lower_consts` (and `lower_parametric_type` with `lower_consts_parametric`) - i.e. *every* call to `lower_type` or `lower_parametric_type` must *also* pass a function for dealing with constants. A const producing a value of the type is the only non-ExtensionOp that needs to be handled, so it's good to remind the user to provide it ;-), you can always pass a callback that panics "there are no constants of this type" (e.g. if you're lowering QB_T !)...

hugr-passes/src/lower_types.rs
```diff
@@ -0,0 +1,756 @@
...
+    /// If the Callback returns None, the new typeargs will be applied to the original op.
+    pub fn lower_parametric_op(
+        &mut self,
+        src: &OpDef,
+        dest_fn: Box<dyn Fn(&[TypeArg]) -> Option<OpReplacement>>,
+    ) {
+        self.param_ops.insert(src.into(), Arc::from(dest_fn));
+    }
+
+    /// Configures this instance to change occurrences consts of type `src_ty`, using
```
cqc-alec, line 178:
```suggestion
    /// Configures this instance to change occurrences of consts of type `src_ty`, using
```

hugr-passes/src/lower_types.rs
cqc-alec: No particularly strong opinion. I guess this method is different from the others in that it's setting a configuration rather than doing something, so I don't mind the inconsistency too much. (Exception: `check_signatures`. Maybe that should use the same pattern.)

hugr-passes/src/lower_types.rs
cqc-alec: Either drop or add tests -- don't mind which!

cqc-alec (commented):
Would `change_types` work as a name?

I think I approve of combining with `lower_consts` as you suggest.

acl-cqc:
I wrote a test for `check_signatures` but eventually realized that it didn't offer much more help debugging than validation does - the error from validation does, at least, identify the source and target of the offending edge, so there are only two nodes to look at.

I also combined lower_consts and lower_type....and then eventually separated them again: I had quite a few `| _ | panic!("No consts")` in tests, and this would be required in production code if you were lowering, say, Qubit. However, I did separate the hashmap and require the `lower_consts(Type,...)` callback to return `Value` rather than Option<Value> as you will have to replace them all. Oh, and removed the Boxes around all callbacks. (Oh, I tried a passing an Option<callback> which might have been the best of both worlds, but had insurmountable problems with type inference if passing `None`)

Re. naming, I am preparing to go with `ReplaceTypes` (I think "Change" is too ambiguous), but the pass does also replaces ops and consts! If we're ok with that then it'll do. `ReplaceTypesConsts` seems a bit odd (or ??)....and whilst `Lower` is nice in some ways it doesn't really tell you what the thing does (it'd only be an aid to *recall*).

cqc-alec:
> Re. naming, I am preparing to go with `ReplaceTypes` (I think "Change" is too ambiguous) [...]

Sounds good to me.

acl-cqc:
I've tried to improve docs a bit too, please have a read ;)

acl-cqc:
Yikes. Got there, adding `pub mod handlers` inside it, usage demonstrated in a test

acl-cqc:
Commit ec2d0ba739c96092d61fa1551f98152c03238a7a is rather horrendous but also seems....necessary? Silently dropping errors doesn't seem good.

cqc-alec (approved):
Looks good, thanks!

hugr-passes/src/replace_types.rs
```diff
@@ -60,24 +61,18 @@ impl OpReplacement {
 }
 
 #[derive(Clone, Default)]
-pub struct LowerTypes {
-    /// Handles simple cases like T1 -> T2.
-    /// If T1 is Copyable and T2 Linear, then error will be raised if we find e.g.
-    /// ArrayOfCopyables(T1). This would require an additional entry for that.
+pub struct ReplaceTypes {
```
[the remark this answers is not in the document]
acl-cqc: Yeah, I was thinking of that one too....I've enabled `warn(missing_docs)` across the module, except for the error type.

PR #1994, opened 2025-03-18
feat(hugr-py): Support envelope compression

### LINKED ISSUE #1979: hugr-py: Support `zstd` compression on hugr envelopes
We could also make `zstd` a default feature on the rust side once this is implemented.

aborgna-q:
Closes #1979 

- Adds a default `zstd` feature to `hugr-rs`, so `hugr-cli` supports it.
- Updates the `validate` conftest util to test both binary and textual encodings.

---

There's multiple python packages which provide bindings for zstd:

- [`zstd`](https://pypi.org/project/zstd), which provides a quite simple encode/decode API. It seems to follow `zstd` releases quite closely.
- [`zstandard`](https://pypi.org/project/zstandard/), a "rich Pythonic interface" to the zstd bindings, listed as an alternative by `zstd`.
- [`pyzstd`](https://pypi.org/project/pyzstd/#files), similar to `zstd`.

I went with the former since it's the most popular one.

There's an open vuln report for this: https://osv.dev/vulnerability/PYSEC-2023-121, but it is outdated and the current version already patches it (since the underlying `zstd` c++ lib already fixed it).

### DIFF
--- a/hugr-py/pyproject.toml
+++ b/hugr-py/pyproject.toml
@@ -31,6 +31,7 @@ dependencies = [
     "pydantic-extra-types>=2.9.0",
     "semver>=3.0.2",
     "graphviz>=0.20.3",
+    "zstd>=1.5.6.6",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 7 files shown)


tatiana-s (approved):
👍

PR #1999, opened 2025-03-19
fix: Remove return from val_or_panic

croyzor:
We shouldn't have been emitting a `ret` in `val_or_panic` - this leads to behaviour where code that comes after the `val_or_panic` section isn't emitted.

For example, the `test_s_to_u` test has been altered to add a constant to the number after its converted. Before this fix it would always fail because the addition was skipped.


croyzor:
> Looks good. Can you remove the libc emission as well? You'll need to pass a `PreludeCodegen` into the extension struct. e.g. https://github.com/CQCL/hugr-qir/blob/4fa0ed8565debf5c540fa2392e0eadb3e59de75d/src/rotation.rs#L25?
> 
> Unfortunately this is a breaking change, but I think we need to do it.

Yes! Sorry, forgot about that one

PR #2003, opened 2025-03-19
feat: Add llvm codegen for collections.static_array

doug-q:


hugr-llvm/src/extension/collections/static_array.rs
```diff
@@ -0,0 +1,517 @@
...
+        let hugr = single_op_hugr(op);
+        check_emission!(hugr, llvm_ctx);
+    }
+
+    #[rstest]
+    #[case(0, StaticArrayValue::try_new("a", usize_t(), (0..10).map(|x| ConstUsize::new(x).into())).unwrap())]
+    #[case(1, StaticArrayValue::try_new("b", float64_type(), (0..10).map(|x| ConstF64::new(x as f64).into())).unwrap())]
+    #[case(2, StaticArrayValue::try_new("c", bool_t(), (0..10).map(|x| Value::from_bool(x % 2 == 0))).unwrap())]
+    #[case(3, StaticArrayValue::try_new("d", option_type(usize_t()).into(), (0..10).map(|x| Value::some([ConstUsize::new(x)]))).unwrap())]
+    fn static_array_const_codegen(
```
cqc-alec, line 432: This and the following test fail when I run
```
cargo test --all-features --workspace
```
with the error
```
thread 'extension::collections::static_array::test::static_array_const_codegen::case_1' panicked at hugr-llvm/src/emit/test.rs:177:48:
Parent node Node(1) has extensions [arithmetic.conversions, arithmetic.float, arithmetic.float.types, arithmetic.int, arithmetic.int.types, collections.array, collections.list, logic, prelude] that are too restrictive for child node Node(4), they must include child extensions [collections.static_array, prelude]
```

I guess we need to annotate them so that they don't run with the `extension_inference` feature?

hugr-llvm/src/extension/collections/static_array.rs
```diff
@@ -0,0 +1,517 @@
...
+            });
+
+        exec_ctx.add_extensions(|ceb| {
+            ceb.add_default_static_array_extensions()
+                .add_default_prelude_extensions()
+                .add_float_extensions()
+        });
+        assert_eq!(expected, exec_ctx.exec_hugr_u64(hugr, "main"));
+    }
+}
```
cqc-alec, line 517: Maybe a test with arrays of arrays would also be good.

cqc-alec (commented):
LGTM, but this is now breaking -- are we OK with that?

cqc-alec (approved):

PR #2007, opened 2025-03-20
feat(hugr-cli): Nicer error when passing a non-envelope file

aborgna-q:
Minor special-casing for CLI message error.

Before:
```
$ hugr validate invalid.hugr
Error decoding HUGR envelope: Bad magic number. expected 0x4855475269484A76 found 0x7B0A20202020226D
```
After:
```
$ hugr validate invalid.hugr
Input file is not a HUGR envelope. Invalid magic number.

Use `--hugr-json` to read a raw HUGR JSON file instead.
```

### DIFF
--- a/hugr-cli/src/hugr_io.rs
+++ b/hugr-cli/src/hugr_io.rs
@@ -1,7 +1,7 @@
 //! Input/output arguments for the HUGR CLI.
 
 use clio::Input;
-use hugr::envelope::read_envelope;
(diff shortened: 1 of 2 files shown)


aborgna-q:
There's no logger currently setup. We just `eprintln` errors after execution depending on the verbosity level.

We should add `env_logger` or something similar, but that's a different PR.

PR #2012, opened 2025-03-20
feat: MakeTuple->UnpackTuple elision pass

### LINKED ISSUE #1810: Move the tuple unpacking rewrite from tket2 to `hugr-passes`
https://github.com/CQCL/tket2/blob/95090a2c9a7b30f89831e16f724c9db815bf6714/tket2/src/passes/tuple_unpack.rs#L16-L20

aborgna-q:
Imports the `MakeTuple -> UnpackTuple` elimination pass from tket2.
This removes tuple unpacks consuming the output of a `MakeTuple`, and deletes the packing operation when possible.

This pass *does not* remove the reverse case, `Unpack->MakeTuple`. Should it be added here? We should avoid producing overlapping rewrites if doing that (the simple `Pack->Unpack->Pack` case is easy, but multiports can make things interesting). 

Requires #2011
Closes #1810

### DIFF
--- a/hugr-passes/src/untuple.rs
+++ b/hugr-passes/src/untuple.rs
@@ -32,9 +32,14 @@ pub enum UntupleRecursive {
 /// by `UnpackTuple`s.
 ///
 /// If the tuple output is consumed by other operations, only the `UnpackTuple`s
-/// are removed and their outputs are connected to the original values accordingly.
+/// are removed and their outputs are connected to the original values
+/// accordingly.
 ///
-/// Removes `MakeTuple` operations that are not consumed by any other operations.
+/// Currently only unpack operations in the same region as the `MakeTuple` are
+/// removed. This may be extended in the future.
+///
+/// Removes `MakeTuple` operations that are not consumed by any other
+/// operations.
 #[derive(Debug, Clone, Default)]
 pub struct UntuplePass {
     /// Whether to traverse the HUGR recursively.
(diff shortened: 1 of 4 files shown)


hugr-passes/src/untuple.rs
```diff
@@ -0,0 +1,394 @@
...
+    let num_unpack_outputs = tuple_types.len() * unpack_nodes.len();
+
+    dbg!(num_unpack_outputs, num_other_outputs);
+
+    let checker = convex_checker.get_or_insert_with(|| TopoConvexChecker::new(hugr));
+
+    let mut nodes = unpack_nodes;
+    nodes.push(pack_node);
+    let subcirc = SiblingSubgraph::try_from_nodes_with_checker(nodes, hugr, checker).unwrap();
+    let subcirc_signature = subcirc.signature(hugr);
```
mark-koch, line 220: afaik this will panic if there are order edges?

hugr-passes/src/untuple.rs
```diff
@@ -0,0 +1,394 @@
...
+) -> SimpleReplacement<T::Node> {
+    let num_unpack_outputs = tuple_types.len() * unpack_nodes.len();
+
+    dbg!(num_unpack_outputs, num_other_outputs);
+
+    let checker = convex_checker.get_or_insert_with(|| TopoConvexChecker::new(hugr));
+
+    let mut nodes = unpack_nodes;
+    nodes.push(pack_node);
+    let subcirc = SiblingSubgraph::try_from_nodes_with_checker(nodes, hugr, checker).unwrap();
```
mark-koch, line 219: Is this `unwrap` safe?

mark-koch: What happens if the make-unpack is an inter-graph edge?

mark-koch (commented):
Thanks! Mostly wondering about edge cases with order and inter-graph edges?

Btw, it would be nice to have to opposite `UnpackTuple -> MakeTuple` elison as well since Guppy sometimes emits those when dealing with structs. Happy for this to be a follow-up though

hugr-passes/src/untuple.rs
aborgna-q: - See other comment about order edges. Unwrap is safe otherwise, as we are just constructing a two-node graph.

- I disabled matching on inter-graph edges to be on the safe side.

mark-koch (approved):

Issue #1521, opened 2024-09-04
Array ops `pop_left` and `pop_right` should validate that the array type size is not 0


doug-q:
I expect that currently we will get panics on https://github.com/CQCL/hugr/blob/f3dafd74f65c1542c83b39f7f12a8170f60ed14d/hugr-core/src/extension/prelude/array.rs#L69

These ops return an `option` type, one assumes that failure here means the array had zero size. As is this is unnecessary: 
* if they array's size is `TypeArg::BoundedNat{n: 0}`, `compute_signature` should return an error (instead of panicking on the line above).
* if the array's size is a `TypeArg::Variable` we fail. We can't do better without arithmetic-in-the-type-system. Note this means that these `pop` functions are not useful inside functions polymorphic on the size of the array.

acl-cqc:
I suspect it won't panic but n-1 will underflow back round to u64_max....

We know size can't be a TypeArg::Variable because the size is a *static* parameter, it has to be in order to be processed by the binary `compute_signature`.

So, we can drop the option and error on the n==0 case, indeed

PR #2021, opened 2025-03-24
docs: Provide docs for array ops, fix bad doc for HugrView::poly_func_type

acl-cqc:


hugr-core/src/std_extensions/collections/array/array_op.rs
```diff
@@ -22,15 +22,32 @@ use super::{array_type, array_type_def, ARRAY_TYPENAME};
...
     get,
+    /// exchanges an element of the array with an external value:
+    /// `set<size,elemty>: array<size,elemty>, index, elemty -> either(elemty, array | elemty, array)`
+    /// tagged for failure/success respectively
     set,
+    /// exchanges two indices within the array: `swap<size,elemty>: array<size, elemty>, index index -> either(array, array)`
+    /// tagged for failure/success respectively
     swap,
+    /// separates leftmost element from the rest of the array.
+    /// `pop_left<SIZE><elemty>: array<SIZE, elemty> -> Option<elemty, array<SIZE-1,elemty>>`
```
acl-cqc, line 42: Note that if SIZE is 0, the u64 will (presumably) wrap, so the result type will be an `Option<array<U64_MAX, elemty>>` which will be None.

Given `SIZE` must be statically known, we could remove the Option:
* `pop_left<0>` could return Option of anything we want, or raise error in `compute_signature`
* `pop_left<1>` could return just the element (no Option nor array)
* `pop_left<x>` when `x>=2` could return the element and rest of array (no Option)

acl-cqc: Hah, yes, OK maybe I've gone too far. See also #1521 though: even if we returned the element and a size-0 array in all cases N>=1, we could still avoid the option by erroring in `compute_signature` for N==0. A lesser step but a good one?

Which is to say, I'll update/correct #1521 a bit and let's go from there

hugr-core/src/std_extensions/collections/array/array_op.rs
```diff
@@ -22,15 +22,36 @@ use super::{array_type, array_type_def, ARRAY_TYPENAME};
...
     swap,
+    /// separates leftmost element from the rest of the array:
+    /// `pop_left<SIZE><elemty>: array<SIZE, elemty> -> Option<elemty, array<SIZE-1, elemty>>`
+    /// where `SIZE` must be known statically (not a variable).
+    /// `None` is returned if the input array was size 0.
     pop_left,
+    /// separates rightmost element from the rest of the array.
+    /// `pop_right<SIZE><elemty>: array<SIZE, elemty> -> Option<elemty, array<SIZE-1, elemty>>`
+    /// where `SIZE` must be known statically (not a variable).
+    /// `None` is returned in the input array was size 0.
```
hugrbot, line 51:
```suggestion
    /// `None` is returned if the input array was size 0.
```

hugrbot (changes requested):

hugr-core/src/std_extensions/collections/array/array_op.rs
```diff
@@ -22,15 +22,36 @@ use super::{array_type, array_type_def, ARRAY_TYPENAME};
...
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, EnumIter, IntoStaticStr, EnumString)]
-#[allow(non_camel_case_types, missing_docs)]
+#[allow(non_camel_case_types)]
 #[non_exhaustive]
 pub enum ArrayOpDef {
+    /// Makes a new array, given distinct inputs equal to its length:
+    /// `new_array<SIZE><elemty>: (elemty)^SIZE -> array<SIZE, elemty>`
+    /// where `SIZE` must be statically known (not a variable)
     new_array,
+    /// copies an element out of the array ([TypeBound::Copyable] elements only):
```
cqc-alec, line 32:
```suggestion
    /// Copies an element out of the array ([TypeBound::Copyable] elements only):
```

hugr-core/src/std_extensions/collections/array/array_op.rs
```diff
@@ -22,15 +22,36 @@ use super::{array_type, array_type_def, ARRAY_TYPENAME};
...
+    /// `get<size,elemty>: array<size, elemty>, index -> option<elemty>`
     get,
+    /// exchanges an element of the array with an external value:
```
cqc-alec, line 35:
```suggestion
    /// Exchanges an element of the array with an external value:
```

hugr-core/src/std_extensions/collections/array/array_op.rs
```diff
@@ -22,15 +22,36 @@ use super::{array_type, array_type_def, ARRAY_TYPENAME};
...
+    /// `set<size, elemty>: array<size, elemty>, index, elemty -> either(elemty, array | elemty, array)`
+    /// tagged for failure/success respectively
     set,
+    /// exchanges two indices within the array:
```
cqc-alec, line 39:
```suggestion
    /// Exchanges two indices within the array:
```

hugr-core/src/std_extensions/collections/array/array_op.rs
```diff
@@ -22,15 +22,36 @@ use super::{array_type, array_type_def, ARRAY_TYPENAME};
...
+    /// `swap<size, elemty>: array<size, elemty>, index, index -> either(array, array)`
+    /// tagged for failure/success respectively
     swap,
+    /// separates leftmost element from the rest of the array:
```
cqc-alec, line 43:
```suggestion
    /// Separates leftmost element from the rest of the array:
```

hugr-core/src/std_extensions/collections/array/array_op.rs
```diff
@@ -22,15 +22,36 @@ use super::{array_type, array_type_def, ARRAY_TYPENAME};
...
+    /// `pop_left<SIZE><elemty>: array<SIZE, elemty> -> Option<elemty, array<SIZE-1, elemty>>`
+    /// where `SIZE` must be known statically (not a variable).
+    /// `None` is returned if the input array was size 0.
     pop_left,
+    /// separates rightmost element from the rest of the array.
```
cqc-alec, line 48:
```suggestion
    /// Separates rightmost element from the rest of the array.
```

hugr-core/src/std_extensions/collections/array/array_op.rs
acl-cqc: Yes, and I pedantically added "the elements at"

hugr-core/src/std_extensions/collections/array/array_op.rs
acl-cqc: Yes, and added "the" (leftmost element)

hugr-core/src/std_extensions/collections/array/array_op.rs
acl-cqc: Yes, and added missing "the"

cqc-alec (approved):

PR #2023, opened 2025-03-25
feat: ReplaceTypes: handlers for array constants + linearization

### LINKED ISSUE #1948: Hugr Type Lowering
* Replace one type with another - a bit like a Substitution but for fixed (maybe only Extension) types not type variables
* If the source type is copyable but the destination type is linear, insert some (specified) `dup` and `free` ops
* For parameterized ops (Extension Ops or those in hugr-core), just replacing in the "type args" should do it
* But, for extension ops that input or output values of the replaced type (not via type parametrization), also need to replace ops

acl-cqc:
Final chapter closes #1948.

`default()` includes all the handlers, `new_empty()` does not.


hugr-passes/src/lower_types/linearize.rs
```diff
@@ -206,19 +219,172 @@ impl Linearizer {
...
+        panic!("Illegal TypeArgs to array: {:?}", args)
+    };
+    // Make a function that maps the linear element to unit
+    let map_fn = {
+        let mut dfb = DFGBuilder::new(inout_sig(ty.clone(), Type::UNIT)).unwrap();
+        let [to_discard] = dfb.input_wires_arr();
+        lin.discard_op(ty)?.add(&mut dfb, [to_discard]).unwrap();
+        let ret = dfb.add_load_value(Value::unary_unit_sum());
+        dfb.finish_hugr_with_outputs([ret]).unwrap()
+    };
```
mark-koch, line 239: This function will be freshly created each time we discard an array, which I expect will lead to code-size problems?

Would it be feasible to store a map of these functions for each elelemt type inside the `Linearizer`?

hugr-passes/src/lower_types/linearize.rs
```diff
@@ -206,19 +219,172 @@ impl Linearizer {
...
+    // Require known length i.e. usable only after monomorphization, due to no-variables limitation
+    // restriction on OpReplacement::CompoundOp
+    let [TypeArg::BoundedNat { n }, TypeArg::Type { ty }] = args else {
+        panic!("Illegal TypeArgs to array: {:?}", args)
+    };
+    let i64_t = INT_TYPES[6].to_owned();
+    let option_sty = option_type(ty.clone());
+    let option_ty = Type::from(option_sty.clone());
+    let option_array = array_type(*n, option_ty.clone());
+    let copy_elem = {
```
mark-koch, line 265: Maybe add a comment on top explaining what this function does, or how we go about copying in general?

hugr-passes/src/lower_types/linearize.rs
```diff
@@ -206,19 +219,172 @@ impl Linearizer {
...
+            .add_dataflow_op(ConvertOpDef::itousize.without_log_width(), [idx])
+            .unwrap()
+            .outputs_arr();
+        let [copy0, copy1] = lin
+            .copy_op(ty)?
+            .add(&mut dfb, [elem])
+            .unwrap()
+            .outputs_arr();
+
+        //let variants = option_ty.variants().cloned().map(|row| row.try_into().unwrap()).collect();
```
mark-koch, line 283:
```suggestion
```

hugr-passes/src/lower_types/linearize.rs
mark-koch: Same for `copy_elem`, `unwrap_elem`, and `fn_none` below

hugr-passes/src/lower_types/linearize.rs
```diff
@@ -360,4 +526,85 @@ mod test {
...
+                Signature::new(array_type(5, usize_t()), type_row![]),
+                [array_in],
+            )
+            .unwrap()
+            .finish_with_outputs([])
+            .unwrap();
+        // and (2) its own output
+        let mut h = dfb.finish_hugr_with_outputs([array_in]).unwrap();
+
+        assert!(lowerer.run(&mut h).unwrap());
```
mark-koch, line 555: Does this also run validation?

acl-cqc: Yes....this is both a feature and also an annoyance, in that without an *explicit* `.validate()` it's easy to miss....sometimes I have written `.validate()` separately and then removed it as redundant :(. #1895 makes this much more obvious via a `validate_if_test` method...

hugr-passes/src/lower_types/linearize.rs
acl-cqc: Ok, we can avoid a load of extra `&mut` parameters by having handlers like that for array generate *ops* for discarding the elements. When the `Linearizer` inserts a copy/discard `NodeTemplate`, it'll scan through the inserted code looking for such ops, and then replace those as well ("recursively" perhaps in that it recurses on itself, but not through the handler). In fact we will probably arrange that *all* linearization is done by first inserting a copy/discard *op* (from ReplaceTypes), and then the Linearizer is a separate pass that just replaces those (probably using a worklist, maintaining an internal cache).

That's all to come, though. The main change will be in #2019 we'll make these callbacks take not a `&Linearizer` but some other structure that provides the same `insert_copy_discard`, and `copy_discard_fn` functions - which for now will delegate to Linearizer. That'll let me merge #2019. Then I'll have another PR (still before this) that changes that structure just to emit single copy/discard ops, and refactor Linearizer to replace those ops (with a mutable cache). Then finally we can get onto this PR. Does that make sense?

mark-koch (approved):

hugr-passes/src/replace_types/handlers.rs
```diff
@@ -0,0 +1,240 @@
...
+use hugr_core::{type_row, Hugr, HugrView};
+use itertools::Itertools;
+
+use super::{
+    CallbackHandler, LinearizeError, Linearizer, NodeTemplate, ReplaceTypes, ReplaceTypesError,
+};
+
+/// Handler for [ListValue] constants that recursively [ReplaceTypes::change_value]s
+/// the elements of the list
+pub fn list_const(
```
mark-koch, line 26: We probably need the same thing for array constants?

acl-cqc: Yeah, done

hugr-passes/src/replace_types/handlers.rs
```diff
@@ -0,0 +1,240 @@
...
+    };
+    let mut vals: Vec<Value> = lv.get_contents().to_vec();
+    let mut ch = false;
+    for v in vals.iter_mut() {
+        ch |= repl.change_value(v)?;
+    }
+    // If none of the values has changed, assume the Type hasn't (Values have a single known type)
+    if !ch {
+        return Ok(None);
+    };
```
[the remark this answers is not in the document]
acl-cqc: Yes, good spot.
`list_const` was originally more for checking the framework did the right thing(s) for constants, rather than being a definitive handler for lists, but given it was `pub` it should have been better ;-). I've fixed this, although I haven't added a test with an empty list - the array_const test does include such a case and the SUT is identical.... (I did try commoning `list_const` and `array_const` up but it needed 3 different `impl Fn`s so was not worth it!)

hugr-passes/src/replace_types/linearize.rs
```diff
@@ -645,4 +651,102 @@ mod test {
             ))
         );
     }
+
+    #[rstest]
+    fn array(#[values(2, 3, 4)] num_outports: usize) {
```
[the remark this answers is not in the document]
acl-cqc: The `loop_const` does some fairly funky things with them, although is not exhaustive

hugr-passes/src/replace_types/handlers.rs
acl-cqc: Yeah, done, with a test.

mark-koch (approved):
Thanks, looks good 👍

> Any thoughts on what to do with the handlers?

I think handlers for std extensions should be included in `default()`. Maybe `ReplaceTypes::empty()` to get one without any handlers?

acl-cqc:
> Thanks, looks good 👍
> 
> > Any thoughts on what to do with the handlers?
> 
> I think handlers for std extensions should be included in `default()`. Maybe `ReplaceTypes::empty()` to get one without any handlers?

That was not quite as neat as I'd have liked, 'coz we can't get the `Default`-derive macro to do the work, but done.

PR #2024, opened 2025-03-25
chore: Enable deprecation warnings on pre-313 python

aborgna-q:
`typing_extensions` defines a `@deprecated` decorator compatible with `python<3.13`

drive-by: Added decorator to `Else.finish`, which said it was deprecated in its docstring but didn't raise any warnings.


ss2165 (approved):
thanks for the drive by

PR #2025, opened 2025-03-25
feat(hugr-llvm): Emit divmod and mod operations

croyzor:
Relates to #1702 .


hugr-llvm/src/extension/int.rs
```diff
@@ -442,6 +738,94 @@ fn emit_int_op<'c, H: HugrView<Node = Node>>(
...
+    numerator: IntValue<'c>,
+    denominator: IntValue<'c>,
+    panic: bool,
+    signed: bool,
+) -> Result<LLVMSumValue<'c>> {
+    let int_ty = numerator.get_type();
+    let zero = int_ty.const_zero();
+    let lower_bounds_check = if signed {
+        ctx.builder()
+            .build_int_compare(IntPredicate::SGT, denominator, zero, "valid_div")
```
aborgna-q, line 769: This caught me by surprise. Why are we doing "signed dividend, unsigned divisor" in div/mod ???

The docs say that the operand may be any non-zero unsigned value,
https://github.com/CQCL/hugr/blob/41905f9a15d9f928492bdbc2ca1edeec3e244e0f/hugr-core/src/std_extensions/arithmetic/int_ops.rs#L194-L195
so to be correct in the lowering we need to allow any non-zero value here and return.

Even more, LLVM's signed remainder is [**not modulo**](https://llvm.org/docs/LangRef.html#srem-instruction); its result has the same sign as the dividend.
To lower things correctly here we need to do a lot of extra logic

$$
\texttt{hugr-divmod-s}(n: signed, m: unsigned) =
$$
$$
\begin{cases}
(0, n) \qquad\qquad\qquad\qquad\qquad\qquad & \text{if } m \geq 2^{N-1} \text{ and } n>0\\
(-1, r-n) & \text{if } m \geq 2^{N-1} \text{ and } n<0\\
(\texttt{llvm-signed-div}(n,m), &\\
\quad \texttt{llvm-signed-rem}(n,m)) & \text{if } m < 2^{N-1} \text{ and } n>0\\
(\texttt{llvm-signed-div}(n,m), 0) & \text{if } m < 2^{N-1} \text{ and } n<0 \text{ and } \texttt{llvm-signed-rem}(n,m)= 0 \\
(\texttt{llvm-signed-div}(n,m)-1, & \\
\quad n-\texttt{llvm-signed-rem}(n,m)) &  \text{if } m < 2^{N-1} \text{ and } n<0 \text{ and } \texttt{llvm-signed-rem}(n,m) \neq 0
\end{cases}
$$

💀

aborgna-q (changes requested):

hugr-llvm/src/extension/int.rs
croyzor: Plus for the 4th case ($$m < 2^{N-1} \text{ and } n < 0 \text{ and } \texttt{llvm-signed-rem(n, m)} $$), we need to negate the result of `llvm-signed-div(n, m)` for the quotient because llvm always returns a positive number

croyzor: Plus the last case should be `(llvm-signed-div(n, m) - 1, m - llvm-signed-rem(n,m))` (`m` instead of `n` in the snd element)

croyzor: Actually the last case should be `llvm-signed-div(n, m) - 1, m + llvm-signed-rem(n, m))` because the result of `llvm-signed-rem` will be negative to match the dividend

hugr-llvm/src/extension/int.rs
```diff
@@ -442,6 +738,94 @@ fn emit_int_op<'c, H: HugrView<Node = Node>>(
...
+        .node
+        .args()
+        .to_owned()
+        .try_into()
+        .expect("Exactly one type argument to int op");
+    let TypeArg::BoundedNat { n: log_width } = arg else {
+        panic!("Unexpected non-nat arg to int op");
+    };
+    log_width
+}
```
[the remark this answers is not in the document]
croyzor: Done in a57618e3

hugr-llvm/src/extension/int.rs
```diff
@@ -435,13 +727,230 @@ fn emit_int_op<'c, H: HugrView<Node = Node>>(
...
+    args: &EmitOpArgs<'_, '_, ExtensionOp, H>,
+    op: &impl NamedOp,
+) -> Result<u64> {
+    let [TypeArg::BoundedNat { n: log_width }] = args.node.args() else {
+        bail!("Expected exactly one BoundedNat parameter to {}", op.name())
+    };
+    Ok(*log_width)
+}
+
+fn make_divmod<'c, H: HugrView<Node = Node>>(
```
aborgna-q, line 749: Add some docs here explaining why we need to make the emission so complicated.
Maybe link back to this PRs discussion.

aborgna-q (approved):
Awesome!

PR #2032, opened 2025-03-26
feat: add `build_expect_sum` to allow specific error messages

ss2165:


ss2165:
good idea on returning a whole error, I will add a From<String> for ConstError that uses the default signal

PR #2038, opened 2025-03-27
chore(hugr-py): Swap `zstd` dependency for `pyzstd`

aborgna-q:
Both libraries offer good support for encoding/decoding zstd, but `zstd` fails to build under PEP517.

Some poetry error hints:
```
Note: This error originates from the build backend, and is likely not a problem with poetry but with zstd (1.5.6.6) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "zstd (==1.5.6.6)"
```

As the libraries are almost interchangeable, let's try using the other one.


hugr-py/tests/test_envelope.py
aborgna-q, line 1: We weren't testing envelope compression, so this tries to do a rountrip under all possible envelope configs

cqc-alec (approved):

PR #2044, opened 2025-03-31
docs: Add usage info to hugr-cli's rustdocs

aborgna-q:
Expands the module docstring of `hugr-cli` so we can link people directly to https://docs.rs/hugr-cli/latest/hugr_cli/ instead of depending on github's README.


qartik (approved):

qartik:
I wonder if there's a way to extract the docstring from the `--help` directly to keep it in sync when changes occur.

aborgna-q:
It's not a straight dump, as there's some markdown formatting around it.
I don't think automating it would be worth it.

PR #2046, opened 2025-03-31
feat!: Handle CallIndirect in constant folding

acl-cqc:
* Extend ValueHandle with a node pointer into the containing Hugr - the previous approach of handling `LoadFunction` by stealing the body of the loaded function, will not work for functions with incoming static edges (i.e. from *other* functions/constants).
* Add new `interpret_call_indirect` method to `trait DFContext` (alongside `interpret_leaf_op`), again defaulting to { }
* Implement for ConstFoldContext by a complete recursion (i.e. when the called function is known).
- [ ] TODO: tests!
- [ ] This recursion may be expensive (and seems more than we do for calls - multiple CallIndirects to *the same function* will each get their own analysis, without aliasing, whereas multiple Calls to the same function get corresponding arguments/intermediates `join`ed together). Should be put it under a switch?

BREAKING CHANGE: mod value_handle (containing ValueHandle and HashedConst) is no longer `pub`.

### DIFF
--- a/hugr-passes/src/const_fold.rs
+++ b/hugr-passes/src/const_fold.rs
@@ -207,13 +207,6 @@ pub fn constant_fold_pass<H: HugrMut>(h: &mut H) {
 
 struct ConstFoldContext<'a, H>(&'a H);
 
-impl<H: HugrView> std::ops::Deref for ConstFoldContext<'_, H> {
-    type Target = H;
-    fn deref(&self) -> &H {
-        self.0
-    }
-}
-
 impl<H: HugrView<Node = Node>> ConstLoader<ValueHandle<H::Node>> for ConstFoldContext<'_, H> {
     type Node = H::Node;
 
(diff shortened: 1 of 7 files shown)


hugr-passes/src/const_fold.rs
```diff
@@ -2,18 +2,15 @@
 //! Constant-folding pass.
 //! An (example) use of the [dataflow analysis framework](super::dataflow).
 
-pub mod value_handle;
+mod value_handle;
```
aborgna-q, line 5: It's quite annoying that we have to make this a breaking change.
It looks like even leaving this as public wouldn't help, since `ValueHandle` is not `non_exhaustive` and added new fields -.-

Is it OK to wait in merging this PR until we accumulate other breaking changes?

aborgna-q (commented):
Looks great, just missing the tests.

> This recursion may be expensive (...). Should be put it under a switch?

You mean to enabling/disabling callindirect folding altogether?
I'd say we can wait and see if it's necessary, but I guess we could add a `ConstFoldConfig` struct here since we're breaking the API already.

hugr-passes/src/const_fold.rs
acl-cqc: Yeah, I was hiding that because I thought that's the way that it should have been, would have avoided this being a breaking change now...

How urgent....I was gonna say, the bit we need first is the `constant_fold_with_hugr` and that's non-breaking so I can break that out and leave the rest as a demo (just tests that the ValueHandle/CallIndirect works would confirm we can do the same approach in/for BRAT).

However....now I realize we need to get the ValueHandle::NodeRef into the BRAT-specific constant-folding code. At present extension-provided constant-folders only get Hugr `Value`s, so rather than adding `constant_fold_with_hugr` I think I need some kind of plugin interface to the constant-folding pass that lets us get at those ValueHandles after all. So I think I'll have to restore `pub`ness and figure out what the plugin interface is.

PR #2059, opened 2025-04-04
feat!: Handle CallIndirect in Dataflow Analysis

acl-cqc:
* PartialValue now has a LoadedFunction variant, created by LoadFunction nodes (only, although other analyses are able to create PartialValues if they want)
   * This requires adding a type parameter to PartialValue for the type of Node, which gets everywhere :-(. The alternative is to restrict all dataflow analysis to work only on HugrViews that use Node.
* Use this to handle CallIndirects *with known targets* (it'll be a single known target or none at all) just like other Calls to the same function
* deprecate (and ignore) `value_from_function`
* Add a new trait `AsConcrete` for the result type of `PartialValue::try_into_concrete` and `PartialSum::try_into_sum`

Note almost no change to constant folding (only to drop impl of `value_from_function`)

BREAKING CHANGE: in dataflow framework, PartialValue now has additional variant; `try_into_concrete` requires the target type to implement AsConcrete


hugr-passes/src/dataflow/results.rs
```diff
@@ -84,12 +87,15 @@ impl<V: AbstractValue, H: HugrView> AnalysisResults<V, H> {
     /// `None` if the analysis did not produce a result for that wire, or if
     ///    the Hugr did not have a [Type](hugr_core::types::Type) for the specified wire
     /// `Some(e)` if [conversion to a concrete value](PartialValue::try_into_concrete) failed with error `e`
-    pub fn try_read_wire_concrete<V2, VE, SE>(
+    #[allow(clippy::type_complexity)]
```
acl-cqc, line 90: Alternative is to define a new
```
trait IntoConcrete<V,N> {
  type ValError;
  type SumError;
  type LoadFuncError;
  fn from_value(V) -> Result<Self, Self::ValError>;
  fn from_sum(Sum<Self>) -> Result<Self, Self::SumError>;
  fn from_loaded_func(LoadedFunction<N>) -> Result<Self, LoadedFunction<N>>;
}
```
and then require `V2: IntoConcrete<V, N>`.

Possibly `ExtractValueError` could then be parametrized by `<V, N, V2: IntoConcrete<N>>` and contain `V2::SumError` and `V2:: ValueError`.

This idea also uses the LoadedFunction struct as the error, I guess I could do that without switching to such an IntoConcrete struct.
Thoughts?

hugr-passes/src/const_fold.rs
```diff
@@ -205,14 +201,8 @@ pub fn constant_fold_pass<H: HugrMut>(h: &mut H) {
...
-struct ConstFoldContext<'a, H>(&'a H);
-
-impl<H: HugrView> std::ops::Deref for ConstFoldContext<'_, H> {
-    type Target = H;
-    fn deref(&self) -> &H {
-        self.0
-    }
-}
+// Probably intend to remove this in a future PR, but not certain, so leaving in for now
+struct ConstFoldContext<'a, H>(#[allow(unused)] &'a H);
```
acl-cqc, line 205: You can see the removal in branch `acl/dataflow_call_indirect2`....

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -362,8 +381,7 @@ fn propagate_leaf_op<V: AbstractValue, H: HugrView>(
             ins.iter().cloned(),
         )])),
         OpType::Input(_) | OpType::Output(_) | OpType::ExitBlock(_) => None, // handled by parent
-        OpType::Call(_) => None,  // handled via Input/Output of FuncDefn
-        OpType::Const(_) => None, // handled by LoadConstant:
+        OpType::Call(_) | OpType::CallIndirect(_) => None, // handled via Input/Output of FuncDefn
```
acl-cqc, line 384: I dropped `Const`: it's not a dataflow node (it's a module/static), so never gets here

hugr-passes/src/dataflow/partial_value.rs
```diff
@@ -400,13 +450,17 @@ impl<V: AbstractValue> Lattice for PartialValue<V> {
                 Some((h3, b)) => (Self::Value(h3), b),
                 None => (Self::Top, true),
             },
+            (Self::LoadedFunction(lf1), Self::LoadedFunction(lf2))
+                if lf1.func_node == lf2.func_node =>
+            {
+                // TODO we should also require TypeArgs to be equal by at the moment these are ignored
```
acl-cqc, line 456: I.e. we are ignoring the TypeArgs in Call nodes too. I'll make an issue, but not this PR.

hugr-passes/src/dataflow/partial_value.rs
```diff
@@ -273,30 +292,32 @@ impl<V: Hash> Hash for PartialSum<V> {
 /// for use in dataflow analysis, including that an instance may be a [PartialSum]
 /// of values of the underlying representation
 #[derive(PartialEq, Clone, Eq, Hash, Debug)]
-pub enum PartialValue<V> {
+pub enum PartialValue<V, N = Node> {
```
acl-cqc, line 295: Not sure about the default N=Node. To get everything to work I ended up searching for a regex of PartialValue with only one thing between the `<>`'s...hence putting in a few explicit `, Node`s into tests. I could retract those and use the default, or remove the default....

acl-cqc: (for now I've removed the explicit `, Node` in tests, but easy to reinstate, and still considering whether to remove the default)

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -362,8 +392,7 @@ fn propagate_leaf_op<V: AbstractValue, H: HugrView>(
...
```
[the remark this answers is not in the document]
acl-cqc: Only by a massive refactor of `propagate_leaf_op`. Which is private, so we could, but I don't feel this is inconsistent - I think it's just like what we have been doing for `Call`

hugr-passes/src/dataflow/partial_value.rs
```diff
@@ -75,9 +86,21 @@ impl<V> PartialSum<V> {
...
         }
     }
+
+    /// Whether this sum might have the specified tag
+    pub fn supports_tag(&self, tag: usize) -> bool {
+        self.0.contains_key(&tag)
+    }
+
+    /// Can this ever occur at runtime? See [PartialValue::contains_bottom]
+    pub fn contains_bottom(&self) -> bool {
```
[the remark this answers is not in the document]
acl-cqc: Yes - "definitely contains bottom" not "might contain bottom". I think the best thing to do is rename, which I *could* do here (this PR is breaking already, of course I could deprecate)...or perhaps a doc update would suffice?

hugr-passes/src/dataflow/partial_value.rs
```diff
@@ -312,33 +333,59 @@ impl<V> PartialValue<V> {
...
+            }
+            PartialValue::PartialSum(ps) => ps.supports_tag(tag),
+            PartialValue::Top => true,
+        }
+    }
+
+    /// A value contains bottom means that it cannot occur during execution:
+    /// it may be an artefact during bootstrapping of the analysis, or else
+    /// the value depends upon a `panic` or a loop that
+    /// [never terminates](super::TailLoopTermination::NeverBreaks).
```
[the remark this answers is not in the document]
acl-cqc: Not yet - e.g. the output of a `LoadConst` inside an unreachable Case does not contain bottom until #1672 (if ever)

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -322,6 +323,35 @@ pub(super) fn run_datalog<V: AbstractValue, H: HugrView>(
             func_call(call, func),
             output_child(func, outp),
             in_wire_value(outp, p, v);
+
+        // CallIndirect --------------------
+        relation indirect_call(H::Node, H::Node); // <Node> is an `IndirectCall` to `FuncDefn` <Node>
```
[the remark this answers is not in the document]
acl-cqc: Great point about it needing to be a lattice, thank you! :). Done.
(Annoying about not having a test to show the difference but I bet it'd have show up sooner or later.)

Polymorphism is not wrong atm. Values from different type-instantiations are just joined together so will likely produce Top. If there's a `LoadNat` - well we don't even handle that yet - but *any* `DFContext::interpret_leaf_op` will just see a `LoadNat<Var0>` and not get the type args, so there's no sensible implementation other than returning Top. It's not ideal, but it's not wrong.

hugr-passes/src/dataflow/partial_value.rs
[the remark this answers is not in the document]
acl-cqc: Happy to deprecate and "rename" to anything we can agree on. `definitely_contains_bottom` seems too long (esp. with `row_` prefix). I don't really like `is_bottom`, would prefer to invert sense and go with `can_occur` (which is already noted in comments!) - but IIUC you don't like that. Other opinions welcome, but a rename with deprecation is not breaking so could follow afterwards.

hugr-passes/src/dataflow/results.rs
acl-cqc: Done!

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -337,13 +371,58 @@ pub(super) fn run_datalog<V: AbstractValue, H: HugrView>(
...
 }
 
+#[derive(PartialEq, Eq, Hash, Clone, PartialOrd)]
+enum LatticeWrapper<T> {
+    Bottom,
+    Value(T),
+    Top,
+}
+
+impl<N: HugrNode> Lattice for LatticeWrapper<N> {
```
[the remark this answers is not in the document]
acl-cqc: Lol yes, I was thinking that myself - it'll do just fine for those TypeArgs to polymorphic functions!!

Issue #1856, opened 2025-01-13
Make Value::Function contain only the Node (index) of a FuncDefn


acl-cqc:
(or FuncDecl, I guess). Then, remove `LoadFunction` - it becomes superceded by `LoadConstant`.

This allows many use cases of "constant Hugrs" via lambda-lifting (turning Lambdas into top-level functions), and "a function pointer" (no captures, not a closure) is the only runtime interpretation of a `Type::Function` that works without dynamic memory allocation.

Ops taking Type::Function thus should generally also take a row of extra arguments. Note that there is still a difference between a `EdgeKind::Value(Type::Function(....))` and `EdgeKind:Static(Type::Function(...))` (following #1594 enabling the latter) i.e. do we perform runtime computation to select the function pointer (this cannot also select any captured values - the functions must have exactly the same signature including any "captures"==extra arguments).

The many use cases of closures requiring dynamic allocation (captures, runtime-generated Hugrs, etc.) should thus be captured by various extension types, with their own `apply` ops etc. There are just too many variations here, and it's not clear which we should favour
* BRAT will need closures that can reference functions in the enclosing Hugr; this requires a linking step when such closures are executed (the approach of copying any functions needed from the outer hugr, into the inner Hugr, doesn't work - the outer Hugr/functions might contain the constant/inner Hugr, thus leading to an infinitely-big Hugr).
* E.g. @doug-q has called for "self-contained Hugrs" that do not require a linking step

I think this corresponds to the hugr-model idea of a value/constant referring to an "isolated" region of the Hugr, where "isolated" means - no incoming nonlocal edges, but static references would be allowed. @zrho right?

acl-cqc:
So maybe we only have `LoadFunction`? We'd lose the ability to have "constants" that were, say, tuples of multiple functions / functions with other things, but that feels minor. (We have also talked about whether we maintain all edge data structures all the time...@zrho)

acl-cqc:
I think we talked about turning hugr-model terms that referred to nodes, into static edges in the hugr. I mean, giving Const nodes an arbitrary number of static inputs would be...quite a change, not sure how straightforward

zrho:
Let my try to formulate the ideas in my head on where this could go.

### Functions as parameters, isolated regions

[...]

PR #2060, opened 2025-04-04
feat: Introduce new enum for constant folding; deprecate Value::Function

acl-cqc:
We've been using `ops::Value` for constant-folding, which is a bit of a shortcut but leads to significant problems: the result of ` LoadFunction` cannot fit into a Value (you'd need to copy all other external functions used by the loaded one inside it, or something - not really practical, and you've lost the understanding that you were calling another function in the same Hugr, too). #2059, which this follows, solves some of these; this tackles the remainder, specifically, feeding "function pointers" from LoadFunction into constant-folding.

The "solution" of extending Value to allow a reference to a node in the containing Hugr was considered in #1856 and was roundly rejected. Instead, this PR adds a separate enum `FoldVal` that looks quite similar to `Value` but adds those references/function-pointers.

I've taken the liberty of *not* including nested Hugrs in `FoldVal`, but rather deprecating `Value::Function` in favour of getting front-ends to lift these into their own FuncDefn's in the Hugr. I could be persuaded not to, and add a nested-Hugr variant to FoldVal, if we really want, but it does seem like probably more effort than it's worth - does anyone really have a good use case for `Value::Function` ?

I've also deprecated the old `constant_fold` and `fold` routines in favour of the new ones, which are currently called `fold2`. Better naming suggestions are welcome....(an alternative might be to define a new version of the ConstantFolder trait with only the new method, make OpDef store one of those, deprecate the old trait, and then `impl<T:ConstantFolder> NewConstantFolder for T` ?)

### DIFF
--- a/hugr-core/src/extension/const_fold.rs
+++ b/hugr-core/src/extension/const_fold.rs
@@ -101,9 +101,10 @@ pub trait ConstFold: Send + Sync {
     ///
     /// Defaults to calling [Self::fold] with those arguments that can be converted ---
     /// [FoldVal::LoadedFunction]s will be lost as these are not representable as [Value]s.
-    fn fold2(&self, type_args: &[TypeArg], inputs: Vec<FoldVal>, outputs: &mut [FoldVal]) {
+    fn fold2(&self, type_args: &[TypeArg], inputs: &[FoldVal], outputs: &mut [FoldVal]) {
         let consts = inputs
-            .into_iter()
+            .iter()
+            .cloned()
             .enumerate()
             .filter_map(|(p, fv)| Some((p.into(), fv.try_into().ok()?)))
             .collect::<Vec<_>>();
--- a/hugr-core/src/extension/const_fold.rs
+++ b/hugr-core/src/extension/const_fold.rs
@@ -2,6 +2,7 @@ use std::fmt::Formatter;
 
 use std::fmt::Debug;
 
+use crate::ops::constant::CustomConst;
 use crate::ops::constant::{OpaqueValue, Sum};
 use crate::ops::Value;
 use crate::types::{SumType, TypeArg};
@@ -10,7 +11,7 @@ use crate::{IncomingPort, Node, OutgoingPort, PortIndex};
 /// Representation of values used for constant folding.
 // Should we be non-exhaustive??
 // No point in parametrizing by HugrNode since then ConstFold would not be dyn/object-safe
-#[derive(Clone, PartialEq, Default)]
+#[derive(Clone, Debug, PartialEq, Default)]
 pub enum FoldVal {
     /// Value is unknown, must assume that it could be anything
     #[default]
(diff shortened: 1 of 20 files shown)


hugr-core/src/extension/const_fold.rs
```diff
@@ -1,19 +1,134 @@
...
 use crate::ops::Value;
-use crate::types::TypeArg;
+use crate::ops::constant::{CustomConst, OpaqueValue, Sum};
+use crate::types::{SumType, TypeArg};
+use crate::{IncomingPort, Node, OutgoingPort, PortIndex};
+
+/// Representation of values used for constant folding.
+/// See [ConstFold], which is used as `dyn` so we cannot parametrize by
+/// [HugrNode](crate::core::HugrNode).
+// Should we be non-exhaustive??
```
ss2165, line 11: Don't think so, "Extension" should be covering the points of extension

hugr-core/src/extension/const_fold.rs
```diff
@@ -1,19 +1,134 @@
...
+    /// Value is unknown, must assume that it could be anything
+    #[default]
+    Unknown,
+    /// A variant of a [SumType]
+    Sum {
+        /// Which variant of the sum type this value is.
+        tag: usize,
+        /// Describes the type of the whole value.
+        // Can we deprecate this immediately? It is only for converting to Value
+        sum_type: SumType,
```
ss2165, line 23: Could move these fields to an inner struct and mark them private i.e. `Sum(SumVal)`

hugr-core/src/extension/const_fold.rs
```diff
@@ -1,19 +1,134 @@
...
+        /// A value for each element (type) within the variant
+        items: Vec<FoldVal>,
+    },
+    /// A constant value defined by an extension
+    Extension(OpaqueValue),
+    /// A function pointer loaded from a [FuncDefn](crate::ops::FuncDefn) or `FuncDecl`
+    LoadedFunction(Node, Vec<TypeArg>), // Deliberately skipping Function(Box<Hugr>) ATM
```
ss2165, line 30: do we need non_exhaustive for `Function(Box<Hugr>)`, or a stub implementation that errors?

hugr-core/src/extension/const_fold.rs
```diff
@@ -1,19 +1,134 @@
...
+    pub fn get_custom_value<T: CustomConst>(&self) -> Option<&T> {
+        let Self::Extension(e) = self else {
+            return None;
+        };
+        e.value().downcast_ref()
+    }
+}
+
+impl TryFrom<FoldVal> for Value {
+    type Error = Option<Node>;
```
ss2165, line 80: please add docstring to explain error type

hugr-core/src/extension/const_fold.rs
```diff
@@ -27,9 +142,29 @@ pub fn fold_out_row(consts: impl IntoIterator<Item = Value>) -> ConstFoldResult
 
 /// Trait implemented by extension operations that can perform constant folding.
 pub trait ConstFold: Send + Sync {
+    /// Given type arguments `type_args` and [`FoldVal`]s for each input,
+    /// update the outputs (these will be initialized to [FoldVal::Unknown]).
+    ///
+    /// Defaults to calling [Self::fold] with those arguments that can be converted ---
+    /// [FoldVal::LoadedFunction]s will be lost as these are not representable as [Value]s.
+    fn fold2(&self, type_args: &[TypeArg], inputs: &[FoldVal], outputs: &mut [FoldVal]) {
```
ss2165, line 150: why does this follow a mutable output pattern? Would some other process have set some values already that this one would leave unchanged?

ss2165 (commented):
Initial thoughts.
There is a lot of deprecation overhead here, not sure how much is worthwhile. If the `fold2` names are kept they may need to become deprecated aliases in the breaking release

PR #2069, opened 2025-04-08
chore: Non-breaking portgraph & pyo3 version bump

aborgna-q:
`portgraph` is about to [make a breaking release](https://github.com/CQCL/portgraph/pull/194), and `pyo3` recently did so.

Normally bumping the dependencies would require us to make a breaking release of `hugr`, but since in both cases there are no code changes required for the update we can do a loose version requirement instead, and let downstream dependents update both dependencies at their own pace.

drive-by: Setup the MSRV CI check to use the minimal dependency versions, to ensure we don't break things in the future. This has de added benefit that we no longer need to pin dependencies in `Cargo.lock` if their newer versions bumped their MSRV (it currently happens with `half` and `zerofrom`)


.github/workflows/ci-rs.yml
```diff
@@ -220,6 +221,55 @@ jobs:
...
+          version: ${{ env.LLVM_VERSION }}
+      - uses: cargo-bins/cargo-binstall@main
+      - name: Install cargo-minimal-versions
+        run: |
+          cargo binstall cargo-hack --force
+          cargo binstall cargo-minimal-versions --force
+      - name: Generate minimal Cargo.lock
+        run: |
+          rm Cargo.lock
+          cargo minimal-versions check --direct
```
aborgna-q, line 257: This is the magic step in the CI job.
It re-generates the `Cargo.lock` using the lowest possible dependency versions, so all following commands run with those

Cargo.toml
```diff
@@ -53,16 +52,15 @@ jsonschema = "0.29.1"
...
-regex = "1.9.5"
+regex = "1.10.6"
 regex-syntax = "0.8.3"
 rstest = "0.24.0"
 semver = "1.0.26"
-serde = "1.0"
+serde = "1.0.195"
 serde_json = "1.0.140"
-serde_yaml = "0.9.19"
+serde_yaml = "0.9.34"
```
aborgna-q, line 63: These version updates fix dependency resolution errors found while compiling with `cargo-minimal-versions`.

mark-koch (approved):

PR #2071, opened 2025-04-08
fix!: Don't expose `HugrMutInternals`

aborgna-q:
`HugrMutInternals` is part of the semi-private traits defined in `hugr-core`.
While most things get re-exported in `hugr`, we `*Internal` traits require you to explicitly declare a dependency on the `-core` package (as we don't want most users to have to interact with them).

For some reason there was a public re-export of the trait in a re-exported module, so it ended up appearing in `hugr` anyways.

BREAKING CHANGE: Removed public re-export of `HugrMutInternals` from `hugr`.


aborgna-q:
Added to the next breaking-change milestone

cqc-alec (approved):

PR #2075, opened 2025-04-10
feat!: Allow generic Nodes in HugrMut insert operations

aborgna-q:
`insert_hugr`, `insert_from_view`, and `insert_subgraph` were written before we made `Node` a type generic, and incorrectly assumed the return type maps were always `hugr::Node`s.

The methods were either unusable or incorrect when using generic `HugrView`s source/targets with non-base node types.

This PR fixes that, and additionally allows us us to have `SiblingSubgraph::extract_subgraph` work for generic `HugrViews`.

BREAKING CHANGE: Added Node type parameters to extraction operations in `HugrMut`.


tatiana-s (approved):
I am not too familiar with this part of the codebase but generally looks good to me 👍

PR #2084, opened 2025-04-15
feat(hugr-py): move in result classes from guppylang

ss2165:
Guppy PR: https://github.com/CQCL/guppylang/pull/918


aborgna-q (commented):
LGTM, I'm just worried about polluting the library's root namespace.

Should we have this in `hugr.qsystems.result`, or `hugr.extension.qsystem` (to mimic the rust crate) instead?

ss2165:
good point  - I like `hugr.qsystem.result`

aborgna-q (approved):

PR #2093, opened 2025-04-16
feat!: remove ExtensionValue

### LINKED ISSUE #1595: Remove values in extensions
Use case isn't clear, can always be replaced with a no-input operation that returns the value. See also #1425

ss2165:
Closes #1595

BREAKING CHANGE: `values` field in `Extension` and `ExtensionValue` struct/class removed in rust and python. Use 0-input ops that return constant values.


acl-cqc:
Hang on - you say "use 0-input ops that return (constant) values", but we still have `Value::Extension` in hugr-core/src/ops/constant.rs?

Issue #2086, opened 2025-04-15
Reduce duplication when linearizing arrays


acl-cqc:
The `linearize_array` handler of #2023 generates identical Hugr `Value::Function` constants every time an array of that element type (regardless of length) is copied/discarded (i.e. for the same #copies) ...we want to be able to share some of these.

PR #2094, opened 2025-04-17
feat!: ReplaceTypes: allow lowering ops into a Call to a function already in the Hugr

acl-cqc:
There are two issues:
* Errors. The previous NodeTemplates still always work, but the Call one can fail if the Hugr doesn't contain the target function node. ATM there is no channel for reporting that error so I've had to panic. Otherwise it's an even-more-breaking change to add an error type to `NodeTemplate::add()` and `NodeTemplate::add_hugr()`. Should we? (I note `HugrMut::connect` panics if the node isn't there, but could make the `NodeTemplate::add` builder method return a BuildError...and propagate that everywhere of course)
* There's a big limitation in `linearize_array` that it'll break if the *element* says it should be copied/discarded via a NodeTemplate::Call, as `linearize_array` puts the elementwise copy/discard function into a *nested Hugr* (`Value::Function`) that won't contain the function. This could be fixed via lifting those to toplevel FuncDefns with name-mangling, but I'd rather leave that for #2086 ....

BREAKING CHANGE: Add new variant NodeTemplate::Call; LinearizeError no longer derives Eq.

### DIFF
--- a/hugr-passes/src/replace_types.rs
+++ b/hugr-passes/src/replace_types.rs
@@ -60,6 +60,9 @@ impl NodeTemplate {
     /// # Panics
     ///
     /// * If `parent` is not in the `hugr`
+    ///
+    /// # Errors
+    ///
     /// * If `self` is a [Self::Call] and the target Node either
     ///    * is neither a [FuncDefn] nor a [FuncDecl]
     ///    * has a [`signature`] which the type-args of the [Self::Call] do not match
--- a/hugr-passes/src/replace_types.rs
+++ b/hugr-passes/src/replace_types.rs
@@ -65,16 +65,16 @@ impl NodeTemplate {
     ///    * has a [`signature`] which the type-args of the [Self::Call] do not match
     ///
     /// [`signature`]: hugr_core::types::PolyFuncType
-    pub fn add_hugr(self, hugr: &mut impl HugrMut, parent: Node) -> Node {
+    pub fn add_hugr(self, hugr: &mut impl HugrMut, parent: Node) -> Result<Node, AddTemplateError> {
         match self {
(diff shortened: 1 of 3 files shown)


ss2165:
1. I think if you're breaking anyway might as well break with a proper error too (i.e. since it will have to go in a breaking release, fixing downstream is low cost)
2. annoying edge case - document and panic on detection if possible

acl-cqc:
Ok, so NodeTemplate::add_hugr already returned an error, but NodeTemplate::add needed one. And, the errors from both needed to be plumbed through ReplaceTypesError and LinearizeError.

I started trying to add my own error for `add_hugr` (see 5daa87b2c56aa0ac209989717ac2ae8e66da614e), but then realized that since I needed to plumb `BuildError`s through from `add` I switched to just use that. The alternative would be to look for all the error cases myself (as in that linked commit, but before calling the builder) and then `unwrap` the builder call....nicer for our users perhaps but seems less maintainable. And I think it's a small point. But I could be arm-twisted...

hugr-passes/src/replace_types.rs
```diff
@@ -48,18 +49,36 @@ pub enum NodeTemplate {
     // TODO: store also a vec<TypeParam>, and update Hugr::validate to take &[TypeParam]s
     // (defaulting to empty list) - see https://github.com/CQCL/hugr/issues/709
     CompoundOp(Box<Hugr>),
-    // TODO allow also Call to a Node in the existing Hugr
-    // (can't see any other way to achieve multiple calls to the same decl.
-    // So client should add the functions before replacement, then remove unused ones afterwards.)
+    /// A Call to a function (already) existing in the Hugr.
```
ss2165, line 52:
```suggestion
    /// A Call to an existing function.
```

ss2165 (approved):
happy to see many unwraps go away

PR #2097, opened 2025-04-17
feat!: Make arrays linear and add copyable value arrays

mark-koch:
* The old `array` type is now called `value_array` and lives in a separate extension
* The default `array` is now a linear type. Additional `clone` and `discard` operations follow in a separate PR
* To avoid code duplication, array operations and values are now defined generically over a new `ArrayKind` trait that is instantiated with `Array` (the linear one) and `VArray` (the copyable one) to generate the `array` and `value_array` extensions

Closes #2066

BREAKING CHANGE: `std.collections.array` is now a linear type, even if the contained elements are copyable. Use the new `std.collections.value_array` for an array with the previous copyable semantics.

PR #2101, opened 2025-04-22
feat: Add value_array to/from array conversion ops

mark-koch:
See #2097 for context


zrho (approved):
LGTM.

We really need to work on the ergonomics of defining operations. The generics trick is clever and avoids a lot of boring code, but hopefully operation definitions will at some point be concise enough that it isn't necessary anymore.

PR #2100, opened 2025-04-22
feat: Add array clone and discard ops

mark-koch:

PR #2109, opened 2025-04-23
feat!: Improved array lowering

### LINKED ISSUE #2066: Define linear array extension
Ops: 
* All operations supported by the regular array
* Clone
* Drop

### LINKED ISSUE #2067: Lower linear arrays to LLVM
Using `malloc`

### LINKED ISSUE #2124: Add array linearisation pass
For turning `value_array`s into regular linear `array`s

mark-koch:
Feature branch for improved array lowering.

* The old `array` type is now called `value_array` and lives in a separate extension
* The default `array` is now a linear type with additional `clone` and `discard` operations
* To avoid code duplication, array operations and values are now defined generically over a new `ArrayKind` trait that is instantiated with `Array` (the linear one) and `VArray` (the copyable one) to generate the `array` and `value_array` extensions
* An `array<n, T>` is now lowered to a fat pointer `{ptr, usize}` where `ptr` is a heap allocated pointer of size at least `n * sizeof(T)` and the `usize` is an offset pointing to the first element (i.e. the first element is at `ptr + offset * sizeof(T)`). The rational behind the additional offset is the `pop_left` operation which bumps the offset instead of mutating the pointer. This way, we can still free the original pointer when the array is discarded after a pop.

Tracked PRs:
* #2097 (closes #2066)
* #2100
* #2101
* #2110
* #2112 (closes #2067)
* #2119
* #2125 (closes #2124)

BREAKING CHANGE: `std.collections.array` is now a linear type, even if the contained elements are copyable. Use the new `std.collections.value_array` for an array with the previous copyable semantics.

BREAKING CHANGE: `std.collections.array.get` now also returns the passed array as an extra output

BREAKING CHANGE: `ArrayOpBuilder` was moved from `hugr_core::std_extensions::collections::array::op_builder` to `hugr_core::std_extensions::collections::array`.


ss2165 (approved):
lets gooooo

PR #2110, opened 2025-04-23
feat!: Make `array.get` give back the passed array

mark-koch:
Since the array type is now linear following #2097, we need to change the signature of the `array.get` op to give the passed array back to the user.

BREAKING CHANGE: `std.collections.array.get` now also returns the passed array as an extra output


hugr-llvm/src/utils/array_op_builder.rs
```diff
@@ -30,13 +30,13 @@ pub trait ArrayOpBuilder<AK: ArrayKind>: Dataflow {
         size: u64,
         input: Wire,
         index: Wire,
-    ) -> Result<Wire, BuildError> {
+    ) -> Result<(Wire, Wire), BuildError> {
```
aborgna-q, line 33: It wouldn't hurt to document the trait method saying which output is which.
(none of the other methods have docs though -.-)

aborgna-q (approved):

Issue #1926, opened 2025-02-24
Replace `HugrInternals::base_hugr` with simpler methods


aborgna-q:
[`HugrInternals::base_hugr`](https://github.com/CQCL/hugr/blob/bb0d54b4e63f6a311eda6f9a83cc221b43ac8269/hugr-core/src/hugr/internal.rs#L32-L34) returns the root `Hugr` used to implement a hugr view.
In practice, this call is used in a handful of cases:

- To access the metadata/optype for a single node.
- To access the `ExtensionRegistry`.
- To access the `Hierarchy` that linked to the `HugrInternals::portgraph`
- To run full-graph validation
- To render mermaid/dot of the full base graph
- In `SiblingGraph`, to avoid nesting views when creating one from a generic `HugrView`.

This works fine for our current definitions, but non-trivial wrappers like the one discussed in https://github.com/CQCL/tket2/pull/778 are not necessarily based on a single base `Hugr`.

All but the last point I listed can be implemented as individual methods on `HugrInternals` instead,
- `metadata(&self, Node) -> Option<NodeMetadataMap>`
- `optype(&self, None) -> Option<OpType>`
- `extension_registry(&self) -> ExtensionRegistry`
- `hierarchy` would return a `Cow<'_, Hierarchy>` so we can either return a reference to the internally stored one, or compute one on-the-fly (for the CircuitHistory usecase).
- Rendering should use the existing `portgraph` directly.
- `SiblingGraph` shouldn't need to do the unwrapping. For non-dyn types all the calls through wrapping layers will hopefully be inlined and optimized away (we should benchmark this).

One drawback to keep in mind is that we no longer get the generic "escape hatch" that lets us recover the outer hugr from a filtered view. This means for example that focusing on a region will hide away non-local nodes outside the parent's hierarchy. I'd argue that those cases should use region-pointers instead. See https://github.com/CQCL/hugr/discussions/1642 https://github.com/CQCL/hugr/issues/1554.

aborgna-q:
There's also a `to_pg_index` method to map the generic node indices, but no `from_pg_index`

PR #2111, opened 2025-04-23
feat: Hugrmut on generic nodes

aborgna-q:
- Allows `HugrMut` to be implemented for `HugrView`s with arbitrary node types
- Removes `HugrMutInternals::hugr_mut(&mut self) -> &mut Hugr`, it can be implemented for more complex types. This is required for #1926, but I haven't touched the read-only side yet.

drive-by: Fix a couple bugs in rewrite implementations that assumed that `SiblingMut` contained transitive children.


hugr-core/src/hugr/hugrmut.rs
```diff
@@ -624,22 +551,16 @@ fn insert_subgraph_internal<N: HugrNode>(
 /// Panic if [`HugrView::valid_node`] fails.
 #[track_caller]
 pub(super) fn panic_invalid_node<H: HugrView + ?Sized>(hugr: &H, node: H::Node) {
-    if !hugr.valid_node(node) {
-        panic!(
-            "Received an invalid node {node} while mutating a HUGR:\n\n {}",
-            hugr.mermaid_string()
-        );
+    if cfg!(debug_assertions) && !hugr.valid_node(node) {
```
mark-koch, line 554: Hmm strictly speaking the docs promise that operations panic if nodes are invalid. Do we need to specify that this is only true in debug builds?

hugr-core/src/hugr/internal.rs
```diff
@@ -294,77 +181,90 @@ pub trait HugrMutInternals: RootTagged<Node = Node> {
...
     /// match the OpType signature.
     ///
     /// Returns the old OpType.
     ///
-    /// TODO: Add a version which ignores input extensions
-    ///
-    /// # Errors
-    ///
-    /// Returns a [`HugrError::InvalidTag`] if this would break the bound
-    /// (`Self::RootHandle`) on the root node's OpTag.
```
mark-koch, line 314: Looks like all impls still do this check. Should the error still be documented?

hugr-core/src/hugr/internal.rs
```diff
@@ -294,77 +181,90 @@ pub trait HugrMutInternals: RootTagged<Node = Node> {
...
-    }
+    fn move_before_sibling(&mut self, node: Self::Node, before: Self::Node);
 
     /// Replace the OpType at node and return the old OpType.
     /// In general this invalidates the ports, which may need to be resized to
     /// match the OpType signature.
     ///
     /// Returns the old OpType.
     ///
-    /// TODO: Add a version which ignores input extensions
```
mark-koch, line 309: Is this TODO no longer relevant?

hugr-core/src/hugr/hugrmut.rs
aborgna-q: The main issue is that now that we don't have the `hugr_mut` shortcut, stacked hugr wrappers will end up running the check at each and every layer.

Maybe we should leave them enabled and benchmark the performance impact later.

PR #2119, opened 2025-04-28
feat!: Generalise array op builers to arbitrary array kinds

mark-koch:
* Renamed `ArrayOpBuilder` trait to `GenericOpBuilder` whose methods are generic over the array implementation
* Defined new `ArrayOpBuilder` and `VArrayOpBuilder` traits that are implemented in terms of the generic version. I moved those into the `array` and `value_array` modules respectively, so they don't come into scope at the same time
* Moved `all_array_ops` ops function out of the tests module so we can reuse it across other tests (in particular getting rid of the code duplication in `hugr-llvm::extension::collections::array::test`

BREAKING CHANGE: `ArrayOpBuilder` was moved from `hugr_core::std_extensions::collections::array::op_builder` to `hugr_core::std_extensions::collections::array`

### DIFF
--- a/hugr-core/src/std_extensions/collections/array.rs
+++ b/hugr-core/src/std_extensions/collections/array.rs
@@ -15,14 +15,15 @@ use std::sync::Arc;
 use delegate::delegate;
 use lazy_static::lazy_static;
 
+use crate::builder::{BuildError, Dataflow};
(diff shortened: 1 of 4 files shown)


aborgna-q (approved):
LGTM. Nice solutions to the duplicated impl problem :)

PR #2113, opened 2025-04-23
feat!: Removed `model_unstable` feature flag.

zrho:
This PR removes the `model_unstable` feature flag across the repository. Closes #2104.

BREAKING CHANGE: Downstream crates need to remove the `model_unstable` feature flag when referencing `hugr` or `hugr-core`.

PR #2120, opened 2025-04-28
feat!: Removed model_unstable feature flag

aborgna-q:
Re-created from #2113, targetting the release branch instead.

BREAKING CHANGE: Downstream crates need to remove the model_unstable feature flag when referencing hugr or hugr-core.


aborgna-q:
@cqc-alec You already did the review on the original PR, so this should be mergable as-is

cqc-alec (approved):

PR #2125, opened 2025-04-28
feat: Add array linearisation pass

### LINKED ISSUE #2124: Add array linearisation pass

mark-koch:
Closes #2124


hugr-passes/src/linearize_array.rs
```diff
@@ -0,0 +1,418 @@
...
+        });
+        for op_def in ArrayOpDef::iter() {
+            pass.replace_parametrized_op(
+                value_array::EXTENSION.get_op(&op_def.name()).unwrap(),
+                move |args| {
+                    // `get` is only allowed for copyable elements. If the elements
+                    // are now linear, we'd have to replace it with a `set`. But what
+                    // should we put in??? For now, let's just error out and make sure
+                    // we're not emitting `get`s for nested value arrays.
+                    if op_def == ArrayOpDef::get && !args[1].as_type().unwrap().copyable() {
```
zrho, line 60: This check avoids `get`s for all linear types, not just nested value arrays, right? The error message should probably reflect that. It'd also be good to document this limitation in some doc comment on the pass.

hugr-passes/src/linearize_array.rs
```diff
@@ -0,0 +1,418 @@
...
+};
+
+/// A HUGR -> HUGR pass that turns 'value_array`s into regular linear `array`s.
+///
+/// To construct a [LinearizeArrayPass] use [Default::default].
+#[derive(Clone)]
+pub struct LinearizeArrayPass(ReplaceTypes);
+
+impl Default for LinearizeArrayPass {
+    fn default() -> Self {
```
zrho, line 37: It's convention in Rust to have a `new()` method in this case as well.

zrho (approved):

hugr-passes/src/linearize_array.rs
mark-koch: Validity checking should ensure that the type was copyable before we ran the pass. The only way for the element to become linear is if it used to contain a `value_array` that we turned into a linear `array`.

I'll add it to the doc 👍

PR #2128, opened 2025-04-28
chore(hugr-llvm): upgrade to inkwell 0.6.0

qartik:
Part of https://github.com/quantinuum-dev/hugrverse/issues/158


DEVELOPMENT.md
```diff
@@ -28,10 +28,10 @@ shell by setting up [direnv](https://devenv.sh/automatic-shell-activation/).
 
 To setup the environment manually you will need:
 
-- Just: https://just.systems/
-- Rust `>=1.75`: https://www.rust-lang.org/tools/install
-- uv `>=0.3`: docs.astral.sh/uv/getting-started/installation
-- Optional: capnproto `>=1.0`: https://capnproto.org/install.html
+- Just: <https://just.systems/>
+- Rust `>=1.75`: <https://www.rust-lang.org/tools/install>
```
cqc-alec, line 32: Shouldn't this change to 1.85?

.github/workflows/ci-rs.yml
```diff
@@ -234,7 +234,7 @@ jobs:
         id: toolchain
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: "1.75"
+          toolchain: "1.85"
```
cqc-alec, line 237: Are we OK with this @aborgna-q ?

DEVELOPMENT.md
qartik: Looks like it's being done in the https://github.com/CQCL/hugr/pull/2136/files#diff-fc2a310fcfedfefb0046def697f932def80cbb1e78c689bc0af3c8eab3e33ece but it's not obvious to me, why that's not merged to `main`.

I think it's best for me to not change it here.

cqc-alec: Yeah that PR is targeting a breaking release branch which will be merged after we've done the next non-breaking release, I think.

qartik: right, so I wonder if this PR should also target that branch?

.github/workflows/ci-rs.yml
aborgna-q: We did this in the MSRV bump in #2136, now on the breaking release branch.
But it shouldn't be done changed, as we are keeping the MSRV for the non-breaking release.

DEVELOPMENT.md
aborgna-q: If it can be included in the non-breaking release then it's fine to target `main` here.

.github/workflows/ci-rs.yml
qartik: I only included it because before this change the CI failed as inkwell's new version seems to require Rust 2024: https://github.com/CQCL/hugr/actions/runs/14716264966/job/41300250543?pr=2128 (see error log below)

Happy to remove the change if it's acceptable for the CI to be bypassed.

```
running `cargo test --no-default-features --no-run` on hugr-llvm (4/7)
  warning: /home/runner/work/hugr/hugr/Cargo.toml: unused manifest key: workspace.lints.rust.unexpected_cfgs.check-cfg
   Downloading crates ...
    Downloaded thiserror-impl v1.0.69
    Downloaded inkwell_internals v0.11.0
  error: failed to download replaced source registry `crates-io`
  
  Caused by:
    failed to parse manifest at `/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/inkwell_internals-0.11.0/Cargo.toml`
  
  Caused by:
    feature `edition2024` is required
  
    The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.75.0 (1d8b05cdd 2023-11-20)).
    Consider trying a newer version of Cargo (this may require the nightly release).
```

qartik: or I can aim to the breaking release.

qartik: Done

PR #2136, opened 2025-04-29
feat!: Bump MSRV to 1.85

aborgna-q:
Bumps the minimum supported rust version to 1.85, and migrates to the 2024 edition.

Most of the changes here are automated. The diff is quite noisy due to some changes in formatting, and me using this opportunity to auto--fix some optional clippy lints. It may be easier to check the changes per-commit.


mark-koch (approved):
Lots of merge conflict fun ahead 🫠

mark-koch:
We could potentially wait to merge this until the other pending breaking changes are in? Or is this needed to go ahead with the tket2 upgrade?

aborgna-q:
Tket2 will need to bump its msrv separatedly (_before_ updating hugr).

This is needed for #2126 though, which is required for other new breaking changes.

I'm happy to wait for PRs that will get merged soon, but this _will_ be problematic no matter when we merge it

aborgna-q:
I think I'll do only the MSRV bump here (which isn't too noisy), and leave the edition change for when there's less PRs going around. (it's also not a breaking change so no need to rush it).

Issue #1496, opened 2024-09-02
`NamedOp::name` should always return scoped names.


aborgna-q:
```rust
// Consider a noop operation
op: Optype = prelude::Noop.to_extension_op().into()

// This returns the fully scoped name 
assert_eq!(<op as NamedOp>::name(), "prelude.Noop");

// But the same operation on the noop definition returns something different
assert_eq!(<NoopDef as NamedOp>::name(), "Noop");

// So the name can't be used as an unique id
assert_ne!(op.name(), NoopDef.name());
```
Since we expect names to be used as unique identifiers for the functions, they should always include the extension ids.

aborgna-q:
Note that this change will be a breaking release. We'll have to rely on workarounds for the current version.

PR #2138, opened 2025-04-29
feat!: Make `NamedOp` private. Add `MakeExtensionOp::name` and `MakeOpDef::opdef_name`

doug-q:
BREAKING CHANGE: `NamedOp` is no longer public. Use `OpType`'s `Display` impl for a short string describing the OpType.
BREAKING CHANGE: New required trait method `opdef_name` on `MakeOpDef`
BREAKING CHANGE: New required trait method `name` on `MakeExtensionOp`

### DIFF
--- a/hugr-core/src/std_extensions/ptr.rs
+++ b/hugr-core/src/std_extensions/ptr.rs
@@ -44,8 +44,7 @@ impl PtrOpDef {
 
 impl MakeOpDef for PtrOpDef {
     fn opdef_id(&self) -> OpName {
-        let s: &str = self.into();
-        s.into()
+        <&'static str>::from(self).into()
     }
 
     fn from_def(op_def: &OpDef) -> Result<Self, OpLoadError>
--- a/hugr-core/src/std_extensions/collections/array/array_op.rs
+++ b/hugr-core/src/std_extensions/collections/array/array_op.rs
@@ -190,8 +190,9 @@ impl<AK: ArrayKind> GenericArrayOpDef<AK> {
 
 impl<AK: ArrayKind> MakeOpDef for GenericArrayOpDef<AK> {
     fn opdef_id(&self) -> OpName {
-        <&Self as Into<&'static str>>::into(self).into()
+        <&'static str>::from(self).into()
     }
+
     fn from_def(op_def: &OpDef) -> Result<Self, OpLoadError>
     where
         Self: Sized,
--- a/hugr-core/src/std_extensions/logic.rs
+++ b/hugr-core/src/std_extensions/logic.rs
@@ -78,7 +78,7 @@ pub enum LogicOp {
 
 impl MakeOpDef for LogicOp {
     fn opdef_id(&self) -> OpName {
-        <&Self as Into<&'static str>>::into(self).into()
+        <&'static str>::from(self).into()
     }
 
     fn init_signature(&self, _extension_ref: &Weak<Extension>) -> SignatureFunc {


hugr-core/src/extension/simple_op.rs
```diff
@@ -29,22 +25,15 @@ pub enum OpLoadError {
...
-}
-
 /// Traits implemented by types which can add themselves to [`Extension`]s as
 /// [`OpDef`]s or load themselves from an [`OpDef`].
 ///
 /// Particularly useful with C-style enums that implement [strum::IntoEnumIterator],
 /// as then all definitions can be added to an extension at once.
-pub trait MakeOpDef: NamedOp {
+pub trait MakeOpDef {
+    /// TODO docs
```
acl-cqc, line 34: In particular, is this qualified or unqualified?

hugr-core/src/extension/simple_op.rs
```diff
@@ -70,7 +59,7 @@ pub trait MakeOpDef: NamedOp {
 
     /// Description of the operation. By default, the same as `self.name()`.
```
acl-cqc, line 60:
```suggestion
    /// Description of the operation. By default, the same as [Self::opdef_name].
```

acl-cqc: Or default to empty string...

hugr-core/src/extension/simple_op.rs
```diff
@@ -150,7 +139,10 @@ pub trait HasDef: MakeExtensionOp {
 
 /// Traits implemented by types which can be loaded from [`ExtensionOp`]s,
 /// i.e. concrete instances of [`OpDef`]s, with defined type arguments.
-pub trait MakeExtensionOp: NamedOp {
+pub trait MakeExtensionOp {
+    /// The name of the operation
```
acl-cqc, line 143: full or unqualified?

hugr-core/src/extension/simple_op.rs
```diff
@@ -188,6 +180,10 @@ pub trait MakeExtensionOp: NamedOp {
 
 /// Blanket implementation for non-polymorphic operations - [OpDef]s with no type parameters.
 impl<T: MakeOpDef> MakeExtensionOp for T {
+    fn name(&self) -> OpName {
+        <Self as MakeOpDef>::opdef_name(self)
```
acl-cqc, line 184: Either `MakeOpDef::opdef_name(self)` or doesn't just `self.opdef_name()` work?

hugr-core/src/extension/simple_op.rs
```diff
@@ -310,6 +306,10 @@ mod test {
     }
 
     impl MakeOpDef for DummyEnum {
+        fn opdef_name(&self) -> OpName {
+            <&Self as Into<&'static str>>::into(self).into()
```
acl-cqc, line 310: as before, `<&'static str>::from(self).into()` is simpler

hugr-core/src/ops/custom.rs
```diff
@@ -136,6 +136,17 @@ impl ExtensionOp {
...
         self.def.extension_id()
     }
+
+    /// Returns the unqualified name of the operation. e.g. 'iadd'
+    ///
+    pub fn unqualified_name(&self) -> &OpNameRef {
+        self.def.name()
+    }
+
+    /// Returns the unqualified name of the operation. e.g. 'arithmetic.iadd'
```
acl-cqc, line 146:
```suggestion
    /// Returns the qualified name of the operation. e.g. 'arithmetic.iadd'
```

hugr-core/src/ops/custom.rs
```diff
@@ -252,9 +263,10 @@ impl OpaqueOp {
 impl NamedOp for OpaqueOp {
     /// The name of the operation.
```
acl-cqc, line 264: Drop comment, it's just a repeat.

hugr-core/src/extension/simple_op.rs
acl-cqc: Even better, rename the method to `qualified_name` or `unqualified_name`

hugr-core/src/ops.rs
```diff
@@ -355,7 +355,7 @@ pub type OpNameRef = str;
 #[enum_dispatch]
 /// Trait for setting name of OpType variants.
 // Separate to OpTrait to allow simple definition via impl_op_name
-pub trait NamedOp {
+pub(crate) trait NamedOp {
     /// The name of the operation.
     fn name(&self) -> OpName;
```
acl-cqc, line 360: At the least, doc whether this is qualified or not, but given the trait is now private, there is no additional disruption if you just rename the method to `qualified_name` or `unqualified_name`, so I'd do that

hugr-core/src/ops/custom.rs
```diff
@@ -136,6 +136,17 @@ impl ExtensionOp {
...
+
+    /// Returns the unqualified name of the operation. e.g. 'iadd'
+    ///
+    pub fn unqualified_name(&self) -> &OpNameRef {
+        self.def.name()
+    }
+
+    /// Returns the unqualified name of the operation. e.g. 'arithmetic.iadd'
+    pub fn qualified_name(&self) -> OpName {
+        self.name()
```
acl-cqc, line 148: Is this NamedOp::name? So we could rename that to NamedOp::qualified_name, and then you'd have to do `NamedOp::qualified_name(self)`, but at least that would be explicit

hugr-core/src/std_extensions/arithmetic/conversions.rs
```diff
@@ -45,6 +45,10 @@ pub enum ConvertOpDef {
 }
 
 impl MakeOpDef for ConvertOpDef {
+    fn opdef_name(&self) -> OpName {
+        <&Self as Into<&'static str>>::into(self).into()
```
acl-cqc, line 49: ditto

hugr-core/src/ops/custom.rs
acl-cqc: Alternatively, doc that `OpName`s are always qualified (right?), and then call this method `name` (i.e. make all `.name()`s qualified, and sometimes there is `.unqualified_name()` too - in that case, you might want to rename to `name_unqualified` because it'll be more obvious browsing the list of autocompletes!)

hugr-core/src/std_extensions/collections/array/array_scan.rs
```diff
@@ -23,17 +23,17 @@ pub const ARRAY_SCAN_OP_ID: OpName = OpName::new_inline("scan");
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
 pub struct ArrayScanDef;
 
-impl NamedOp for ArrayScanDef {
-    fn name(&self) -> OpName {
-        ARRAY_SCAN_OP_ID
+impl From<&ArrayScanDef> for &'static str {
```
acl-cqc, line 26: I see, so is that how the `<&Self as Into<&'static str>>:into(self)` works. This is a bit of an odd declaration - do we parametrize anything `<T>...where &T: Into<&'static str>`? I am wondering if maybe we should just put this directly into the impl of `MakeOpDef::opdef_name` and do away with this rather odd conversion; what purpose does it serve?

(There are also things like `ToString` which are more standard - I haven't seen why we need a trait separate from MakeOpDef, but if we do, then how about using that?)

hugr-core/src/std_extensions/collections/static_array.rs
```diff
@@ -218,6 +218,11 @@ impl StaticArrayOpDef {
 }
 
 impl MakeOpDef for StaticArrayOpDef {
+    fn opdef_name(&self) -> OpName {
+        let s: &str = self.into();
```
acl-cqc, line 222: Make standardize this with the others (suggest `<&'static str>::from(self).into()`), but see prev. comment, why do we have that made `impl From<...> for &'static str` anyway?

hugr-core/src/std_extensions/collections/array/array_scan.rs
acl-cqc: Ok, this seems to be the only place where we have to do this, so we must be getting this odd conversion some other way for the other places??

hugr-core/src/std_extensions/ptr.rs
```diff
@@ -234,8 +236,8 @@ pub(crate) mod test {
 
     use super::*;
     use crate::std_extensions::arithmetic::float_types::float64_type;
-    fn get_opdef(op: impl NamedOp) -> Option<&'static Arc<OpDef>> {
-        EXTENSION.get_op(&op.name())
+    fn get_opdef(op: impl Into<&'static str>) -> Option<&'static Arc<OpDef>> {
```
acl-cqc, line 239: Ah, I think this is the (only) place where we're parametrizing over types that have this odd conversion. (So you can pass an actual `&'static str` here, or a reference to an op-thing that has that conversion). However, this is only a test method - consider using `ToString` here?

acl-cqc (commented):
Thanks @doug - generally I like this, but I think we should standardize more the method naming - some places we have `qualified_name`, others `name` yet these are equivalent (I think).

You haven't mentioned why we are doing this - does this close #1496 ?

Also if the goal is to hide NamedOp, but then publically use `&T: Into<&'static str>` then I think that would be a retrograde step, but I don't think the latter really is intended for public use, so I think it would be better to clean that up instead...

so I'm 90% of the way to hitting approve but I think I'd like to see what you can do about those!

acl-cqc:
If the into-static-string thing works with that `impl From<&ArrayScanDef>` removed, then yeah, that's fine for a test method!

cqc-alec (approved):

PR #2143, opened 2025-04-29
feat!: Add Visibility to FuncDefn/FuncDecl.

### LINKED ISSUE #1752: Decide on which functions are "public"
We don't currently have a notion of "public" functions. 

In the spec, "main" is defined as the "entry point", but surely there are cases where the HUGR is a library and some subset of the top-level functions are "public".

The notion of publicity is important at least for:
* removing unused functions (a public function and all it's transitive callees should not be deleted)
* dataflow analysis (for a public function we cannot assume we know all callsites)

Maybe a `FuncDecl` is public, or maybe it is a private alias to a public function. It's not clear to me when the publicity of a FuncDecl would matter. (I suppose when deleting unused functions?

LLVM has the concept of "linkage" which contains this binary notion of publicity. Linkage has lots of details allowing it to model weak symbols etc.

We could:
* add a `public: bool` field to `FuncDefn`, 
* establish a convention where `hugr.public: true` metadata means public.
* leave it unspecified, anything that cares should take a set of nodes identifying all public funcs.

### LINKED ISSUE #2354: Add visibility specifier to FuncDefn (+friends)
`pub: bool` would be the poor mans way, but we could do better with a (presumably non-exhaustive) enum for future extensibility

acl-cqc:
Note: modified to come after #2412 as I want those Schema checks in place to check this works!
closes #2354, #1752.
Includes cherry-pick of #2343.

* Add `enum Visibility`. In fact *2 (hugr-model version is not serde-able so cannot be used in hugr-core).
* `FuncDefn` and `FuncDecl` gain this. (A good use-case for a private FuncDecl is building cyclic functions via builder `define_declaration`.)
   * I've not added to `AliasDecl` or `AliasDefn` - I could, it would be consistent, but we hardly use them....
* `validate` checks that the *public* children of a Module have unique names (except multiple FuncDecls may alias).

* `FuncDefn::new` defaults to private (also builder `define_function` / `FunctionBuilder::new`, and json-deserialize in both python and rust). There is a `_vis` taking an explicit Visibility parameter.
* `FuncDecl::new` defaults to public (also builder `declare_function`, and json-deserialize*2). Again there is `_vis`.
* hugr-model import/export + roundtrip + AST printing via new keyword `pub`. Note since `hugr-core` does not have visibility on aliases or extension operations I've defaulted these to private to avoid any change to textual output.
* add field to hugr-py, using "string"-enum (`Literal["Public", "Private"]`)
* I've attempted to add hugr-model hugr-py bindings, but there are no tests of these yet.

Note all changes to hugr-passes delayed until a follow-up PR.

BREAKING CHANGE: hugr-model: Symbol has an extra field


acl-cqc:
> So when a FuncDefn with public == true is "lifted" to a module the public flag "turns on"? This seems a bit dangerous. Lifting functions to modules is very normal and useful.

Yes, good point.

> I don't see a better path though. This is a case for emitting a warning (non-module-child has link name)

I don't really see why this shouldn't be an error - the link-name is pointless, you can never link to it. (I only didn't do that because it requires another bit of validation code.)

> > * An alternative would be to have a separate field `link_name: Option<String>`. Preferences? Factory/builder methods might default to using the same string for both kinds of name. However, `name` then has no use in the code, so might be better replaced by name/description metadata....?
> 
> This is my preference. I expect the "String" to grow in complexity over time, linking is complicated. One might want to link two hugrs together, i.e. one uses a "public" function of the other, but that function doesn't want to be actually exposed as a public symbol after conversion to llvm. There is weak linking, ODR. I expect we will want more than zero of these features.
> 
> Yes name can be moved to metadata. Or it can stay where it is because it's Very Important Metadata.

Yup ok, let's keep `name` but not enforce uniqueness, and add optional linkname, with uniqueness requirement

> Nice to have: it would be great if we were able to be unconcerned with naming generated functions (e.g. monomorphise)

So monomorphise will have to mangle the link-name (if there is one), but could leave the name alone if we want...

> * multiple FuncDecls with the same signature should be able to share link names (the two funcdecls are aliased)

yup, good point, let's not require deduplication, but checking same signature, indeed

> * It should be valid for FuncDecls to not have link names. Obviously you can't codegen this yet, but I expect such HUGRs to be useful.

This is the one I haven't really seen the use case for. Seems to me that whereas a `FuncDefn` has both name and optionally link-name, a `FuncDecl` has *only* (but always) link-name. What else does a FuncDecl do except open the way for linking?

> Perhaps we should mark FuncDefn (and everything else?) #[non_exhaustive] ?

And require use of factory methods....yes, I think that's good :+1
>
> Do you think we should change "FuncDecls must be children of module" here?

Yes, I think they should be children of root (to make them easy for the linker to find), my use case in #2108 requires linking a DFG-rooted Hugr into another. (Its own linking code, but good for this to be as consistent as possible.)

> I do not think we should attempt to link Consts. One can just write a function that returns the const. Yes we need const de-duplication, but we need it without linking too.

Not convinced about this, but definitely happy to drop from this PR, will do that.

zrho:
Yes, there's an argument to allow for `name`s on non-public things. This keeps names as the general mechanism to refer to symbols such as functions, but allows to internally go for a nameless representation when convenient, e.g. while rewriting. The philosophy then would be that everything has a name, but for some things we don't care what it is.

The human readable title should be optional. If `hugr-core` had a nicer interface for metadata already, it probably wouldn't even need to be blessed metadata.

hugr-core/tests/snapshots/model__roundtrip_alias.snap
```diff
@@ -1,6 +1,6 @@
 ---
 source: hugr-core/tests/model.rs
-expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-alias.edn\"))"
+expression: ast
```
acl-cqc, line 3: This looks wrong?! No idea how I've broken this atm, will investigate but tips most welcome

hugr-core/tests/snapshots/model__roundtrip_cond.snap
```diff
@@ -17,6 +17,7 @@ expression: ast
 (import arithmetic.int.types.int)
 
 (declare-operation
+  pub
```
acl-cqc, line 20: Ok these are annoying. Any objection to changing the visibility for Operation symbols to private? (Obvs there is no equivalent in hugr-core.)

hugr-core/src/core.rs
```diff
@@ -238,6 +238,49 @@ impl<N: HugrNode> std::fmt::Display for Wire<N> {
...
+    /// Function is hidden, for use within the hugr only
+    Private,
+}
+
+impl From<hugr_model::v0::Visibility> for Visibility {
+    fn from(value: hugr_model::v0::Visibility) -> Self {
+        match value {
+            hugr_model::v0::Visibility::Private => Self::Private,
+            hugr_model::v0::Visibility::Public => Self::Public,
+            _ => unimplemented!(),
```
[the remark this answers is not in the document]
acl-cqc: `hugr_model::Visibility` is non-exhaustive, so we have to have this. Alternatively, we could decide not to mark it as non-exhaustive...opinions vary on whether that's a good idea

hugr-passes/src/dead_funcs.rs
```diff
@@ -124,16 +140,42 @@ impl<H: HugrMut<Node = Node>> ComposablePass<H> for RemoveDeadFuncsPass {
...
+/// * otherwise, the [HugrView::entrypoint] itself will.
+///
+/// # Errors
+/// * If any node in `entry_points` is not a [`FuncDefn`]
+///
+/// [`Call`]: hugr_core::ops::OpType::Call
+/// [`FuncDefn`]: hugr_core::ops::OpType::FuncDefn
+/// [`LoadFunction`]: hugr_core::ops::OpType::LoadFunction
+/// [`Module`]: hugr_core::ops::OpType::Module
+pub fn remove_dead_funcs_vis(
```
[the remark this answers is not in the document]
acl-cqc: Similarly....`remove_dead_funcs_pass` ??

hugr-core/src/core.rs
```diff
@@ -238,6 +238,49 @@ impl<N: HugrNode> std::fmt::Display for Wire<N> {
     }
 }
 
+/// Marks [FuncDefn](crate::ops::FuncDefn)s and [FuncDecl](crate::ops::FuncDecl)s as
+/// to whether they should be considered for linking, and as reachable (starting points)
+/// for optimization/analysis.
+#[derive(
+    Copy,
```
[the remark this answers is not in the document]
acl-cqc: Yup, done.

hugr-passes/src/dead_funcs.rs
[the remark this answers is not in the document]
acl-cqc: Ok, `remove_dead_funcs2` it is

hugr-core/src/builder/dataflow.rs
```diff
@@ -152,17 +147,44 @@ impl<B, T> DFGWrapper<B, T> {
...
-        let op = ops::FuncDefn::new(name, signature);
+        Self::new_vis(name, signature, Visibility::Private)
+    }
+
+    /// Initialize a builder for a FuncDefn-rooted HUGR; the function will be [Visibility::Public].
+    ///
+    /// # Errors
+    ///
+    /// Error in adding DFG child nodes.
+    pub fn new_pub(
```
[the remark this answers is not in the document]
acl-cqc: Removed - there is `new` and `new_vis`

hugr-core/src/ops/module.rs
```diff
@@ -57,14 +56,39 @@ pub struct FuncDefn {
     #[cfg_attr(test, proptest(strategy = "any_nonempty_string()"))]
     name: String,
     signature: PolyFuncType,
+    #[serde(default = "priv_vis")]
```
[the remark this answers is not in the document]
acl-cqc: Counterargument would be that if they continue using old/deprecated analysis methods that don't take account of visibility, it doesn't matter if we import `main` as private - and they can fix the visibility of their `main` when they switch to non-deprecated analysis.

specification/schema/hugr_schema_live.json
```diff
@@ -676,12 +676,21 @@
...
+                    "title": "Visibility",
+                    "type": "string"
                 }
             },
             "required": [
                 "parent",
                 "name",
-                "signature"
+                "signature",
+                "visibility"
```
aborgna-q, line 693: Use a default (public?) fall-back instead of requiring the new visibility field, so the format remains backwards-compatible.

aborgna-q (changes requested):

specification/schema/hugr_schema_live.json
acl-cqc: I view the defaults used when deserializing as a legacy backwards-compatibility mechanism. AFAICS serde correctly applies the defaults and there is no problem loading an "old" Hugr w/out visibility - at least outside the envelope mechanism, I'm working on a test using production/envelope-loading code.

Do you think that marking old hugrs as "not schema compliant" prevents loading them? I mean, do we require old Hugrs to be schema-compliant (not just loadable) in order to say backwards compatibility is maintained? (Do we really need the schema to encompass all supported legacy formats?) Your answer suggests "yes" to at least the last two questions here, is that really what you mean?

aborgna-q: The fields must be marked as default-if-not-present to be correctly loaded, no?
- For pydantic, [add `visibility = Field(default=...)`](https://docs.pydantic.dev/2.0/usage/fields/#default-values) on the definition
- For serde, [add `#[serde(default)]`](https://serde.rs/attr-default.html) on the field
- You added a fallback on `hugr-model` already, so that should work fine

acl-cqc: Added a test using `Hugr::load` of a FuncDecl and a FuncDefn without any visibility. Defaults correctly applied....?

aborgna-q: Ugh, it looks like that's [undocumented serde behaviour](https://www.github.com/serde-rs/serde/issues/2214#issuecomment-1627645095).
It doesn't hurt to be explicit though...

I tested a toy example for pydantic, and I'm pretty sure that we require the default there.
- ```py
    ValidationError: 1 validation error for Test
    b
      Field required [type=missing, input_value={'a': 1}, input_type=dict]
        For further information visit https://errors.pydantic.dev/2.11/v/missing
   ```
- https://docs.pydantic.dev/latest/errors/validation_errors/#missing

aborgna-q (approved):

PR #2148, opened 2025-05-02
feat!: Entrypoints in `hugr-py`

aborgna-q:
Adds entrypoint definitions to `hugr-py`. Has similar caveats to #2147.

The main issue here was builders that initialized a hugr for a dataflow op without knowing its output types. The automatic machinery that wraps the op in a function definition needs to connect the outputs to the new function's output once we know the types, so I had to add some extra machinery to `DataflowOp` to signal back to the hugr that it should connect things.
This feels a bit hacky, perhaps there's a cleaner way to do it...

Blocked by #2147.

BREAKING CHANGE: Hugrs now define an `entrypoint` in addition to a module root.


hugr-py/src/hugr/hugr/render.py
ss2165, line 1: I assume these changes have been tested (maybe attach an image to the PR?)

hugr-py/src/hugr/hugr/base.py
```diff
@@ -80,26 +93,126 @@ class Hugr(Mapping[Node, NodeData], Generic[OpVarCov]):
...
 
-    #: Root node of the HUGR.
-    root: Node
+    # The module at the root of the HUGR.
+    module_root: Node
+    # Entrypoint node for the HUGR.
+    #
+    # Most traversals and rewrite operations start from this node.
+    #
+    # This node may be of any optype, and is a descendant of the module
```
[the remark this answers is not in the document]
aborgna-q: I was also missing the change we did on the rust side restricting entrypoints to only region parents (#2173).

I fixed that, so now the logic is a bit different.

hugr-py/src/hugr/hugr/render.py
aborgna-q: We do [test snapshots](https://github.com/CQCL/hugr/blob/f280627c7d5b5bee9f6808aa52240b5508e7501a/hugr-py/tests/conftest.py#L171-L175) of these.

Here's a render of `test_basic_cfg`. I added a little change to highlight the entrypoint region too.
hugr dot

aborgna-q: ...we should probably find a less awful color tbh

ss2165: maybe add a follow up task for "entrypoint only" rendering, should make debugging a lot easier

ss2165 (approved):

PR #2149, opened 2025-05-04
feat!: Generic SimpleReplacement over boundary maps

### LINKED ISSUE #2098: SimpleReplacement invalidates output nodes
Currently, the invalidation set of SimpleReplacement includes the nodes past the output boundaries. This is required for the invalidation check to work on replacements that act on an empty subgraph (otherwise the invalidation set is empty and thus would not conflict with a replacement that encloses the empty subgraph - wrong!). As far as I can see, this is the _only_ reason for adding those output nodes to the invalidation set. Or am I missing something?

## Issues with this

- this is a conservative overapproximation of the invalidation set and thus excludes valid combinations of replacements
- it means the new set of nodes after replacement isn't "old nodes - invalidated nodes + new nodes". This would be a very useful formula to have when computing replacements that stack on top of previous ones (PersistentHugr)
- for me this "hack" smells of the fact that we're thinking about it the wrong way.

## Suggested fix?

I _think_ this problem goes away if the invalidation set is instead given by wires instead of nodes. This needs to be thought through better though and an API for this isn't immediately obvious (I think you'd need to distinguish incoming wires to the subgraph from outgoing wires?)

### LINKED ISSUE #2146: Invalid `SimpleReplacement` should fail
`SimpleReplacement` should be capable of checking whether applying a patch will result in an invalid HUGR (e.g. copy a linear wire), but it doesn't. Examples of invalid replacements that could be defined

- Missing nodes in the subject subgraph to be deleted
- Invalid boundary (e.g. re-use of same linear resource twice)

IMO, the first one is the more problematic one. In particular, I would like to rely on the following invariant

> If a wire is in the output boundary of the subgraph being replaced, then the source of the wire is in the node deletion set.

The same invariant also applies at the input boundary for linear values. This is not enforced at the moment. An example of this issue is given below

### Example

Consider the following subject HUGR:

```mermaid
graph LR
    subgraph 0 ["(0) DFG"]
        direction LR
        1["(1) Input"]
        2["(2) Output"]
        3["(3) test.quantum.H"]
        1--"0:0<br>qubit"-->2
        1--"1:0<br>qubit"-->3
        3--"0:1<br>qubit"-->2
    end
```

and a `SimpleReplacement` given by the replacement HUGR

```mermaid
graph LR
    subgraph 0 ["(0) DFG"]
        direction LR
        1["(1) Input"]
        2["(2) Output"]
        3["(3) test.quantum.QDiscard"]
        4["(4) test.quantum.H"]
        1--"0:0<br>qubit"-->3
        1--"1:0<br>qubit"-->4
        4--"0:0<br>qubit"-->2
    end
```

The result of applying the replacement on the entire subject is then

```mermaid
graph LR
    subgraph 0 ["(0) DFG"]
        direction LR
        1["(1) Input"]
        2["(2) Output"]
        7["(7) test.quantum.QDiscard"]
        8["(8) test.quantum.H"]
        1--"0:0<br>qubit"-->2
        1--"0:0<br>qubit"-->7
        1--"1:0<br>qubit"-->8
        8--"0:1<br>qubit"-->2
    end
```

which is obviously invalid, but `SimpleReplacement` will not complain.

#### Code

```rust
    #[test]
    fn test_invalid_replacement_copy_qubit() {
        // Create a DFG with a boolean wire and a qubit wire
        let (mut subject, subject_h_gate) = {
            let mut builder = DFGBuilder::new(Signature::new_endo(vec![qb_t(), qb_t()])).unwrap();
            let [wire1, wire2] = builder.input_wires_arr();
            let h_gate = builder.add_dataflow_op(h_gate(), vec![wire2]).unwrap();
            (
                builder
                    .finish_hugr_with_outputs([wire1, h_gate.out_wire(0)])
                    .unwrap(),
                h_gate.node(),
            )
        };

        // Create a replacement DFG that applies a QDiscard to the first qubit and a Hadamard to the second
        let (repl, repl_h_gate, repl_discard_gate) = {
            let mut repl_builder =
                DFGBuilder::new(Signature::new(vec![qb_t(), qb_t()], vec![qb_t()])).unwrap();
            let [wire1, wire2] = repl_builder.input_wires_arr();
            let discard_gate = repl_builder
                .add_dataflow_op(q_discard(), vec![wire1])
                .unwrap();
            let h_gate = repl_builder.add_dataflow_op(h_gate(), vec![wire2]).unwrap();
            (
                repl_builder
                    .finish_hugr_with_outputs([h_gate.out_wire(0)])
                    .unwrap(),
                h_gate.node(),
                discard_gate.node(),
            )
        };

        // Get the input and output nodes
        let [_input, subject_output] = subject.get_io(subject.root()).unwrap();

        // Create the subgraph (empty is invalid, so we just remove a dummy H gate)
        let subgraph = SiblingSubgraph::try_from_nodes(vec![subject_h_gate], &subject).unwrap();

        // Set up the port mappings
        let mut nu_inp = HashMap::new();
        let mut nu_out = HashMap::new();

        // Map the input wires to the QDiscard and Hadamard gates' inputs
        nu_inp.insert(
            (repl_discard_gate, IncomingPort::from(0)),
            (subject_output, IncomingPort::from(0)),
        );
        nu_inp.insert(
            (repl_h_gate, IncomingPort::from(0)),
            (subject_h_gate, IncomingPort::from(0)),
        );
        nu_out.insert(
            (subject_output, IncomingPort::from(1)),
            IncomingPort::from(0),
        );

        // Create the replacement
        let r = SimpleReplacement {
            subgraph,
            replacement: repl,
            nu_inp,
            nu_out,
        };

        r.apply(&mut subject).unwrap_err(); // This should return an error
    }
```

lmondada:
Currently, `SimpleReplacement` stores its output boundary map `nu_out` by referring to nodes in the host graph that are outside of the deleted subgraph. This forces the invalidation set of the replacements to include nodes past the output, forbidding simultaneous adjacent replacements.

This PR fixes this by allowing the boundary maps to be a generic type, as long as it implements the newly introduced `BoundaryMap` trait. In particular `nu_out`, can be defined in terms of outgoing ports on nodes in the subgraph, thus reducing the required invalidation set. The current way of defining SimpleReplacement remains valid (and is the most general case), but I find using maps on outgoing ports more intuitive for most common cases.

This PR also adds validity checks on construction of `SimpleReplacement`s, which is why the PR is marked as breaking. It could be made non-breaking if we depracated `fn new` instead. Let me know if that is desired.

Closes #2098
Closes #2146

BREAKING CHANGE: Renamed `SimpleReplacement::new` to `SimpleReplacement::new_unsafe`. Consider using the new `SimpleReplacement::try_new` instead.


lmondada:
Actually, I think I can achieve what I want in a simpler way. Will try again tomorrow.

PR #2151, opened 2025-05-06
feat!: Accept outgoing ports in SimpleReplacement nu_out

lmondada:
Currently, SimpleReplacement stores its output boundary map nu_out by referring to nodes outside the deleted subgraph in the host graph. This forces the invalidation set of the replacements to include nodes past the output, forbidding simultaneous adjacent replacements.

This PR fixes this by allowing the keys of `nu_out` (i.e. the ports on the output boundary of the subgraph) to be either incoming ports (as before), or outgoing ports (in which case this is equivalent to specifying the map on all incoming ports linked to the given outgoing ports). The latter is less general but covers most use cases and reduces the size of the invalidation set.

Closes https://github.com/CQCL/hugr/issues/2098

BREAKING CHANGE: Generalised arguments to [`SimpleReplacement::new`] and [`SimpleReplacement::map_host_output`] to allow for outgoing ports. Previous type signatures remain valid, however, type inference may fail.

### DIFF
--- a/hugr-core/src/hugr/patch/simple_replace.rs
+++ b/hugr-core/src/hugr/patch/simple_replace.rs
@@ -1134,7 +1134,7 @@ pub(in crate::hugr::patch) mod test {
             .nodes()
             .find(|node: &Node| *h.get_optype(*node) == cx_gate().into())
             .unwrap();
-        let s: Vec<Node> = vec![h_node_cx].into_iter().collect();
+        let s = vec![h_node_cx];
         // 2. Construct a new DFG-rooted hugr for the replacement
         let n: Hugr = dfg_hugr2;
         // 3. Construct the input and output matchings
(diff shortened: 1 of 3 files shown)


hugr-core/src/hugr/patch/simple_replace.rs
```diff
@@ -977,6 +1122,110 @@ pub(in crate::hugr::patch) mod test {
...
+        dfg_hugr2: Hugr,
+    ) {
+        let mut h: Hugr = simple_hugr;
+
+        // 1. Locate the CX in h
+        let h_node_cx: Node = h
+            .nodes()
+            .find(|node: &Node| *h.get_optype(*node) == cx_gate().into())
+            .unwrap();
+        let s: Vec<Node> = vec![h_node_cx].into_iter().collect();
```
mark-koch, line 1137:
```suggestion
        let s = vec![h_node_cx];
```
?

mark-koch (approved):
The implementation looks good to me, but I'm missing the broader context to tell if the thing itself is a good idea 😅 A second opinion on this might be useful

lmondada:
> A second opinion on this might be useful

@aborgna-q, ok? 🙃

aborgna-q:
Looks great to me!

PR #2152, opened 2025-05-06
feat: export mangle name function

ss2165:
for use in tket2/other downstream compilation tools

### DIFF
--- a/hugr-passes/src/monomorphize.rs
+++ b/hugr-passes/src/monomorphize.rs
@@ -306,20 +306,19 @@ fn write_type_arg_str(arg: &TypeArg, f: &mut std::fmt::Formatter<'_>) -> std::fm
     }
 }
 
-/// Produce a mangled name for a monorphized instance of a function.
+/// Produce a mangled name for a monomorphized instance of a function.
 ///
-/// We do our best to generate unique names. Our strategy is to pick out '$' as
+/// Best effort to generate unique names. The strategy is to pick out '$' as
 /// a special character.
 ///
-/// We:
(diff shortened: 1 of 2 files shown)


hugr-passes/src/monomorphize.rs
```diff
@@ -306,6 +306,8 @@ fn write_type_arg_str(arg: &TypeArg, f: &mut std::fmt::Formatter<'_>) -> std::fm
     }
 }
 
+/// Produce a mangled name for a monorphized instance of a function.
```
cqc-alec, line 309:
```suggestion
/// Produce a mangled name for a monomorphized instance of a function.
```

hugr-passes/src/monomorphize.rs
```diff
@@ -318,7 +320,7 @@ fn write_type_arg_str(arg: &TypeArg, f: &mut std::fmt::Formatter<'_>) -> std::fm
 ///    simple obvious thing.
 ///  - For all TypeArg Constructors we choose a short prefix (e.g. `t` for type)
 ///    and use "t({arg})" as the string representation of that arg.
-fn mangle_name(name: &str, type_args: impl AsRef<[TypeArg]>) -> String {
+pub fn mangle_name(name: &str, type_args: impl AsRef<[TypeArg]>) -> String {
```
cqc-alec, line 323: The docstring could be improved for public consumption:
- The bullet list isn't grammatically coherent. I suggest making each bullet a sentence, punctuated as such.
- Is the first bullet actually right? It doesn't seem to match the implementation below.
- What is the "simple obvious thing"?

cqc-alec (approved):

PR #2155, opened 2025-05-06
feat!: Validate any HugrView, make errors generic

aborgna-q:
Removes the dependency on `base_hugr` when validating views.
Now the validation code runs directly on any given `HugrView`.

To make this work I had to add `<N: HugrNode>` parameters to multiple error types.
This makes sense, as now validating a non-`Node` hugr will return errors with the correct indices.

BREAKING CHANGE: Added node parameters to multiple error types, including the one returned by `HugrView::validate`.


aborgna-q:
The changes here are mostly just noise due to adding the extra node type param to the errors. There is no logic change.

zrho (approved):

PR #2164, opened 2025-05-07
feat!: Return a node mapping in HugrInternals::region_portgraph

### LINKED ISSUE #2158: `HugrInternals::region_portgraph` should return a map from/to portgraph node idxs
This will allow us to replace `to_portgraph_node` and `from_portgraph_node`, which require global indexing, with local indices.

The map should be defined as a trait, so identity maps can be generated in `O(1)`.
See [`hugr::views::ExtractionResult`](https://github.com/CQCL/hugr/blob/e61b048f61a84d43a91589b427eb9c769d5fca69/hugr-core/src/hugr/views.rs#L486) for a similar case.

aborgna-q:
Closes #2158

Annoyingly, we still need to return a `FlatRegion` (or some other concrete wrapper) here so that _references_ to the graph implements some of petgraph's traits (like `IntoNodeIdentifiers`) which are normally implemented for `&'_ G`. There's no way to write that in a GAT constraint :/
We'll have to wait for the big petgraph rewrite to clean that.

BREAKING CHANGE: `HugrInternals::region_portgraph` now also returns a node mapping between hugr and portgraph nodes

### DIFF
--- a/hugr-core/src/hugr/internal.rs
+++ b/hugr-core/src/hugr/internal.rs
@@ -46,7 +46,7 @@ pub trait HugrInternals {
     // when doing so we are unable to restrict the type to implement petgraph's
     // traits over references (e.g. `&MyGraph : IntoNodeIdentifiers`, which is
     // needed if we want to use petgraph's algorithms on the region graph).
-    // This won't be solvable until we don't do the big petgraph refactor -.-
+    // This won't be solvable until we do the big petgraph refactor -.-
     // In the meantime, just wrap the portgraph in a `FlatRegion` as needed.
     fn region_portgraph(
         &self,
--- a/hugr-core/src/hugr/internal.rs
+++ b/hugr-core/src/hugr/internal.rs
@@ -24,19 +24,37 @@ pub trait HugrInternals {
     where
         Self: 'p;
 
+    /// The portgraph graph structure returned by [`HugrInternals::region_portgraph`].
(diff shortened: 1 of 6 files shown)


mark-koch (approved):

lmondada (approved):
Looks amazing, thanks @aborgna-q!

I've tried to remove the methods
```rust
    #[inline]
    fn to_portgraph_node(&self, node: impl NodeHandle<Self::Node>) -> portgraph::NodeIndex {
        node.node().into_portgraph()
    }
    #[inline]
    fn from_portgraph_node(&self, index: portgraph::NodeIndex) -> Self::Node {
        index.into()
    }
```
in `HugrInternals` and all their uses, but I've hit a roadblock in the implementation of `HugrMut::insert_from_view`. This turns the view into a portgraph view and then inserts that.

Do you have a plan on how to do it? In any case, it seems big enough to be its own PR.

aborgna-q:
There's a TODO in `insert_hugr_internal`:
> TODO: We may need to replicate `portgraph::LinkMut::insert_graph` at the
> hugr level to avoid calling `hierarchy()` here.

https://github.com/CQCL/hugr/blob/851e3964ce556e70b8a837f66952cfe7e7f9e1c7/hugr-core/src/hugr/hugrmut.rs#L569-L571

We'll need to implement that to remove those uses :/

PR #2167, opened 2025-05-07
feat!: Only expose envelope serialization of hugrs and packages

### LINKED ISSUE #2159: Remove `impl {De,}Serialize for Hugr`
We should be using envelopes instead.

We should keep the `serialize::Versioned` definition and add some helper method so that envelopes can still use the json format, but it shouldn't be exposed so directly for `Hugr`.

We should instead have `store`/`load` methods as we do with packages.

This could be a breaking serialization change if we decide to store nested hugrs as envelopes rather than raw hugr jsons.

aborgna-q:
Closes #2159

Now `Hugr`s expose `load`/`load_str`/`store`/`store_str` methods instead of implementing `serde` traits, mirroring `Package`s. This uses envelopes interternally, so we no longer support row hugr jsons.

`hugr-cli` keeps the backwards compatibility for the moment, by simply wrapping the jsons with the appropriate header when using the `--hugr-json` flag.

For users that need it, we define a [`serde_with`](https://docs.rs/serde_with/3.12.0/serde_with/index.html) helper to be able to include hugrs and packages inside serde-encoded structs.
The implementation serializes the hugr into an envelope, and includes it as a string/bytes value. For backwards compatibility, we temporarily also try to read Hugr JSONs as a fallback.
```rust
use serde::{Deserialize, Serialize};
use serde_json::json;
use serde_with::{serde_as};
use hugr_core::Hugr;
use hugr_core::package::Package;
use hugr_core::envelope::serde_with::AsStringEnvelope;

#[serde_as]
#[derive(Deserialize, Serialize)]
struct A {
    #[serde_as(as = "AsStringEnvelope")]
    package: Package,
    #[serde_as(as = "Vec<AsStringEnvelope>")]
    hugrs: Vec<Hugr>,
}
```

BREAKING CHANGE: `Hugr` and `Package` no longer implement serde traits. Use envelopes instead. See `AsStringEnvelope` for backwards compatibility.


aborgna-q:
There's some time added to the serialization benchmarks since now we're measuring both extension encoding and resolution, but I'm not sure if that accounts for all the performance hit :/

hugr-core/src/envelope/package_json.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    let PackageDeser {
+        modules,
+        extensions: pkg_extensions,
+    } = serde_json::from_value::<PackageDeser>(val.clone())?;
+    let mut modules = modules.into_iter().map(|h| h.0).collect_vec();
+
+    // TODO: We don't currently store transitive extension dependencies in the
+    // package's extensions. For example, if we use a `collections.list` const
+    // value but don't use anything `prelude` we would not include `prelude` in
```
ss2165, line 27:
```suggestion
    // value but don't use anything in `prelude` we would not include `prelude` in
```

hugr-core/src/hugr/serialize/upgrade/testcases/empty_hugr.json
ss2165, line 1: where are these json files used? Should they be .hugr files?

resources/test/hugr-0.hugr
ss2165, line 1: I'm surprised this one didn't need wrapping?

ss2165 (commented):
just a few queries

hugr-core/src/hugr/serialize/upgrade/testcases/empty_hugr.json
aborgna-q: These are tests for the `serde` definitions, tested directly using `serde_json`:
https://github.com/CQCL/hugr/blob/7c101d9765a10e0c4302f0445736a60360b96766/hugr-core/src/hugr/serialize/upgrade/test.rs#L58-L70

resources/test/hugr-0.hugr
aborgna-q: The test just checked that we got an `Err(_)`, so it was accepting the missing-envelope-header error.

ss2165 (approved):

PR #2172, opened 2025-05-08
feat: Export and import entrypoints via metadata in `hugr-model`.

zrho:
This PR adds code to import and export the entrypoint field of modules to the `hugr-model` representation. It assumes that only regions should be made the entrypoint. The PR also fixes some bugs in the parser that prevented packages with multiple modules to be parsed correctly.


.clj-kondo/.cache/v1/lock
aborgna-q, line 1: Drop / ignore this file

hugr-core/tests/model.rs
```diff
@@ -9,6 +9,7 @@ use hugr_model::v0 as model;
 fn roundtrip(source: &str) -> String {
     let bump = model::bumpalo::Bump::new();
     let package_ast = model::ast::Package::from_str(source).unwrap();
+    println!("{:#?}", package_ast);
```
aborgna-q, line 12:
```suggestion
```

hugr-model/src/v0/ast/hugr.pest
```diff
@@ -1,11 +1,35 @@
...
-COMMENT     = _{ ";" ~ (!("\n") ~ ANY)* ~ "\n" }
-identifier  = @{ (ASCII_ALPHA | "_" | "-") ~ (ASCII_ALPHANUMERIC | "_" | "-")* }
-symbol_name = @{ identifier ~ ("." ~ identifier)* }
-link_name   = @{ "%" ~ (ASCII_ALPHANUMERIC | "_" | "-")* }
+WHITESPACE         = _{ " " | "\t" | "\r" | "\n" }
+COMMENT            = _{ ";" ~ (!("\n") ~ ANY)* ~ "\n" }
+identifier         = @{ identifier_initial ~ identifier_rest* }
+identifier_initial = @{ ASCII_ALPHA | "_" | "-" }
+identifier_rest    = @{ ASCII_ALPHANUMERIC | "_" | "-" }
+symbol_name        = @{ !(reserved ~ !(identifier_rest | ".")) ~ (identifier ~ ("." ~ identifier)*) }
```
aborgna-q, line 6: Are we fine with allowing things like `mod.foo` even if `mod` is reserved?

aborgna-q (approved):
Nice!

.clj-kondo/.cache/v1/lock
zrho: Ah, I missed one! I had changed the way I managed my dotfiles which messed up the global gitignore, leading to this PR originally having thousands of spurious lines in irrelevant files...

hugr-model/src/v0/ast/hugr.pest
zrho: We currently are and I added this rule to make the implementation match that, but we should at some point have a good look at the rules for identifiers.

PR #2176, opened 2025-05-08
fix: Remove deleted nodes from node_map in `SimpleReplacement`

lmondada:


hugr-core/src/hugr/patch/simple_replace.rs
```diff
@@ -447,23 +447,33 @@ impl<N: HugrNode> PatchHugrMut for SimpleReplacement<N> {
...
+        for node in repl_io {
+            let node_h = node_map[&node];
+            h.remove_node(node_h);
+            node_map.remove(&node);
         }
-        // make all replacement top level children children of the parent
-        for &child in &replace_children[2..] {
+
+        // make all (remaining) replacement top level children children of the parent
+        for child in h.children(new_entrypoint).collect_vec() {
```
aborgna-q, line 470:
```suggestion
        for child in h.children(new_entrypoint) {
```

aborgna-q (approved):

hugr-core/src/hugr/patch/simple_replace.rs
lmondada: no, I've had to `collect` because `h` is mutated in the loop body

PR #2180, opened 2025-05-09
refactor!: Removed global portgraph-related methods from `HugrInternals`

### LINKED ISSUE #1926: Replace `HugrInternals::base_hugr` with simpler methods

### LINKED ISSUE #2141: Remove `HugrInternals::hierarchy`?
After removing some default impls, we now only use `hierarchy()` for
- Verification
- Some extra cases that'll get dropped with #2029 

#1 is an issue I'll tackle in a different PR (it's also one of the last callers of `base_hugr`, so it needs some fixing), and the other one shouldn't be an issue.

Should I maybe just remove the method? (probs do the same as with `base_hugr` and deprecate it for now)

_Originally posted by @aborgna-q in https://github.com/CQCL/hugr/pull/2126#discussion_r2066482687_

aborgna-q:
Removes
- `base_hugr`
- `portgraph`
- `hierarchy`
- `to_portgraph_node`
- `from_portgraph_node`
from the `HugrInternals` trait.

There were some last uses of these methods that I had to replace, but it should all be mostly transparent. Main changes
- Added a `HugrView` type parameter to `ComposablePass` (required #2179)
- Added some logic to pick a parent for the convex checker in `SiblingSubgraph`
- `impl<H> pv::NodeIndexable for PetgraphWrapper<H>` now has some bounds on the node type of `H`. These reflect what was currently allowed, but we may want to add a new trait for `Node ⇔ usize` at some point.

Closes #1926
Closes #2141
Depends on #2179

BREAKING CHANGE: Removed global portgraph-related methods from `HugrInternals`


hugr-core/src/hugr.rs
```diff
@@ -307,6 +308,25 @@ impl Hugr {
...
+    pub(crate) fn to_portgraph_node(&self, node: impl NodeHandle<Node>) -> portgraph::NodeIndex {
+        node.node().into_portgraph()
+    }
+
+    /// Translate a node index in [`Hugr::portgraph`] into a HUGR node.
+    #[allow(clippy::wrong_self_convention)]
+    #[inline]
+    pub(crate) fn from_portgraph_node(&self, index: portgraph::NodeIndex) -> Node {
+        index.into()
+    }
```
lmondada, line 329: I don't think these Node <-> NodeIndex conversions belong into `impl Hugr`. If I remember correctly, we added them here as a "hack" when the `Self::Node` AT was added to `HugrView`. Now might be a good time to remove them?

hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -217,7 +222,20 @@ impl<N: HugrNode> SiblingSubgraph<N> {
...
         hugr: &impl HugrView<Node = N>,
     ) -> Result<Self, InvalidSubgraph<N>> {
-        let checker = TopoConvexChecker::new(hugr);
+        let nodes = nodes.into();
+        let Some(node) = nodes.first() else {
+            return Err(InvalidSubgraph::EmptySubgraph);
+        };
+        let parent = hugr
+            .get_parent(*node)
+            .ok_or(InvalidSubgraph::NoSharedParent {
```
lmondada, line 231: Could you add an enum variant e.g. InvalidSubgraph::OrphanNode? Currently, the error message you'd get would be confusing (also useful in pick_parent)

lmondada (approved):

lmondada (approved):
:fire: :fire: :fire:

I've taken the liberty to push a commit to your branch, feel free to remove it if there's something wrong with it.

Issue #2185, opened 2025-05-09
Package does not store transitively required extensions


aborgna-q:
Extension definitions may require other extensions to be well-defined.

See for example the `sample_hugr` here.
https://github.com/CQCL/hugr/blob/0f78fab2205ea46591455b8270214221f41a9b53/hugr-core/src/hugr/views/tests.rs#L24-L35

The HUGR uses `prelude` and `test.quantum`.
When encoding a package with extensions we need to store the whole `test.quantum` definition, which includes operations with float inputs. For this to be valid, we need to include `arithmetic.float.types` in the package too.

PR #2186, opened 2025-05-09
test: check envelope roundtrips rather than json in `HugrView::verify`

aborgna-q:
`verify` was calling an internal serialization test helpers to check that roundtrip encoding was correct. With the latest changes, this wasn't checking the actual user-facing envelope encoding.

~This PR currently fails some checks due to #2185~

### DIFF
--- a/hugr-core/src/envelope.rs
+++ b/hugr-core/src/envelope.rs
@@ -319,14 +319,77 @@ fn encode_model<'h>(
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod test {
     use super::*;
     use cool_asserts::assert_matches;
     use rstest::rstest;
+    use std::borrow::Cow;
     use std::io::BufReader;
 
     use crate::builder::test::{multi_module_package, simple_package};
     use crate::extension::PRELUDE_REGISTRY;
+    use crate::hugr::test::check_hugr_equality;
+    use crate::std_extensions::STD_REG;
+    use crate::HugrView;
+
+    /// Returns an `ExtensionRegistry` with the extensions from both
+    /// sets. Avoids cloning if the first one already contains all
+    /// extensions from the second one.
+    fn join_extensions<'a>(
+        extensions: &'a ExtensionRegistry,
+        other: &ExtensionRegistry,
+    ) -> Cow<'a, ExtensionRegistry> {
+        if other.iter().all(|e| extensions.contains(e.name())) {
+            Cow::Borrowed(extensions)
+        } else {
+            let mut extensions = extensions.clone();
+            extensions.extend(other);
+            Cow::Owned(extensions)
+        }
+    }
+
+    /// Serialize and deserialize a HUGR into a binary envelope, and check that the
+    /// result is the same as the original.
+    ///
+    /// We do not compare the before and after `Hugr`s for
+    /// equality directly, because impls of `CustomConst` are not required to implement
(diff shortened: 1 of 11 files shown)


hugr-core/src/envelope.rs
```diff
@@ -319,14 +319,77 @@ fn encode_model<'h>(
...
+
+    /// Serialize and deserialize a HUGR into a text envelope, and check that the
+    /// result is the same as the original.
+    ///
+    /// We do not compare the before and after `Hugr`s for
+    /// equality directly, because impls of `CustomConst` are not required to implement
+    /// equality checking.
+    ///
+    /// Returns the deserialized HUGR.
+    pub(crate) fn check_hugr_text_roundtrip(hugr: &Hugr) -> Hugr {
```
ss2165, line 381: nit: both of these could call a shared implementation that takes an envelope config as function argument

hugr-core/src/hugr.rs
```diff
@@ -489,14 +489,63 @@ fn make_module_hugr(root_op: OpType, nodes: usize, ports: usize) -> Option<Hugr>
...
+    use super::*;
 
     use crate::envelope::{EnvelopeError, PackageEncodingError};
+    use crate::ops::OpaqueOp;
     use crate::test_file;
     use cool_asserts::assert_matches;
+    use portgraph::LinkView;
+
+    /// Check that two HUGRs are equivalent, up to node renumbering.
+    pub(crate) fn check_hugr_equality(lhs: &Hugr, rhs: &Hugr) {
```
ss2165, line 504: I think no changes to the logic here?

hugr-core/src/ops/custom.rs
```diff
@@ -168,7 +168,10 @@ impl From<ExtensionOp> for OpaqueOp {
 
 impl PartialEq for ExtensionOp {
     fn eq(&self, other: &Self) -> bool {
-        Arc::<OpDef>::ptr_eq(&self.def, &other.def) && self.args == other.args
+        self.args() == other.args()
```
ss2165, line 171: hmm I thought OpDef pointer equality was something we wanted to take advantage of for performance. I can see the reasoning here but are we ok with losing that here?

justfile
```diff
@@ -10,25 +10,25 @@ setup:
...
 # Run the pre-commit checks.
 check:
-    HUGR_TEST_SCHEMA=1 uv run pre-commit run --all-files
+    uv run pre-commit run --all-files
 
 # Run all the tests.
 test: test-rust test-python
 # Run all rust tests.
-test-rust:
+test-rust *TEST_ARGS:
```
ss2165, line 18: 🙏 blessed drive-by

ss2165 (commented):
Main query is about the pointer equality

hugr-core/src/hugr.rs
aborgna-q: I had to canonicalize both sides instead of only `lhs`, but the rest of the logic remains the same.

hugr-core/src/ops/custom.rs
aborgna-q: I added back the shortcircuit when the pointers are the same, but we still need to check everything else if they are not (we may have multiple instantiations of the same extensions around)

ss2165: sounds good

ss2165 (approved):

PR #2188, opened 2025-05-09
feat: Define text-model envelope formats

aborgna-q:
Adds experimental model s-expressions to the supported envelope formats

### DIFF
--- a/hugr-core/src/envelope.rs
+++ b/hugr-core/src/envelope.rs
@@ -311,8 +311,12 @@ fn decode_model_ast(
 
     let mut extension_registry = extension_registry.clone();
     if format == EnvelopeFormat::ModelTextWithExtensions {
-        let extra_extensions: Vec<Extension> =
-            serde_json::from_reader::<_, Vec<Extension>>(&mut stream)?;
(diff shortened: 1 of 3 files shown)


hugr-core/src/envelope.rs
```diff
@@ -273,6 +287,50 @@ fn decode_model(
...
 }
 
+/// Read a HUGR model text payload from a reader.
+///
+/// Parameters:
+/// - `stream`: The reader to read the envelope from.
+/// - `extension_registry`: An extension registry with additional extensions to use when
+///   decoding the HUGR, if they are not already included in the package.
+/// - `format`: The format of the payload.
+fn decode_model_ast(
```
ss2165, line 297: there's no test coverage of this function?

aborgna-q: Yeah, the tests only checked the roundtrip of the header.

I added a new envelope roundtrip test for al model kinds, for empty/single/multi-hugr packages.

And also fixed a bug with the decoding :)

ss2165 (approved):
thanks!

PR #2192, opened 2025-05-11
fix!: Use unique region_portgraph in convexity check

lmondada:
Previously, `region_portgraph` was called twice: once to construct the convexity checker (e.g. in `try_from_nodes`), and once to actually construct the subgraph (in `try_new_with_checker`). As a result, if the results from `region_portgraph` were not deterministic, the checker would not be valid and the convexity check might fail. I've had to narrow down the argument type, but there is currently only one type of convexity checker used, so this should not impact anyone.

BREAKING CHANGE: `SiblingSubgraph` only accepts `TopoConvexChecker` as convexity checker type

### DIFF
--- a/hugr-core/src/hugr/views/sibling_subgraph.rs
+++ b/hugr-core/src/hugr/views/sibling_subgraph.rs
@@ -178,13 +178,12 @@ impl<N: HugrNode> SiblingSubgraph<N> {
         inputs: IncomingPorts<N>,
         outputs: OutgoingPorts<N>,
         hugr: &H,
-        checker: &impl ConvexChecker,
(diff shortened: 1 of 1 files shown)


aborgna-q (approved):
Uh. I didn't consider that the region graph's nodes may have non-deterministic IDs but it looks alright since we always return the node map.

Maybe we should add some comment in region_portgraph's docs?

lmondada:
> Maybe we should add some comment in region_portgraph's docs?

I think it's fine if it's non-deterministic? (as long as we fix issues like this one, but there shouldn't be too many of these)

PR #2195, opened 2025-05-12
chore: Edition 2024

### LINKED ISSUE #2144: Migrate to rust's `edition2024`
This is a nice-to have for the 0.20 release, but not required as it's shouldn't be a breaking change.

aborgna-q:
Migrates the package to rust's `edition 2024`. Closes #2144 
This PR's changes come almost completely from automated tools.

It may be easier to look at them commit by commit. In order,
1. run `cargo fix edition` and remove some backwards-compatibility changes (e.g. matching on `expr_2021` instead of `expr` in macro_rules).
2. run `cargo fmt`
3. fix new lints coming in `rust 1.87`.

The last commit is the noisiest one, and totally optional:
4. `clippy --fix`es some additional non-default lints,
  - Ensuring code links in the rustdocs are formated using backticks.
  - Adding some `must_use`s
  - Replacing matches on `bool` with if/else
  - Removing unnecessary `into_iter` calls
  - etc.
  
I'll make sure to `.git-ignore-revs` this change to avoid polluting the git history.


aborgna-q:
Looks like the new clippy version releasing next thursday has a new link for result error types that are too big.
The lint fails with a lot of our error definitions -.-'
I temporarily disabled it, but we may want to address it before the release...

Additionally, while fiddling with that I realised that `hugr-passes` and `hugr-llvm` weren't using the workspace lints, and enabling them causes warnings due to missing docs 🫠

PR #2202, opened 2025-05-12
feat: PersistentHugr implements HugrView

### LINKED ISSUE #2095: Implement HugrView & Rewrite traits on `PersistentHugr`

lmondada:
Closes #2095

### DIFF
--- a/hugr-core/src/hugr/internal.rs
+++ b/hugr-core/src/hugr/internal.rs
@@ -6,6 +6,7 @@ use std::sync::OnceLock;
 use itertools::Itertools;
 use portgraph::{LinkMut, LinkView, MultiPortGraph, PortMut, PortOffset, PortView};
 
+use crate::core::HugrNode;
 use crate::extension::ExtensionRegistry;
 use crate::{Direction, Hugr, Node};
 
@@ -88,6 +89,21 @@ impl PortgraphNodeMap<Node> for DefaultPGNodeMap {
     #[inline]
     fn from_portgraph(&self, node: portgraph::NodeIndex) -> Node {
         node.into()
+    }
+}
+
+impl<N: HugrNode> PortgraphNodeMap<N> for std::collections::HashMap<N, Node> {
+    #[inline]
+    fn to_portgraph(&self, node: N) -> portgraph::NodeIndex {
+        self[&node].into_portgraph()
+    }
+
+    #[inline]
+    fn from_portgraph(&self, node: portgraph::NodeIndex) -> N {
+        let node = node.into();
+        self.iter()
+            .find_map(|(&k, &v)| (v == node).then_some(k))
+            .expect("Portgraph node not found in map")
     }
 }
 
--- a/hugr-core/src/hugr/patch/simple_replace.rs
+++ b/hugr-core/src/hugr/patch/simple_replace.rs
@@ -7,7 +7,7 @@ use crate::hugr::hugrmut::InsertionResult;
 use crate::hugr::views::SiblingSubgraph;
 use crate::hugr::{HugrMut, HugrView};
 use crate::ops::{OpTag, OpTrait, OpType};
-use crate::{Hugr, IncomingPort, Node, OutgoingPort, Port};
+use crate::{Direction, Hugr, IncomingPort, Node, OutgoingPort, Port};
 
 use derive_more::derive::From;
 use itertools::{Either, Itertools};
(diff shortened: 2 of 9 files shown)


hugr-core/src/hugr/patch/simple_replace.rs
```diff
@@ -93,6 +93,22 @@ impl<N: HugrNode> OutputBoundaryMap<N> {
         .into_iter()
     }
 
+    /// The keys of the map.
+    ///
+    /// These will be either outgoing or incoming ports, depending on the
```
acl-cqc, line 98: This makes me wonder - is this really an *Output*BoundaryMap, rather than just a BoundaryMap?

lmondada: This was deleted by the SimpleReplacement simplification PR (#2208)

hugr-core/src/hugr/internal.rs
```diff
@@ -91,6 +92,21 @@ impl PortgraphNodeMap<Node> for DefaultPGNodeMap {
...
+        self[&node].into_portgraph()
+    }
+
+    #[inline]
+    fn from_portgraph(&self, node: portgraph::NodeIndex) -> N {
+        let node = node.into();
+        self.iter()
+            .find_map(|(&k, &v)| (v == node).then_some(k))
+            .expect("Portgraph node not found in map")
+    }
```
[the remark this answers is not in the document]
lmondada: I think this is fine for the time being. We can always upgrade it to a bijective map if we want, but I suspect this will be used rather rarely -- and on portgraphs that are a single region of the HUGR, so relatively small.

We do not have `bimap` as a dependency at the moment and I don't want to add my own implementation of it until it's required.

lmondada:
I've rebased on top of main and have simplified/improved the logic to find linked incoming/outgoing ports. In doing so, I've also established a more stable API for querying the boundary edges of `SimpleReplacement`s, see https://github.com/CQCL/hugr/pull/2231

Apologies I think this will require a complete re-review. The combination of the changes + rebase makes diff-ing changes between before and now difficult. The main changes are the methods
- `PersistentHugr::get_single_outgoing_port`
- `PersistentHugr::get_all_incoming_ports`

lmondada:
Question: I've implemented a partial subset of `HugrView` in `parents_view.rs`. I think this is fine because the type is private.

The alternative would be to introduce a new trait that captures the parts of `HugrView` that I need, but this new trait would have to be public as it would be exposed through the `SimpleReplacement` API. I worry that this makes the API less readable (as the user will need to be aware of one more trait), even if we provide blanket implementations.

Let me know which approach you prefer

lmondada:
> Question: I've implemented a partial subset of HugrView in parents_view.rs. I think this is fine because the type is private.

btw this is also the reason why coverage is low. Everything else is well-covered. I could add dummy for the missing implementations to increase coverage artificially, but I don't think that's a useful thing to do...

hugr-core/src/hugr/persistent/trait_impls.rs
```diff
@@ -0,0 +1,487 @@
...
+    type RegionPortgraphNodes = HashMap<PatchNode, Node>;
+
+    fn region_portgraph(
+        &self,
+        parent: Self::Node,
+    ) -> (
+        portgraph::view::FlatRegion<'_, Self::RegionPortgraph<'_>>,
+        Self::RegionPortgraphNodes,
+    ) {
+        // TODO: this is currently not very efficient
```
aborgna-q, line 56: I imagine this would require a custom portgraph implementation.
Open an issue and link it from here.

hugr-core/src/hugr/persistent/trait_impls.rs
```diff
@@ -0,0 +1,487 @@
...
+    fn node_metadata_map(&self, node: Self::Node) -> &crate::hugr::NodeMetadataMap {
+        self.as_state_space().node_metadata_map(node)
+    }
+}
+
+// TODO: A lot of these implementations (especially the ones relating to node
+// hierarchies) are very inefficient as they (often unnecessarily) construct
+// the whole extracted HUGR in memory. We are currently prioritizing correctness
+// and clarity over performance and will optimise some of these operations in
+// the future as bottlenecks are encountered.
```
aborgna-q, line 78: :+1:

Open an issue!

hugr-core/src/hugr/persistent/trait_impls.rs
```diff
@@ -0,0 +1,487 @@
...
+impl HugrView for PersistentHugr {
+    fn entrypoint(&self) -> Self::Node {
+        // The entrypoint remains unchanged throughout the patch history, and is
+        // found in the base hugr.
+        let entry = self.base_hugr().entrypoint();
+        let node = PatchNode(self.base(), entry);
+
+        assert!(self.contains_node(node), "invalid entrypoint");
```
aborgna-q, line 86: Maybe `debug_assert`, since this only depends on the local impl? (Same with `module_root`)

aborgna-q (approved):
Awesome!
I only have some nitpicks

PR #2207, opened 2025-05-13
fix: Import and export JSON metadata on module roots.

zrho:
This PR ensures that JSON metadata on a module root is imported and exported appropriately.


aborgna-q (approved):
It'd be nice to have a test, but LGTM

PR #2208, opened 2025-05-13
feat!: Remove boundary map in SimpleReplacement

lmondada:
`SimpleReplacement` is now so much simpler!!

I've just gone ahead and removed the `nu_inp` and `nu_out` maps of `SimpleReplacement`. Makes defining replacements so much less verbose! But you will probably (rightfully!) be sceptical that this can work.

The crux is: there is already a boundary map defined in the `subgraph: SiblingSubgraph`. My claim is: **We can restrict `nu_inp` and `nu_out` to be bijective maps between the boundary as defined by `SiblingSubgraph` and the ports at the Input/Output nodes of the replacement**. And once the map is bijective, we might as well restrict it to the canonical identity map, mapping the i-th port in the subgraph to the i-th port in the replacement.

To convince yourself of that:

### nu_inp: (Node, IncomingPort) -> (HostNode, IncomingPort)

Non-injective case: that’s using a value from the host HUGR multiple times (copying). This is covered by having a `replacement` HUGR that copies a value at the input.

Non-surjective case: that’s discarding a value from host. This is covered by having a `replacement` that discards a value at the input.

### nu_out: (HostNode, IncomingPort) -> (Node, IncomingPort)

Non-injective case: that’s using a value produced by `replacement` multiple times in the host HUGR. This is replaced by copying the value in the replacement to multiple outputs

Non-surjective case: that’s discarding a value from `replacement`. Replaced by not linking that value to the output node.

### "The proof is convincing, but what about the following case?"

Consider a subgraph given by nodes `1` and `2` and consider a wire with an outgoing port in `1` and two incoming ports, one in `2` and one in `3`.

```mermaid
flowchart LR
 subgraph G["Subgraph"]
        1
        2
  end
    1 --> 2 & 3
```

The previous version of `nu_out` was able to handle this case, by specifying that the incoming port in `3` is a boundary output port, whereas the incoming port in `2` is not. The new version is not, because the output boundary is now the outgoing port in `1`. Is the proof broken? No!! In fact, whilst the maps `nu_inp` and `nu_out` can express this case, the `subgraph` field of `SimpleReplacement` **cannot** represent this subgraph (it is "non-convex", according to our definition of convexity). So this PR is in no way a regression in functionality. I will open a PR to handle this case in `SiblingSubgraph`, at which point we can update `SimpleReplacement` to handle it too.

BREAKING CHANGE: `SimpleReplacement::new` was replaced by `SimpleReplacement::try_new`. `SimpleReplacement::{map_host_output, map_replacement_input}` take and return opposite port directions.

### DIFF
--- a/hugr-core/src/hugr/patch/simple_replace.rs
+++ b/hugr-core/src/hugr/patch/simple_replace.rs
@@ -4,10 +4,11 @@ use std::collections::HashMap;
 
 use crate::core::HugrNode;
 use crate::hugr::hugrmut::InsertionResult;
-use crate::hugr::views::SiblingSubgraph;
 pub use crate::hugr::views::sibling_subgraph::InvalidReplacement;
+use crate::hugr::views::SiblingSubgraph;
 use crate::hugr::{HugrMut, HugrView};
 use crate::ops::{OpTag, OpTrait, OpType};
+use crate::types::EdgeKind;
 use crate::{Hugr, IncomingPort, Node, OutgoingPort, Port, PortIndex};
 
 use itertools::Itertools;
@@ -51,9 +52,13 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
         replacement: Hugr,
     ) -> Result<Self, InvalidReplacement> {
         let subgraph_sig = subgraph.signature(host);
-        let repl_sig = replacement
-            .inner_function_type()
-            .expect("replacement is a DFG");
+        let repl_sig =
+            replacement
+                .inner_function_type()
+                .ok_or(InvalidReplacement::InvalidDataflowGraph {
+                    node: replacement.entrypoint(),
+                    op: replacement.get_optype(replacement.entrypoint()).to_owned(),
+                })?;
         if subgraph_sig != repl_sig {
             return Err(InvalidReplacement::InvalidSignature {
                 expected: subgraph_sig,
@@ -107,10 +112,10 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
     }
 
     /// Get the input and output nodes of the replacement hugr.
-    pub fn get_replacement_io(&self) -> Result<[Node; 2], SimpleReplacementError> {
+    pub fn get_replacement_io(&self) -> [Node; 2] {
         self.replacement
             .get_io(self.replacement.entrypoint())
-            .ok_or(SimpleReplacementError::InvalidParentNode())
+            .expect("replacement is a DFG")
     }
 
     /// Get all edges that the replacement would add from outgoing ports in
@@ -144,7 +149,7 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
             });
 
         // The incoming ports at the input boundary of `replacement`
-        let [repl_inp, _] = self.get_replacement_io().expect("replacement is a DFG");
+        let [repl_inp, _] = self.get_replacement_io();
         let repl_incoming_ports = self
             .replacement
             .node_outputs(repl_inp)
@@ -202,7 +207,7 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
                 });
 
         // The outgoing ports at the output boundary of `replacement`
-        let [_, repl_out] = self.get_replacement_io().expect("replacement is a DFG");
+        let [_, repl_out] = self.get_replacement_io();
         let repl_outgoing_ports = self
             .replacement
             .node_inputs(repl_out)
@@ -257,7 +262,7 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
             HostPort<HostNode, IncomingPort>,
         ),
     > + 'a {
-        let [repl_in, repl_out] = self.get_replacement_io().expect("replacement is a DFG");
+        let [repl_in, repl_out] = self.get_replacement_io();
 
         let empty_wires = self
             .replacement
@@ -322,7 +327,7 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
             .iter()
             .position(|&node_port| node_port == (node, port))?;
         let incoming_port: IncomingPort = pos.into();
-        let [_, rep_output] = self.get_replacement_io().expect("replacement is a DFG");
+        let [_, rep_output] = self.get_replacement_io();
         Some(ReplacementPort(rep_output, incoming_port))
     }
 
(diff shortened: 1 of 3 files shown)


ss2165:
>  we might as well restrict it to the canonical identity map

this would prevent say using the same replacement with different maps for alternative wirings. Perhaps we should offer a utility that can wrap a replacement with the require re-wiring/copy/discards? Urgency of this depends on how breaking this PR is downstream.

hugr-core/src/hugr/patch/simple_replace.rs
```diff
@@ -514,16 +495,12 @@ pub enum SimpleReplacementError {
...
-) -> impl Iterator<Item = (N, IncomingPort)> + 'a {
-    match port.as_directed() {
-        Either::Left(incoming) => Either::Left(std::iter::once((node, incoming))),
-        Either::Right(outgoing) => Either::Right(hugr.linked_inputs(node, outgoing)),
-    }
-    .into_iter()
+    port: impl Into<Port>,
+) -> bool {
+    hugr.get_optype(node).port_kind(port).unwrap().is_value()
 }
```
ss2165, line 504: the unwrap is scary
```suggestion
    hugr.get_optype(node).port_kind(port).as_ref().is_some_and(EdgeKind::is_value)
}
```

hugr-core/src/hugr/patch/simple_replace.rs
```diff
@@ -1146,156 +1040,48 @@ pub(in crate::hugr::patch) mod test {
         assert_eq!(h.entry_descendants().count(), 6);
     }
 
-    #[rstest]
```
ss2165, line 1149: what's happened to these two tests

hugr-core/src/hugr/patch/simple_replace.rs
```diff
@@ -851,8 +823,8 @@ pub(in crate::hugr::patch) mod test {
         let r = SimpleReplacement {
             subgraph: SiblingSubgraph::try_from_nodes(s, &h).unwrap(),
             replacement: n,
-            nu_inp,
-            nu_out: nu_out.into(),
+            // nu_inp,
```
ss2165, line 826: remove comments (and below)

hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -733,8 +691,8 @@ fn validate_subgraph<H: HugrView>(
     }) {
         return Err(InvalidSubgraph::NotConvex);
     }
-    // Check that every outgoing port of a node in the subgraph whose target is not in the subgraph
-    // belongs to outputs.
+    // Check that every outgoing port of a node in the subgraph whose target is not
```
ss2165, line 694: nit: lots of noise from docstring re-formatting, would have been nicer as a pre-cursor

hugr-core/src/hugr/patch/simple_replace.rs
```diff
@@ -230,36 +132,55 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
...
-                        .single_linked_output(*rem_inp_node, *rem_inp_port)
-                        .unwrap();
+            .map(|in_ports| *in_ports.first().expect("non-empty input partition"))
+            .map(|(node, in_port)| {
+                host.single_linked_output(node, in_port)
+                    .expect("valid dfg wire")
+            });
+
+        // The incoming ports at the input boundary of `replacement`
+        let [repl_inp, _] = self.get_replacement_io().expect("replacement is a DFG");
```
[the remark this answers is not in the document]
lmondada: thanks!

hugr-core/src/hugr/patch/simple_replace.rs
```diff
@@ -310,55 +257,96 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
...
+                (repl_out_node == repl_in).then_some((repl_out_port, repl_in_port))
+            });
+
+        // The outgoing ports at the input boundary of `subgraph`
+        let subgraph_input_boundary = self
+            .subgraph
+            .incoming_ports()
+            .iter()
+            .map(|node_ports| {
+                let (node, port) = *node_ports.first().expect("non-empty boundary partition");
```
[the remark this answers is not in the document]
lmondada: Yep. This is already the case: it is prevented at construction time by `SiblingSubgraph` (see validate_subgraph) function in `sibling_subgraph.rs`.

hugr-core/src/hugr/patch/simple_replace.rs
lmondada: They had been added in #2151 and covered the `OutputBoundaryMap` struct (now deleted), as well as the multiple ways in which the boundary map could be defined. I've now realised that we can get rid of all of that.

hugr-core/src/hugr/patch/simple_replace.rs
lmondada: done, sorry for missing these

hugr-core/src/hugr/views/sibling_subgraph.rs
lmondada: sorry, bad github pre-commit formatting on my side, I've undone these changes, the diff should look clean now.

lmondada:
> this would prevent say using the same replacement with different maps for alternative wirings. Perhaps we should offer a utility that can wrap a replacement with the require re-wiring/copy/discards? Urgency of this depends on how breaking this PR is downstream.

I like this idea a lot. I would imagine something like:

```rust
impl Hugr {
    /// Change the function type of a DFG HUGR by reordering/copying/discarding
    /// inputs and outputs of self.
    /// 
    /// Both `new_to_old_inputs` and `old_to_new_outputs` must be bijective when
    /// restricted to the set of linear values.
    fn change_function_type(
        &self,
        new_to_old_inputs: impl Fn(OutgoingPort) -> OutgoingPort,
        old_to_new_outputs: impl Fn(IncomingPort) -> IncomingPort,
    ) {
    }
}
```

I think it makes more sense to have this in a separate PR. Currently, all instances of `SimpleReplacements` are constructed using `SiblingSubgraph::create_simple_replacement` and so no code is broken by these breaking changes (I've checked in `hugr` and `tket2`)

lmondada:
I've created:
- https://github.com/CQCL/hugr/issues/2211
- https://github.com/CQCL/hugr/issues/2212

ss2165 (approved):
🟥 🟥 🟥

PR #2209, opened 2025-05-13
feat: Export macro for hugr serde wrappers with custom extensions

aborgna-q:
The `AsStringEnvelope` wrapper works well enough, but doesn't allow defining which extension sets to use when loading the Package/Hugrs.

Unfortunately, there is no way to define custom extension registers due to the way in which `serde_with` works.
The solution here is to add a public macro that implements the required traits.

This is needed for the ECC sets definition in `tket2`, which encode hugrs containing tket2-extension operations inside a serde struct.

### DIFF
--- a/hugr-core/src/hugr/serialize.rs
+++ b/hugr-core/src/hugr/serialize.rs
@@ -166,6 +166,21 @@ impl Hugr {
     }
 }
 
+/// Deerialize the HUGR using a serde decoder.
+///
+/// This API is unstable API and will be removed in the future.
+#[deprecated(
+    since = "0.20.0",
+    note = "This API is unstable and will be removed in the future.
+            Use `Hugr::load` or the `AsStringEnvelope` adaptor instead."
+)]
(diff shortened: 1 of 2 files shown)


hugr-core/src/hugr/serialize.rs
```diff
@@ -166,6 +166,21 @@ impl Hugr {
...
 
+/// Deerialize the HUGR using a serde decoder.
+///
+/// This API is unstable API and will be removed in the future.
+#[deprecated(
+    since = "0.20.0",
+    note = "This API is unstable and will be removed in the future.
+            Use `Hugr::load` or the `AsStringEnvelope` adaptor instead."
+)]
+pub fn serde_deserialize_hugr<'de, D>(deserializer: D) -> Result<Hugr, D::Error>
```
croyzor, line 177: IIUC this is a new deprecated method to replace the old `Hugr::serde_deserialize` - my concern is that the discoverability of this relative to using the old method isn't high enough to justify it

croyzor (approved):

PR #2216, opened 2025-05-14
test: Check hugr json serializations against the schema (again)

aborgna-q:
Removed in #2186

### DIFF
--- a/justfile
+++ b/justfile
@@ -10,7 +10,7 @@ setup:
 
 # Run the pre-commit checks.
 check:
-    uv run pre-commit run --all-files
+    HUGR_TEST_SCHEMA=1 uv run pre-commit run --all-files
 
 # Run all the tests.
 test: test-rust test-python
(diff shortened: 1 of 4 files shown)


justfile
```diff
@@ -10,7 +10,7 @@ setup:
 
 # Run the pre-commit checks.
 check:
-    uv run pre-commit run --all-files
+    HUGR_TEST_SCHEMA=1 uv run pre-commit run --all-files
```
acl-cqc, line 13: This feels quite heavyweight for a pre-commit, but I don't really object

acl-cqc (commented):
Thanks @aborgna-q - Lemme just check why the roundtrip schema validation wasn't catching my change, because this looks like the same change...

justfile
aborgna-q: This is just the rule ran on `just check`. `pre-commit` has it's own config that doesn't set the envvar.

aborgna-q:
> So whilst I am minded to approve this PR as is (that is - this PR doesn't look like it introduces too much overhead to me!)

So, should we merge it?

acl-cqc:
> So, should we merge it?

I think we should understand why these kind of checks don't catch the malformed json's mentioned above; if these aren't schema violations, why not / what are they?

Thinking we have schema checks when actually they don't catch what look to me like schema errors, could well be worse than *knowing* we don't have schema checks....

acl-cqc (approved):
Yup, looks good, thanks Agustin!

PR #2224, opened 2025-05-15
feat: Add `Type::used_extensions`

aborgna-q:
We already have `OpType::used_extensions` and `Signature::used_extensions`.

This PR exposes the method for `Type` too.

### DIFF
--- a/hugr-core/src/extension/resolution.rs
+++ b/hugr-core/src/extension/resolution.rs
@@ -23,7 +23,7 @@ mod weak_registry;
 pub use weak_registry::WeakExtensionRegistry;
 
 pub(crate) use ops::{collect_op_extension, resolve_op_extensions};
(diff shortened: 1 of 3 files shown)


hugr-core/src/extension/resolution.rs
```diff
@@ -212,6 +213,18 @@ pub enum ExtensionCollectionError<N: HugrNode = Node> {
         /// The missing extensions.
         missing_extensions: Vec<ExtensionId>,
     },
+    /// A signature requires an extension that is not in the given registry.
+    #[display(
+        "Type {typ} contains custom types for which have lost the reference to their defining extensions. Dropped extensions: {}",
```
cqc-alec, line 218:
```suggestion
        "Type {typ} contains custom types which have lost the reference to their defining extensions. Dropped extensions: {}",
```
?

cqc-alec (approved):

Issue #2212, opened 2025-05-13
More general subgraphs in `SiblingSubgraph`


lmondada:
`SiblingSubgraph` is currently limited to convex nodes-induced HUGR subgraphs.

Other than obviously non-convex cases, it also prohibits subgraphs such as:

A subgraph given by nodes `1` and `2` and a wire with
- an outgoing port in `1` and
- two incoming ports, one in `2` and one in `3`.

The outgoing port of `1` should be part of the subgraph (along with the node `2` and its incoming port), however, node `3` and its incoming port should not be in the subgraph:

```mermaid
flowchart LR
 subgraph G["Subgraph"]
        1
        2
  end
    1 --> 2 & 3
```

PR #2231, opened 2025-05-16
feat: Add boundary edge traversal in SimpleReplacement

lmondada:
I believe these methods finally expose the "right" way to think of the boundary between a replacement and its subgraph, in particular this won't be broken as we extend `SiblingSubgraph` to express more types of subgraphs (see issue #2212).

The idea is simple: for any replacement, there must be links between ports that are just outside the `subgraph` in host, and ports that are linked to the input and output nodes of the `replacement`. These links are precisely the edges that will need to be added when the replacement is applied, between the existing host HUGR and the newly added nodes of the `replacement`. Specifically these edges link:
- outgoing ports in `host` that are "just before the subgraph" (i.e. ports linked to an incoming port of `subgraph`) to incoming ports in `replacement` that are linked to the input node, and
- incoming ports in `host` that are "just after the subgraph" (i.e. ports linked to an outgoing port of `subgraph`) to outgoing ports in `replacement` that are linked to the output node.

### DIFF
--- a/hugr-core/src/hugr/patch/port_types.rs
+++ b/hugr-core/src/hugr/patch/port_types.rs
@@ -10,7 +10,7 @@ use derive_more::From;
 ///
 /// This is used to represent boundary edges that will be added between the host
 /// and replacement graphs when applying a rewrite.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum BoundaryPort<HostNode, P> {
     /// A port in the host graph.
     Host(HostNode, P),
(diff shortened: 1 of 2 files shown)


hugr-core/src/hugr/patch/port_types.rs
```diff
@@ -8,9 +8,9 @@ use derive_more::From;
 
 /// A port in either the host or replacement graph.
 ///
-/// This is used to represent boundary edges that will be added between the host and
-/// replacement graphs when applying a rewrite.
-#[derive(Debug, Clone, Copy)]
+/// This is used to represent boundary edges that will be added between the host
+/// and replacement graphs when applying a rewrite.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
```
aborgna-q, line 13: Add `Hash` too ? (same on `HostPort`)

aborgna-q (approved):
👍

PR #2232, opened 2025-05-16
feat: add HierarchyTester for O(1) subtree-containment; rewrite validation

acl-cqc:
I thought it was worth reifying HierarchyTester as there are more uses upcoming and it seems a good thing to have. I'm not sure if there are other places already existing where we could use it...?

However I admit the validation rewrite is not as compelling as I had hoped, due to the requirement for `containing_child`
* The `containing_child` for Ext edges, I expect to eventually disappear, as I think we are planning to remove the requirement for these Order edges (some other utility will compute the necessary topsort order that puts sources before nodes containing their targets)
* The `containing_child` for Dom edges could be removed by computing the same start/stop indices as used in HierarchyTester, but *taking dominance into account* i.e. by traversing the dominator tree. No reason we can't do this now, but I haven't done in in this PR.

### DIFF
--- a/hugr-core/src/hierarchy.rs
+++ b/hugr-core/src/hierarchy.rs
@@ -13,13 +13,11 @@ type NodeData<N> = (usize, usize, Option<N>);
 
 /// Caches enough information on the hierarchy of an immutably-held Hugr
 /// to allow efficient querying of [Self::is_ancestor_of] and [Self::nearest_enclosing_funcdefn].
-/// Also supports [Self::which_child_contains] (less efficiently).
 #[derive(Clone, Debug)]
 pub struct HierarchyTester<'a, H: HugrView> {
-    /// This both allows us to access the Hugr, but also guarantees the Hugr isn't
-    /// changing beneath our back to invalidate the results.
+    #[allow(unused)] // Just make sure the Hugr isn't changing behind our back
     hugr: &'a H,
-    entry_exit: HashMap<H::Node, NodeData<H::Node>> // for every node beneath entrypoint
+    entry_exit: HashMap<H::Node, NodeData<H::Node>>, // for every node beneath entrypoint
 }
 
 impl<'a, H: HugrView> HierarchyTester<'a, H> {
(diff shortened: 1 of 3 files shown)


hugr-core/src/hugr/validate.rs
```diff
@@ -40,7 +41,11 @@ impl<'a, H: HugrView> ValidationContext<'a, H> {
     /// Create a new validation context.
     pub fn new(hugr: &'a H) -> Self {
         let dominators = HashMap::new();
-        Self { hugr, dominators }
+        Self {
+            hugr,
+            hierarchy_tester: None,
```
acl-cqc, line 46: We could also just compute this now, rather than if-needed. Whereas dominators are only needed for `Dom` edges, the hierarchy tester is needed for both `Dom` and `Ext` edges, so more likely. (One might argue that any Hugr that doesn't need it, will probably be so small that computing it is insignificant??)

hugr-core/src/hierarchy.rs
```diff
@@ -0,0 +1,150 @@
...
+/// The entry and exit indices (inclusive, note every entry number is different);
+/// and the nearest strictly-enclosing FuncDefn.
+type NodeData<N> = (usize, usize, Option<N>);
+
+/// Caches enough information on the hierarchy of an immutably-held Hugr
+/// to allow efficient querying of [Self::is_ancestor_of] and [Self::nearest_enclosing_funcdefn].
+#[derive(Clone, Debug)]
+pub struct HierarchyTester<'a, H: HugrView> {
+    #[allow(unused)] // Just make sure the Hugr isn't changing behind our back
+    hugr: &'a H,
```
aborgna-q, line 19: What if we _want_ to mutate the hugr though? Not all changes affect the hierarchy.
(See e.g. petgraph's `Topo` iterator)

hugr-core/src/hierarchy.rs
```diff
@@ -0,0 +1,150 @@
...
+
+use std::collections::{HashMap, hash_map::Entry};
+
+use itertools::Itertools;
+
+use crate::HugrView;
+
+/// The entry and exit indices (inclusive, note every entry number is different);
+/// and the nearest strictly-enclosing FuncDefn.
+type NodeData<N> = (usize, usize, Option<N>);
```
aborgna-q, line 12: This should be a struct with named fields.

hugr-core/src/hierarchy.rs
aborgna-q, line 1: An alternative implementation (maybe a 2nd definition) is lazily computing these values.

See portgraph's [`Region::is_descendant`](https://github.com/CQCL/portgraph/blob/f27bc79300cd56c8e5b0b44638602b24f844e28d/src/view/region.rs#L46-L49), which has an amortized `O(1)` complexity too, and here would let us add nodes after the fact.

acl-cqc: The issue here is that I don't see a way to update it incrementally. (I presume you don't mean, invalidate the cache *in case of mutation* and then update? If we want to detect all mutation and completely invalidate the cache, we could do that.)

This is a bit like the info in `Region::is_descendant` but with the disadvantage that (AFAICS) it has to be computed completely for all nodes, and the advantage that it gives ancestry information for *all subtrees*.

The only thing I can see is to record subtree size (which is, nonetheless, a complete traversal), I think this would let us compute *dominators* lazily - an extension, but if we made `is_descendant` also class as a descendant any *dominated* BB sibling. I think this could be made to work, on the basis that re-traversing the control flow subgraph in a different order (i.e. respecting dominance) would not affect the traversal indices of any node outside the CFG-subtree. Even that could be quite tricky to do at the right time to avoid surprising changes in `is_descendant` results.

Issue #1906, opened 2025-02-04
Remove extension sets


ss2165:
### Discussed in https://github.com/CQCL/hugr/discussions/1894

<div type='discussions-op-text'>

<sup>Originally posted by **zrho** January 27, 2025</sup>
Extension sets add a non-trivial amount of complexity that is pervasive throughout many parts of the compiler. While I understand the motivation, I am not convinced that they carry their weight. I propose that we consider simplifying the compiler by removing extension sets.</div>

acl-cqc:
So the idea here is that if an op "requires runtime support", instead you'll feed in an extra input of a custom type that means "the extension is present" (0-bits at runtime).

This is a bit like the order/effect types of https://github.com/quantinuum-dev/hugrverse/issues/116 BUT
* can be freely copied (? probably?) as we don't want to totally-order every operation using an extension. (We could use copy/merge but that's a PITA). Hence, probably also discarded, and is an input to each op but not an output.
* cannot be returned through an Output node. (This is necessary for type-checking to guarantee the properties we want!)

There is a big downside though.
* It'll be annoying to build graphs like this...I mean, this is an IR, we can live with some ugliness, but
* It'll be particularly annoying to turn on/off (like we do with the `extension inference` feature ATM), because adding/not-adding all those ports will affect indexing etc. all over the place
   * And transforming ops (adding/removing such inputs) will be more difficult too.....one goal of an IR being, to ease transformation.....

acl-cqc:
....but metadata cannot be applied to arbitrary points in a type tree, only to nodes. This might be something that we can handle in `hugr-model` (@zrho ?) - in which case, if we remove extension sets now, we probably can't bring them back until HugrV2 arrives. (That might be ok...)

Issue #1225, opened 2024-06-25
In serialisation we do not distinguish between unknown `CustomConst` serialised value and invalid but known `CustomConst` serialised values


doug-q:
We use the `typetag` crate to manage serialisation and deserialisation of `dyn CustomConst` values. When deserialisation of a `dyn CustomConst` fails, we are not able to distinguish between:
1) The `impl CustomConst` is not linked into the current program, so the typetag handler for that type is not present
2) The serialisation of the known `impl CustomConst` was invalid. (e.g. see https://github.com/CQCL/hugr/issues/1091)

Currently we do not fail to deserialise, and store a `CustomSerialized` in the Hugr.

We should fail to deserialise in the 2) case, and continue to succeed with a `CustomSerialized` in the 1) case. 

`typetag` does not offer any facility to distinguish these cases. I see three options:
* Attempt to parse the deserialisation error message.
* Attempt to contribute this feature to `typetag`. It's unclear whether it would be accepted.
* Write (proc-macro) wrappers for `typetag::serde` (e.g. `hugr::custom_const::serde`) that add this feature. This could be done by registering an `inventory` of type names (as typetag does) in parallel, and and inspecting this during serialisation, and failing in a detectable way if deserialisation of the content failed when the tag is present in our inventory.
* abandon typetag. Extensions declare their Consts. add `Serialize + for<'a> Deserialize<'a>` to super traits of `CustomConst`. Thread `&ExtensionRegistry` through where it is needed. remove `#[cfg(not(miri))]` on tests that typetag breaks.

ss2165:
may be addressed by #2296

PR #2233, opened 2025-05-19
fix(py): deprecate extensions sets in values

ss2165:
This was missed in #1906 

It was detected downstream when tket2 generated HUGRs could not be read in.
We need round trip tests that go Python -> serialised -> rust -> deserialized -> Python, might need a special binding in the validation code to do this.

### DIFF
--- a/hugr-py/src/hugr/_serialization/ops.py
+++ b/hugr-py/src/hugr/_serialization/ops.py
@@ -16,7 +16,6 @@ from . import tys as stys
 from .tys import (
     ConfiguredBaseModel,
     ExtensionId,
-    ExtensionSet,
     FunctionType,
     PolyFuncType,
     SumType,
@@ -109,7 +108,6 @@ class CustomValue(BaseValue):
     """An extension constant value, that can check it is of a given [CustomType]."""
 
     v: Literal["Extension"] = Field(default="Extension", title="ValueTag")
-    extensions: ExtensionSet
     typ: Type
     value: CustomConst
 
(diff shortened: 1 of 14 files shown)


hugr-py/src/hugr/_serialization/ops.py
```diff
@@ -109,7 +108,6 @@ class CustomValue(BaseValue):
     """An extension constant value, that can check it is of a given [CustomType]."""
 
     v: Literal["Extension"] = Field(default="Extension", title="ValueTag")
-    extensions: ExtensionSet
```
aborgna-q, line 112: I'm surprised to see that there's no reference to the defining extension in the encoded `Values`, but I guess it's fine as long as all possible values are pre-defined (either via `typetag` in rust, or with the explicit literals here).

hugr-py/tests/serialization/test_extension.py
```diff
@@ -98,7 +98,6 @@ def test_extension():
         version=SemanticVersion(0, 1, 0),
         name="ext",
         types={"foo": type_def},
-        values={},
```
croyzor, line 101: This change I don't understand - it seems out of place?

ss2165: yeah it's a drive by, values in extensions were also removed but this test wasn't updated

hugr-py/src/hugr/_serialization/ops.py
ss2165: Values point to types and types point to extensions. We can have many different encodings of values for the same type, and indeed this relies on a static definition being able to resolve the value, which is not great. Related: https://github.com/CQCL/hugr/issues/1225.

croyzor (approved):

PR #2238, opened 2025-05-20
fix: handle order port case for missing dataflow ops

### LINKED ISSUE #1968: `Hugr.port_kind` fails for state-order ports

ss2165:
Fixes #1968

drive-by `Call` now inherits from `DataflowOp`

### DIFF
--- a/hugr-py/src/hugr/ops.py
+++ b/hugr-py/src/hugr/ops.py
@@ -79,17 +79,6 @@ class Op(Protocol):
         return str(self)
 
 
-def _sig_port_type(sig: tys.FunctionType, port: InPort | OutPort) -> tys.Type:
-    """Get the type of the given dataflow port given the signature of the operation."""
-    if port.offset == -1:
-        # Order port
-        msg = "Order port has no type."
-        raise ValueError(msg)
-    if port.direction == Direction.INCOMING:
-        return sig.input[port.offset]
-    return sig.output[port.offset]
-
-
 @runtime_checkable
 class DataflowOp(Op, Protocol):
     """Abstract dataflow operation. Can be assumed to have a signature and Value-
@@ -142,7 +131,17 @@ class DataflowOp(Op, Protocol):
             Bool
 
         """
-        return _sig_port_type(self.outer_signature(), port)
+        sig = self.outer_signature()
+        if port.offset == -1:
+            # Order port
+            msg = "Order port has no type."
+            raise ValueError(msg)
+        try:
+            if port.direction == Direction.INCOMING:
+                return sig.input[port.offset]
+            return sig.output[port.offset]
+        except IndexError as e:
+            raise self._invalid_port(port) from e
 
     def __call__(self, *args) -> Command:
         """Calling with incoming :class:`Wire` arguments returns a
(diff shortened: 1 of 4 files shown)


hugr-py/tests/test_hugr_build.py
```diff
@@ -299,17 +299,18 @@ def test_invalid_recursive_function() -> None:
         f_recursive.set_outputs(f_recursive.input_node[0])
 
 
-def test_higher_order() -> None:
+def test_higher_order(snapshot) -> None:
```
ss2165, line 302: issue was originally found as a rendering failure so including that here to make sure rendering now works

hugr-py/src/hugr/ops.py
```diff
@@ -1317,6 +1319,8 @@ def port_kind(self, port: InPort | OutPort) -> tys.Kind:
         match port:
             case InPort(_, offset) if offset == self._function_port_offset():
                 return tys.FunctionKind(self.signature)
+            case InPort(_, -1) | OutPort(_, -1):
```
acl-cqc, line 1322: So for example, can this `def port_kind` check only the first case, and then forward to `super().port_kind` (or `DataflowOp.port_kind) for all the others?

acl-cqc (commented):
Thanks @ss2165! :). I can see this works but I have to wonder whether we need four places handling -1 = Order. (The existing code in DataflowOp for this case having been overridden by LoadFunc/LoadConst, hence the bug.)

Is there a possibility of making `LoadFunc` and `LoadConst` defer to `DataflowOp.port_kind` for cases that aren't specific to those, i.e. so they don't have to handle order ports themselves?

I'm a little puzzled why `Call` is not a `DataflowOp` but `LoadFunc` and `LoadConst` both are...that is a by-the-way however.

hugr-py/src/hugr/ops.py
ss2165: So we can in this case, but not in the case of the `Load` functions because they also assert that other ports are invalid.

hugr-py/tests/test_hugr_build.py
```diff
@@ -299,17 +299,18 @@ def test_invalid_recursive_function() -> None:
...
     noop_fn = Dfg(tys.Qubit)
     noop_fn.set_outputs(noop_fn.add(ops.Noop()(noop_fn.input_node[0])))
 
     d = Dfg(tys.Qubit)
     (q,) = d.inputs()
     f_val = d.load(val.Function(noop_fn.hugr))
     call = d.add(ops.CallIndirect()(f_val, q))[0]
+    d.add_state_order(f_val, call.node)
```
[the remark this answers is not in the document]
ss2165: done

hugr-py/src/hugr/ops.py
```diff
@@ -975,6 +975,8 @@ def port_kind(self, port: InPort | OutPort) -> tys.Kind:
                 return tys.ConstKind(self.type_)
             case OutPort(_, 0):
```
acl-cqc, line 976:
```suggestion
            case _:
               return DataflowOp.port_kind(self, port)
```

hugr-py/src/hugr/ops.py
```diff
@@ -1410,6 +1418,8 @@ def port_kind(self, port: InPort | OutPort) -> tys.Kind:
                 return tys.FunctionKind(self.signature)
             case OutPort(_, 0):
                 return tys.ValueKind(self.instantiation)
+            case InPort(_, -1) | OutPort(_, -1):
```
acl-cqc, line 1421: Agreed this one is harder / rather less of a win, but can't you do:
```
match port:
   case InPort(_, 0):
      return tys.FunctionKind(self.signature)
  case _:
      return DataflowOp.port_kind(self, port)

hugr-py/src/hugr/ops.py
ss2165: I prefer to leave this explicit rather than re-direct to a function that doesn't know that we've already narrowed down to a particular case, it is more readable

hugr-py/src/hugr/ops.py
ss2165: yes you're right, I just needed to make sure the invalid port error is raised correctly for `DataflowOp`, now done.

Issue #2211, opened 2025-05-13
Add `change_function_type` for HugrView DFGs


lmondada:
`SiblingSubgraph::create_simple_replacement` as well as `SimpleReplacement::try_new` expect the type signature of the HUGR subgraph to match the type signature of the replacement DFG HUGR.

It would be useful to have a utility function that can reorder/copy/discard inputs and outputs of DFG graphs so that replacement graphs can be adjusted to match the subgraph signature.

Something like:

```rust
impl HugrView {
    /// Change the function type of a DFG HUGR by reordering/copying/discarding
    /// inputs and outputs of self.
    /// 
    /// Both `new_to_old_inputs` and `old_to_new_outputs` must be bijective when
    /// restricted to the set of linear values.
    fn change_function_type(
        &self,
        new_to_old_inputs: impl Fn(OutgoingPort) -> OutgoingPort,
        old_to_new_outputs: impl Fn(IncomingPort) -> IncomingPort,
    ) -> Hugr {
    }
}
```

PR #2239, opened 2025-05-20
feat: Add signature map function for DFGs

lmondada:
Closes https://github.com/CQCL/hugr/issues/2211


cqc-alec (approved):
This looks good. Apart from some trivial changes to error messages, the only request I have is for a couple more tests: one to demonstrate "recursively updating the signatures of all ancestors of the entrypoint"; and one where the input and output wires are each permuted in a non-self-inverse manner, e.g. a cyclic permutation of order 3.

lmondada:
Thanks for the review @cqc-alec, I've added the two tests as you requested.

PR #2251, opened 2025-05-23
refactor: tidies to PersistentHugr

acl-cqc:
I mistakenly made these as comments on #2168 because GitHub's "show changes from last review" includes changes merged in from main - the code I was commenting on actually went into main as part of #2202 . So, nothing major, but separating out.


lmondada (approved):
Thanks Alan!

Issue #2254, opened 2025-05-26
Dataflow analysis does not deal with incoming nonlocal edges


acl-cqc:
If you pass a Hugr with a non-root entrypoint, and there are edges from outside the entrypoint, we should report `Top` for those edges (if they are value edges - Constant/Function edges will need similar-but-different handling). Currently you'll get `Bottom`, which could lead to wrong (unsound) analysis results.

PR #2255, opened 2025-05-26
fix: Dataflow analysis produces unsound results on Hugrs with entrypoint

acl-cqc:
The bigger issue here is #2254, which this PR *does not solve*; it is a workaround.

However, it at least avoids unsoundness, by running the analysis on *all* nodes in the Hugr, rather than only those beneath the entrypoint. While this loses the ability to run analysis *faster* by restricting to within entry_descendants (that would be the full solution to #2254, not done here), it still
* allows using the entrypoint to identify the starting point to which inputs are fed in [`Machine::run`]
* provdes *more accurate results* than if we ran only within the subtree
* reduces memory after the analysis is complete (by returning only filtered results)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,7 +1,7 @@
 # for syntax see https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax
 
 # base rule, hugrverse team is codeowner for whole repo
-* @CQCL/hugrverse
+* @CQCL/hugr
 
 # Add specific rules below.
 # Lower the rule, higher the precedence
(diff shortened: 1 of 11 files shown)


hugr-passes/src/dataflow/test.rs
```diff
@@ -467,16 +466,22 @@ fn test_region() {
         Some(pv_false())
     );
 
-    // Do not provide a value on the second input (constant false in the whole hugr, above)
```
acl-cqc, line 470: This was not testing what was intended: not passing the second input, just makes `Machine::run` / `Machine::prepopulate_inputs` provide `Top` on the missing input, so there is no connection with (or reliance on the edge from) the graph outside the entrypoint. Hence, replacing the second input with a nonlocal edge.

PR #2263, opened 2025-05-27
fix: canonicalize_nodes sometimes mangles the entrypoint

### LINKED ISSUE #2262: canonicalize_nodes does not always preserve entrypoint
Found this following a failure in roundtripping.
```
#[test]
    fn canonicalize_nodes() {
        use crate::builder::{
            Container, Dataflow, DataflowSubContainer, HugrBuilder, ModuleBuilder,
        };
        use crate::extension::prelude::bool_t;
        use crate::ops::handle::{FuncID, NodeHandle};
        use crate::types::Signature;

        let sig = Signature::new(vec![bool_t(); 2], bool_t());
        let mut mb = ModuleBuilder::new();
        let mut fa = mb.define_function("a", sig.clone()).unwrap();
        let f_id = FuncID::<true>::from(fa.container_node());
        let mut dfg = fa.dfg_builder(sig.clone(), fa.input_wires()).unwrap();
        let call = dfg.call(&f_id, &[], dfg.input_wires()).unwrap();
        let dfg = dfg.finish_with_outputs(call.outputs()).unwrap();
        let fa = fa.finish_with_outputs(dfg.outputs()).unwrap();
        let fb = mb.define_function("b", sig).unwrap();
        let [fst, _] = fb.input_wires_arr();
        let fb = fb.finish_with_outputs([fst]).unwrap();
        let mut h = mb.finish_hugr().unwrap();
        h.set_entrypoint(dfg.node());
        let static_in = h.get_optype(call.node()).static_input_port().unwrap();
        let static_out = h.get_optype(fb.node()).static_output_port().unwrap();
        assert_eq!(
            h.single_linked_output(call.node(), static_in)
                .map(|(n, _p)| n),
            Some(fa.node())
        );
        h.disconnect(call.node(), static_in);
        h.connect(fb.node(), static_out, call.node(), static_in);

        fn find_dfg(h: &Hugr) -> Node {
            h.nodes()
                .filter(|n| h.get_optype(*n).is_dfg())
                .exactly_one()
                .ok()
                .unwrap()
        }
        assert_eq!(find_dfg(&h), dfg.node());
        assert_eq!(h.entrypoint(), dfg.node());

        h.canonicalize_nodes(|_, _| ());
        assert_ne!(h.entrypoint(), find_dfg(&h)); // OOOPS
    }
```
investigating the mermaid string shows the entrypoint has "moved" to the Input node of `fa`.

acl-cqc:
closes #2262

I left the `use`s local to the test, as the test seems very different to all the others. Should we keep it? (It seems fragile - i.e. somewhat lucky to have reproduced the bug in the first place.) If so, should we turn it into serialized etc.? (This feels like more effort than it's worth...)


acl-cqc:
> I **think** this should work, but I'm slightly confused why it's done this way.
> 
> If I understand this correctly, `ordered` can be seen as a map from new node id to old node id. Therefore the new entrypoint is the node with id `target` such that `ordered[target] == self.entrypoint`. You can do one check for this just after the `let mut source = \\ ...` line, before you follow the permutation chain. Am I missing something?

Thanks - yes, you are right; `ordered` is not mutated (I've refactored to make that clearer). So the old check was kind of correct, if we had done it before following the permutation chain, rather than after. (I had been scratching my head over this a bit before, thank you for figuring it out!)

I've simplified using that, and made some of the logic a bit more explicit using an Option. And to increase confidence I've generalized the test to include more cases...how's that? :)

zrho (approved):
LGTM!

Issue #2240, opened 2025-05-20
Dynamic directed-graph Convexity Checking


aborgna-q:
Skip to _"This issues's problem"_ for the problem definition.

## Background: Directed-graph Convexity Checking

Given a **directed** graph with nodes $V$ and edges $E$ and a set of nodes $S \subseteq V$, we say $S$ is **convex** if all paths between two nodes in $S$ contain _only_ nodes in $S$.
For example, given a graph
```mermaid
graph LR
    inp0[" "] --> n0
    n0 --> n1
    n1 --> out0[" "]
    inp1[" "] --> n2
    n0 --> n3
    n2 --> n3
    n3 --> n4
    n3 --> n1
    n4 --> out1[" "]

    classDef highlight fill:#843334;
    classDef void fill:#0000,stroke:#0000;
    class n0,n1 highlight;
    class inp0,inp1,out0,out1 void;
```
the subgraph $\{n_0, n_1\}$ is _not_ convex since the path $n_0 \to n_3 \to n_1$ contains an extraneous node.

The convexity property is critical when optimizing the graph representation of a program by rewriting parts of it, since it ensures we do not break causality requirements via edges that are not part of the graph being replaced.

### A simple algorithm

A simple algorithm for detecting convexity requires exploring all the successors for each edge that connects out of the subgraph (known as the subgraph _output boundary_).

Take the edge $n_0 \to n_2$ from before, that connects a node in the subgraph with node outside of it. From that edge we can reach all of $\{n_1, n_3, n_4\}$.
Notice that while the edge's immediate successor was outside the $S$ we can reach a node in $S$ from it, so there's a path that demonstrates non-convexity.

In general, checking convexity for a subset of nodes requires exploring all the successors from the output boundary of the set. If we can reach the subset again from any of its outputs, then we know it is not convex. This has a runtime complexity of $O(|V| + |E|)$
### Speedup via offline pre-computation

When we plan to check convexity for multiple subgraphs repeatedly, the process can be sped up with some pre-calculation:
- First, we compute a topological ordering $T: V \to \mathbb{N}$ for the whole graph.
- For a subset $S$, if there exists a path $u_o \to \dots \to v \to \dots \to u_i$ where $u_o, u_i$ are in $S$ and $v$, we know that $v$ must be between the other nodes in the topological order, so in general $min_{u \in S} T(u) \le T(v) \le max_{u \in S} T(u)$.
[...]

PR #2276, opened 2025-05-30
feat: Add DynamicTopoConvexChecker and tests

gluonhiggs:
Resolve issue #2240 
## Description

I have added the logic for `DynamicTopoConvexCheck` in the `convexity.rs` based on the paper (instead of leveraging `petgraph`'s `OrderMap` because it may require extra modifications to `petgraph` and is less flexible). I have also adapted it to the repository, and a variety of simple test cases is covered.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@ proptest-regressions/
 # Devenv
 .devenv*
 devenv.local.nix
-.envrc
 
 # Coverage report
 lcov.info
@@ -41,6 +40,7 @@ htmlcov/
 .venv
 env/
 venv/
+.envrc
 
 # ruff
 .ruff_cache


lmondada:
Hi @gluonhiggs, thanks for your contribution!

I should be able to get to your PR later today :)

gluonhiggs:
> Hi @gluonhiggs, thank you again for your contribution! I'm very glad you've looked at this problem.
> 
> There is currently a flaw in your PR: because of an unimplemented function (`fn port_to_node`), your code logic in `fn is_convex` is skipped (unless both inputs and outputs are empty) and the method will always return true. See my comments below.
> 
> On a more high-level note: can I ask you to open this same PR but in the `portgraph` repo? As you've designed this `ConvexChecker`, there is no reason for this to be in `Hugr` and not in [`portgraph`](https://docs.rs/portgraph/latest/portgraph/). I could copy your files over myself, but it's better if you do it, so you get the GitHub attribution you deserve. Apologies for the extra hassle.
> 
> Thanks again for your effort and I look forward to your fixes!

I have created a PR [here](https://github.com/CQCL/portgraph/pull/212) on `portgraph`. I thought it was simply just copy paste from `hugr` to `portgraph` :joy:

lmondada:
Thank you!

PR #2283, opened 2025-06-02
fix: use envelopes for `FixedHugr` encoding

### LINKED ISSUE #2282: Match python `FixedHugr` encoding to rust
The fixedhugr encoding is using deprecated json fallback in Python.

ss2165:
Closes #2282


aborgna-q:
Should we try to keep the loading part backwards-compatible? (The storing should be fine, as long as all users are on hugr-rs 0.20

ss2165:
> Should we try to keep the loading part backwards-compatible?


I don't think it is worth it, this isn't a part of the code-base being used

PR #2288, opened 2025-06-03
perf(py): mutable `Node` to avoid linear update cost

ss2165:
`list.index` call to update parent was very costly for nodes with many children. This was only necessary because Node was frozen, this can be avoided with a custom hash.


aborgna-q (approved):
Nice!

I also see uses of `replace` on a `Node` in `./hugr/build/dfg.py:234`

PR #2289, opened 2025-06-03
feat!: Added float and bytes literal to core and python bindings.

### LINKED ISSUE #2292: Floating point and bytes literals in `hugr-core`.
In addition to strings and natural numbers, the literals of `hugr-model` support floats and byte strings. These are useful, among other things, to express constants as terms. `hugr-core` currently does not support these literals in `TypeArg`s.

zrho:
This PR adds floating point and bytes literals to the core and Python representation of terms to make it closer to terms in `hugr-model`.

BREAKING CHANGE: Added `TypeParam`s and `TypeArg`s corresponding to floats and bytes.

Closes #2292.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1239,6 +1239,7 @@ dependencies = [
 name = "hugr-core"
 version = "0.20.1"
 dependencies = [
+ "base64",
  "cgmath",
  "cool_asserts",
  "delegate",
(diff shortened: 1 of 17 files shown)


ss2165 (changes requested):
Changes look good, please add tests for python and rust

ss2165 (approved):
thanks!

PR #2290, opened 2025-06-03
feat: Rewrite for peeling a TailLoop

### LINKED ISSUE #2107: Rewrite for peeling a TailLoop
Useful for https://github.com/CQCL/tket2/issues/632
There is some template code in #1834, but this should be changed to use `HugrMut::copy_descendants`, as #1934

acl-cqc:
closes #2107 

Uses `HugrMut::copy_descendants` to copy incoming edges (without copying their sources)


aborgna-q (approved):
Logic looks great. Just a small comment about the invalidation set.

PR #2315, opened 2025-06-09
fix: print inner model errors

ss2165:


zrho:
The inner errors are deliberately not printed since the inner errors are accessible via the `.source()` chain of errors. The PR #2315 makes `hugr-cli` report the chain of errors.

ss2165:
Closing in favour of #2318

PR #2317, opened 2025-06-09
test: enable Model as default BINARY envelope format

ss2165:


hugr-py/tests/conftest.py
```diff
@@ -149,8 +149,10 @@ def validate(
     # TODO: Use envelopes instead of legacy hugr-json
```
aborgna-q, line 149: drive-by
```suggestion
```

aborgna-q (approved):
:shipit:

PR #2318, opened 2025-06-09
feat!: Better error reporting in `hugr-cli`.

zrho:
This PR streamlines the error reporting in `hugr-cli`, displaying error source trails and setting up the tracing crate. For that, some error trait implementations were migrated to `thiserror`. The PR also simplifies the `clap` structures by pulling out the verbosity level from the individual commands. The commands should use tracing macros instead of relying on filtering for verbosity themselves.

BREAKING CHANGE: Reorganised `clap` commands and errors in `hugr-cli` which is only visible when it is used as a library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1223,6 +1223,7 @@ dependencies = [
 name = "hugr-cli"
 version = "0.20.1"
 dependencies = [
+ "anyhow",
  "assert_cmd",
  "assert_fs",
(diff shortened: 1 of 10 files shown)


aborgna-q:
What's the reason for swapping `derive_more::Error` with `thiserror`?

I was hoping we'd be able to drop the latter at some point :/

PR #2319, opened 2025-06-09
fix: Fixed invalid extension name in test.

zrho:
The Python testing suite defined an extension named `pytest.quantum,`. The `,` is not allowed as part of an extension name, making the model import fail.


acl-cqc (approved):
LGTM, thanks Lukas! I guess this was never fed to Rust's `IdentList::new` which would surely have caught this....

PR #2323, opened 2025-06-10
fix: Fixed bug in python model export name mangling.

zrho:
This PR fixes a bug in the Python model export: `Call` nodes referred to themselves instead of the function to be called.


ss2165:
Testing is in progress here: https://github.com/CQCL/hugr/pull/2317

PR #2331, opened 2025-06-10
feat(cli): convert sub-command for converting envelope formats

### LINKED ISSUE #2307: Add hugr-cli `convert` subcommand
Add a subcommand that reads a hugr and outputs it with the given envelope parameters.

In the future we could also allow reading older envelope versions here, and port update them as needed.

ss2165:
Closes #2307


hugr-cli/src/convert.rs
```diff
@@ -0,0 +1,86 @@
...
+                },
+                None => env_config.format, // Use input format if not specified
+            };
+            EnvelopeConfig::new(format)
+        };
+
+        // Create the envelope config
+        // Configure compression
+        if let Some(level) = self.compression_level {
+            config = config.with_zstd(ZstdConfig::new(level));
```
acl-cqc, line 78: I don't see `self.compress` used - should this be
```
if self.compress {
   config = config.with_zstd(ZstdConfig::new(self.compression_level.unwrap_or_default()))
}
```
?

hugr-cli/src/convert.rs
```diff
@@ -0,0 +1,86 @@
...
+pub struct ConvertArgs {
+    /// Hugr input.
+    #[command(flatten)]
+    pub input_args: HugrInputArgs,
+
+    /// Output file. Use '-' for stdout.
+    #[clap(short, long, value_parser, default_value = "-")]
+    pub output: Output,
+
+    /// Output format. One of: json, model, model-exts, model-text, model-text-exts
```
acl-cqc, line 25: Mention cannot be combined with `text` or `binary`. I'm mildly confused whether an appropriate `format` can give the same result as `text` / `binary` or these are all different, but the former makes sense if we're just trying to be non-committal about what the defaults are ;).

hugr-cli/src/convert.rs
```diff
@@ -0,0 +1,86 @@
...
+                    "model-text" => EnvelopeFormat::ModelText,
+                    "model-text-exts" => EnvelopeFormat::ModelTextWithExtensions,
+                    _ => Err(CliError::InvalidFormat(fmt.clone()))?,
+                },
+                None => env_config.format, // Use input format if not specified
+            };
+            EnvelopeConfig::new(format)
+        };
+
+        // Create the envelope config
```
acl-cqc, line 75: nit: this comment belongs earlier, if it's needed at all

hugr-cli/src/hugr_io.rs
```diff
@@ -46,15 +46,23 @@ impl HugrInputArgs {
...
+    /// configuration and the package encoded within.
+    ///
+    /// If [`HugrInputArgs::hugr_json`] is `true`, [`HugrInputArgs::get_hugr`] should be called instead as
+    /// reading the input as a package will fail.
+    pub fn get_envelope(&mut self) -> Result<(EnvelopeConfig, Package), CliError> {
         let extensions = self.load_extensions()?;
         let buffer = BufReader::new(&mut self.input);
         match read_envelope(buffer, &extensions) {
-            Ok((_, pkg)) => Ok(pkg),
+            Ok(env) => Ok(env),
```
acl-cqc, line 61: consider `map_err` here

hugr-cli/src/hugr_io.rs
```diff
@@ -46,15 +46,23 @@ impl HugrInputArgs {
     /// If [`HugrInputArgs::hugr_json`] is `true`, [`HugrInputArgs::get_hugr`] should be called instead as
     /// reading the input as a package will fail.
     pub fn get_package(&mut self) -> Result<Package, CliError> {
+        self.get_envelope().map(|(_, package)| package)
+    }
+
+    /// Read a hugr envelope from the input and return the envelope
+    /// configuration and the package encoded within.
+    ///
+    /// If [`HugrInputArgs::hugr_json`] is `true`, [`HugrInputArgs::get_hugr`] should be called instead as
```
acl-cqc, line 55: This is a slightly awkward API but I admit I can't see how to make it better given clap....does this belong in an `/// # Errors` ?

hugr-cli/tests/convert.rs
```diff
@@ -0,0 +1,262 @@
...
+    let output = convert_cmd.assert().success().get_output().to_owned();
+    let stdout = output.stdout.clone();
+
+    let reader = BufReader::new(stdout.as_slice());
+    let registry = ExtensionRegistry::default();
+    let (config, _) = read_envelope(reader, &registry).expect("Failed to read output envelope");
+
+    // Verify it's a binary format (not ASCII printable)
+    assert!(!config.format.ascii_printable());
+    assert!(config.zstd.is_some());
```
acl-cqc, line 238: This is because binary defaults to some level of compression?

hugr-cli/src/hugr_io.rs
acl-cqc: Perhaps first line of comment should be "Read a hugr envelope containing a model and....." or similar

hugr-cli/tests/convert.rs
```diff
@@ -0,0 +1,262 @@
...
+fn test_convert_with_compression(test_envelope_file: NamedTempFile, mut convert_cmd: Command) {
+    // Create output file
+    let output_file = assert_fs::NamedTempFile::new("output.hugr").unwrap();
+
+    convert_cmd.args([
+        test_envelope_file.path().to_str().unwrap(),
+        "-o",
+        output_file.path().to_str().unwrap(),
+        "--compress",
+        "--compression-level",
```
acl-cqc, line 141: You could add a param `#[values([], ["--compression-level", "5"])] compress_flags: impl IntoIterator<Item=&'static str>` and generalize this, and test the output has been compressed

acl-cqc (approved):
Thanks Seyon - good tests - generally LGTM, tho not sure the logic re. compression flags is correct.

hugr-cli/src/convert.rs
ss2165: clap should enforce this because `compression_level` "requires" `compress`, but no harm in making it more explicit

hugr-cli/src/hugr_io.rs
ss2165: the top line is correct `hugr_json` being true implies the payload is _not_ envelope encoded. Envelope encoded does not necessarily imply using model.

PR #2332, opened 2025-06-10
test(py): check roundtrip through CLI

ss2165:


ss2165:
> Why do we only do this for packages, and not Hugrs?

Good question! fixed, which unearthed a bug that has been fixed

PR #2334, opened 2025-06-11
fix!: Fixed bugs in model CFG handling and improved CFG signatures

zrho:
This PR fixes some bugs and a design error related to model CFGs. The signature of a control flow region and of a dataflow block is now of the form `(core.ctrl ?inputs ?outputs)` where `?inputs` and `?outputs` are lists of lists of types. Each item in the outer list corresponds to a branch and the items in the inner lists of the types of the values that are transmitted in that branch.

BREAKING CHANGE: The model CFG signature types were changed.

### DIFF
--- a/hugr-core/src/export.rs
+++ b/hugr-core/src/export.rs
@@ -14,7 +14,7 @@ use crate::{
     },
     types::{
         CustomType, EdgeKind, FuncTypeBase, MaybeRV, PolyFuncTypeBase, RowVariable, SumType,
-        TypeArg, TypeBase, TypeBound, TypeEnum, TypeRow,
+        TypeArg, TypeBase, TypeBound, TypeEnum,
         type_param::{TypeArgVariable, TypeParam},
         type_row::TypeRowBase,
     },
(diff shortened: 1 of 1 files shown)


hugr-core/src/export.rs
```diff
@@ -742,16 +740,14 @@ impl<'a> Context<'a> {
 
             let mut wrap_ctrl = |types: &TypeRow| {
```
croyzor, line 741: This function is now poorly named - maybe it can even be removed?

hugr-core/tests/snapshots/model__roundtrip_cfg.snap
```diff
@@ -22,10 +22,9 @@ expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-cfg.
     (cfg [%0] [%1]
       (signature (core.fn [?0] [?0]))
       (cfg [%2] [%3]
-        (signature (core.fn [(core.ctrl [?0])] [(core.ctrl [?0])]))
+        (signature (core.ctrl [[?0]] [[?0]]))
```
croyzor, line 25: Why do these need two sets of square brackets?

zrho: The outer list is the list of branches. The inner lists are the lists of values for those branches.

hugr-core/tests/snapshots/model__roundtrip_cfg.snap
```diff
@@ -1,6 +1,6 @@
 ---
 source: hugr-core/tests/model.rs
-expression: "roundtrip(include_str!(\"../../hugr-model/tests/fixtures/model-cfg.edn\"))"
+expression: ast
```
croyzor, line 3: Not sure what's going on here?

zrho: There was some refactoring in an earlier PR that changed how the snapshot test is called. The snapshot itself didn't change, and so the snapshot file was still considered valid. Now the snapshot changed, which also led to a change in the metadata that captures how the snapshot test is called.

PR #2336, opened 2025-06-11
fix: update CallGraph and remove_dead_funcs for module-only FuncDefns

acl-cqc:
This should have been part of #2256 but some confusion dates from #2147.

Making a callgraph for only the functions beneath the entrypoint makes no sense - that's either all the functions (if the entrypoint is the root) or at most one of them (if the entrypoint is one of the funcdefns). Of course, we didn't test with any Hugrs that had a non-module-root entrypoint.... so the sensible thing to do now, seems to be for callgraph to ignore the entrypoint. (We can optimize it looking for FuncDefns tho, they are now easy to find.)

Dead function removal allowed specifying starting points for analysis (top-level FuncDefns) but only if the Hugr's entrypoint was the module. There seems no good reason not to allow specifying module-level FuncDefns even if the entrypoint is lower, and similarly, we should use the entrypoint as  one starting point (perhaps among many) for analysis...

(One could rename `entry_points` used by analysis to "starting_points" but that would be breaking)

### DIFF
--- a/hugr-passes/src/call_graph.rs
+++ b/hugr-passes/src/call_graph.rs
@@ -26,7 +26,7 @@ pub enum CallGraphNode<N = Node> {
 }
 
 /// Details the [`Call`]s and [`LoadFunction`]s in a Hugr.
-/// 
+///
 /// Each node in the `CallGraph` corresponds to a [`FuncDefn`] or [`FuncDecl`] in the Hugr;
 /// each edge corresponds to a [`Call`]/[`LoadFunction`] of the edge's target, contained in
 /// the edge's source.


hugr-passes/src/dead_funcs.rs
```diff
@@ -35,30 +35,17 @@ fn reachable_funcs<'a, H: HugrView>(
...
-        for n in entry_points {
-            if !h.get_optype(n).is_func_defn() || h.get_parent(n) != Some(h.entrypoint()) {
-                return Err(RemoveDeadFuncsError::InvalidEntryPoint { node: n });
-            }
-            d.stack.push(cg.node_index(n).unwrap());
-        }
-        d
-    } else {
-        if let Some(n) = entry_points.next() {
-            // Can't be a child of the module root as there isn't a module root!
```
acl-cqc, line 53: This comment has been bad since #2147 ...

ss2165 (approved):
nice 👍

PR #2300, opened 2025-06-03
feat: Add serial data types for SimpleReplacement and PersistentHugr

lmondada:
This PR provides serde serialization/deserialization support for `SimpleReplacement` as well as the main types in the `PersistentHugr` implementation as discussed in #2253.

This can also be used as a template for other types that require ser/deser implementations in the future.

Closes https://github.com/CQCL/hugr/issues/2253

PR #2344, opened 2025-06-16
feat(persistent): Add serialisation for CommitStateSpace

### LINKED ISSUE #2299: Replace PointerEqResolver by hash-based resolver
Currently `PersistentHugr` uses `PointerEqResolver` which generates NodeIDs non-deterministically. Replace this with a deterministic hash-based resolver.

lmondada:
This PR adds serialisation for the `CommitStateSpace` type, following the pattern of #2300. To make the serialisation/deserialisation deterministic, a new `Resolver` is introduced.

Resolvers define a notion of equivalence between commits in the state space. The simplest resolver is `PointerEqResolver`, which considers commits to be equivalent if they point to the same data in memory. The "key" to that commit is the memory pointer, and so is non-deterministic.

The new resolver determines commit equivalence by computing a hash of the serialised data. This is not very efficient and will have to be improved in the future, but it is correct and convenient to use for the time being. For this purpose, a fast, reproducible and platform-independent hash is used (`wyhash`).

To accomodate multiple resolves, `CommitStateSpace` and `PersistentHugr` had to be made generic over the resolver type.

Closes #2299

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1351,6 +1351,7 @@ dependencies = [
  "delegate",
  "derive_more 1.0.0",
  "hugr-core",
+ "insta",
  "itertools 0.14.0",
  "lazy_static",
  "petgraph 0.8.1",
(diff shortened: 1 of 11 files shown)


mark-koch (approved):
Looks good to me 👍

PR #2364, opened 2025-06-23
ci: ensure std extension versions are updated if they change

### LINKED ISSUE #2363: ci: check extension versions are updated if serialization changes

ss2165:
Closes #2363


scripts/check_extension_versions.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def get_changed_files() -> list[Path]:
+    """Get list of changed extension files in the PR"""
+    # Use git to get the list of files changed compared to main
+    cmd = [
+        "git",
+        "diff",
+        "--name-only",
+        "origin/main",
```
aborgna-q, line 16: This will compare against
- The latest `main`, in PRs and the merge queue.
- The same commit, in `push`-to-main runs

For the former, we should pass the PR's target commit instead so it's safer.
```yaml
${{ github.event.pull_request.base.sha || github.event.merge_group.base.sha }}
```

For the runs on merged commits, I say we should compare against the latest release, so we can ensure that fixes work.
```bash
git describe --abbrev=0 --match="hugr-v[0-9]*"
```

aborgna-q: Or probably 
```
git describe --abbrev=0 --match="hugr-v[0-9]*" $TARGET_SHA
```
in the PR case, since there may have been a release in `main` 🤔

aborgna-q (approved):

PR #2367, opened 2025-06-24
feat: add toposort to HUGR-py

CalMacCQ:
Basic Toposort implementation with [Kahn's algorithm](https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm).

This is my first code contribution to hugr-py so any pointers on how to write a good test would be appreciated. Spent quite some time trying to figure out the graph builder.

### DIFF
--- a/hugr-py/src/hugr/hugr/base.py
+++ b/hugr-py/src/hugr/hugr/base.py
@@ -237,7 +237,7 @@ class Hugr(Mapping[Node, NodeData], Generic[OpVarCov]):
             parent: The parent node of the region to sort.
 
         Raises:
-            ValueError: If the hugr contains a cycle.
+            ValueError: If the region contains a cycle.
 
         Examples:
             >>> from hugr.build.tracked_dfg import TrackedDfg


hugr-py/tests/test_hugr_build.py
```diff
@@ -404,3 +405,19 @@ def test_option() -> None:
...
+    q = f.add_op(H, q).out(0)
+    b0 = f.add_op(Not, b)
+    b1 = b
+
+    f.set_outputs(q, b0, b1)
+
+    validate(Package([f.hugr], [QUANTUM_EXT]))
+
+    sorted_nodes = f.hugr.get_sorted_nodes(f.hugr.module_root)
+    assert len(list(sorted_nodes)) == len(list(f.hugr))
```
CalMacCQ, line 423: This is a bad test. Will replace with asserting that the sort is a valid topological ordering

aborgna-q (dismissed):

hugr-py/src/hugr/hugr/base.py
```diff
@@ -227,6 +227,28 @@ def nodes(self) -> Iterable[tuple[Node, NodeData]]:
...
+            if visit_dict[node] == 0:
+                queue.put(node)
+
+        while not queue.empty():
+            new_node = queue.get()
+            yield new_node
+
+            for node in self.children(new_node):
+                if self.has_link(new_node.out(1), node.inp(1)):
+                    visit_dict[node] -= 1
```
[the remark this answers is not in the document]
CalMacCQ: Thanks for your help.

just to clarify here `node` is the node that you pop from the queue? I've used `new_node` for that. I think I might switch to your naming for clarity

CalMacCQ:
[This implementation](https://github.com/CQCL/hugr-phase-fold/commit/463a36c76d10b1e0c69b9b9fbbda56e6d3e91776) handles non-local edges

aborgna-q:
> We may also want to check that there are no cycles

If `queue` is empty but `visit_dict` still has elements then there are things that we couldn't reach, so there is a loop and we should raise an error.

hugr-py/src/hugr/hugr/base.py
```diff
@@ -227,6 +227,43 @@ def nodes(self) -> Iterable[tuple[Node, NodeData]]:
         """
         return self.items()
 
+    def sort_region_nodes(self, parent: Node) -> Iterator[Node]:
```
aborgna-q, line 230: "sorted", to use python parlance

hugr-py/src/hugr/hugr/base.py
```diff
@@ -227,6 +227,43 @@ def nodes(self) -> Iterable[tuple[Node, NodeData]]:
...
+        """Iterator over topological ordering of all the hugr nodes within a region."""
```
aborgna-q, line 231: Add `Args` and `Raises` to the docs.
Maybe also an `Example` (see the other methods in this file).

And mention that external edges get ignored.

hugr-py/tests/test_hugr_build.py
```diff
@@ -404,3 +405,47 @@ def test_option() -> None:
     dfg.set_outputs(b)
 
     validate(dfg.hugr)
+
+
+# a helper for the toposort tests
+def _build_test_hugrfn() -> Function:
```
aborgna-q, line 411: You can mark this as a fixture,
```suggestion
@pytest.fixture
def simple_function() -> Function:
    """A helper for the toposort tests."""
```

And then use it as a parameter to the tests
```python
def test_toposort(simple_function: Function) -> None: ...
```

hugr-py/tests/test_hugr_build.py
```diff
@@ -404,3 +405,47 @@ def test_option() -> None:
...
+
+    f.set_outputs(q, nnot, b)
+    return f
+
+
+# https://github.com/CQCL/hugr/issues/2350
+def test_toposort() -> None:
+    f = _build_test_hugrfn()
+
+    validate(Package([f.hugr], [QUANTUM_EXT]))
```
aborgna-q, line 428: This should go in the function building function rather than the toposort test

hugr-py/tests/test_hugr_build.py
```diff
@@ -404,3 +405,47 @@ def test_option() -> None:
...
+
+    nodes = list(f.hugr)
+    func_node = nodes[1]
+
+    sorted_nodes = list(f.hugr.sort_region_nodes(func_node))
+    assert set(sorted_nodes) == {f.input_node, f.output_node, nodes[4], nodes[5]}
```
aborgna-q, line 434: I guess the "safer" way of checking this rather than indexing into the nodes array is something like
```suggestion
    assert set(sorted_nodes) == set(f.hugr.children(f))
```

hugr-py/tests/test_hugr_build.py
```diff
@@ -404,3 +405,47 @@ def test_option() -> None:
...
+    assert sorted_nodes[0] == f.input_node
+    assert sorted_nodes[-1] == f.output_node
+
+
+def test_toposort_error() -> None:
+    # Test that we get an error if we toposort an invalid hugr containing a cycle
+    f = _build_test_hugrfn()
+
+    nodes = list(f.hugr)
+    func_node = nodes[1]
```
aborgna-q, line 444: `Function` implements `ToNode`, you can use it directly to refer to the region node,
```python
list(f.hugr.sort_region_nodes(f))
```

hugr-py/tests/test_hugr_build.py
```diff
@@ -404,3 +405,47 @@ def test_option() -> None:
...
+
+    # Add a loop, invalidating the HUGR
+    f.hugr.add_link(nodes[4].out_port(), nodes[4].inp(0))
```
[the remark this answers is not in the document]
CalMacCQ: Hmmm I get this error

```
simple_fn = Function(parent_node=Node(1))

    def test_toposort_error(simple_fn: Function) -> None:
        # Test that we get an error if we toposort an invalid hugr containing a cycle
        nodes = list(simple_fn.hugr)
        func_node = nodes[1]
    
        # Add a loop, invalidating the HUGR
>       simple_fn.hugr.add_link(simple_fn.input_node, simple_fn.output_node)

hugr-py/tests/test_hugr_build.py:443: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
hugr-py/src/hugr/hugr/base.py:499: in add_link
    src_sub = self._unused_sub_offset(src)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Hugr(module_root=Node(0), entrypoint=Node(1), _nodes=[NodeData(op=Module(), parent=None, metadata={}), NodeData(op=Fun...), _SubPort(port=OutPort(Node(2), 0), sub_offset=1): _SubPort(port=InPort(Node(3), 2), sub_offset=0)}), _free_nodes=[])
port = Node(2)

    def _unused_sub_offset(self, port: P) -> _SubPort[P]:
        d: dict[_SO, _SI] | dict[_SI, _SO]
        match port:
            case OutPort(_):
                d = self._links.fwd
            case InPort(_):
                d = self._links.bck
        sub_port = _SubPort(port)
>       while sub_port in d:
E       UnboundLocalError: cannot access local variable 'd' where it is not associated with a value
```

Source

```python
def test_toposort_error(simple_fn: Function) -> None:
    # Test that we get an error if we toposort an invalid hugr containing a cycle
    nodes = list(simple_fn.hugr)
    func_node = nodes[1]

    # Add a loop, invalidating the HUGR
    simple_fn.hugr.add_link(simple_fn.input_node, simple_fn.output_node)
    with pytest.raises(
        ValueError, match="Graph contains a cycle. No topological ordering exists."
    ):
        list(simple_fn.hugr.sort_region_nodes(func_node))
```

hugr-py/src/hugr/hugr/base.py
CalMacCQ: I slightly prefer `sort` as its a verb rather than an adjective but python does have a builtin `sorted` its true.

PR #2369, opened 2025-06-24
refactor!: remove deprecated runtime extension errors

ss2165:
BREAKING CHANGE: Unused errors to do with runtime extension requirement checking have been removed (deprecated in v0.20.0).


lmondada (approved):
nit: I would write the changelog entry as something like

> Removed unused errors `ValidationError::ExtensionError` (deprecated in v0.20) and `PackageEncodingError::RuntimeExtensionResolution`

PR #2378, opened 2025-06-26
fix: Use List instead of Tuple in conversions for TypeArg/TypeRow

lmondada:
Conversions introduced in #2366 were expecting `Term::Tuples` instead of `Term::List`. This fixes this.

### DIFF
--- a/.github/workflows/ci-py.yml
+++ b/.github/workflows/ci-py.yml
@@ -200,14 +200,13 @@ jobs:
             # Check against latest tag on the target branch
             # When not on a pull request, base_ref should be empty so we default to HEAD
             if [ -z "$TARGET_REF" ]; then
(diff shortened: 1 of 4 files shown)


hugr-core/src/types/type_row.rs
```diff
@@ -333,7 +334,7 @@ mod test {
         // Test successful conversion with Tuple
```
zrho, line 334: This comment is out of date.

hugr-core/src/types/type_row.rs
```diff
@@ -349,7 +350,7 @@ mod test {
...
```
zrho, line 350: Also this one.

zrho (approved):

PR #2403, opened 2025-07-03
ci: support releases from release/* branches

ss2165:
Following https://github.com/googleapis/release-please-action?tab=readme-ov-file#supporting-multiple-release-branches


aborgna-q (approved):
👍 
I think the wheel publishing PR will work automatically with the release trigger

PR #2411, opened 2025-07-04
feat: Type of constants in `core` `Term`s.

### LINKED ISSUE #2304: `TypeParam` for constants.
`hugr-model` represents constants for a runtime type `?t` as terms of the static type `(core.const ?t)`. To ease the migration from JSON encoded to term encoded constants, we should add the ability to represent the type of constants in `TypeParam`.

This could be done, as it is in `hugr-model`, via custom constructors. However, to avoid having to figure out custom `TypeParam`s first (#2296), we can special case this (for now) as follows:

```rust
pub enum TypeParam {
    // ...

    /// The type of runtime constants.
    Const {
        runtime_type: Type,
    }
}
```

In particular, resolving this issue would require:
 - Add a `Const` variant to `TypeParam` as above.
 - Add a `ConstTypeParam` to the Python bindings.
 - Write the Rust `model` to `core` import code that imports `(core.const ?t)` as `TypeParam::Const`.
 - Write the Rust`core` to `model` export code that exports `TypeParam::Const` to `(core.const ?t)`.
 - Write the Python `core` to `model` export code that exports `ConstTypeParam` to `(core.const ?t)`.
 - Update the spec accordingly.

zrho:
This PR adds a `Term::Const` variant, representing the type of constants for a given runtime type. This corresponds to the `core.const` type in `hugr-model`.

Closes #2304.

### DIFF
--- a/hugr-core/src/types/type_param.rs
+++ b/hugr-core/src/types/type_param.rs
@@ -141,7 +141,14 @@ pub enum Term {
     Variable(TermVar),
 
     /// The type of constants for a runtime type.
-    Const(Box<Type>),
+    ///
+    /// A constant is a compile time description of a runtime value. The runtime
+    /// value is constructed when the constant is loaded.
+    ///
+    /// Constants are distinct from the runtime values that they describe. In
+    /// particular, as part of the term language, constants can be freely copied
+    /// or destroyed even when they describe a non-linear runtime value.
+    ConstType(Box<Type>),
 }
 
 impl Term {
@@ -174,7 +181,7 @@ impl Term {
 
     /// Creates a new [`Term::Const`] from a runtime type.
     pub fn new_const(ty: impl Into<Type>) -> Self {
-        Self::Const(Box::new(ty.into()))
+        Self::ConstType(Box::new(ty.into()))
     }
 
     /// Checks if this term is a supertype of another.
@@ -377,7 +384,7 @@ impl Term {
             Term::ListType(item_type) => item_type.validate(var_decls),
             Term::TupleType(item_types) => item_types.validate(var_decls),
             Term::StaticType => Ok(()),
-            Term::Const(ty) => ty.validate(var_decls),
+            Term::ConstType(ty) => ty.validate(var_decls),
         }
     }
 
@@ -441,7 +448,7 @@ impl Term {
             Term::ListType(item_type) => Term::new_list_type(item_type.substitute(t)),
             Term::TupleType(item_types) => Term::new_list_type(item_types.substitute(t)),
             Term::StaticType => self.clone(),
-            Term::Const(ty) => Term::new_const(ty.substitute1(t)),
+            Term::ConstType(ty) => Term::new_const(ty.substitute1(t)),
         }
     }
 
@@ -603,7 +610,7 @@ impl Transformable for Term {
             Term::StaticType => Ok(false),
             TypeArg::ListConcat(lists) => lists.transform(tr),
             TypeArg::TupleConcat(tuples) => tuples.transform(tr),
-            Term::Const(ty) => ty.transform(tr),
+            Term::ConstType(ty) => ty.transform(tr),
         }
     }
 }
@@ -696,7 +703,7 @@ pub fn check_term_type(term: &Term, type_: &Term) -> Result<(), TermTypeError> {
         (Term::ListType { .. }, Term::StaticType) => Ok(()),
         (Term::TupleType(_), Term::StaticType) => Ok(()),
         (Term::RuntimeType(_), Term::StaticType) => Ok(()),
-        (Term::Const(_), Term::StaticType) => Ok(()),
+        (Term::ConstType(_), Term::StaticType) => Ok(()),
 
         _ => Err(TermTypeError::TypeMismatch {
             term: Box::new(term.clone()),
--- a/hugr-core/src/types/type_param.rs
+++ b/hugr-core/src/types/type_param.rs
@@ -179,7 +179,7 @@ impl Term {
         Self::TupleType(Box::new(item_types.into()))
     }
 
-    /// Creates a new [`Term::Const`] from a runtime type.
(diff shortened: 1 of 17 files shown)


hugr-core/src/types/type_param.rs
```diff
@@ -139,6 +139,9 @@ pub enum Term {
     /// - see [`Term::new_var_use`]
     #[display("{_0}")]
     Variable(TermVar),
+
+    /// The type of constants for a runtime type.
```
acl-cqc, line 143:
```suggestion
    /// The type of constants, i.e. terms that can be turned into runtime values of a given (runtime) type.
```

hugr-core/src/types/type_param.rs
```diff
@@ -139,6 +139,9 @@ pub enum Term {
...
+    Const(Box<Type>),
```
acl-cqc, line 144: I reckon this should be called `ConstType`, paralleling `BytesType` or `RuntimeType`. (In hindsight, using a `Param` suffix, rather than `...Type`, would have been more similar to the python!).

There may never be a `Const` constructor (i.e. in the family of TypeArgs, a term that is/produces a runtime value) - rather, there will likely be a number of different ones, but I think we can still try to be consistent wrt. naming.

hugr-core/src/types/type_param.rs
```diff
@@ -685,6 +696,7 @@ pub fn check_term_type(term: &Term, type_: &Term) -> Result<(), TermTypeError> {
         (Term::ListType { .. }, Term::StaticType) => Ok(()),
         (Term::TupleType(_), Term::StaticType) => Ok(()),
         (Term::RuntimeType(_), Term::StaticType) => Ok(()),
+        (Term::Const(_), Term::StaticType) => Ok(()),
```
acl-cqc, line 700: I guess we do not want to say that e.g. Term::BoundedNat is a const (e.g. of Runtime type `usize`) but rather wait for a different term that does that?

acl-cqc (commented):
Generally fine, a couple of suggestions - naming?

And, it's a bit of a shocker to have only 35% coverage! Obviously one cannot get very far without any terms *of type* Term::Const(....) but I think at the least you could add some roundtrip serialization tests and things like that?

hugr-core/src/types/type_param.rs
acl-cqc: Noting your example of qubit, I think we definitely want to say something stronger here - these are not static representations of the runtime values; they can (effectively) be *arbitrary recipes* or factories, procedures for production of.

(I think it's ok to continue to call them ConstType given "constants" are *the common usage* so long as we doc that really this is *a more general concept* than a constant)

(For example, there would not necessarily even be a guarantee that two runtime values produced by the same recipe would be equal, would there? I mean, there is not necessarily a notion of equality, so...)

hugr-core/src/types/type_param.rs
```diff
@@ -139,6 +139,16 @@ pub enum Term {
     /// - see [`Term::new_var_use`]
     #[display("{_0}")]
     Variable(TermVar),
+
+    /// The type of constants for a runtime type.
+    ///
+    /// A constant is a compile time description of a runtime value. The runtime
```
acl-cqc, line 145: So here - it does not describe *a runtime value* - maybe a population of them (e.g. floats produced according to a distribution, right?) but it's weak and/or presumptive to say that the constant describes an (individual) value coming from it.

(You also can't go backwards, from runtime value to constant, but that's a side-issue)

hugr-core/src/types/type_param.rs
```diff
@@ -169,6 +179,11 @@ impl Term {
         Self::TupleType(Box::new(item_types.into()))
     }
 
+    /// Creates a new [`Term::ConstType`] from a runtime type.
```
acl-cqc, line 182: nit: perhaps 
```suggestion
    /// Creates a new [`Term::ConstType`] for a runtime type.
```

hugr-py/tests/test_hugr_build.py
```diff
@@ -292,6 +292,20 @@ def test_literals() -> None:
     validate(mod.hugr)
 
 
+def test_const_type() -> None:
+    mod = Module()
+
+    mod.declare_function(
+        "const_type",
+        tys.PolyFuncType(
```
acl-cqc, line 300: right - nice example - statically parametric over a recipe for qubits, at runtime just produces qubits. Just to be clear - this is not like the one in model__roundtrip_const.snap, where the function's output is *the constant* (recipe) not the runtime type produced by it.

hugr-core/tests/snapshots/model__roundtrip_const.snap
```diff
@@ -94,3 +96,9 @@ expression: ast
           "{\"c\":\"ConstUnknown\",\"v\":{\"value\":1.0}}"))
       [] [%1]
       (signature (core.fn [] [arithmetic.float.types.float64])))))
+
+(declare-func
+  public
+  example.const_as_param
+  (param ?0 (core.const arithmetic.float.types.float64))
+  (core.fn [] [?0]))
```
acl-cqc, line 104: Just to be clear - not sure if you can have a comment here or in `model-const.edn`? - the function here returns the constant recipe, not any runtime float....right?

hugr-py/tests/test_tys.py
```diff
@@ -101,6 +102,7 @@ def test_tys_sum_str(ty: Type, string: str, repr_str: str):
             "(Linear, Nat(3))",
         ),
         (ListParam(StringParam()), "[String]"),
+        (ConstParam(Qubit), "Const(Qubit)"),
```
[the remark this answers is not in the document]
acl-cqc: Ok, indeed good you picked that one :). Let's be more explicit in the doc-comments on declaration of `ConstType` (I've commented there too), then.

acl-cqc (approved):
Thanks @zrho, sorry for dropping the ball. Nice! :) Just see comments on the doc for `ConstType`.

PR #2412, opened 2025-07-04
fix: Make JSON Schema checks actually work again

### LINKED ISSUE #2401: JSON Schema validation is not working
On main, this test added to `hugr-core/src/hugr/serialize/test.rs` passes:
```
#[test]
fn not_a_hugr() {
    let val = serde_json::Value::from("Hello, world!");
    NamedSchema::check_schemas(&val, get_testing_schemas(true));
}
```

Clearly it shouldn't! Adding this to `testing_hugr_schema_live.json` causes the schema violation to be detected (similarly for `testing_hugr_schema_strict_live.json`):
```
"$ref": "#/$defs/TestingHugr"
```
....but, lo, there are a bunch of other new failures in:
```
    hugr::serialize::test::proptest::prop_roundtrip_opdef
    hugr::serialize::test::proptest::prop_roundtrip_optype
    hugr::serialize::test::proptest::prop_roundtrip_poly_func_type
    hugr::serialize::test::proptest::prop_roundtrip_type
    hugr::serialize::test::proptest::prop_roundtrip_value
```

Moreover, the "non-testing" schemas will also need some kind of update - I conjecture possibly `"$ref": "#/$defs/Package"` (or a `oneOf` including package and SerialHugr?)

acl-cqc:
fixes #2401
* Update generate_schema.py to add an entry to the generated schema describing what members of `#/$defs` the actual value (i.e. root) should conform to.
* Add a test that schema validation fails when expected (by adding/removing fields).
* Some changes to `check_schemas` and related (crate) methods to expose errors etc.

* Fix various failures in proptests stemming from
   * #2309 (the `Arbitrary` instance can generate "TypeArg"s where "TypeParam"s are expected or vice versa, and these do not confirm to the JSON schema).
   * #2360 - I'm not clear what *well-typed* ArrayOrTermSer::Term can fit the JSON schema, but our `Arbitrary`s generate plenty that are not well-typed.
...I've "fixed" the proptests by filtering out non-schema-compliant instances using `prop_filter`, this is not great, the new situation is detailed in #2415 for further cleaning up.


hugr-core/src/hugr/serialize/test.rs
```diff
@@ -225,6 +230,77 @@ fn check_testing_roundtrip(t: impl Into<SerTestingLatest>) {
...
+}
+
+#[rstest]
+#[case(&TESTING_SCHEMA, &TESTING_SCHEMA_STRICT, test_schema_val(), ["optype"])]
+#[case(&SCHEMA, &SCHEMA_STRICT, schema_val(), [])]
+fn extra_and_missing_fields<const L: usize>(
+    #[case] lax_schema: &'static NamedSchema,
+    #[case] strict_schema: &'static NamedSchema,
+    #[case] mut val: serde_json::Value,
+    #[case] target_loc: [&'static str; L],
```
acl-cqc, line 269: Hmmm, I guess `: impl IntoIter<Item=&'static str> + Clone` would probably work, I didn't think of the `+ Clone`

PR #2413, opened 2025-07-07
fix: map IntValue to unsigned repr when serializing

### LINKED ISSUE #2409: bug(py): `IntVal` doesn't map integer values to u64 bounds
So e.g. for negative number loading fails

ss2165:
to match Rust definition
Closes #2409

### DIFF
--- a/hugr-py/src/hugr/std/int.py
+++ b/hugr-py/src/hugr/std/int.py
@@ -52,16 +52,39 @@ def _int_tv(index: int) -> tys.ExtType:
 INT_T = int_t(5)
 
 
+def _to_unsigned(val: int, bits: int) -> int:
+    """Convert a signed integer to its unsigned representation
+    in twos-complement form.
+
+    Positive integers are unchanged, while negative integers
+    are converted by adding 2^bits to the value.
+
+    Raises ValueError if the value is out of range for the given bit width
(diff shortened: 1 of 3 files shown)


hugr-py/src/hugr/std/int.py
```diff
@@ -52,16 +52,33 @@ def _int_tv(index: int) -> tys.ExtType:
...
+
+    if val < 0:
+        return (1 << bits) + val
+    return val
+
+
 @dataclass
 class IntVal(val.ExtensionValue):
-    """Custom value for an integer."""
+    """Custom value for a signedinteger."""
```
aborgna-q, line 73:
```suggestion
    """Custom value for a signed integer."""
```

hugr-py/src/hugr/std/int.py
```diff
@@ -52,16 +52,33 @@ def _int_tv(index: int) -> tys.ExtType:
 INT_T = int_t(5)
 
 
+def _to_unsigned(val: int, bits: int) -> int:
+    """Convert a signed integer to its unsigned representation
+    in twos-complement form.
```
aborgna-q, line 57: Mention that we are wrapping the negative values.

Otherwise, the "unsigned representation of a signed value" is just the identity, and undefined for negatives.

aborgna-q (approved):

PR #2420, opened 2025-07-08
refactor!: Reduce error type sizes

### LINKED ISSUE #2196: Reduce size of rust errors
As mentioned in https://github.com/CQCL/hugr/pull/2195#issuecomment-2872250588, the new clippy version releasing with 1.87 includes a new lint marking error types that are too big:
https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err

This fails with a lot of our definitions.
We should probably `Box` them when possible. The cost of the extra indirection is non important since error handling should not normally be in critical paths.

aborgna-q:
Closes #2196

> the new clippy version releasing with 1.87 includes a new lint marking error types that are too big:
> https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
>
> This fails with a lot of our definitions. We should probably Box them when possible. The cost of the extra indirection is non important since error handling should not normally be in critical paths.

This PR removes the clippy lint ignores from the workspace config.

BREAKING CHANGE: Wrapped some error fields in `Box`es to reduce their footprint.

### DIFF
--- a/hugr-passes/src/dataflow/datalog.rs
+++ b/hugr-passes/src/dataflow/datalog.rs
@@ -51,7 +51,11 @@ impl<H: HugrView, V: AbstractValue> Machine<H, V> {
     /// or [Conditional](hugr_core::ops::Conditional)).
     /// Any inputs not given values by `in_values`, are set to [`PartialValue::Top`].
     /// Multiple calls for the same `parent` will `join` values for corresponding ports.
-    #[expect(clippy::result_large_err)]
+    #[expect(
+        clippy::result_large_err,
+        reason = "Not called recursively and not a performance bottleneck"
+    )]
+    #[inline]
     pub fn prepopulate_inputs(
         &mut self,
         parent: H::Node,


aborgna-q:
Note that this caused transitive lint errors in dependencies too, as any function returning a hugr error or enum wrapping them did also produce a warning.

hugr-passes/src/dataflow/datalog.rs
```diff
@@ -51,6 +51,7 @@ impl<H: HugrView, V: AbstractValue> Machine<H, V> {
     /// or [Conditional](hugr_core::ops::Conditional)).
     /// Any inputs not given values by `in_values`, are set to [`PartialValue::Top`].
     /// Multiple calls for the same `parent` will `join` values for corresponding ports.
+    #[expect(clippy::result_large_err)]
```
ss2165, line 54: too annoying to change?

ss2165 (approved):

hugr-passes/src/dataflow/datalog.rs
aborgna-q: The function is only called directly and non-recursively, with its results unwrapped immediately.
Even if the call doesn't get inlined, increasing the stack pointer by an extra 100kB shouldn't change much performance-wise.

I'll add a reason

PR #2429, opened 2025-07-10
fix: Fixed export of `Call` and `LoadConst` nodes in `hugr-py`.

zrho:
`Call` and `LoadConst` nodes have static inputs which were erroneously exported in `hugr-py`. The function to be called and the constant to be loaded are arguments to the operations in `hugr-model`, so the static input isn't needed. Before this fix, this caused import errors when the same function would be called twice.

### DIFF
--- a/hugr-py/src/hugr/model/export.py
+++ b/hugr-py/src/hugr/model/export.py
@@ -204,6 +204,11 @@ class ModelExport:
                     error = f"Call node {node} is not connected to a function."
                     raise ValueError(error)
 
+                # We ignore the static input edge since the function is passed
+                # as an argument instead.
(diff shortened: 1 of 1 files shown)


hugr-py/src/hugr/model/export.py
```diff
@@ -204,6 +204,11 @@ def export_node(
                     error = f"Call node {node} is not connected to a function."
                     raise ValueError(error)
 
+                # We ignore the static input edge since the function is passed
```
acl-cqc, line 207:
```suggestion
                # We ignore the static input edge (last) since the function is passed
```

acl-cqc (approved):
Looks good, like the assert. Tests?

PR #2434, opened 2025-07-11
feat: Export entrypoint metadata in Python and fix bug in import

zrho:
This PR ensures that the `core.entrypoint` metadata is exported in the Python bindings. It also fixes a bug in `hugr-core` import that added children of the root node to the entrypoint instead of the root.


ss2165:
Either split in to two PRs or include both changes in PR title (otherwise second is lost in changelog)

PR #2435, opened 2025-07-11
feat: ReplaceTypes allows linearizing inside Op replacements

acl-cqc:
* Add `replace_parametrized_op_with` and `replace_op_with` which allow specifying a ReplacementOptions struct
* For now, ReplacementOptions has only one option, to allow running the linearizer on all outports of the replacement
* Test demonstrates - of course the replacement Hugr is invalid (unconnected port) before linearization.

### DIFF
--- a/hugr-passes/src/replace_types.rs
+++ b/hugr-passes/src/replace_types.rs
@@ -204,10 +204,7 @@ pub struct ReplaceTypes {
     param_types: HashMap<ParametricType, Arc<dyn Fn(&[TypeArg]) -> Option<Type>>>,
     linearize: DelegatingLinearizer,
     op_map: HashMap<OpHashWrapper, NodeTemplate>,
-    param_ops: HashMap<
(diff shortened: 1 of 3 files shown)


hugr-passes/src/replace_types.rs
```diff
@@ -171,6 +171,40 @@ fn call<H: HugrView<Node = Node>>(
...
+    process_recursive: bool,
+    linearize: bool,
+}
+
+impl ReplacementOptions {
+    /// Specifies that the replacement should be processed by the same [ReplaceTypes].
+    /// This increases compositionality (in that replacements for different ops do not
+    /// need to account for each other), but would lead to an infinite loop if e.g.
+    /// changing an op for a DFG containing an instance of the same op. Also, note
+    /// that if the recursive processing changes the signature of the replacement,
```
acl-cqc, line 189: I've doc'd this gotcha, see #2442

PR #2436, opened 2025-07-11
test: Better python roundtrip tests, and lots of fixes

aborgna-q:
TODO: Describe the fixes


hugr-py/tests/conftest.py
```diff
@@ -146,45 +147,120 @@ def validate(
...
+                for m1, m2 in zip(loaded.modules, modules, strict=True):
+                    h1_hash = _NodeHash.hash_hugr(m1, "original")
+                    h2_hash = _NodeHash.hash_hugr(m2, "loaded")
+                    assert (
+                        h1_hash == h2_hash
+                    ), f"HUGRs are not the same for {write_fmt} -> {load_fmt}"
+
+
+@dataclass(frozen=True, order=True)
+class _NodeHash:
```
mark-koch, line 204: Should the hash also include the links? E.g. a list of successor hashes here?

hugr-py/src/hugr/ops.py
```diff
@@ -934,6 +942,8 @@ def port_kind(self, port: InPort | OutPort) -> tys.Kind:
         match port:
             case OutPort(_, 0):
                 return tys.ConstKind(self.val.type_())
+            case InPort(_, -1) | OutPort(_, -1):
+                return tys.OrderKind()
```
mark-koch, line 946: Do we allow order edges on consts?

mark-koch (approved):
Looks good 👍 

I'm not sure if #2439 is fixed by this, could you add a test? Also not sure about #2438?

hugr-py/tests/conftest.py
aborgna-q: That would break with non-dag regions.
Since this is a hash of the whole hugr, including CFGs, I had avoid node orderings

hugr-py/tests/test_val.py
```diff
@@ -76,6 +76,7 @@ def test_val_sum_str(value: Sum, string: str, repr_str: str):
     assert repr(sum_val) == repr_str
 
 
+@pytest.mark.skip("FIXME: static array value does not roundtrip-serialize correctly")
```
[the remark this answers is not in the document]
aborgna-q: Fixed the issue and removed the skip.

PR #2442, opened 2025-07-15
feat: ReplaceTypes: recurse on replacements, much deprecation

### LINKED ISSUE #2668: refactor: ReplaceTypes linearization interface
Since #2435 we have developed several places where we jump through hoops to construct an invalid hugr before handing it over for linearization. The execution test for #2666 also faced similar problems.

There are at least a couple of options:
* Rollback #2435, and go down the route of passing an `&ReplaceTypes` to all the callbacks instead. This might seem a more sensible API than ReplacementOptions, but still raises the question of how awkward it is to make the invalid hugr (e.g. using a builder) before giving it to that ReplaceTypes. Moreover it doesn't really help with the execution test.
* Add a "linearizer" extension in hugr-passes, with a drop op (perhaps also copy) - this can replace the existing one in guppy/tket2. Then we'd need a ReplacementOptions that runs the ReplaceTypes on the Hugr (NodeTemplate) returned by the callback (i.e. containing that drop op) - #2442 is aimed at doing this part. Ideally that option might even become the default ReplacementOptions (in time, that's a behaviour-breaking change); and we'd like the rule for lowering drop (using `ReplacementOptions::with_linearization` on an invalid hugr...thankfully we'd only need this in one place) to be included in `ReplaceTypes::default()` or maybe some `ReplaceTypes::new_with_drop()`

acl-cqc:
closes #2668 (by a combination of the two routes, minus the extension).

There are two(/3) goals here:
* firstly, to improve compositionality - if you configure ReplaceTypes to convert type A to type B, and also type B to type C, then if it sees an A it "should" produce a C. (This is not the case at present, you'll get a B.) Likewise for op-replacements containing ops that themselves would be replaced.
* secondly, to allow lowering "drop" without creating an invalid hugr - instead, you can now call `copy_discard_op(ty, 0)`, as demonstrated in the test.
* finally, as a driveby, fixes bug where linearization was applied (only) if the op *or any previous op in the subtree* had been changed

Note that the first is a breaking behaviour change (potentially very breaking, i.e. infinite loop), so to keep this PR non-breaking, I add new methods `set_replace(_parametrized,)_(op,type)` that process the RHS recursively. To satisfy the second goal, the op-replacing callback is passed the ReplaceTypes (hence rolling two breaking signature changes into one PR).. The naming of the new methods is a little odd, largely because the first choice `replace_blah` is already taken, and we keep those methods (with non-recursive behaviour) but deprecated.

(If later we really want to keep non-recursive replacement, we can add that later, but there are enough methods right now!)

Note for consts, there's no real equivalent to "recursive" replacement, because there is no way for the ReplaceTypes to look inside CustomConsts like lists, they are opaque. Instead, recursion must be done manually on the elements, as per e.g. `array_const`: https://github.com/CQCL/hugr/blob/fcd102645c0af644a29d55453e97ee9894d9aa69/hugr-passes/src/replace_types/handlers.rs#L75

### DIFF
--- a/hugr-passes/src/replace_types.rs
+++ b/hugr-passes/src/replace_types.rs
@@ -580,16 +580,18 @@ impl ReplaceTypes {
         root: Node,
         linearize_unchanged_ops: bool,
     ) -> Result<bool, ReplaceTypesError> {
-        let mut changed = false;
-        for n in hugr.descendants(root).collect::<Vec<_>>() {
+        let mut descs = hugr.descendants(root).collect::<Vec<_>>().into_iter();
+        assert_eq!(descs.next(), Some(root));
+        let mut changed = self.change_node(hugr, root)?;
+        // Do not linearize the root's outputs - that's done by the caller if appropriate,
+        // as any copy/discard ops would be *outside* the root
+        for n in descs {
             if self.change_node(hugr, n)? {
                 changed = true;
             } else if !linearize_unchanged_ops {
                 continue;
             }
-            if n != root {
-                self.linearize_outputs(hugr, n)?;
-            }
+            self.linearize_outputs(hugr, n)?;
         }
         Ok(changed)
     }
@@ -680,7 +682,8 @@ impl ReplaceTypes {
                         .map_err(|e| ReplaceTypesError::AddTemplateError(n, Box::new(e)))?;
                     if opts.process_recursive {
                         self.change_subtree(hugr, n, opts.linearize_unchanged)?;
-                        // change_subtree does not linearize it's root, but that's done by our caller
+                        // change_subtree does not linearize its root, just as change_node
+                        // does not linearize the node it's called on; our caller does.
                     } else if opts.linearize_unchanged {
                         let mut descs = hugr.descendants(n);
                         assert_eq!(descs.next(), Some(n));


hugr-passes/src/replace_types.rs
```diff
@@ -171,20 +171,31 @@ fn call<H: HugrView<Node = Node>>(
...
-/// Options for how the replacement for an op is processed.
+// TODO also need to apply to replacement consts:
+/// Options for how a replacement (op or type) is processed.
 ///
-/// May be specified by [ReplaceTypes::replace_op_with] and [ReplaceTypes::replace_parametrized_op_with].
-/// Otherwise (the default), replacements are inserted as is (without further processing).
+/// May be specified by
+/// [ReplaceTypes::replace_op_with], [ReplaceTypes::replace_parametrized_op_with],
+/// [ReplaceTypes::replace_type_opts] or [ReplaceTypes::replace_parametrized_type_opts].
+/// Otherwise (the default), replacements are inserted as given (without further processing).
```
acl-cqc: Of course this is only relevant to the deprecated methods...so I could deprecate this struct. That's pretty messy though as there are a lot of uses ATM (supporting the interface where clients can specify one)...it'll easier to clean up the `ReplaceTypes` struct and remove all the references to this when the deprecated methods are removed, and THEN we can deprecate this struct (which will be unused)

hugr-passes/src/replace_types.rs
```diff
@@ -255,19 +272,25 @@ impl TypeTransformer for ReplaceTypes {
     type Err = ReplaceTypesError;
 
     fn apply_custom(&self, ct: &CustomType) -> Result<Option<Type>, Self::Err> {
-        Ok(if let Some(res) = self.type_map.get(ct) {
-            Some(res.clone())
-        } else if let Some(dest_fn) = self.param_types.get(&ct.into()) {
+        let mut ty = None;
```
mark-koch: nit: maybe `ty_and_opts` or something similar is clearer (especially since you shadow `ty` below)?

hugr-passes/src/replace_types.rs
```diff
@@ -620,12 +775,7 @@ impl<H: HugrMut<Node = Node>> ComposablePass<H> for ReplaceTypes {
         };
         let mut changed = false;
         for region_root in regions {
-            for n in hugr.descendants(*region_root).collect::<Vec<_>>() {
-                changed |= self.change_node(hugr, n)?;
-                if n != hugr.entrypoint() && changed {
-                    self.linearize_outputs(hugr, n)?;
-                }
-            }
+            changed |= self.change_subtree(hugr, *region_root, false)?;
```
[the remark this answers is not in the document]
acl-cqc: Oh, sorry, I completely misunderstand....you mean the last parameter!

Issue #2445, opened 2025-07-17
hugr-py does not encode FuncDefn visibility in model export


aborgna-q:
#2436 added better roundtrip tests that write the test hugr in a given format, convert them to a different one using `hugr convert` if necessary, and load them back.

These work fine for `json -> json`, but if we enable writing hugrs in `model-exts` format then many tests fail with errors like
```
FAILED hugr-py/tests/test_hugr_build.py::test_simple_id - RuntimeError:   2025-07-17T10:11:34.474358Z ERROR hugr: Error decoding HUGR envelope.

Caused by:
    0: failed to import hugr
    1: import failed in context: `define-func` node with id 0
    2: No visibility for FuncDefn
    at hugr-cli/src/main.rs:38
```
That means we successfully produce an encoded model envelope, but it fails to load from the rust side in the `hugr convert` call.

To reproduce this, add `model-exts` to this list
https://github.com/CQCL/hugr/blob/1a301eb818401c314d4d7bac40698ec2e73babe7/hugr-py/tests/conftest.py#L189-L191
And run with `just test-python -k test_simple_id`.

aborgna-q:
Looks like this error may only be happening locally for me.
The tests are encoding the model envelopes using an older version of the rust bindings -.-

PR #2447, opened 2025-07-17
feat!: Use binary envelopes for operation lower_func encoding 

aborgna-q:
Blocked by #2445.

Changes the format used to encode the `LowerFunc::FixedHugr`s in an extension's operation from string-encoded envelopes to base64-encoded binary envelopes.

Adds an `envelope::serde_with::AsBinaryEnvelope` wrapper similar to `AsStringEnvelope`, that lets us include base64-encoded envelopes in a serde definition.

- hugr-py does not yet support loading binary envelopes, so we ignore lowering functions when loading extensions.
- The deserializer can still load older extensions. We check for the envelope magic numbers on the encoded string and skip the base64 deserialization if found.
- Older hugr versions will **not** be able to read extensions encoded this way, and will fail with an "invalid magic" error.

drive-by: Adds a custom deserializing function for `LowerFunc` so errors found while decoding the fixed hugr get communicated back up. The default `serde` derive produced an opaque "no variants match" error due to using `#[serde(untagged)]`.

BREAKING CHANGE: Lowering functions in extension operations are now encoded as binary envelopes. Older hugr versions will error out when trying to load them.

### DIFF
--- a/hugr-core/src/envelope/serde_with.rs
+++ b/hugr-core/src/envelope/serde_with.rs
@@ -329,12 +329,12 @@ macro_rules! impl_serde_as_binary_envelope {
                             // skip the base64 decoding.
                             let reader = std::io::Cursor::new(value.as_bytes());
                             $crate::package::Package::load(reader, Some(extensions))
-                                .map_err(serde::de::Error::custom)
+                                .map_err(|e| serde::de::Error::custom(format!("{e:?}")))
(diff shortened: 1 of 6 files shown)


ss2165:
Would be nice to have a bit more coverage

ss2165 (approved):
beautiful test

PR #2453, opened 2025-07-18
feat: Detect and fail on unrecognised envelope flags

aborgna-q:
When reading an envelope header flags, we ignored any unrecognised bits.

This change adds a new error to both the python and rust decoders, that ensures we detect and report this unexpected flags (nudging the user to update their HUGR version).

This will let us add new config flags in the future that change the payload format, having older versions gracefully fail to read them.

### DIFF
--- a/hugr-core/src/envelope.rs
+++ b/hugr-core/src/envelope.rs
@@ -215,6 +215,15 @@ pub enum EnvelopeError {
         /// Optionally, the feature required to support this format.
         feature: Option<&'static str>,
     },
+    /// The specified payload format is not supported.
+    #[error(
+        "The envelope configuration has unknown {}. Please update your HUGR version.",
+        if flag_ids.len() == 1 {format!("flag #{}", flag_ids[0])} else {format!("flags {}", flag_ids.iter().join(", "))}
+    )]
+    FlagUnsupported {
+        /// The unrecognized flag bits.
+        flag_ids: Vec<usize>,
+    },
     /// Not all envelope formats can be represented as ASCII.
     ///
     /// This error is used when trying to store the envelope into a string.
--- a/hugr-core/src/envelope.rs
+++ b/hugr-core/src/envelope.rs
@@ -215,15 +215,6 @@ pub enum EnvelopeError {
         /// Optionally, the feature required to support this format.
         feature: Option<&'static str>,
     },
-    /// The specified payload format is not supported.
-    #[error(
-        "The envelope configuration has unknown {}. Please update your HUGR version.",
-        if flag_ids.len() == 1 {format!("flag #{}", flag_ids[0])} else {format!("flags {}", flag_ids.iter().join(", "))}
-    )]
-    FlagUnsupported {
-        /// The unrecognized flag bits.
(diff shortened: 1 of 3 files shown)


hugr-core/src/envelope.rs
```diff
@@ -300,6 +300,15 @@ pub enum EnvelopeError {
         #[from]
         source: crate::extension::ExtensionRegistryLoadError,
     },
+    /// The specified payload format is not supported.
+    #[error(
+        "The envelope configuration has unknown {}. Please update your HUGR version.",
+        if flag_ids.len() == 1 {format!("flag #{}", flag_ids[0])} else {format!("flags {}", flag_ids.iter().join(", "))}
```
lmondada, line 306: Nit: you're using # in the singular case. You can use it too in the plural by changing the join argument to ", #" as well as adding one to the string

hugr-core/src/envelope/header.rs
```diff
@@ -224,8 +232,10 @@ impl EnvelopeHeader {
         let format_bytes = [self.format as u8];
         writer.write_all(&format_bytes)?;
         // Next is the flags byte.
-        let mut flags = 0b01000000u8;
-        flags |= u8::from(self.zstd);
+        let mut flags = DEFAULT_FLAGS;
+        if self.zstd {
```
lmondada, line 236: Thanks that's much clearer!

hugr-core/src/envelope/header.rs
```diff
@@ -259,7 +269,16 @@ impl EnvelopeHeader {
         // Next is the flags byte.
         let mut flags_bytes = [0; 1];
         reader.read_exact(&mut flags_bytes)?;
-        let zstd = flags_bytes[0] & 0x1 != 0;
+        let flags: u8 = flags_bytes[0];
+
+        let zstd = flags & ZSTD_FLAG != 0;
+
+        // Check if there's any unrecognized flags.
+        let other_flags = (flags ^ DEFAULT_FLAGS) & !ZSTD_FLAG;
```
lmondada, line 277: I'd find it easier to read as

```
let other_flags = flags & !(DEFAULT_FLAGS | ZSTD_FLAG);
```

hugr-py/src/hugr/envelope.py
```diff
@@ -46,6 +46,12 @@
 # This is a hard-coded magic number that identifies the start of a HUGR envelope.
 MAGIC_NUMBERS = b"HUGRiHJv"
 
+# The all-unset header flags configuration.
+# Bit 7 is always set to ensure we have a printable ASCII character.
```
lmondada, line 50: If you can't reuse the same consts as un Rust, can you maybe add a comment about which file their Rust counterparts are defined in?

lmondada (approved):

hugr-core/src/envelope/header.rs
aborgna-q: It's not quite the same. DEFAULT_FLAGS has an always true value, so if there's a zero there we have to complain too.
(I should make that more explicit)

PR #2459, opened 2025-07-21
ci: Disable zstd for miri tests

### LINKED ISSUE #2457: 💥 Unsoundness check fail on main
The unsoundness check for `CQCL/hugr` failed.

[Please investigate](https://github.com/CQCL/hugr/actions/runs/16408402159).

aborgna-q:
Fixes #2457

Ci run: https://github.com/CQCL/hugr/actions/runs/16420064018/job/46395888590

drive-by: Use `nextest` for miri runs. This way we can use multiple threads for testing, and we also report each test duration in the log (so we know if any is particularly long and should be skipped).
    See https://nexte.st/docs/integrations/miri/#benefits
    
drive-by: Skip tests that take over 5mins to run (See linked CI run, most test take 10-20s).
  In general any test defining new extensions seems to be really slow. Things dealing with `Term::is_supertype` also seem slow.
  I don't know if that's because of miri, or because of Hugr...
image

With the improved parallelism and skipped tests the unsoundness test now takes <1h rather than 5h :)

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,7 @@
+
+
+[profile.default-miri]
+# Fail if tests take more than 5 mins.
+# Those tests should be skipped in `.github/workflows/unsondness.yml`.
+slow-timeout = { period = "60s", terminate-after = 5 }
+fail-fast = false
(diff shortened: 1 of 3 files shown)


aborgna-q:
We should definitely try it for CI.

My only worry for local dev is adding another tool to setup but it's easy to `cargo binstall` it, so I'm ambivalent.

PR #2465, opened 2025-07-23
fix: Ensure SumTypes have the same json encoding in -rs and -py

### LINKED ISSUE #2454: `hugr-core` and `hugr-py` JSON serialize sum values differently
There is a mismatch in how `hugr-core` and `hugr-py` serialize sum values as JSON.

When `hugr-py` serializes a `SumValue`, the `typ` is a `SumType` which includes a `t` property (with the comment "This seems to be required for nested discriminated unions to work"). When `hugr-core` serializes a `hugr_core::ops::constant::Sum`, it gets converted into a `hugr_core::ops::constant::SerialSum` whose `typ` is an `Option<hugr_core::types::SumType>`. But `hugr_core::types::SumType` does not include a `t` property. The `t` property is only included on the Rust side when the `hugr_core::types::SumType` is first wrapped into a `hugr_core::types::TypeBase`. The `hugr_core::types::TypeBase` on serialization gets converted into a `hugr_core::types::serialize::SerSimpleType`, which has the `t` tag.

This bug showed up in the roundtrip tests for the `hugr-model` based serialization but it does not for the `json` serialization. I suspect this is due to the builtin support for arrays in `hugr-model` making the bug visible by actually looking into the value. Meanwhile the `json` serialization path deserializes the value as a `hugr_core::ops::constant::OpaqueValue` pointing at a `hugr_core::ops::constant::custom::CustomSerialized` that just contains the value as a `serde_json::Value`. Therefore roundtripping to `json` can simply parrot back the exact JSON that Python provided, even if `hugr-core` would have produced different JSON.

aborgna-q:
Fixes #2454

drive-by: Compare encoded const values as JSON dicts instead of strings in the python roundtrip tests, so they ignore field order.

### DIFF
--- a/hugr-core/src/types.rs
+++ b/hugr-core/src/types.rs
@@ -4,7 +4,7 @@ mod check;
 pub mod custom;
 mod poly_func;
 mod row_var;
-mod serialize;
+pub(crate) mod serialize;
 mod signature;
 pub mod type_param;
 pub mod type_row;
--- a/hugr-core/src/types/serialize.rs
+++ b/hugr-core/src/types/serialize.rs
@@ -14,7 +14,7 @@ use crate::types::{Term, Type};
 
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug)]
 #[serde(tag = "t")]
-pub(super) enum SerSimpleType {
+pub(crate) enum SerSimpleType {
     Q,
     I,
     G(Box<FuncValueType>),
--- a/hugr-core/src/ops/constant.rs
+++ b/hugr-core/src/ops/constant.rs
@@ -9,6 +9,7 @@ use std::hash::{Hash, Hasher};
 use super::{NamedOp, OpName, OpTrait, StaticTag};
 use super::{OpTag, OpType};
 use crate::envelope::serde_with::AsStringEnvelope;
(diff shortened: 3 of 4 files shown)


hugr-core/src/types.rs
```diff
@@ -4,7 +4,7 @@ mod check;
 pub mod custom;
 mod poly_func;
 mod row_var;
-mod serialize;
+pub(crate) mod serialize;
```
lmondada, line 7: I leave it up to you, but you could keep the stronger encapsulation that comes with a private `serialize` module by moving `impl TryFrom<SerialSum> for Sum` from `hugr-core/src/ops/constant.rs` into `hugr-core/src/types/serialize.rs`

hugr-core/src/types/serialize.rs
```diff
@@ -14,7 +14,7 @@ use crate::types::{Term, Type};
 
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug)]
 #[serde(tag = "t")]
-pub(super) enum SerSimpleType {
+pub(crate) enum SerSimpleType {
```
lmondada, line 17: see comment above

lmondada (approved):

PR #2467, opened 2025-07-24
feat: Non-region entrypoints in `hugr-model`.

zrho:
This PR allows to mark any node as an entrypoint in `hugr-model`.


aborgna-q (approved):
Can confirm that this works with the roundtrip tests in #2449.

Am I right to say that this change is backwards-compatible? (The import code didn't change, so both entrypoint metadata positions are valid)

PR #2468, opened 2025-07-24
feat: Version number in hugr binary format.

zrho:
This PR adds a (major, minor) version number to the binary hugr format and validates it on import.


hugr-model/capnp/hugr-v0.capnp
```diff
@@ -20,6 +20,8 @@ using LinkIndex = UInt32;
 
 struct Package {
     modules @0 :List(Module);
+    versionMajor @1 :UInt32;
+    versionMinor @2 :UInt32;
```
aborgna-q, line 24: Put these in a standalone `Version` struct?

hugr-model/src/v0/mod.rs
```diff
@@ -91,6 +91,12 @@ use smol_str::SmolStr;
 use std::sync::Arc;
 use table::LinkIndex;
 
+/// The current major version of the `hugr-model` format.
+pub const VERSION_MAJOR: u32 = 0;
+
+/// The current minor version of the `hugr-model` format.
+pub const VERSION_MINOR: u32 = 0;
+
```
aborgna-q, line 99: We should add the current version to the main crate doc, and explain the update/compatibility policy.

I'd also like to enforce we update the version in CI, so it would be better to have a single source of truth for these. Could we have a `version.txt` that get's `include!`d and parsed here?

aborgna-q: Perhaps a name like `MODEL_VERSION` may be more clear

hugr-model/src/version.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub minor: u32,
+}
+
+impl Version {
+    /// The current version of the HUGR model format.
+    pub fn current() -> Self {
+        static VERSION: LazyLock<Version> = LazyLock::new(|| {
+            include_str!("../version.txt")
+                .trim()
+                .parse()
```
aborgna-q, line 21: There's a `const trim_ascii`, but `const u32::from_ascii` is still in nightly u.u

aborgna-q (approved):
LGTM 👍

PR #2469, opened 2025-07-24
ci: Move file change checks to a composite action

aborgna-q:
Cleans up the workflows by moving the change detection and its override logic to a composite action.


.github/actions/check-changes/action.yml
```diff
@@ -0,0 +1,54 @@
...
+          echo "Overriding due to label 'run-ci-checks'"
+          echo "out=true" >> $GITHUB_OUTPUT
+        elif [ "$DEFAULT_BRANCH" == "true" ]; then
+          echo "Overriding due to running on the default branch"
+          echo "out=true" >> $GITHUB_OUTPUT
+        fi
+      env:
+        OVERRIDE_LABEL: ${{ github.event_name == 'pull_request' && contains( github.event.pull_request.labels.*.name, 'run-ci-checks') }}
+        DEFAULT_BRANCH: ${{ github.ref_name == github.event.repository.default_branch }}
+    - uses: dorny/paths-filter@v3
```
lmondada, line 51: can you skip this step if override is on?

```
if: steps.override.outputs.out != 'true'
```

it might be that you cannot skip it because then the references to `steps.filter.outputs.xyz` would be undefined, I'm not sure what the behaviour of the runners is in this case

.github/workflows/ci-py.yml
```diff
@@ -23,21 +23,17 @@ jobs:
...
     outputs:
-      python: ${{ github.ref_name == github.event.repository.default_branch || steps.filter.outputs.python }}
-      extensions: ${{ github.ref_name == github.event.repository.default_branch || steps.filter.outputs.std-extensions }}
+      python: ${{ steps.filter.outputs.python }}
+      extensions: ${{ steps.filter.outputs.llvm }}
     steps:
       - uses: actions/checkout@v4
-      - uses: dorny/paths-filter@v3
+      - uses: ./.github/actions/check-changes
         id: filter
```
lmondada, line 36: I'd rename this id to `check-changes`

lmondada (approved):

.github/actions/check-changes/action.yml
aborgna-q: Undefined is false-y, so it should work.
There's short-circuiting, so we can invert the `||` terms and it should also work.

PR #2471, opened 2025-07-24
feat: Use semver crate for -model version, and include in docs

aborgna-q:
image

We can bump the patch version in the file to mark forward/backward compatible changes (e.g. docs), and it will get ignored by the compatibility checks.


ss2165 (approved):
thank you

aborgna-q:
(Ignoring the semver-checks warnings since it's removing an unpublished struct)

PR #2472, opened 2025-07-24
ci: Avoid `release-plz` closing PRs unnecessarily

aborgna-q:
If two instances of the `release-plz` workflow ran at the same time (due to multiple rust PRs being merged together), the newer one may try to update the release branch with stale ref info, and resort to re-creating the release PR instead.

This change avoids that situation by ensuring no two instances of the workflow run concurrently for the same branch.

- Stale info error: https://github.com/CQCL/hugr/actions/runs/16497826514/job/46647892322#step:5:305
- PR closed for no reason: https://github.com/CQCL/hugr/pull/2443


aborgna-q:
Uhm, this may actually end up cancelling runs after a release PR is merged.

Since we set [`release_always = false`](https://release-plz.dev/docs/config#the-release_always-field) in the config, later commits to `main` will not trigger the release so we may end up missing the publication 🤔

aborgna-q:
Ok, I split the PR-creating step from the registry-publish step and only added the concurrency limit to the former.
This should work directly now.

croyzor (approved):

Issue #2356, opened 2025-06-17
Link Module-rooted Hugrs (w/ configurable policies)


acl-cqc:
Using #2354 we can decide what FuncDecls(/FuncDefns(?)) in the inserted Hugr are actually proxies (or replacements?) for FuncDefns(/FuncDecls) in the target and provide directives to #2355 to wire up the inserted nodes accordingly.

Note the `insert_hugr`-like idea of inserting the entrypoint subtree into a specified location, deferred until #2517.

PR #2481, opened 2025-07-25
feat: [RFC] linking by name when inserting Hugrs

acl-cqc:
UPDATE: needs splitting; this should address #2356 only, not #2517.
----
Follows #2285, with a number of breaking changes to functions introduced in that PR (which thankfully was never merged - there are no breakages against `main`!):
* `insert_hugr_with_defns` renamed to `insert_hugr_link_nodes`, similarly `insert_from_view_with_defns`, + builder meths
* The values in the map are now `enum NodeLinkingDirective`, and allow a newly-inserted module-child (e.g. FuncDefn) to replace any number of existing module-children (e.g. FuncDecls).
* New methods `insert_hugr_link_names` and `insert_from_view_link_names` take a `NameLinkingPolicy` allowing to automatically determine what (public) FuncDefns should replace what FuncDecls/FuncDefns according to their names. There is a wide range of behaviours available here.
   * No tests yet! So this is not ready. But I wanted feedback on the API. 
* `insert_hugr` and `insert_from_view` now do some linking-by-name on public functions by default. However, these have no way to report an error, so I've kept them restricted to creating invalid hugrs if they can't do linking properly. This is very much open to debate, see comments...

- [ ] Big TODO, more tests
- [ ] Currently I am thinking we can resolve much of the `insert_hugr` dilemma if we allow linking between the target Hugr's module-children, and the new nodes from the inserted hugr, *even if* the latter are not beneath the module root (i.e. are beneath a bogus second module node). Hence *e.g.* a call in a host Hugr to a FuncDecl (in the correct place) could become a call to a FuncDefn *in the wrong place* (beneath a second module node). This feels a bit like we are making more stuff invalid but may avoid the special casing. However, I'm still not sure that it isn't better just to make a valid hugr...(i.e. inserting a module beneath the module-root, doesn't - it just inserts the *children* of the former, beneath the latter).


hugr-core/src/hugr/hugrmut.rs
```diff
@@ -180,102 +181,218 @@ pub trait HugrMut: HugrMutInternals {
...
     /// specified `parent` in this Hugr, and `children` of the Module root of `other`
-    /// are either inserted with their subtrees or linked according to their [InsertDefnMode].
+    /// are either inserted with their subtrees or linked according to their [NodeLinkingDirective].
     ///
     /// # Errors
     ///
     /// * If `children` are not `children` of the root of `other`
     /// * If `other`s entrypoint is among `children`, or descends from an element
-    ///   of `children` with [InsertDefnMode::Add]
+    ///   of `children` with [NodeLinkingDirective::Add]
```
acl-cqc, line 237: So there are sane behaviours for inserting module-entrypoint Hugrs here: either set parent=None; or set children==empty (and accept 2*module nodes). Almost anything else raises error (as the entrypoint-subtree needs copying into two places), which is OK.

hugr-core/src/hugr/hugrmut.rs
```diff
@@ -180,102 +181,218 @@ pub trait HugrMut: HugrMutInternals {
...
+    ///   that will be added according to the [NameLinkingPolicy]
+    ///
+    /// # Panics
+    ///
+    /// If `parent` is `Some` but not in the graph.
+    ///
+    /// [FuncDefn]: crate::ops::FuncDefn
+    /// [MultipleImplHandling::ErrorDontInsert]: crate::hugr::linking::MultipleImplHandling::ErrorDontInsert
+    #[allow(clippy::type_complexity)]
+    fn insert_hugr_link_names(
```
acl-cqc, line 271: This is also fine for module-entrypoint hugrs: again either set parent=None, or use NameLikingPolicy::AddNone (and you get two module nodes). Pretty much anything else should cause an error so I think that's ok?

hugr-core/src/hugr/hugrmut.rs
```diff
@@ -180,102 +181,218 @@ pub trait HugrMut: HugrMutInternals {
...
-    /// nonlocal edges (including [Const] / [Function]) into the entrypoint-subtree.
-    /// (See [Self::insert_from_view_with_defns].)
+    /// Public functions (children of the module-root of `other`) are also copied under the
+    /// module-root of `self`, with linking according to [NameLinkingPolicy::default_for_view]:
+    /// * Public [FuncDecl]s in `other` are replaced by public functions in `self` with the same
+    ///   name and signature
+    /// * Public [FuncDefn]s in `other` are replaced by public [FuncDefn]s in `self` with the
+    ///   same name and signature, but replace any such public [FuncDecl]s.
+    ///
+    /// (ALAN NOTE: alternatively we could just use NameLinkingPolicy::AddNone ?
```
acl-cqc, line 299: The idea that `insert_from_view` should not insert private functions came from the previous PR; you still have the private functions as you still have the view you were inserting from. However, you have made it much harder to track the edges, and it seems a bit inconsistent to copy public functions but not private, so I'm not sure anymore....

hugr-core/src/hugr/hugrmut.rs
```diff
@@ -180,102 +181,218 @@ pub trait HugrMut: HugrMutInternals {
...
+    /// [FuncDefn]: crate::ops::FuncDefn
+    /// [FuncDecl]: crate::ops::FuncDecl
     fn insert_from_view<H: HugrView>(
         &mut self,
         root: Self::Node,
         other: &H,
     ) -> InsertionResult<H::Node, Self::Node> {
-        self.insert_from_view_with_defns(root, other, HashMap::new())
-            .expect("No defns being inserted so no possibility of error")
+        // ALAN TODO: use empty NodeLinkingPolicy if other.entrypoint() == other.module_root() ?
```
acl-cqc, line 321: i.e. much like `insert_hugr`; when that's more settled I'll do the same thing here

acl-cqc:
Thanks @doug-q - yes will separate into two, also respin to follow #2521.

PR #2485, opened 2025-07-28
fix(py): Python builder for function declarations did not create static output port

zrho:
The Python builder for function declarations did not create a static output port when the function declaration is not used. This caused the json to json roundtrip test to fail, since the json deserialisation does add the static output port.


ss2165 (approved):
nice catch

PR #2487, opened 2025-07-28
feat: Support LineConvexChecker

lmondada:


croyzor (approved):

lmondada:
An alternative is to play it safe (i.e. define `TopoConvexChecker` not as a type alias but a newtype wrapping `ConvexChecker`), and then already create the breaking PR that will undo this (as well as generalising the `fn` signatures currently marked as TODO). Then we'd leave the breaking PR hanging around until the next breaking release?

There shouldn't be too many merge conflicts arising, given that it only touches this one file.

aborgna-q:
Relevant semver-checks issue: [obi1kenobi/cargo-semver-checks#609 (comment)](https://www.github.com/obi1kenobi/cargo-semver-checks/issues/609#issuecomment-1856463681).

I think we're fine overwriting the breaking change marker here.
We should do the overriden patch release soon after merging this to avoid headaches later.

lmondada:
> Relevant semver-checks issue: [obi1kenobi/cargo-semver-checks#609 (comment)](https://www.github.com/obi1kenobi/cargo-semver-checks/issues/609#issuecomment-1856463681).

Our case is slightly different: the type alias and struct are both defined within this crate, so the justification does not apply. However, in our case, the aliased struct isn't equal to the old one (it is equal once some of the generics are fixed).

Do you think I should open an issue in semver-checks?

aborgna-q:
Uhm, you're right though it's quite the edge case.
May be worth reporting it if you have some free at some point.

PR #2488, opened 2025-07-28
build: relax typing-extensions constraint

qartik:
selene-core requires a later version, see https://github.com/CQCL/selene/blob/ef2cd14b27757b805872a5d42d3dbb3e20ce038d/selene-core/pyproject.toml#L11

The current constraint is too strict.

### DIFF
--- a/hugr-py/pyproject.toml
+++ b/hugr-py/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "pydantic>=2.8,<2.12",
     "pydantic-extra-types>=2.9.0",
     "semver>=3.0.2",
-    "typing-extensions~=4.12",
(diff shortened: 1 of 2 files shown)


lmondada:
I'm confused, shouldn't `~=4.12` mean `==4.*, >= 4.12` as per https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release?

In which case, it should be compatible with `~=4.14`... I'm not trying to nitpick, I just don't know if there is a reason we didn't allow `< 4.12` and I am unsure whether the relaxation would be useful to you.

qartik:
Yes, you are right. Looks like I confused it to mean `>=4.12,<4.13`

PR #2497, opened 2025-07-30
feat: Add option to render a subgraph of a hugr.

cqc-alec:
For moderately sized hugrs, rendering the whole hugr produces sprawling images that are difficult to analyse. This feature allows one to select a region to render.

Example output:
[h_10.pdf](https://github.com/user-attachments/files/21509633/h_10.pdf)

### DIFF
--- a/hugr-py/src/hugr/hugr/render.py
+++ b/hugr-py/src/hugr/hugr/render.py
@@ -92,9 +92,20 @@ class DotRenderer:
 
     def __init__(self, config: RenderConfig | None = None) -> None:
         self.config = config or RenderConfig()
+        self.nodes: set[Node] = set()
 
-    def render(self, hugr: Hugr) -> Digraph:
-        """Render a HUGR to a graphviz dot object."""
+    def render(self, hugr: Hugr, root: Node | None = None) -> Digraph:
+        """Render a HUGR to a graphviz dot object.
+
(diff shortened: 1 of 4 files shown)


hugr-py/src/hugr/hugr/render.py
```diff
@@ -266,6 +284,7 @@ def _viz_node(self, node: Node, hugr: Hugr, graph: Digraph) -> None:
                     self._viz_node(child, hugr, sub)
                 html_label = self._format_html_label(**label_config)
                 sub.node(f"{node.idx}", shape="plain", label=f"<{html_label}>")
+                self.nodes.add(node)
```
mark-koch, line 287: Maybe hoist this out of the if statement since you're doing the same in both branches?

mark-koch (commented):
Thanks, that's a very useful feature 👍

I'm not a big fan of the default dot rendering for external nodes:
image
Could you add those nodes manually and use `self._format_html_label` to get a matching style?

Also, would be good to add a snapshot test (e.g. see `test_html_labels` in `hugr-py/tests/test_hugr_build.py`)

hugr-py/src/hugr/hugr/render.py
cqc-alec: Done.

PR #2521, opened 2025-08-13
feat: add trait+funcs for linking Hugrs explicitly by Node

### LINKED ISSUE #2355: Linking with explicit Node directives (no names)
detailing how the inserted hugr should be linked with the target/host

acl-cqc:
Closes #2355.

This adds a new `trait HugrLinking: HugrMut` with methods `insert_link_hugr_by_node` and `insert_link_view_by_node` that allow inserting a Hugr/View into `self` - the entrypoint subtree (anywhere) and/or the module-children (under the module-root of `self`). An error is raised if you're trying to insert the entrypoint-subtree into two places. The added nodes can be linked i.e. static edges added from module-children to uses, both from new to existing or vice versa. The new methods are default-implemented on top of `HugrMut::insert_`(`view_`)`forest`, and there's a blanket impl for any `HugrMut`, so the trait is just to keep these methods separated out.

Note that it would be possible to deprecate `insert_hugr` now, although I have not: `insert_link_hugr_by_node(other, Some(parent), HashMap::new())` does the same job (explicitly not copying any module-children), and returns an error, which is probably what we want.


hugr-core/src/builder.rs
```diff
@@ -177,6 +178,16 @@ pub enum BuildError {
...
 
+    /// From [Dataflow::add_hugr_with_wires_link_nodes]
+    #[error{"In inserting Hugr: {0}"}]
+    HugrInsertionError(#[from] NodeLinkingError<Node, Node>),
+
+    /// From [Dataflow::add_hugr_view_with_wires_link_nodes].
+    /// Note that because the type of node in the [NodeLinkingError] depends
+    /// upon the view being inserted, we convert the error to a string here.
+    #[error("In inserting HugrView: {0}")]
+    HugrViewInsertionError(String),
```
acl-cqc, line 189: I considered a single error variant with a `NodeLinkingError<String, Node>` but would have needed a NodeLinkingError::map to convert the unknown-nodes into strings

hugr-core/src/hugr/linking.rs
```diff
@@ -0,0 +1,452 @@
...
+    }
+    Ok(trns)
+}
+
+fn link_by_node<SN: HugrNode, TGT: LinkHugr + ?Sized>(
+    hugr: &mut TGT,
+    transfers: Transfers<SN, TGT::Node>,
+    node_map: &mut HashMap<SN, TGT::Node>,
+) {
+    // Resolve `use_existing` first in case the existing node is also replaced by
```
acl-cqc, line 244: No test yet, coming up...

Alternatively we could error if the same existing-node is target of both a `replace` and a `UseExisting`.

hugr-core/src/builder.rs
```diff
@@ -352,4 +363,33 @@ pub(crate) mod test {
         );
         hugr
     }
+
+    /// A DFG-entrypoint Hugr (no inputs, one bool_t output) containing two calls,
+    /// to a FuncDefn and a FuncDecl each bool_t->bool_t, and their handles.
```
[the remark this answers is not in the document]
lmondada: oh i see you just copied this over. Might be worth fixing anyways :)

hugr-core/src/hugr/hugrmut.rs
```diff
@@ -833,39 +840,7 @@ mod test {
                     .chain(call2.then_some((decl.node(), h.module_root())).into_iter()),
             );
             h.insert_forest(insert, roots).unwrap();
```
lmondada, line 842: this is in your previous PR I think, but anyways:

in the docs of `fn insert_forest`, can you change any reference to `root_parents` to `roots`, to match the argument name?

lmondada: I was going to make another comment about using rstest instead of the `for` loop, but I see that this seems already taken care of in #2518.

hugr-core/src/hugr/linking.rs
```diff
@@ -0,0 +1,520 @@
...
+
+use itertools::Either;
+
+use crate::{
+    Hugr, HugrView, Node,
+    core::HugrNode,
+    hugr::{HugrMut, internal::HugrMutInternals},
+};
+
+/// Methods for linking Hugrs, i.e. merging the Hugrs and adding edges between old and inserted nodes.
```
lmondada, line 13: A bit shorter so it fits in 80chars

```suggestion
/// Methods that merge Hugrs, adding edges between old and inserted nodes.
```

hugr-core/src/hugr/linking.rs
```diff
@@ -0,0 +1,520 @@
...
+    /// * If `children` are not `children` of the root of `other`
+    /// * If `parent` is Some, and `other.entrypoint()` is either
+    ///   * among `children`, or
+    ///   * descends from an element of `children` with [NodeLinkingDirective::Add]
+    ///
+    /// # Panics
+    ///
+    /// If `parent` is `Some` but not in the graph.
+    #[allow(clippy::type_complexity)]
+    fn insert_from_view_link_nodes<H: HugrView>(
```
lmondada, line 36: These names make sense, but I wonder if we could shorten them a bit -- or do you plan on renaming to `insert_hugr` and `insert_from_view` once those are deleted?

I can suggest:
- `insert_link_from_view`
- `insert_link_hugr`

I think they are just as clear and sound like natural "evolutions" of the current API. Up to you :)

hugr-core/src/hugr/linking.rs
```diff
@@ -0,0 +1,520 @@
...
+    core::HugrNode,
+    hugr::{HugrMut, internal::HugrMutInternals},
+};
+
+/// Methods for linking Hugrs, i.e. merging the Hugrs and adding edges between old and inserted nodes.
+///
+/// This is done by module-children from the inserted (source) Hugr replacing, or being replaced by,
+/// module-children already in the target Hugr; static edges from the replaced node,
+/// are transferred to come from the replacing node, and the replaced node(/subtree) then deleted.
+pub trait LinkHugr: HugrMut {
```
lmondada, line 18: I'm fine with leaving `insert_hugr` and `insert_from_view` as they are for the moment, but could you add a note in their docs pointing users here if they are considering using them?

E.g. say something like "more specialised methods that handle copying multiple module root children and preserving edge connectivity can be found in [`LinkHugr`].

hugr-core/src/hugr/linking.rs
```diff
@@ -0,0 +1,520 @@
...
+    Hugr, HugrView, Node,
+    core::HugrNode,
+    hugr::{HugrMut, internal::HugrMutInternals},
+};
+
+/// Methods for linking Hugrs, i.e. merging the Hugrs and adding edges between old and inserted nodes.
+///
+/// This is done by module-children from the inserted (source) Hugr replacing, or being replaced by,
+/// module-children already in the target Hugr; static edges from the replaced node,
+/// are transferred to come from the replacing node, and the replaced node(/subtree) then deleted.
```
lmondada, line 17: Have you thought of other names for this trait? What I dislike about `LinkHugr` is that it sounds like `LinkView` in portgraph, which is read-only; a lot of our traits for mutation have a `Mut` suffix, but I cannot think of a good name that would keep this convention.

`InsertHugr` or `HugrInsert` are pretty straight-forward, though? Maybe `InsertHugrMut` if you want to highlight that this is a "specialisation" of `HugrMut`...

Also, is there a reason this cannot be part of the `HugrMut` trait? Or at least provide a blanket implementation for any `H: HugrMut`?

I think there is a tradeoff between avoiding huge bloated traits and on the other hand a proliferation of traits... I find more traits make API exploration much harder as the available methods might get missed by Intellisense. A pointer from the `HugrMut` docs to here would already help with that (see my other comment).

hugr-core/src/hugr/linking.rs
```diff
@@ -0,0 +1,520 @@
...
+        let mut h = simple_dfg_hugr();
+        h.insert_from_view(h.entrypoint(), &insert);
+        check_calls_defn_decl(&h, false, false);
+
+        let mut h = simple_dfg_hugr();
+        h.insert_hugr(h.entrypoint(), insert);
+        check_calls_defn_decl(&h, false, false);
+
+        // Specify which decls to transfer. No real "linking" here though.
+        for (call1, call2) in [(false, false), (false, true), (true, false), (true, true)] {
```
lmondada, line 307: here you could use rstest #[case] or #[values] as in the `test_insert_forest` case :)

lmondada (changes requested):
The design looks solid, I like it! My main discussion points are... naming, as usual :)

hugr-core/src/hugr/hugrmut.rs
acl-cqc: Yes, sorry, I'm rather behind on merging, hoping 2518 goes in soon!

acl-cqc:
Ok, #2518 now merged into main, and I hope that's fixed a number of issues here. As mentioned, I've gone with `trait HugrLinking` and `add_hugr_link_nodes` / `add_view_link_nodes` (and builder versions have extra `with_wires`, except ModuleBuilder - I've just added two new methods there too).

hugr-core/src/builder/module.rs
```diff
@@ -220,17 +221,46 @@ impl<T: AsMut<Hugr> + AsRef<Hugr>> ModuleBuilder<T> {
 
         Ok(AliasID::new(node, name, bound))
     }
+
+    /// Adds some module-children of another Hugr to this module, with
```
lmondada, line 225:
```suggestion
    /// Add some module-children of another Hugr to this module, with
```

hugr-core/src/builder/module.rs
```diff
@@ -220,17 +221,46 @@ impl<T: AsMut<Hugr> + AsRef<Hugr>> ModuleBuilder<T> {
...
+    /// this map is empty, nothing is added.
+    pub fn link_hugr_by_node(
+        &mut self,
+        other: Hugr,
+        children: NodeLinkingDirectives<Node, Node>,
+    ) -> Result<InsertedForest, NodeLinkingError> {
+        self.hugr_mut().add_hugr_link_nodes(None, other, children)
+    }
+
+    /// Copies module-children from a HugrView to this module, with
```
lmondada, line 239:
```suggestion
    /// Copy module-children from a HugrView to this module, with
```

hugr-core/src/builder/module.rs
lmondada: slightly subjective, but I think we _mostly_ use the imperative form.

hugr-core/src/hugr/linking.rs
lmondada: Excellent, I think `HugrLinking` is very good.

hugr-core/src/hugr/linking.rs
```diff
@@ -33,12 +33,12 @@ pub trait LinkHugr: HugrMut {
     ///
     /// If `parent` is `Some` but not in the graph.
     #[allow(clippy::type_complexity)]
-    fn insert_from_view_link_nodes<H: HugrView>(
+    fn add_view_link_nodes<H: HugrView>(
```
lmondada, line 36: Ok, good name proposals!

1. I prefer `insert` to `add` (`add` sounds to me like adding an element to a container, whereas `insert` is a more complex process?) I admit this is subjective, so up to you.

2. Given your explanation, I think a good suffix for the names would be `_by_nodes`. Names with `by` are pretty idiomatic.

3. If you say linking by name will be the most common use case, then I am not fussed if the method names for linking explicitly by node is quite verbose. As you suggest I would just drop the `by_nodes` suffix for the link-by-name versions.

So maybe `insert_link_view_by_nodes` and `insert_link_hugr_by_nodes`? I don't know at this point 😅

lmondada (approved):

hugr-core/src/hugr/linking.rs
acl-cqc: Ok, so
* I think you're right about `insert` rather than `add`: I think "insert" is "put the entrypoint-subtree under a new parent" whereas "link" is "do stuff with the module children" (only). Using "add" for "maybe insert" is a faux pas, will just go with insert.
* Let's also establish the "link....by_nodes" convention (foreshadowing upcoming+shorter "link...." that's by name), replacing current "...._link_nodes".
* I also think rejecting names for being too long may be a bad move here. There's an old idea that the more a method does, the longer its name should be. These methods do quite a lot, so long names are at least ok....
* Hence: I propose: `HugrMut::insert_link_hugr_by_node` (and s/hugr/view/), Dataflow builder `add_link_hugr_by_node` (where add in the builder means insert elsewhere but nvm), Module builder `link_hugr_by_node` (without the add: it *only* does stuff with module children).

Will enact shortly unless you object @lmondada...

acl-cqc: oh, by_node singular or plural? I think I favour singular

PR #2528, opened 2025-08-25
feat: SiblingSubgraph supports function calls

lmondada:
ended up being a bit more work than expected! With this PR, `SiblingSubgraph` supports incoming boundary static edges that link a function declared/defined outside of the subgraph to a call within the subgraph.

I've taken the opportunity to clean up the docs and error handling for any non-value (and non-function static edges).


hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -598,6 +629,90 @@ impl<N: HugrNode> SiblingSubgraph<N> {
...
+    inputs.retain_mut(|calls| {
+        let Some(&(n, p)) = calls.first() else {
+            return true;
+        };
+        let op = hugr.get_optype(n);
+        if op.static_input_port() == Some(p)
+            && op
+                .static_input()
+                .expect("static input exists")
+                .is_function()
```
aborgna-q, line 647: Not necessary for this PR, would we need to change much to also support external const loads?
`Load`s have an `EdgeKind::Const` static input, but the rest seems quite similar to function calls.

lmondada: ah good point! It would probably be straightforward. I'll open an issue for that.

PR #2529, opened 2025-08-25
feat: Add method to link Hugr modules (linking pt3)

### LINKED ISSUE #2356: Link Module-rooted Hugrs (w/ configurable policies)

acl-cqc:
closes #2356, adding new methods `link_module` and `link_module_view`.
* These ignore the entrypoint.
* But take a 'policy' that acts on names of public functions to determine how to link the two Hugrs.

Does *not* use callgraph/reachability; this is coming in #2555.

There are a few other suggestions for even further into the future in #2687.

### DIFF
--- a/hugr-core/src/hugr/linking.rs
+++ b/hugr-core/src/hugr/linking.rs
@@ -321,22 +321,10 @@ pub enum NameLinkingError<SN: Display, TN: Display + std::fmt::Debug> {
 }
 
 impl NameLinkingPolicy {
-    /// Makes a new instance that specifies to handle
-    /// [signature conflicts](Self::on_signature_conflict) by failing with an error and
-    /// multiple [FuncDefn]s according to `multi_defn`.
-    ///
-    /// [FuncDefn]: crate::ops::FuncDefn
-    pub fn err_on_conflict(multi_defn: impl Into<OnMultiDefn>) -> Self {
-        Self {
-            multi_defn: multi_defn.into(),
-            sig_conflict: OnNewFunc::RaiseError,
-        }
-    }
-
     /// Makes a new instance that specifies to keep both decls/defns when (for the same name)
     /// they have different signatures or when both are defns. Thus, an error is never raised;
     /// a (potentially-invalid) Hugr is always produced.
-    pub fn keep_both_invalid() -> Self {
+    pub fn new_keep_both_invalid() -> Self {
         Self {
             multi_defn: OnMultiDefn::NewFunc(OnNewFunc::Add),
             sig_conflict: OnNewFunc::Add,
@@ -424,7 +412,10 @@ impl NameLinkingPolicy {
 
 impl Default for NameLinkingPolicy {
     fn default() -> Self {
-        Self::err_on_conflict(OnNewFunc::RaiseError)
+        Self {
+            sig_conflict: OnNewFunc::RaiseError,
+            multi_defn: OnNewFunc::RaiseError.into(),
+        }
     }
 }
 
@@ -1013,7 +1004,7 @@ mod test {
         let new_sig = Signature::new_endo(INT_TYPES[3].clone());
         let (inserted, inserted_fn) = mk_def_or_decl("foo", new_sig.clone(), inserted_defn);
 
-        let pol = NameLinkingPolicy::err_on_conflict(OnNewFunc::RaiseError);
+        let pol = NameLinkingPolicy::default();
         let mut host = orig_host.clone();
         let res = host.link_module_view(&inserted, &pol);
         assert_eq!(host, orig_host); // Did nothing
@@ -1059,7 +1050,7 @@ mod test {
         let mut host = backup.clone();
         let inserted = build_hugr(11);
 
-        let pol = NameLinkingPolicy::keep_both_invalid().on_multiple_defn(multi_defn);
+        let pol = NameLinkingPolicy::new_keep_both_invalid().on_multiple_defn(multi_defn);
         let res = host.link_module(inserted, &pol);
         if multi_defn == OnNewFunc::RaiseError.into() {
             assert!(matches!(res, Err(NameLinkingError::MultipleDefn(n, _, _)) if n == "foo"));


hugr-core/src/hugr/linking.rs
```diff
@@ -185,12 +248,302 @@ impl<TN> NodeLinkingDirective<TN> {
     }
 }
 
+/// Describes ways to link a "Source" Hugr being inserted into a target Hugr.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct NameLinkingPolicy {
+    // TODO: consider pub-funcs-to-add? (With others, optionally filtered by callgraph, made private)
```
acl-cqc, line 254: I can remove these various (non-doc) comments about future ideas...

hugr-core/src/hugr/linking.rs
```diff
@@ -185,12 +248,302 @@ impl<TN> NodeLinkingDirective<TN> {
...
+    Add,
+}
+
+/// What to do when both target and inserted Hugr
+/// have a [Visibility::Public] FuncDefn with the same name and signature.
+///
+/// [Visibility::Public]: crate::Visibility::Public
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
+#[non_exhaustive] // could consider e.g. disconnections
+pub enum MultipleImplHandling {
```
acl-cqc, line 287: A possibility is to call this just `NewImplHandling` with the idea of adding in the future an option `::ConvertToDecls`, this would avoid adding new code (except for the entrypoint subtree, when the next PR adds the feature to insert that alongside linking). Existing enum variants would then be `PreferExisting`, `PreferNew`, and a wrapper around `NewFuncHandling` (which allows to e.g. error upon any new impl)

hugr-core/src/hugr/linking.rs
```diff
@@ -185,12 +248,302 @@ impl<TN> NodeLinkingDirective<TN> {
...
+    AddFunctionContainingEntrypoint(SN, NodeLinkingDirective<TN>),
+}
+
+impl NameLinkingPolicy {
+    /// Makes a new instance that specifies to handle
+    /// [signature conflicts](Self::on_signature_conflict) by failing with an error and
+    /// multiple [FuncDefn]s according to `multi_impls`.
+    ///
+    /// [FuncDefn]: crate::ops::FuncDefn
+    pub fn err_on_conflict(multi_impls: impl Into<MultipleImplHandling>) -> Self {
```
acl-cqc, line 337: Names for constructors, given that there'll be significant extension to this `NameLinkingPolicy` class in the future, are tricky and these are a bit dubious....

hugr-core/src/hugr/linking.rs
```diff
@@ -185,12 +248,302 @@ impl<TN> NodeLinkingDirective<TN> {
...
+            multi_impls: MultipleImplHandling::NewFunc(NewFuncHandling::Add),
+            sig_conflict: NewFuncHandling::Add,
+        }
+    }
+
+    /// Sets how to behave when both target and inserted Hugr have a
+    /// [Public] function with the same name but different signatures.
+    ///
+    /// [Public]: crate::Visibility::Public
+    pub fn on_signature_conflict(mut self, sc: NewFuncHandling) -> Self {
```
acl-cqc, line 358: I felt that the short expression style `NameLinkingPolicy::foo().on_xyz(bar)` was preferable for configuration rather than having to `{let pol = NameLinkingPolicy::foo(); *pol.xyz() = bar; pol }` over several lines. Any counter-opinions?

hugr-core/src/hugr/linking.rs
```diff
@@ -105,6 +107,67 @@ pub trait HugrLinking: HugrMut {
         link_by_node(self, transfers, &mut inserted.node_map);
         Ok(inserted)
     }
+
+    /// Insert module-children from another Hugr into this one according to a [NameLinkingPolicy].
+    ///
+    /// All [Visibility::Public] module-children are inserted, or linked, according to the
+    /// specified policy; private children will also be inserted, at least including all those
+    /// used by the copied public children.
+    // Yes at present we copy all private children, i.e. a safe over-approximation!
```
acl-cqc, line 116: I'm debating whether to
1. specify a vague "at least" bound on behaviour in the docs; and then quietly change the actual behaviour (from the maximum set within that bound, to the minimum) in the next PR
2. Add an extra flag in [the next PR](#2555) to preserve the exact behaviour introduced in this PR (i.e. add all the private functions, even unused ones, unless overridden). #2555 currently takes this latter approach and it's not super-difficult but mildly annoying. (And do people really want those private funcs to be added?)

Thoughts welcome! :-)

lmondada: Not adding unused private functions sounds like the right thing to do. Let's not bloat the Hugr unless we have a very good reason to.

If we do come up with a use case in the future for adding those private functions, it's an easy change.

Regarding docs: it's up to you. Behaviour change is not breaking, so even if there's a release in between it would not matter.

hugr-core/src/hugr/linking.rs
lmondada: I think these comments would be better suited as an issue than as comments here.

It's more searchable and you can be as verbose as you want without bloating the source code.

hugr-core/src/hugr/linking.rs
```diff
@@ -185,12 +248,302 @@ impl<TN> NodeLinkingDirective<TN> {
...
+
+/// Specifies what to do with a function in some situation - used in
+/// * [NameLinkingPolicy::on_signature_conflict]
+/// * [MultipleImplHandling::NewFunc]
+///
+/// [FuncDefn]: crate::ops::FuncDefn
+/// [Visibility::Public]: crate::Visibility::Public
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
+#[non_exhaustive] // could consider e.g. disconnections
+pub enum NewFuncHandling {
```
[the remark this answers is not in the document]
lmondada: Ah but you have further enums with the same naming pattern below...

hugr-core/src/hugr/linking.rs
lmondada: `ImplMergeOptions`?

lmondada: You can probably tell by now I dont find "Handling" particularly descriptive 😂

hugr-core/src/hugr/linking.rs
```diff
@@ -185,12 +248,296 @@ impl<TN> NodeLinkingDirective<TN> {
...
+#[derive(Clone, Debug, thiserror::Error, PartialEq)]
+pub enum NameLinkingError<SN: Display, TN: Display + std::fmt::Debug> {
+    /// Both source and target contained a [FuncDefn] (public and with same name
+    /// and signature).
+    ///
+    /// [FuncDefn]: crate::ops::FuncDefn
+    #[error("Source ({_1}) and target ({_2}) both contained FuncDefn with same public name {_0}")]
+    MultipleImpls(String, SN, TN),
+    /// Source and target containing public functions with conflicting signatures
+    // TODO ALAN Should we indicate which were decls or defns? via an extra enum?
```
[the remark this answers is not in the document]
lmondada: I think the error is clear enough as is.

hugr-core/src/hugr/linking.rs
```diff
@@ -185,12 +248,296 @@ impl<TN> NodeLinkingDirective<TN> {
...
+    /// Tells how to behave when both target and inserted Hugr have a
+    /// [FuncDefn](crate::ops::FuncDefn) with the same name and signature.
+    pub fn get_multiple_impls(&self) -> MultipleImplHandling {
+        self.multi_impls
+    }
+
+    /// Computes how this policy will act on a specified source (inserted) and target
+    /// (host) Hugr.
+    #[allow(clippy::type_complexity)]
+    pub fn to_node_linking<T: HugrView + ?Sized, S: HugrView>(
```
lmondada, line 387: Nit: presumably S does not need to be Sized either

lmondada (commented):
I'll finish my review later. Here's the first half

hugr-core/src/hugr/linking.rs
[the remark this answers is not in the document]
acl-cqc: Maybe something like FuncAddResult? I know it's not a rust `enum Result {Ok, Err}` but it's actually not that far off - those are basically the two options available now (`::Add` and `::RaiseError`), but I'll be adding `::AddIfReached` and `::ErrorIfReached` in the next PR #2555, and there'll be an `::MakePrivate` (really `::AddIfReachedButMakePrivate`) after that....)

`FuncAddOutcome` ??

hugr-core/src/hugr/linking.rs
acl-cqc: Looking at the other suggestions...`ImplAdd`Something? `DefnAddOutcome`? (These options are appropriate for impls/defns only not other things as that's the only situation where you have two things that can be merged)

`DefnMerging::ConvertToDecls` (later PR) is not great because that's about adding Defns not merging them...

`BlahAddPolicy` is not ridiculous - a NameLinkingPolicy is then largely composed of other/sub policies. But `Policy` may make this simple thing sound more complex than it is...

hugr-core/src/hugr/linking.rs
acl-cqc: https://github.com/CQCL/hugr/issues/2687

hugr-core/src/hugr/linking.rs
acl-cqc: My view here is that having to specify `+ ?Sized` is the annoying non-default thing that should be minimized....unfortunately `trait LinkHugr`s default impls don't compile without the first one.

hugr-core/src/hugr/linking.rs
```diff
@@ -522,4 +874,219 @@ mod test {
...
+    fn call_targets<H: HugrView>(h: &H) -> HashMap<H::Node, H::Node> {
+        h.nodes()
+            .filter(|n| h.get_optype(*n).is_call())
+            .map(|n| (n, h.static_source(n).unwrap()))
+            .collect()
+    }
+
+    #[rstest]
+    fn combines_decls_defn(
+        #[values(OnNewFunc::RaiseError, OnNewFunc::Add)] sig_conflict: OnNewFunc,
```
acl-cqc: This mass of `#[values....]` is because none of these actually matter, the test is the same in all cases...

I'd like to define a pytest-like "parametrized fixture" `#[fixture] fn any_policy(#[values....] ....) -> NameLinkingPolicy` and then have every test using `any_policy` get run multiple times, but `rstest` doesn't offer such a facility. There are crates that do: https://crates.io/crates/rstest_reuse (a small addition) and https://crates.io/crates/rustest (a complete replacement of rstest)....

doug-q (approved):
Looks Great!

Just nits.

hugr-core/src/hugr/linking.rs
```diff
@@ -185,12 +248,292 @@ impl<TN> NodeLinkingDirective<TN> {
...
+pub enum LinkAction<TN> {
+    /// Just apply the specified [NodeLinkingDirective].
+    LinkNode(#[from] NodeLinkingDirective<TN>),
+}
+
+/// Details the concrete actions to link a specific source Hugr into a specific target Hugr.
+///
+/// Computed from a [NameLinkingPolicy] and contains all actions required to implement
+/// that policy (for those specific Hugrs).
+pub type LinkActions<SN, TN> = HashMap<SN, LinkAction<TN>>;
```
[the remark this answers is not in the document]
acl-cqc: I've switched this one to BTreeMap because it's externally visible and so you might print it. The ordering should not affect the result of linking (unless you hand-construct a map of actions which modify each other or something), so I've not switched any of the *other* HashMaps....

PR #2531, opened 2025-08-26
feat: move CallGraph into hugr-core

acl-cqc:
In the short term, this is likely wanted for an upcoming PR on linking.
In the (only-slightly-)longer term, we're likely to want this to include static function type-arguments, where a Call/LoadFunction can refer to multiple functions (e.g. `foo<bar>(runtime_args)`), which will probably mean adding variants to `enum CallGraphEdge`. So....

* There is a possibility I could derive `#[non_exhaustive] enum CallGraphEdge` *now*, making that (anticipated, hypothetical) future change less breaking. (Probably still breaking, though!)
* This has the cost of making this change breaking now, which it presently is not (because of deprecated re-export in `hugr-passes` - **I think rust-semver checks is wrong** here).
* However, rather than re-export, I could *duplicate* call_graph.rs, with the hugr-core version having `non_exhaustive` but hugr-passes not.
* I'm not sure this is worth it: static function type-args are likely to be breaking anyway, and the non-exhaustive just delays the point of user-code needing an update until, most likely, it sees a new instance *at runtime* rather than compile-time....
* **Thoughts please?**

### DIFF
--- a/hugr-core/src/call_graph.rs
+++ b/hugr-core/src/call_graph.rs
@@ -1,8 +1,9 @@
 //! Data structure for call graphs of a Hugr
 use std::collections::HashMap;
 
-use hugr_core::{HugrView, Node, core::HugrNode, ops::OpType};
 use petgraph::Graph;
+
+use crate::{HugrView, Node, core::HugrNode, ops::OpType};
 
 /// Weight for an edge in a [`CallGraph`]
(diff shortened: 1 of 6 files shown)


lmondada:
I'm confused - why does semver complain on this change?

PR #2534, opened 2025-08-27
feat(persistent): Redesign CommitStateSpace, bound Commit lifetime

lmondada:
This PR updates the relrc dependency to v0.5.0 and completely redesigns the `CommitStateSpace` struct. The result is much cleaner:

- `PersistentHugr` is an immutable data type following the design of the types of the [`im` crate](https://docs.rs/im/latest/im/), i.e. it is a mutable Rust type that is cheap to clone and mutate (immutability -- if desired -- is achieved by cloning before mutating).
- `Commit` is a "transaction", i.e. one atomic mutation that was applied to a `PersistentHugr`. It has an ID, obtainable through `commit.id()` and a lifetime parameter that ensures it does not stay in scope beyond the lifetime of the `PersistentHugr`
- Each `PersistentHugr` belongs to a unique `CommitStateSpace`. A `CommitStateSpace` is a registry of all commits of the `PersistentHugr`s within that state space. It keeps the map between commits and IDs.

This is a breaking change, but does not trigger a breaking HUGR release because it is hidden behind an unstable feature.

### DIFF
--- a/hugr-core/src/std_extensions/collections/borrow_array.rs
+++ b/hugr-core/src/std_extensions/collections/borrow_array.rs
@@ -45,7 +45,7 @@ pub const BORROW_ARRAY_VALUENAME: TypeName = TypeName::new_inline("borrow_array"
 /// Reported unique name of the extension
 pub const EXTENSION_ID: ExtensionId = ExtensionId::new_unchecked("collections.borrow_arr");
 /// Extension version.
-pub const VERSION: semver::Version = semver::Version::new(0, 1, 1);
+pub const VERSION: semver::Version = semver::Version::new(0, 1, 2);
 
 /// A linear, unsafe, fixed-length collection of values.
 ///
@@ -123,6 +123,8 @@ pub enum BArrayUnsafeOpDef {
     discard_all_borrowed,
     /// `new_all_borrowed<size, elem_ty>: () -> borrow_array<size, elem_ty>`
     new_all_borrowed,
+    /// is_borrowed<N, T>: borrow_array<N, T>, usize -> bool, borrow_array<N, T>
+    is_borrowed,
 }
 
 impl BArrayUnsafeOpDef {
@@ -166,6 +168,13 @@ impl BArrayUnsafeOpDef {
             Self::new_all_borrowed => {
                 PolyFuncTypeRV::new(params, FuncValueType::new(type_row![], vec![array_ty]))
             }
+            Self::is_borrowed => PolyFuncTypeRV::new(
+                params,
+                FuncValueType::new(
+                    vec![array_ty.clone(), usize_t],
+                    vec![crate::extension::prelude::bool_t(), array_ty],
+                ),
+            ),
         }
         .into()
     }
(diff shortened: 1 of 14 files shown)


hugr-persistent/src/state_space.rs
```diff
@@ -74,561 +61,161 @@ mod hidden {
...
-    pub(super) graph: HistoryGraph<CommitData, (), R>,
-    /// The unique root of the commit graph.
-    ///
-    /// The only commit in the graph with variant [`CommitData::Base`]. All
-    /// other commits are [`CommitData::Replacement`]s, and are descendants
-    /// of this.
-    pub(super) base_commit: CommitId,
+#[repr(transparent)]
+pub struct CommitStateSpace {
+    registry: Rc<RefCell<Registry<CommitData, ()>>>,
```
acl-cqc, line 85: Does the `Rc` really make sense here? That means you can cheaply `clone` a CommitStateSpace, but the clones will then share the state hidden inside the RefCell. I would not expect cloning a CommitStateSpace to be something you want to do often....???

(One can imagine a "fork" operation but TBH I'm not really sure why you would want to do that either ;-) unless Ids are scarce or something)

hugr-persistent/src/persistent_hugr.rs
```diff
@@ -211,32 +52,45 @@ impl<'a> From<&'a RelRc<CommitData, ()>> for &'a Commit {
...
-    /// Invariant: all commits are "compatible", meaning that no two patches
-    /// invalidate the same node.
-    state_space: CommitStateSpace<R>,
+    /// Invariants:
+    ///  - all commits are "compatible", meaning that no two patches invalidate
+    ///    the same node.
+    ///  - there is a unique commit of variant [`CommitData::Base`] and its ID
+    ///    is `base_commit_id`.
+    graph: HistoryGraph<CommitData, ()>,
+    /// The unique root of the commit graph.
```
acl-cqc, line 69: "(Cache of)" perhaps in that it could be computed (by following the parent pointer from any commit, result would be the same regardless of which commit you started from)

hugr-persistent/src/persistent_hugr.rs
```diff
@@ -211,32 +52,45 @@ impl<'a> From<&'a RelRc<CommitData, ()>> for &'a Commit {
...
     ///
-    /// Invariant: all commits are "compatible", meaning that no two patches
-    /// invalidate the same node.
-    state_space: CommitStateSpace<R>,
+    /// Invariants:
+    ///  - all commits are "compatible", meaning that no two patches invalidate
+    ///    the same node.
+    ///  - there is a unique commit of variant [`CommitData::Base`] and its ID
+    ///    is `base_commit_id`.
+    graph: HistoryGraph<CommitData, ()>,
```
acl-cqc, line 68: So this is the bit that might be expensive to clone. I'm not terribly clear from the docs of HistoryGraph as to whether it's cheap, or whether it's actually a complete+owned petgraph constructed from (a cache of) the parent pointers of some set of commits. It might be worth nothing that a persistent (im::) list/set of "leaf commits" would store the same information (perhaps in less accessible form) and really would be cheap to clone...

lmondada: Yes agreed; I will create an issue for this, to be picked up when this becomes a bottleneck.

lmondada: https://github.com/CQCL/tket2/issues/1141

PR #2541, opened 2025-09-04
feat!: Add NodeType trait as hugr base trait

lmondada:
This is a small change but "pretty breaking". Some bullet points
- I am usually against fragmenting the API into more traits. My main issue is that it makes API discovery and intellisense worse. In this case, as it is a super trait of many common traits, you rarely need to actually import this new trait directly, so the fragmentation is less of a concern.
- I do think that the break up makes sense here: having an assoicated Node type is a much weaker constraint than implementing `HugrInternals` (which e.g. requires a view into an underlying portgraph)
- Seyon is "not against" it, and no-one else has expressed strong opinions
- Users won't notice this breaking change, unless they implemented `HugrInternals` themselves or explicitly refer to the `<Self as HugrInternals>::Node` type in macros or type aliases (in which case they'd probably get the side benefit of being able to loosen their type bounds).

BREAKING CHANGE: The HugrInternals::Node associated type was moved to the dedicated trait NodeType.

### DIFF
--- a/hugr-core/src/hugr/internal.rs
+++ b/hugr-core/src/hugr/internal.rs
@@ -15,21 +15,27 @@ use super::views::{panic_invalid_node, panic_invalid_non_entrypoint};
 use super::{NodeMetadataMap, OpType};
 use crate::ops::handle::NodeHandle;
 
+/// Base trait specifying the node type associated with Hugr implementations.
+///
+/// Base trait of [`HugrInternals`], and thus by transitivity of [`HugrView`],
+/// [`HugrMut`](crate::hugr::HugrMut) etc.
+pub trait NodeType {
+    /// The type of nodes in the Hugr.
+    type Node: HugrNode;
+}
+
 /// Trait for accessing the internals of a Hugr(View).
 ///
 /// Specifically, this trait provides access to the underlying portgraph
 /// view.
-pub trait HugrInternals {
+pub trait HugrInternals: NodeType {
     /// The portgraph graph structure returned by [`HugrInternals::region_portgraph`].
     type RegionPortgraph<'p>: LinkView<LinkEndpoint: Eq, PortOffsetBase = u32> + Clone + 'p
     where
         Self: 'p;
 
-    /// The type of nodes in the Hugr.
-    type Node: Copy + Ord + std::fmt::Debug + std::fmt::Display + std::hash::Hash;
-
-    /// A mapping between HUGR nodes and portgraph nodes in the graph returned by
-    /// [`HugrInternals::region_portgraph`].
+    /// A mapping between HUGR nodes and portgraph nodes in the graph returned
+    /// by [`HugrInternals::region_portgraph`].
     type RegionPortgraphNodes: PortgraphNodeMap<Self::Node>;
 
     /// Returns a flat portgraph view of a region in the HUGR, and a mapping between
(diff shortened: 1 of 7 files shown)


acl-cqc:
> > Firstly, I have to query why we want to do this
> 
> **`Circuit`, `SiblingSubgraph` and `ResourceScope` do not implement HugrInternals**
> 
> There are types that could implement `NodeType` (or `HasNode`) that do not implement `HugrInternals`, such as `SiblingSubgraph` (not a valid Hugr) or tket's Circuit. These types could provide `NodeType` implementations. Admittedly, these types could also implement a much broader almost-a-hugr-but-not-quite trait, but I think that is a separate, much broader discussion.
> 
> **As a result, there are traits (`Patch`, `Rewriter`) that cannot be implemented by these types**

As I say, if you only know a type H: NodeType, that doesn't let you do anything with an H. (So, I think that broader discussion is important; I'm not sure we can get very far without it.)

Otherwise the scenario where this is useful looks something like....
```
fn foo<H:NodeType>(p: Patch<P>, h: H) {
  ....
  p.apply(h)
  ...
}
```
There really isn't much else you can write here! Is that really useful?

I think the real problem here is the impedance mismatch between PatchVerification (parametric over *any* HugrView with the right node, so we can verify by using hugrview methods to inspect, but not mutate) and Patch (parametric over an impl of HugrView so we can restrict to an impl with mutation methods but not necessarily HugrMut).

I mean, if you have:
```
trait Patch<H: NodeType>: PatchVerification<H::Node> { .... }
```
then PatchVerification has to be implemented for HugrView's but the Patch can be applied to things that aren't HugrView, i.e., your Patch can be applied to things on which it can't be verified. That doesn't seem right....

So, one way be to remove the prerequisite of Patch that it support PatchVerification; we might end up with (in both cases `H` may be a concrete type rather than a parameter):
* anywhere that's currently `<H:HugrView>(p: impl Patch<H>)` instead becomes `<H:HugrView>(p: impl Patch<H> + PatchVerification<H::Node>)`
* In new cases, *either*
   * `<H: MyOtherNodeType>(p: impl Patch<H> + OtherVerification<H::Node>)` (where `OtherVerification` has polymorphic method `verify(h: &impl MyOtherNodeType<Node=Self::Node>>)` - assuming `MyOtherNodeType` supplies non-HugrView inspection methods), *or*
   * `<H: MyOtherNodeType>(p: impl Patch<H> + MonoVerification<H>)` where MonoVerification has non-polymorphic `verify(h: &H)`

I find the latter a bit restrictive but I guess the question would be, does is support your use case?

I quickly end up  thinking that Patch should also be parametrized by `N: HugrNode>, and support applying to any concrete type supporting `(Base)HugrMut<N>`, where BaseHugrMut is some simplified form of HugrMut that we can implement for e.g. SiblingSubgraph, PersistentHugr (via transactions), or wherever is actually required (??)

lmondada:
Agree that there is a mismatch between `Patch` and `PatchVerification`.

Here's another example where `HasNode` would be useful: the `CircuitHash` trait of tket2, implemented for Circuits:

```rust
pub trait CircuitHash {
    /// Compute the hash of a dataflow container node.
    ///
    /// We compute a hash for each node from its operation and the hash of
    /// the predecessors. The hash of the circuit corresponds to the hash of its
    /// output node.
    ///
    /// This hash is independent from the operation traversal order.
    ///
    /// Adapted from Quartz (Apache 2.0)
    /// <https://github.com/quantum-compiler/quartz/blob/2e13eb7ffb3c5c5fe96cf5b4246f4fd7512e111e/src/quartz/tasograph/tasograph.cpp#L410>
    fn circuit_hash(&self, node: Node) -> Result<u64, HashError>;
}
```

On my branch I've been able to generalise this to non-`Node` type, using `pub trait CircuitHash: HasNode`, instead of having to make this trait parametric on the node type.

lmondada:
regarding your example:

> fn foo<H:NodeType>(p: Patch<P>, h: H) {
>  ....
>  p.apply(h)
>  ...
>}

it becomes useful if `H` isn't generic but a specific type: say a `ResourceScope`. The following function makes absolute sense:

```
fn foo(circ: ResourceScope, p: impl Patch<ResourceScope>) {...}
```

PR #2568, opened 2025-09-22
test: Add framework for LLVM execution tests involving panics

mark-koch:
The default LLVM lowering for panic ops just aborts the process. This makes testing of panic behaviour (e.g. for the new borrow array ops) impossible. This PR adds an alternative `TestPanicPreludeCodegen` that calls into a small panic runtime to safely exit programs and report the error message, instead of aborting the process.

### DIFF
--- a/hugr-llvm/build.rs
+++ b/hugr-llvm/build.rs
@@ -8,6 +8,7 @@ fn main() {
     compile_panic_runtime();
 }
 
+#[cfg(feature = "test-utils")]
 fn compile_panic_runtime() {
     println!("cargo::rerun-if-changed=src/emit/test/panic_runtime.c");
     cc::Build::new()


hugr-llvm/src/emit/test/panic_runtime.c
```diff
@@ -0,0 +1,57 @@
...
+// which launches the program. Calls to `panic_exit` will exit the program
+// and make `trampoline` return prematurely. Note that the runtime will
+// perform a direct jump to the exit and no unrolling is performed. This
+// is fine for our use case since stack frames from hugr-llvm generated
+// bytecode don't require cleanup. However, using `panic_exit` to jump over
+// Rust frames is undefined behaviour.
+//
+// Note that the implementation relies on setjmp-longjmp (SJLJ) jumping to
+// exit the program. As SJLJ is not available in Rust, this had to be
+// implemented in C in linked into the Rust binary.
```
aborgna-q, line 18: Did you try [`sjlj`](https://docs.rs/sjlj/0.1.3/sjlj/index.html) ? If we could skip the C lib it would make the rest of the diff simpler...

mark-koch: My understanding is that crates like this have undefined behaviour: 

> We would also like to specify conditions under which longjmp and pthread_exit may safely deallocate Rust frames. This RFC specifies that frames deallocated in this way [must be POFs](https://rust-lang.github.io/rfcs/2945-c-unwind-abi.html#reference-level-explanation). However, this condition is merely necessary rather than sufficient to ensure well-defined behavior.
>
> Within the context of this RFC and in discussions among members of the [FFI-unwind project group](https://github.com/rust-lang/project-ffi-unwind), this class of formally-undefined behavior which we plan to define in future RFCs is referred to as “TBD behavior”.

https://rust-lang.github.io/rfcs/2945-c-unwind-abi.html#unresolved-questions

hugr-llvm/build.rs
```diff
@@ -0,0 +1,16 @@
+//! The hugr-llvm build script.
+//!
+//! For debug builds, we compile the `test_panic_runtime` library to enable
+//! LLVM execution tests involving panics.
+
+fn main() {
+    println!("cargo::rerun-if-changed=src/emit/test/panic_runtime.c");
+    match std::env::var("PROFILE").unwrap().as_str() {
+        "debug" | "bench" => {
```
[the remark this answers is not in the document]
mark-koch: That worked, thanks!

aborgna-q (approved):

PR #2591, opened 2025-09-26
feat: Normalize CFGs

### LINKED ISSUE #945: CFG Normalisations
As per spec....
* If the Exit node has a single predecessor, and that predecessor has only one successor (i.e. the Exit node), then ops in the predecessor get moved *outside the CFG*. (This is like a special case of #561, as we can remove the predecessor, but we have to put things outside the CFG rather than in a new block.)
* If the CFG has only an exit node, then the whole CFG can be removed

The second is trivial but the common factor is these require making changes *outside* the CFG node itself.

Combined, these mean that a CFG of a single BB leading to the exit, can be removed altogether

### LINKED ISSUE #2556: Define a ComposablePass for merge_basic_blocks
Should it have an option to chose between recursive and top-region-only?
See [UntuplePass's recursive option](https://github.com/CQCL/hugr/blob/ef06013b916661b9a53e4125e5a1855d4b87cfcb/hugr-passes/src/untuple.rs#L73).

acl-cqc:
closes #945, closes #2556.

* Add module `normalize_cfgs` and a new `merge_basic_blocks` (with improved API) within
* Normalizations (in fn normalize_cfg and NormalizeCFGPass):
   * Single block CFG (Entry -> Exit) becomes a DFG
   * If Entry node has no internal predecessors and only one successor -> move contents outside/before CFG
   * If Exit node has only one predecessor -> move contents of that node outside/after CFG (and make it become the exit node)
* Deprecate the old `merge_basic_blocks` as the interface is annoying (panicking, using `RootCheckable`).
* While about it, generalize InlineDFG over NodeType

Does not address: "inlining" arbitrary CFG into a BB (requires splitting the outer BB into parts before/after the inner CFG); emptying (but preserving) a *branching* entry block (if no predecessors).


hugr-passes/src/merge_bbs.rs
```diff
@@ -49,11 +59,204 @@ where
...
+///   outside/after CFG.
+///    * If that predecessor is the entry block, then convert the CFG to a DFG.
+///      This will be reported via [NormalizeCFGResult::CFGToDFG]
+///
+///
+/// # Errors
+///
+/// [NormalizeCFGError::NotCFG] If the entrypoint is not a CFG
+#[allow(deprecated)] // inline/combine/refactor with merge_bbs, or just hide latter
+pub fn normalize_cfg<H: HugrMut<Node = Node>>(
```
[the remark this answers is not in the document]
acl-cqc: Yes, done

hugr-passes/src/merge_bbs.rs
```diff
@@ -49,11 +59,204 @@ where
...
+
+            // Inputs to entry block Output node go instead to CFG
+            let (entry_results, orders) = take_inputs(h, entry_output);
+            h.remove_node(entry_output);
+            wire_unpack_first(h, entry_results, orders, cfg_node);
+
+            // Transfer remaining entry children - including any used to compute the predicate
+            while let Some(n) = h.first_child(entry) {
+                h.set_parent(n, cfg_parent);
+            }
```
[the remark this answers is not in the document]
acl-cqc: Done in 317b5968215cf3cc3768fdfbd244266c7660410c, can revert if you disagree but I think it is more uniform

hugr-passes/src/normalize_cfg.rs
```diff
@@ -0,0 +1,1033 @@
...
+pub enum NormalizeCFGResult {
+    /// The entire [`CFG`] was converted into a [`DFG`].
+    ///
+    /// The entrypoint node id is preserved after the conversion, but now identifies
+    /// the new [`DFG`].
+    CFGToDFG,
+    /// The CFG was preserved, but the entry or exit blocks may have changed.
+    #[allow(missing_docs)]
+    CFGPreserved {
+        entry_changed: bool,
```
acl-cqc, line 87: would like to add 'num blocks merged' here, I guess could do that by changing signature of `merge_basic_blocks` and then making the `merge_bbs::` version (currently a re-export) just throw away this information

hugr-passes/src/normalize_cfg.rs
```diff
@@ -0,0 +1,1033 @@
...
+    // 1. If the entry block has only one successor, and no predecessors, then move contents
+    // outside/before CFG.
+    // (Note if the entry block has multiple successors, but no predecessors, we *could* move
+    // its contents outside/before the CFG, but would need to keep an empty/identity entry
+    // block - we do not do this here.)
+    let mut entry_changed = false;
+    #[allow(clippy::match_result_ok)] // let Ok(...) without .ok() borrows `h`
+    if let Some(succ) = h.output_neighbours(entry).exactly_one().ok() {
+        if succ == exit {
+            // Any (basic-block) predecessors of `entry` are unreachable, so allow/ignore
```
acl-cqc, line 179: This seemed a neat trick: less checks on our part, more optimization. However I'm starting to wonder: it requires its own test and makes `ComposablePass::run` much harder. Perhaps a better strategy would be to enhance DeadCodeElimination to remove unreachable blocks (currently it assumes all BBs are reachable if the CFG is!) and insist here.

hugr-passes/src/normalize_cfg.rs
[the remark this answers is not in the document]
aborgna-q: That works 👍

hugr-passes/src/normalize_cfg.rs
```diff
@@ -0,0 +1,1033 @@
...
+    type Result = HashMap<H::Node, NormalizeCFGResult>;
+
+    fn run(&self, hugr: &mut H) -> Result<Self::Result, Self::Error> {
+        let cfgs = if self.cfgs.is_empty() {
+            let mut v = hugr
+                .entry_descendants()
+                .filter(|n| hugr.get_optype(*n).is_cfg())
+                .collect::<Vec<_>>();
+            // Process inner CFGs first, in case they are removed
+            // (if they are in an unreachable block when the Entry node has only the Exit as successor)
```
acl-cqc, line 127: Sadly even this current behaviour means that you can get results saying what was done to a CFG that was later removed from the Hugr altogether as the whole thing was unreachable...see https://github.com/CQCL/hugr/pull/2591/files#r2390401557

hugr-passes/src/normalize_cfg.rs
```diff
@@ -49,11 +59,204 @@ where
...
+            for blk in children_to_remove {
+                h.remove_subtree(blk);
+            }
+            while let Some(ch) = h.first_child(entry) {
+                h.set_parent(ch, cfg_node);
+            }
+            h.remove_node(entry);
+            let cfg_ty = h.optype_mut(cfg_node);
+            let OpType::CFG(CFG { signature }) = std::mem::take(cfg_ty) else {
+                panic!()
```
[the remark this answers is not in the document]
acl-cqc: Commoned up several like this into a helper function. (Sadly can't be a closure as it has to be polymorphic over lifetime)

hugr-passes/src/lib.rs
```diff
@@ -17,10 +17,17 @@ pub use inline_funcs::inline_acyclic;
...
 pub use linearize_array::LinearizeArrayPass;
 pub mod lower;
-pub mod merge_bbs;
 mod monomorphize;
+pub mod normalize_cfg;
 pub mod untuple;
 
+/// Merge basic blocks. Subset of [normalize_cfg], use the latter.
+#[deprecated(note = "Use normalize_cfg", since = "0.15.1")]
+pub mod merge_bbs {
```
aborgna-q, line 26: I don't think modules can be deprecated, but at least it's nice to have this as a reminder to delete the re-export.

hugr-passes/src/normalize_cfg.rs
```diff
@@ -0,0 +1,1031 @@
...
+    }
+}
+
+/// Normalize a CFG in a Hugr:
+/// * Merge consecutive basic blocks i.e. where a BB has only a single successor which
+///   has no predecessors
+/// * If the entry block has only one successor, and no predecessors, then move its contents
+///   outside/before CFG.
+/// * (Similarly) if the exit block has only one predecessor, then move contents
+///   outside/after CFG.
```
aborgna-q, line 148: Mention that the moved contents are now in a DFG block

aborgna-q (approved):
Agreed with removing the optimized special case, but the rest looks gtm

hugr-passes/src/normalize_cfg.rs
acl-cqc: You might want me to move the inline module in hugr-passes/src/lib.rs out into `merge_bbs.rs` again, can do.

hugr-passes/src/normalize_cfg.rs
acl-cqc: Not quite as good as I hoped - one could still have entirely disconnected BBs that raise the same issue - but again, it seems more uniform. Raised https://github.com/CQCL/hugr/issues/2597.

hugr-passes/src/lib.rs
```diff
@@ -17,10 +17,34 @@ pub use inline_funcs::inline_acyclic;
 pub mod linearize_array;
 pub use linearize_array::LinearizeArrayPass;
 pub mod lower;
-pub mod merge_bbs;
 mod monomorphize;
+pub mod normalize_cfgs;
 pub mod untuple;
 
+/// Merge basic blocks. Subset of [normalize_cfgs], use the latter.
+#[deprecated(note = "Use normalize_cfgs", since = "0.15.1")]
```
aborgna-q, line 25: Mixed up the version with tket's
```suggestion
#[deprecated(note = "Use normalize_cfgs", since = "0.23.0")]
```

hugr-passes/src/lib.rs
```diff
@@ -17,10 +17,34 @@ pub use inline_funcs::inline_acyclic;
...
+    /// Merge any basic blocks that are direct children of the specified CFG
+    /// i.e. where a basic block B has a single successor B' whose only predecessor
+    /// is B, B and B' can be combined.
+    ///
+    /// # Panics
+    ///
+    /// If the `entrypoint` of `cfg` is not an [OpType::CFG]
+    ///
+    /// [OpType::CFG]: hugr_core::ops::OpType::CFG
+    #[deprecated(note = "Use version in normalize_cfgs", since = "0.15.1")]
```
aborgna-q, line 39: Mixed up the version with tket's
```suggestion
    #[deprecated(note = "Use version in normalize_cfgs", since = "0.23.0")]
```

aborgna-q (approved):
LGTM

hugr-passes/src/lib.rs
acl-cqc: lol, should have twigged that was low...15 was ages ago! This looks ahead/predicts the next number, right...

PR #2611, opened 2025-10-06
chore: use cargo-nextest in justfile and CI

ss2165:
drive-by: update devenv, cargo and uv

Non-devenv developers will need to install cargo-nextest


justfile
```diff
@@ -2,12 +2,19 @@
 help:
     @just --list --justfile {{justfile()}}
 
+
+_check_nextest_installed:
+    #!/usr/bin/env bash
+    if ! cargo nextest --version >/dev/null 2>&1; then
+        echo "cargo nextest needs to be installed to use `just test` https://nexte.st/docs/installation/pre-built-binaries/" >&2
+        exit 1
+    fi
```
aborgna-q, line 11: Should we call this from `test-rust` too? The check is cheap

PR #2612, opened 2025-10-07
fix(test): No extension serialisation in persistent-hugr testing

lmondada:
This changes the behaviour of persistent-hugr hashing in testing only. Commit hashes will no longer change (and tests will no longer break) when HUGR extensions are changed.


ss2165 (approved):
thanks

PR #2636, opened 2025-10-27
feat(py):  `ComposablePass` protocol and `ComposedPass` for hugr-py

### LINKED ISSUE #2617: Define a composable pass interface in Python
We need a interface that allows us to define passes to be performed in sequence. 

We could have some `ComposablePass` protocol which is implemented by some bound pytket passes and possibly some hugr passes. 

Note there is already a way to compose passes in Rust. See [this qsystem test](https://github.com/CQCL/tket2/blob/main/tket-qsystem/tests/guppy_opt.rs#L98).

CalMacCQ:
closes #2617 

Mostly follows Agustin's suggestion see [eeb4c50](https://github.com/CQCL/hugr/pull/2636/commits/eeb4c50880c599efb35e477b26123b2e635d4c4e)

I'm also added some `to/from_dict` methods as well as a `supported_ops` property. EDIT: removed these additional methods for now.

BEGIN_COMMIT_OVERRIDE

feat: ComposablePass protocol and ComposedPass for hugr-py (unstable) ([#2636](https://github.com/CQCL/hugr/pull/2636))

END_COMMIT_OVERRIDE

### DIFF
--- a/hugr-py/tests/test_passes.py
+++ b/hugr-py/tests/test_passes.py
@@ -1,0 +1,26 @@
+from hugr.hugr.base import Hugr
+from hugr.passes.composable_pass import ComposablePass, ComposedPass
+
+
+def test_composable_pass() -> None:
+    class MyDummyPass(ComposablePass):
+        def __call__(self, hugr: Hugr) -> None:
+            return self(hugr)
+
+        def then(self, other: ComposablePass) -> ComposablePass:
+            return ComposedPass([self, other])
+
+        @property
+        def name(self) -> str:
+            return "Dummy"
+
+    dummy = MyDummyPass()
+
+    composed = dummy.then(dummy)
+
+    my_composed_pass = ComposedPass([dummy, dummy])
+
+    assert my_composed_pass.passes == [dummy, dummy]
+    assert isinstance(my_composed_pass, ComposablePass)
+    assert isinstance(composed, ComposablePass)
+    assert dummy.name == "Dummy"


hugr-py/src/hugr/hugr/pass.py
```diff
@@ -0,0 +1,31 @@
...
+    def __call__(self, hugr: Hugr) -> None: ...
+
+    def then(self, other: Self) -> Self: ...
+
+    def with_entrypoint(self, node: Node) -> Self: ...
+
+    @classmethod
+    def from_dict(cls, dictionary: dict) -> Self: ...
+
+    def to_dict(self) -> dict: ...
```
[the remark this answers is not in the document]
CalMacCQ: What would be an example of a pass with a non-encodable context? A user defined pass where the user passes a function into the builder?

CalMacCQ:
After chatting to Alec it may be worth removing the `supported_ops` property. In TKET1 having a similar property for supported gates causes issues when we compose passes with mismatching `GateSetPredicate`s. Its been a bit of a maintence headache. What do you think @aborgna-q?

hugr-py/src/hugr/hugr/__init__.py
```diff
@@ -1,6 +1,7 @@
 """The main HUGR structure."""
 
 from .base import Hugr, NodeData
+from .composable_pass import ComposablePass
```
CalMacCQ: Open to better suggestions for this module name. I opted to not use `pass.py` as `pass` is a keyword in Python.

acl-cqc (commented):
Looks good thank you @CalMacCQ. The main issue I remember from the `hugr_passes` ComposablePass was wanting to pass a single runtime boolean (whether to validate or not) around all the passes...which could be plumbed in, but we didn't.

In Python...I guess we could add an `options: Options | None = None` parameter to `run`, and indeed, we could do that later, so probably no real need to do that now.

hugr-py/src/hugr/passes/composable_pass.py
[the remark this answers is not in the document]
CalMacCQ: Done in [01b7126](https://github.com/CQCL/hugr/pull/2636/commits/01b71269baf9a60c61dd28f7a367f02b29a767cf)

Some maybe the `isinstance` checks are not needed. I also made `ComposablePass` runtime checkable.

hugr-py/tests/test_passes.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def test_composable_pass() -> None:
+    class MyDummyPass(ComposablePass):
+        def __call__(self, hugr: Hugr) -> None:
+            return self(hugr)
+
+        def then(self, other: ComposablePass) -> ComposablePass:
+            return ComposedPass([self, other])
+
```
aborgna-q: Let it use the default implementation (we want to test that rather than test code)
```suggestion
```

aborgna-q: #2688

PR #2639, opened 2025-10-27
feat: track package descriptions when loading

ss2165:
# Goals

- Track descriptions during package loading such that partial descriptions can be provided even if an error is returned
- Provide a `describe` cli subcommand for high-level descriptions of a package, primarily for debug purposes https://github.com/CQCL/hugr/pull/2650. Since we want a description even if some loading stage fails (e.g. extension resolution), tracking partial descriptions is a requirement.

## Secondary goals
- Unify loading approaches across json/model, centralise more validation logic
- Reduce burden on EnvelopeError - more local errors
- Replace existing debug extraction (generator, used extensions, envelopeconfig) with single description

# Approach

In order to make this PR non-breaking I have deprecated existing reading functions, and introduced new functions that return a full description on successful read, and a partial when error. The deprecated functions are most of the missing test coverage.

Reading naturally falls in to two stages: header and payload. If the header fails no description is valid, so that error does not come with a description. Payload errors can come with descriptions.

In order to build descriptions while loading I introduce the `Reader` struct to track data across the reading stages. This is handed off to the model `Context` when that stage takes over.

## Partial descriptions

Descriptions are defined in description.rs. They are intended to be small amounts of simple data, so all fields are pub. All fields except the envelope header are optional, a value of None indicates the field has not been set yet.

## Appendix: JSON
Since the JSON format is effectively legacy I did not focus much on incremental description tracking, it is more difficult since serde likes to read all at once. 

I found a surprise performance regression in JSON loading. In commit [af01043](https://github.com/CQCL/hugr/pull/2639/commits/af010430e24dab156746b0800eda419a8fcccce8) I attempted to address this by removing an unnecessary two stage serde deser, which has swung things in the other direction, improving performance from prior.

## Appendix: EnvelopeError
EnvelopeError is used as a catch all for all reading and writing errors. This PR goes some way to introducing new more specific errors for reaeding. EnvelopeError should really in future have the reading errors be taken out of it and renamed to `WriteError` or similar, then all the conversions from the reading errors can be removed. Removing the deprecated read functions is a pre-requisite for this. The merging of this PR should be accompanied by issues for these follow ups.

### DIFF
--- a/hugr-core/src/envelope.rs
+++ b/hugr-core/src/envelope.rs
@@ -171,6 +171,12 @@ pub fn read_envelope(
 /// - `reader`: The reader to read the envelope from.
 /// - `registry`: An extension registry with additional extensions to use when
 ///   decoding the HUGR, if they are not already included in the package.
+///
+/// # Errors
+/// - [`ReadError::EnvelopeHeader`] if there was an error reading the envelope header.
+/// - [`ReadError::Payload`] if there was an error reading the package payload,
+///   including a partial description of the envelope read before the error occurred.
+///
 // TODO deprecate and rename to read_envelope when removing old version
 pub fn read_described_envelope(
     reader: impl BufRead,
@@ -384,7 +390,7 @@ pub enum EnvelopeError {
     /// Error raised while checking for breaking extension version mismatch.
     #[deprecated(since = "0.24.1")]
     #[error(transparent)]
-    #[allow(deprecated)]
+    #[expect(deprecated)]
     ExtensionVersion {
         /// The source error.
         #[from]
--- a/hugr-core/src/envelope/description.rs
+++ b/hugr-core/src/envelope/description.rs
@@ -104,14 +104,15 @@ impl PackageDesc {
         Some(generators.join(", "))
     }
 
-    /// Returns an iterator over the module descriptions.
+    /// Returns an iterator over the module descriptions. Modules with
+    /// expected but missing descriptions yield `None`.
     pub fn modules(&self) -> impl Iterator<Item = &Option<ModuleDesc>> {
         self.modules.iter()
     }
 
-    /// Returns an iterator over the packaged extension descriptions.
-    pub fn packaged_extensions(&self) -> impl Iterator<Item = &Option<ExtensionDesc>> {
-        self.packaged_extensions.iter()
+    /// Returns an iterator over the packaged extension descriptions. Missing extensions are skipped.
+    pub fn packaged_extensions(&self) -> impl Iterator<Item = &ExtensionDesc> {
+        self.packaged_extensions.iter().flatten()
     }
 }
 
@@ -145,7 +146,7 @@ impl<E: AsRef<crate::Extension>> From<&E> for ExtensionDesc {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, serde::Serialize)]
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 /// Description of the entrypoint of a module.
 pub struct Entrypoint {
     /// Node id of the entrypoint.
@@ -184,7 +185,7 @@ where
     serializer.serialize_str(op_string(op_type).as_str())
 }
 
-#[derive(Debug, Clone, PartialEq, Default, serde::Serialize)]
+#[derive(Debug, Clone, PartialEq, Default, serde::Serialize, serde::Deserialize)]
 /// High-level description of a module in a HUGR package.
 pub struct ModuleDesc {
     /// Number of nodes in the module.
@@ -401,11 +402,7 @@ mod test {
     ) {
         empty_package_desc.set_packaged_extension(0, test_extension.clone());
         assert_eq!(
-            empty_package_desc
-                .packaged_extensions()
-                .next()
-                .unwrap()
-                .as_ref(),
+            empty_package_desc.packaged_extensions().next(),
             Some(&test_extension)
         );
     }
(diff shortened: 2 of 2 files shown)


hugr-core/src/envelope/description.rs
```diff
@@ -0,0 +1,567 @@
...
+#[derive(Debug, Clone, PartialEq, Default, serde::Serialize)]
+/// High-level description of a module in a HUGR package.
+pub struct ModuleDesc {
+    /// Number of nodes in the module.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub num_nodes: Option<usize>,
+    /// The entrypoint node and the corresponding operation type.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub entrypoint: Option<Entrypoint>,
+    /// Extensions used in the module computed while resolving, expected to be a subset of `used_extensions_metadata`.
```
aborgna-q:
```suggestion
    /// Extensions used in the module computed while resolving, expected to be a subset of `used_extensions_generator`.
```
?
The `_metadata` is only used in local variable names.

hugr-core/src/envelope/description.rs
```diff
@@ -0,0 +1,567 @@
...
+            .modules
+            .iter()
+            .flatten()
+            .flat_map(|m| m.generator.clone())
+            .collect();
+        if generators.is_empty() {
+            return None;
+        }
+
+        Some(generators.join(", "))
```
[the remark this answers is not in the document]
ss2165: yes now that describe can give a per-module generator field we can dedup here

hugr-core/src/envelope/description.rs
```diff
@@ -0,0 +1,563 @@
...
+
+impl Entrypoint {
+    /// Create a new entrypoint description.
+    pub fn new(node: Node, optype: OpType) -> Self {
+        Self { node, optype }
+    }
+}
+
+/// Get a string representation of an OpType for description purposes.
+pub fn op_string(op: &OpType) -> String {
```
doug-q: I think these are user visible strings. 
I wonder if it would be better to just use `Display` here? that's mostly what you're doing, I wonder if tweaking you do should be done in Display. That would probably be breaking.

I suggest a snapshot test of this. We have experienced buggy Display getting to users before which is pretty embarrassing.

doug-q (commented):
This is big, I'll do another pass.

I like this a lot

hugr-core/src/envelope/description.rs
```diff
@@ -0,0 +1,563 @@
...
+    }
+
+    /// Returns the number of packaged extensions in the package.
+    pub fn n_packaged_extensions(&self) -> usize {
+        self.packaged_extensions.len()
+    }
+
+    /// Returns the generator(s) of the package modules, if any.
+    /// Concatenates multiple generators with commas.
+    pub fn generator(&self) -> Option<String> {
```
doug-q: Maybe it doesn't matter, but this will "break" for generators that include commas.

hugr-core/src/envelope/description.rs
```diff
@@ -0,0 +1,563 @@
...
+        Some(generators.join(", "))
+    }
+
+    /// Returns an iterator over the module descriptions.
+    pub fn modules(&self) -> impl Iterator<Item = &Option<ModuleDesc>> {
+        self.modules.iter()
+    }
+
+    /// Returns an iterator over the packaged extension descriptions.
+    pub fn packaged_extensions(&self) -> impl Iterator<Item = &Option<ExtensionDesc>> {
```
doug-q: Wouldn't this be more useful with `Item = &ExtensionDesc`?

hugr-core/src/envelope/description.rs
```diff
@@ -0,0 +1,563 @@
...
+    /// Create a new extension description.
+    pub fn new(name: impl ToString, version: impl Into<Version>) -> Self {
+        Self {
+            name: name.to_string(),
+            version: version.into(),
+        }
+    }
+}
+
+impl<E: AsRef<crate::Extension>> From<&E> for ExtensionDesc {
```
doug-q: If we are using AsRef<Extension> on our interface we should really `impl AsRef<Extension> for Extension` e.g. https://doc.rust-lang.org/std/primitive.str.html#impl-AsRef%3Cstr%3E-for-str

hugr-core/src/envelope/description.rs
```diff
@@ -0,0 +1,563 @@
...
+    fn from(ext: &E) -> Self {
+        let ext = ext.as_ref();
+        Self {
+            name: ext.name.to_string(),
+            version: ext.version.clone(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, serde::Serialize)]
```
doug-q: I would like Deserialize

hugr-core/src/envelope.rs
```diff
@@ -150,24 +151,65 @@ impl<E: std::fmt::Display> WithGenerator<E> {
...
+    registry: &ExtensionRegistry,
+) -> Result<(PackageDesc, Package), ReadError> {
+    let reader = EnvelopeReader::new(reader, registry).map_err(Box::new)?;
+    let (desc, res) = reader.read();
+    match res {
+        Ok(pkg) => Ok((desc, pkg)),
+        Err(e) => Err(ReadError::Payload {
+            source: Box::new(e),
+            partial_description: desc,
+        }),
```
doug-q: I think this is the main point of the whole operation, give a partial description on error.
Is it tested? (coverage checker broken for me right now). 
This behaviour isn't described in docs (maybe docs on ReadError are enough)

PR #2649, opened 2025-10-28
feat(hugr-passes)!: normalize_cfgs inlines entry DFG

acl-cqc:
Breaking followup to #2633.
* normalize_cfgs moves the entry block outside the CFG *even if there are outgoing `Dom` edges* by inlining the DFG (so the entry block children become siblings of the CFG), as this makes said edges into valid `Ext` edges.

Sadly, I don't see a good/corresponding treatment for the exit dfg;  it doesn't help that the source block of the `Dom` edges must dominate the exit (and hence we could break the CFG just before that source block and start another CFG), because even then there could be backedges to the source (new CFG entry) block, so we can't necessarily lift it outside. Options thus seem to be:
   1. thread the values from the source block through the whole CFG
   2. add extra outputs to the CFG, and thus to each predecessor of the new exit block - but that requires either finding the corresponding Tag, or also adding new inputs to any other successors of those predecessors (and any other predecessors of those, and so on)

...and I've not tried either of those here.

BREAKING CHANGE: NormalizeCFGResult specifies entry_nodes_moved not entry_dfg (as no DFG is inserted).

### DIFF
--- a/.github/workflows/ci-py.yml
+++ b/.github/workflows/ci-py.yml
@@ -52,7 +52,7 @@ jobs:
         uses: mozilla-actions/sccache-action@v0.0.9
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@v6
+        uses: astral-sh/setup-uv@v7
         with:
           version: ${{ env.UV_VERSION }}
           enable-cache: true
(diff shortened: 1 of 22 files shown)


hugr-passes/src/normalize_cfgs.rs
```diff
@@ -885,19 +883,17 @@ mod test {
         Ok(())
     }
 
-    #[rstest]
-    fn nested_cfgs_pass(#[values(true, false)] nonlocal: bool) {
+    #[test]
+    fn nested_cfgs_pass() {
```
acl-cqc: I suppose I could keep the parameterization, the two cases just become a lot more similar because they both get transformed now! ?

PR #2660, opened 2025-11-03
chore: cargo update - yanked smol_str

ss2165:


aborgna-q (approved):
The issue in the cargo-dist release was that we were using a rustc pre-installed in the `aarch64-pc-windows-msvc` runner that failed to build the (yanked) smol_str version in the lockfile.

I couldn't find info on what rustc version is actually being used, and if it will detect that the version now in the lockfile has a `msrv` that's too high for it...
(Cargo's msrv-aware `resolver v3` only [got released in rust 1.84](https://releases.rs/docs/1.84.0/#cargo))

PR #2666, opened 2025-11-04
fix: BorrowArray discard handler allows elements to be borrowed

acl-cqc:
fixes https://github.com/CQCL/guppylang/issues/1330

This is a mass of Hugr construction; the alternative of adding a new op to BorrowArray would do much the same in LLVM, and I don't think the increase in Hugr size should be very much - compared to a standard linear array discard, we have TailLoop rather than ArrayScan, with merely an extra conditional around the discarding of each element (and nodes moved into the main Hugr rather than a Value::Function, so it *looks* bigger).

Note there is no need to do this for BorrowArray::discard (the copyable-elements-only function), as that doesn't check borrowedness; copyable elements don't need discard functions so we just deallocate the lot without looking what's there.

I added hugr-passes as a dev dependency of hugr-llvm to get an execution test, and verified switching-round the senses of conditionals made this fail, which gives *some* confidence it's correct ;)

### DIFF
--- a/hugr-passes/src/replace_types/handlers.rs
+++ b/hugr-passes/src/replace_types/handlers.rs
@@ -451,13 +451,13 @@ pub fn copy_discard_borrow_array(
                 .unwrap();
             {
                 // borrowed - do nothing
-                let borrowed_case = cond2.case_builder(0).unwrap();
+                let borrowed_case = cond2.case_builder(1).unwrap();
                 let [arr] = borrowed_case.input_wires_arr();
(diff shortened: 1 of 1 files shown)


aborgna-q (approved):
> I had to add hugr-passes as a dev dependency of hugr-llvm

Should be fine, It makes more sense than testing llvm code on hugr-passes. If needed, we can move things to integration tests later.

> I suggest we make the "drop" op public as part of some hugr-passes extension

Maybe open an issue?

PR #2677, opened 2025-11-11
feat(cli, python): programmatic interface to cli with python bindings

### LINKED ISSUE #2671: Distribute cli with python wheel

ss2165:
Closes #2671

## Goals
1. Have `hugr` cli available in environment when installed via uv (`uvx hugr --help`)
2. Allow use of cli functionality from python package without spawning subprocess

## Implications
- Either ship binary with wheel or bind the function directly - chose the latter as it is more flexible and avoids a bunch of build faff
- CLI uses argc + stdin for input and stdout for output (by default), we need a way to do this via binding i/o to avoid subprocess

## Primary outcomes
- Cli invocation bound via pyo3 to python "package script" with same name as binary, meeting goal 1
- hugr-cli refactored to read inputs from a reader and write outputs to a writer when appropriate, satisfying goal 2. Uses an "override" pattern - when i/o is specified it overrides the corresponding cli-arg

## Secondary outcomes
- CLI main.rs contents moved to lib.rs, allowing use by library dependants
- Make tracing a default feature for hugr-cli (not used by hugr-py)
- `cli` module in python to wrap core sub-commands and parameters, with simply typed interfaces
- Python test cli use replaced with bound function calls - meaning no need to build cli before running python tests
- Model loading in python via conversion to json uses cli convert function rather than dedicated binding
- Dedicated cargo profile for python wheel for smaller wheels (very open to critique on this)
- Pydantic models to parse `describe` json output
- Remove parallel pytest since it is now faster without it

## Possible issues
- Versioning: the cli version is pinned to the rust crates, the python package is not. Meaning `uvx hugr --version` will report a different value to `uv --with hugr run python -c "import hugr; print(hugr.__version__)"`. Are we ok with this?

### DIFF
--- a/hugr-py/src/hugr/cli.py
+++ b/hugr-py/src/hugr/cli.py
@@ -23,31 +23,36 @@ __all__ = [
 ]
 
 
+def _add_input_args(
+    args: list[str], no_std: bool, extensions: list[str] | None
+) -> list[str]:
+    """Add common HugrInputArgs parameters to the argument list."""
+    if no_std:
+        args.append("--no-std")
+    if extensions is not None:
+        for ext in extensions:
+            args.extend(["--extensions", ext])
+    return args
+
+
 def validate(
     hugr_bytes: bytes,
     *,
-    quiet: bool = False,
     no_std: bool = False,
-    extensions: str | None = None,
+    extensions: list[str] | None = None,
 ) -> None:
     """Validate a HUGR package.
 
     Args:
         hugr_bytes: The HUGR package as bytes.
-        quiet: Suppress success messages (default: False).
-        no_std: Don't include the standard extension (default: False).
-        extensions: Path to additional extensions registry file.
+        no_std: Don't use standard extensions when validating hugrs.
+            Prelude is still used (default: False).
+        extensions: Paths to additional serialised extensions needed to load the HUGR.
 
     Raises:
         ValueError: On validation failure.
     """
-    args = ["validate"]
-    if quiet:
-        args.extend(["-q"])
-    if no_std:
-        args.append("--no-std")
-    if extensions:
-        args.extend(["--extensions", extensions])
+    args = _add_input_args(["validate"], no_std, extensions)
     cli_with_io(args, hugr_bytes)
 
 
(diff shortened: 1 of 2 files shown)


hugr-cli/src/main.rs
```diff
@@ -1,73 +1,6 @@
...
-        _ => Err(anyhow!("Unknown command")),
-    };
-
-    if let Err(err) = result {
-        error!("{:?}", err);
-        std::process::exit(1);
-    }
-}
-
-fn run_external(args: Vec<OsString>) -> Result<()> {
```
ss2165: moved unchanged to lib.rs

hugr-py/src/hugr/cli.py
```diff
@@ -0,0 +1,259 @@
...
+    Args:
+        hugr_bytes: The HUGR package as bytes.
+        packaged_extensions: Show packaged extensions (default: False).
+        no_resolved_extensions: Hide resolved extensions (default: False).
+        public_symbols: Show public symbols (default: False).
+        generator_claimed_extensions: Show generator claimed extensions
+            (default: False).
+
+    Returns:
+        Text description of the package..
```
cqc-alec:
```suggestion
        Text description of the package.
```

cqc-alec (commented):
Looks good, just a few comments.

hugr-py/rust/lib.rs
```diff
@@ -75,6 +61,43 @@ fn current_model_version() -> (u64, u64, u64) {
...
+/// * `args` - Command line arguments as a list of strings, not including the executable name.
+/// * `input_bytes` - Optional input data as bytes (e.g., a HUGR package)
+///
+/// # Returns
+///
+/// Returns the command output as bytes, maybe empty.
+/// Raises an exception on error.
+///
+/// Errors or tracing may still be printed to stderr as normal.
+/// ```
```
aborgna-q:
```suggestion
///
```

hugr-py/src/hugr/cli.py
aborgna-q: It may be a bit confusing to call this module `cli` when including it in the python lib.

From a user standpoint, there's no connection between `hugr.cli.describe` or `hugr.cli.mermaid` and a command line interface.

ss2165: I personally like making it clear that it relies on the cli interface, it makes the idiosyncrasies of the interface easier to understand (e.g. why pass bytes?). The module docstring explains why it is this way.

PR #2681, opened 2025-11-13
feat: return description output to python on error

ss2165:


cqc-alec (commented):
Just need to fix the CI error...

PR #2697, opened 2025-11-18
feat: Make `ComposablePass.__call__` return a Hugr

CalMacCQ:
Implementing the revisions to `ComposablePass` and `ComposedPass` as discussed with Agustin. 

I still find the `_apply` and `_apply_inplace` rather unintuitive (although now I realise that it will work). Looking at the default implementations in the `ComposablePass` protocol these look circular but as you're always overrididing at least one of the `_apply` or `_apply_inplace` methods whenever you implement the protocol.

Would be interested to know if there are suggestions to improve this.

### DIFF
--- a/hugr-core/src/extension/prelude.rs
+++ b/hugr-core/src/extension/prelude.rs
@@ -731,8 +731,8 @@ impl MakeRegisteredOp for MakeTuple {
         PRELUDE_ID.clone()
     }
 
-    fn extension_ref(&self) -> Weak<Extension> {
-        Arc::downgrade(&PRELUDE)
+    fn extension_ref(&self) -> Arc<Extension> {
+        PRELUDE.clone()
     }
 }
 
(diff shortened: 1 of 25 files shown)


hugr-py/src/hugr/passes/_composable_pass.py
```diff
@@ -16,9 +17,23 @@
 class ComposablePass(Protocol):
     """A Protocol which represents a composable Hugr transformation."""
 
-    def __call__(self, hugr: Hugr) -> None:
+    def __call__(self, hugr: Hugr, inplace: bool = True) -> Hugr:
```
aborgna-q:
```suggestion
    def __call__(self, hugr: Hugr, *, inplace: bool = True) -> Hugr:
```

hugr-py/src/hugr/passes/_composable_pass.py
```diff
@@ -48,7 +63,8 @@ class ComposedPass(ComposablePass):
 
     passes: list[ComposablePass]
 
-    def __call__(self, hugr: Hugr):
+    def __call__(self, hugr: Hugr, inplace: bool = True) -> Hugr:
         """Call all of the passes in sequence."""
         for comp_pass in self.passes:
-            comp_pass(hugr)
+            comp_pass(hugr, inplace)
```
aborgna-q: if `inplace = False` this is discarding the updated hugrs

aborgna-q (approved):
#2703 builds on this, so lets merge and iterate over it

PR #2698, opened 2025-11-18
feat: add hugr-core StaticGraph, deprecate hugr-passes CallGraph

acl-cqc:
Respin of #2531, marking the node/edge enums in the new StaticGraph as `non-exhaustive` and adding Consts + LoadConstants.

(Lukas originally planned to make something like this to reflect references from Terms to functions, and non-exhaustive allows extending this StaticGraph for that purpose - if we do go that way, however, I have a different plan, can describe if relevant. Rather I think non-exhaustive is just a reasonable allowance for future extension to new use-cases hopefully without damaging existing use-cases.)

This is taken from #2555, i.e. which uses it (although there it is still called CallGraph in hugr-core).

Naming: could be ModuleGraph ?? Other ideas welcome.

I will add a test with some Consts + use of `out_edges` (tho will be supplied in #2555 !)

### DIFF
--- a/hugr-core/src/module_graph.rs
+++ b/hugr-core/src/module_graph.rs
@@ -154,3 +154,57 @@ impl<N: HugrNode> ModuleGraph<N> {
         })
     }
 }
+
+#[cfg(test)]
+mod test {
+    use itertools::Itertools as _;
+
+    use crate::builder::{
+        Container, Dataflow, DataflowSubContainer, HugrBuilder, ModuleBuilder, endo_sig, inout_sig,
+    };
+    use crate::extension::prelude::{ConstUsize, usize_t};
+    use crate::ops::{Value, handle::NodeHandle};
+
+    use super::*;
+
+    #[test]
+    fn edges() {
+        let mut mb = ModuleBuilder::new();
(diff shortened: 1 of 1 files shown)


hugr-core/src/static_graph.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// a [`FuncDefn`](OpType::FuncDefn). Note that it will not be a [Module](OpType::Module)
+    /// either, as such a node could not have edges, so is not represented in the petgraph.
+    NonFuncEntrypoint,
+    /// petgraph-node corresponds to a constant; will have no outgoing edges, and incoming
+    /// edges will be [StaticEdge::LoadConstant]
+    Const(N),
+}
+
+/// Details the [`FuncDefn`]s, [`FuncDecl`]s and module-level [`Const`]s in a Hugr,
+/// in a Hugr, along with the [`Call`]s, [`LoadFunction`]s, and [`LoadConstant`]s connecting them.
```
mark-koch:
```suggestion
/// along with the [`Call`]s, [`LoadFunction`]s, and [`LoadConstant`]s connecting them.
```

PR #2703, opened 2025-11-20
feat: Result type for ComposablePasses

aborgna-q:
Includes an idea for simplifying the protocol's `_apply`/`_apply_inline` from #2697 by providing a helper function instead (859c8111b498fb21532274ef862318d4eaf0a876).

### DIFF
--- a/hugr-py/src/hugr/passes/_composable_pass.py
+++ b/hugr-py/src/hugr/passes/_composable_pass.py
@@ -72,7 +72,7 @@ def implement_pass_run(
             pass_result.hugr = hugr
             if pass_result.modified:
                 hugr._overwrite_hugr(pass_result.hugr)
-                pass_result.original_dirty = True
+                pass_result.inplace = True
             return pass_result
     elif not inplace:
         if copy_call is not None:
@@ -80,7 +80,7 @@ def implement_pass_run(
         elif inplace_call is not None:
             new_hugr = deepcopy(hugr)
             pass_result = inplace_call(new_hugr)
-            pass_result.original_dirty = False
+            pass_result.inplace = False
             return pass_result
 
     msg = (
@@ -106,7 +106,7 @@ class ComposedPass(ComposablePass):
 
     def run(self, hugr: Hugr, *, inplace: bool = True) -> PassResult:
         def apply(inplace: bool, hugr: Hugr) -> PassResult:
-            pass_result = PassResult(hugr=hugr)
+            pass_result = PassResult(hugr=hugr, inplace=inplace)
             for comp_pass in self.passes:
                 new_result = comp_pass.run(pass_result.hugr, inplace=inplace)
                 pass_result = pass_result.then(new_result)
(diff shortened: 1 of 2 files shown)


CalMacCQ (approved):
It looks nice, thanks. I find this more intuntive than having `_apply` and `_apply_inplace` methods (at least one of which must be overriden when the protocol is implemented.

One question for clarification and also one nit.

hugr-py/src/hugr/passes/_composable_pass.py
```diff
@@ -6,55 +6,83 @@
...
+
+    :param pass_: The pass being run. Used for error messages.
+    :param hugr: The Hugr to apply the pass to.
+    :param inplace: Whether to apply the pass inplace.
+    :param inplace_call: The method to apply the pass inplace.
+    :param copy_call: The method to apply the pass by copying the Hugr.
+    :return: The result of the pass application.
+    :raises ValueError: If neither `inplace_call` nor `copy_call` is provided.
+    """
+    if inplace and inplace_call is not None:
```
acl-cqc: nit: would this be easier to read as two nested ifs?
```
if inplace:
   if inplace_call is not None:
      ...
   else:
      ...
else: # copying
   if copy_call is not None:
      ...
   else:
      ...
```
I mean, turn the outer `else` into `else if`, sure

hugr-py/tests/test_passes.py
```diff
@@ -1,30 +1,105 @@
...
+                    hugr,
+                    result=None,
+                    inline=True,
+                    # Say that we modified the HUGR even though we didn't
+                    modified=True,
+                ),
             )
 
-    dummy = MyDummyPass()
+    class MyDummyCopyPass(ComposablePass):
```
cqc-alec:
```suggestion
    class DummyCopyPass(ComposablePass):
```

hugr-py/src/hugr/passes/_composable_pass.py
```diff
@@ -89,24 +93,90 @@ class ComposedPass(ComposablePass):
...
+    Includes a flag indicating whether the passes modified the HUGR, and an
+    arbitrary result object for each pass.
+
+    In some cases, `modified` may be set to `True` even if the pass did not
+    modify the program.
+
+    :attr hugr: The transformed Hugr.
+    :attr original_dirty: Whether the original HUGR was modified by the pass.
+    :attr modified: Whether the pass made changes to the HUGR.
+    :attr results: The result of each applied pass, as a tuple of the pass and
```
cqc-alec:
```suggestion
    :attr results: The result of each applied pass, as a tuple of the pass name and
```

hugr-py/src/hugr/passes/_composable_pass.py
```diff
@@ -89,24 +93,90 @@ class ComposedPass(ComposablePass):
...
+        the result.
+    """
+
+    hugr: Hugr
+    original_dirty: bool = False
+    modified: bool = False
+    results: list[tuple[PassName, Any]] = field(default_factory=list)
```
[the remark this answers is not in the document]
aborgna-q: Mostly for debuggability, otherwise `ComposedPass.run` would return a `List[Any]` of arbitrary payloads without much indication of what came from where.

The result may also be serialized, so the original `ComposedPass` and its list of passes may be lost.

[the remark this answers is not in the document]
acl-cqc: I thought including the pass name was quite a nice solution!

PR #2715, opened 2025-11-26
fix: set hugr field of `PassResult` correctly

CalMacCQ:


acl-cqc (approved):
Looks good thanks Callum!

Bonus points for improving the test that it would have spotted the problem before we fixed it ;-) :-)

PR #2719, opened 2025-11-26
chore(release-py): bring hugr-py-v0.14.4 updates to main

acl-cqc:
Modelled on #2708 (for 0.14.3) but this time including `.release-please-manifest.json`...oooh, should I? Will that confuse release-please when we do make a (breaking) release actually from `main` ??


.release-please-manifest.json
```diff
@@ -1,3 +1,3 @@
 {
-    "hugr-py": "0.14.2"
+    "hugr-py": "0.14.4"
```
aborgna-q: release-please uses this to decide which commits to consider, looking at the release tag.

This patch tag is out-of-tree, so the automation wouldn't be able to find the right range.

We should leave it at `0.14.2`, and remove the already-picked changelog lines manually.

```suggestion
    "hugr-py": "0.14.2"
```

PR #2720, opened 2025-11-26
docs: Improve instructions for patch release

acl-cqc:


DEVELOPMENT.md
```diff
@@ -240,10 +240,19 @@ release-plz release --git-token $GITHUB_TOKEN
...
 the existing release PRs for examples on how to do this. Once the branch is
 ready, create a draft PR so that the release team can review it.
 
 The wheel building process and publication to PyPI is handled by the CI.
 Just create a [github release](https://github.com/CQCL/hugr/releases/new) from the **unmerged** branch.
 The release tag should follow the format used in the previous releases, e.g. `hugr-py-v0.1.1`.
+
+**Release-please method** Make sure the name of your branch begins `release/` so
+the release-please workflow automatically runs on it. This will create the draft
+PR with ChangeLog updates. When approved, merge, and this will trigger release-please to actually make the release.
```
acl-cqc: This is only 90% what I did. I created the branch without the `release/` prefix so had to manually run release-please. Then I created the same branch with that prefix, and I think #2718 was automatically created as expected...but the release PR that I merged was the first one, with me manually changing the PR target branch.

So I'm 95% sure this is correct / all you need to do, but not quite the final 5%....

PR #2743, opened 2025-12-09
chore: Use working syntax for nextest install check in justfile setup

maximilianruesch:
The nextest install check during `just setup` was broken since it used a just recipe invocation syntax that is not valid. The `setup` recipe has been updated to use a "post" command, which is equivalent to running the named recipe as a last command in the parent recipe. See https://github.com/casey/just/discussions/818#discussioncomment-11854830 for more information.

Discovered during Hugr setup.


doug-q (approved):
Thanks. I merged main because this wasn't building for me, msrv had not been updated.

PR #2755, opened 2025-12-12
feat!: Type-safe access for node metadata

### LINKED ISSUE #2651: Centralise `core` metadata keys in a module
Metadata keys are defined as constants in random places, so hard to keep track of. Centralise them. E.g.:
https://github.com/CQCL/hugr/blob/01e7ac0000875ff08a902dca356dfc6002857a50/hugr-core/src/envelope.rs#L63

### LINKED ISSUE #2753: Typed node metadata
Define a trait
```rust
trait Metadata {
    type PayloadType: serde::de::DeserializeOwned + serde::ser::Serialize;
    const KEY: &'static str;
}
```

and modify the hugrview accessors so we get type safety for known metadata values,

```rust
trait HugrView {

    fn get_metadata<T: Metadata>(&self, node: Node) -> Option<T::PayloadType> {
        self.get_metadata_any(node).and_then(|value| {
            serde_json::from_value(value).ok()
        }
    }

    fn get_metadata_any(&self, node: Node) -> Option<serde_json::Value>;
}
```

aborgna-q:
Closes #2753 
Closes #2651 

Adds a `Metadata` trait that can be used for type-safe access of metadata entries on the rust library.
Moves the existing metadata keys definitions to `crate::metadata`.

To access a metadata value, the new methods take the metadata marker as a type parameter,
```rust
struct SomeUsizeMetadata;
impl Metadata for SomeUsizeMetadata {
    type Type<'hugr> = usize;
    const KEY: &'static str = "some.key";
}

fn HugrView::get_metadata<M: Metadata>(&self, node: Self::Node) -> Option<<M as Metadata>::Type<'_>>;

hugr.set_metadata::<SomeUsizeMetadata>(node, 42);
let val: Some(usize) = hugr.get_metadata::<SomeUsizeMetadata>(node);
```
Note that we always need to explicitly set the metadata marker when calling `get_metadata`, since the marker cannot be derived from the return type.

An alternative design would have been to return the Metadata object itself,
```rust
fn HugrView::get_metadata<M: Metadata>(&self, node: Self::Node) -> Option<Metadata<'_>>;
```
but in most cases we are just storing intrinsic types for payloads, so wrapping them into new structs is just extra boilerplate. We are already getting the type safety from explicitly stating the markers on the method calls.

BREAKING CHANGE: `get`/`set_metadata` methods now use the new `Metadata` trait. Use `get_metadata_any` for untyped access.

### DIFF
--- a/hugr-core/src/envelope/description.rs
+++ b/hugr-core/src/envelope/description.rs
@@ -1,6 +1,4 @@
 //! Description of the contents of a HUGR envelope used for debugging and error reporting.
-use std::collections::HashMap;
-
 use crate::envelope::HugrUsedExtensions;
 use crate::metadata;
 use crate::{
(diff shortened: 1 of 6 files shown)


hugr-core/src/envelope/description.rs
```diff
@@ -156,6 +162,130 @@ impl<E: AsRef<crate::Extension>> From<&E> for ExtensionDesc {
     }
 }
 
+/// Description of the generator that defined the module.
+///
+/// These are stored at the module root node metadata under the [`crate::metadata::HugrGenerator`] entry.
+#[derive(derive_more::Display, Debug, Clone, PartialEq)]
+pub enum GeneratorDesc {
```
aborgna-q: I defined an actual type to deserialize, since the logic read either a json string _or_ a json object and handled them separately.

PR #2760, opened 2025-12-15
refactor!: Delete ValueArray

### LINKED ISSUE #2398: Delete `value_array`

acl-cqc:
Joint PR with @tatiana-s
closes #2398

Removes ValueArray and conversion to Array.

The handlers for linearizing (non-linear code using) Array remain, as these do not actually depend on ValueArray, and we still need the `num_outports==0` case for Guppy's `drop` op, so I'm retaining the code for arbitrary #outports even though this is not used "in production".

BREAKING CHANGE: ValueArray is gone, as is LinearizeArrayPass

### DIFF
--- a/hugr-passes/src/replace_types.rs
+++ b/hugr-passes/src/replace_types.rs
@@ -1089,7 +1089,7 @@ mod test {
                 .collect_vec(),
             ["get", "itousize", "panic"]
         );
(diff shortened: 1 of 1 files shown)


doug-q (approved):
This is scary. Downstream compilers should be updated.

PR #2765, opened 2025-12-17
feat(hugr-py): Define typed Metadata protocol

### LINKED ISSUE #2757: Export core metadata keys in `hugr-py`
Follow up to https://github.com/Quantinuum/hugr/pull/2755, which added type-safe metadata to the rust lib and centralized the definitions.

We should also define a `hugr.metadata` submodule in `hugr-py` and re-export the keys from the rust side using the bindings.
Can we define some safe `Protocol` for these?

aborgna-q:
Adds a `Metadata` protocol similar to the one in the rust side, and exports the rust key definitions.

~Depends on #2764~

Closes #2757

### DIFF
--- a/hugr-py/Cargo.toml
+++ b/hugr-py/Cargo.toml
@@ -20,8 +20,6 @@ test = false
 bench = false
 
 [dependencies]
-
-
 bumpalo = { workspace = true, features = ["collections"] }
 hugr-core = { version = "0.25.5", path = "../hugr-core", default-features = false }
 hugr-cli = { version = "0.25.5", path = "../hugr-cli", default-features = false }


hugr-py/rust/lib.rs
```diff
@@ -1,12 +1,15 @@
...
 use pyo3::pymodule;
 
 #[pymodule]
 mod _hugr {
-
     #[pymodule_export]
     use super::model::model;
+
+    #[pymodule_export]
+    use super::metadata::metadata;
```
maximilianruesch: Depending on whether you usually order these uses alphabetically, this one is in the wrong place (since "me..." < "mo...")

hugr-py/src/hugr/hugr/base.py
```diff
@@ -964,8 +969,8 @@ def _serialize_link(
             serial_idx = len(nodes)
 
             # non contiguous indices will be erased
-            nodes.append(data._to_serial(Node(serial_idx, {})))
-            metadata.append(data.metadata if data.metadata else None)
+            nodes.append(data._to_serial(Node(serial_idx, NodeMetadata())))
+            metadata.append(data.metadata._dict if data.metadata else None)
```
maximilianruesch: Similar question about private member access as https://github.com/Quantinuum/hugr/pull/2765/changes#r2631034862

hugr-py/src/hugr/hugr/base.py
```diff
@@ -921,7 +926,7 @@ def insert_hugr(self, hugr: Hugr, parent: ToNode | None = None) -> dict[Node, No
                 node_data.op,
                 node_parent,
                 num_outs=node_data._num_outs,
-                metadata=node_data.metadata,
+                metadata=node_data.metadata.as_dict(),
```
maximilianruesch: If you convert this to a dict (even though it will be later reconverted into a NodeMetadata again) for copying reasons, this should be made explicit e.g. through a `clone` or `copy` function on `NodeMetadata`. This `.as_dict()` call here is too easy to (accidentally) remove, especially since both a dict and `NodeMetadata` are accepted.

hugr-py/src/hugr/hugr/base.py
```diff
@@ -344,16 +345,20 @@ def _add_node(
...
-        metadata: dict[str, Any] | None = None,
+        metadata: dict[str, Any] | NodeMetadata | None = None,
     ) -> Node:
         parent = parent.to_node() if parent else None
-        node_data = NodeData(op, parent, metadata=metadata or {})
+
+        if metadata is None or isinstance(metadata, dict):
+            metadata = NodeMetadata(metadata)
+
+        node_data = NodeData(op, parent, metadata=metadata)
```
maximilianruesch: How is the metadata of `NodeData` connected to that of `Node`? What is the difference?

hugr-py/src/hugr/hugr/base.py
```diff
@@ -474,7 +479,7 @@ def delete_node(self, node: ToNode) -> NodeData | None:
         weight, self._nodes[node.idx] = self._nodes[node.idx], None
 
         # Free up the metadata dictionary
-        node._metadata = {}
+        node._metadata = NodeMetadata()
```
maximilianruesch: Would it be useful to "free" by assigning `None` instead of a fresh object and only creating a new instance when needed (aka when reusing the freed node)? This ensures at the usage site that there is a cleared metadata attached here, instead of making this part of the contract of `delete_node`.

I do also understand the argument of "but I want to have type safety that as long as the node object exists it has a valid pointer to some Metadata instance", so choose your evil / good.

hugr-py/src/hugr/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+
+    KEY = rust_metadata.HUGR_GENERATOR
+
+    @classmethod
+    def to_json(cls, value: GeneratorDesc) -> dict[str, str]:
+        return value._to_json()
+
+    @classmethod
+    def from_json(cls, value: Any) -> GeneratorDesc:
+        return GeneratorDesc._from_json(value)
```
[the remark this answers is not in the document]
aborgna-q: 👍 

For native-type metadata (ints, strings, lists, and dicts) there's no need to override these methods. But yeah, it'd be nice to have a _deserialize_/_serialize` 
I guess we could use pydantic for that? It's a bit annoying that we're starting from the already-loaded object rather than a json string.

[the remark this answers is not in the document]
maximilianruesch: I have not yet used pydantic, but it strikes me as very powerful. If we adopt it, we should assess whether there is considerable overhead to serialization / deserialization that is not ignorable in the context of `hugr-py`.

hugr-py/src/hugr/hugr/base.py
```diff
@@ -1022,12 +1027,11 @@ def _from_serial(cls, serial: SerialHugr) -> Hugr:
         """Load a HUGR from a serialized form."""
         assert serial.nodes, "The encoded Hugr is empty"
 
-        def get_meta(idx: int) -> dict[str, Any]:
-            if not serial.metadata:
-                return {}
-            if idx < len(serial.metadata):
-                return serial.metadata[idx] or {}
-            return {}
```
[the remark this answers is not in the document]
aborgna-q: It was a leftover from when the constructor took a `dict[str | Metadata]`, to avoid the invariant type mismatchs.

I changed all the direct accesses.

hugr-py/src/hugr/hugr/base.py
aborgna-q: Removed the `as_dict` and added a shallow `copy`.
Do you think that's correct, or should we be doing a `deepcopy`? (I'm not too familiar with python expected patterns)

hugr-py/src/hugr/hugr/base.py
aborgna-q: Yeah, I think it's better for objects to always be correct by construction (with things always initialized).
Making `_metadata` nullable would require adding extra checks to each access.

I agree it's just preference though

maximilianruesch (changes requested):

PR #2770, opened 2025-12-19
refactor: Deprecate Value::Function and inline_constant_functions

### LINKED ISSUE #2353: Deprecate Value::Function
When we have some form of linking (ideally #2517),  DFG-entrypoint Hugrs can carry functions along with them (beneath the module root), and Value::Function's can then be replaced by FuncDefn + LoadFunction. (As long as the Value::Function isn't in a tuple, admittedly. That will probably have to wait for hugr-model.)

acl-cqc:
closes #2353

### DIFF
--- a/hugr-core/src/ops/constant.rs
+++ b/hugr-core/src/ops/constant.rs
@@ -376,10 +376,10 @@ impl Value {
     ///
     /// Returns an error if the Hugr root node does not define a function.
     #[deprecated(note = "Flatten and lift contents to a FuncDefn", since = "0.25.0")]
+    #[expect(deprecated)] // Remove along with Value::Function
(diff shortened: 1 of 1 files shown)


hugr-passes/src/dataflow.rs
```diff
@@ -69,6 +69,7 @@ pub trait ConstLoader<V> {
 
     /// Produces an abstract value from a Hugr in a [`Value::Function`], if possible.
     /// The default just returns `None`, which will be interpreted as [`PartialValue::Top`].
+    #[deprecated(note = "Remove along with Value::Function")]
```
aborgna-q:
```suggestion
    #[deprecated(note = "Remove along with Value::Function", since = "0.25.0")]
```

aborgna-q (commented):
There are some `#![allow(deprecated)]`, can't those use `expect`?

Issue #2771, opened 2025-12-19
Migrate passes to use the new PassScope


aborgna-q:
See #2772 

```rs
/// Set the scope configuration used to run the pass.
///
/// See [`PassScope`] for more details.
///
/// In `hugr 0.25.*`, this configuration is only a guidance, and may be
/// ignored by the pass.
///
/// From `hugr >=0.26.0`, passes must respect the scope configuration.
fn ComposablePass::with_scope(self, scope: &PassScope) -> Self;
```

- [ ] `ConstantFoldPass`
- [ ] `DeadCodeElimPass`
- [ ] `RemoveDeadFuncsPass`
- [ ] `InlineDFGsPass`
- [ ] `MonomorphizePass`
- [ ] `LocalizeEdges`
- [ ] `NormalizeCFGPass`
- [ ] `RedundantOrderEdgesPass`
- [ ] `ReplaceTypes`
- [ ] `UntuplePass`

See #2871, #2836

PR #2772, opened 2025-12-19
feat: Define pass application scopes

### LINKED ISSUE #2748: Optimization Scope
For controlling what *interfaces* an optimization must preserve (typically functions, but also perhaps the entrypoint if a DFG)

We may want ability to tell passes not to waste compute time on optimizing other parts, but this is less of an issue - few classical compilers bother very much! Instead the preferred way is to split your Hugr into several, optimize them separately/differently, and then link them back together.

Hence, propose to combine these two concepts into a single enum:
```rs
#[non_exhaustive]
enum OptimizationScope<N> {
   SpecificFuncs (Vec<N>), // heavily discouraged because serialization renumbers - maybe remove altogether and require setting visibility??
    // maybe SpecificPublicFunctions(Vec<String>), etc. etc.

   AllPublic (include_entrypoint: bool), // true has no effect if entrypoint==module-root

   All (include_entrypoint: bool), // true has no effect if entrypoint==module-root

   Aggressive, // if entrypoint is module root, then AllPublic(false); else just entrypoint

   Local, // use entrypoint, do not recurse (possibly error on module-entrypoint?)
}
```
`Default` == `All(true)` as this is pretty conservative. Indeed, **perhaps too conservative**: `All` means DeadFunctionRemoval can't remove unused private functions! (It makes no sense to allow removal of functions but not alteration; a pass could create a new function, change all calls to use that, and remove the old one!)

Add ComposablePass::set_scope (breaking change as new pass method)

Note SpecificFuncs requires them all to be top-level/module children, so cannot express arbitrary entrypoint

There is some method `to_nodes<H: HugrView>(&OptimizationScope, &H) -> Vec<H::Node>` (all elements of which are *not* guaranteed to be module children), but this might be hidden. (Or, this might want to return 2*Vec: places to examine/transform, and funcs whose semantics to preserve - the former must be a superset of the latter but usually plus reachable funcs - e.g. if we are preserving interface of public funcs, we might still butcher up the private ones according to how they are used.)

aborgna-q:
Defines `PassScope` configurations in both `hugr-passes` and `hugr-py` that let users define set the parts of the hugr that a pass should optimize.

Since each pass must be carefully modified to support the new config, this PR only adds the definitions (and a `ComposablePass::with_scope` method), and states that passes are not required to follow the configuration (for now).
This will let us make the required changes incrementally as non-breaking changes. I opened an issue to track the implementations. #2771 

Closes #2748. See discussion about pass properties there. We left the `NamedFunctions` variant to be defined later.

Python tests included adding optional `visibility` param to `Module.define_function` and `declare_function`.

### DIFF
--- a/hugr-py/tests/test_scope.py
+++ b/hugr-py/tests/test_scope.py
@@ -22,7 +22,7 @@ class TestHugr:
 
 
 @pytest.fixture
-def test_hug() -> TestHugr:
+def test_hugr() -> TestHugr:
     mod = Module()
     module_root = mod.hugr.module_root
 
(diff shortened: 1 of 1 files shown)


acl-cqc (approved):
Agustin, myself and Callum all approve, so never mind who's PR it is or who wrote the code, that seems like enough!

PR #2781, opened 2025-12-29
feat: Re-export hugr_core::hugr::linking from hugr::

aborgna-q:


acl-cqc (approved):
Yes good plan, thanks!

Issue #2341, opened 2025-06-16
Merge `Type` and `Term` in `hugr-core`.


zrho:
Includes removing Row Variables and `<RV>` parametrization from Types, instead using Term::ListConcat/TupleConcat.
* Add Term::RuntimeSum - i.e. equivalent to Type::Sum, storing a SumType. (Serializes to `core.adt`, code already present)
* SumType could become either
   * A Term of type `Term::ListType(Term::ListType(Term::RuntimeType(x)))` if we want to allow the number of *variants* to be extensible, or
   * A `Vec<Term>` each of type `Term::ListType(Term::RuntimeType(x))` if we want fixed #variants but variable number of elements in each variant - this is equivalent to current SumType storing a `Vec<Vec<Type<RV>`
   * May also want the Simple/Unit(num variants) option for efficient storage
   * May want to cache the computed TypeBound inside the SumType, as we do not have Term/TermEnum (like current Type/TypeEnum) so cannot cache the bound there
* Add Term::RuntimeFunc - to support row vars, this stores a Term for inputs and another for outputs, each of type `Term::ListType(Term::RuntimeType(x))`
* Add `Term::RuntimeExtension` i.e. paralleling current Type::Extension (but currently for runtime types only. In #2296 we'll make this into Term::Extension that can be any term, so perhaps best to call it that now, even if a little inconsistent.)
* Each of these `check_term_type`s against `Term::RuntimeType` (of appropriate TypeBound)
* Remove Term::Runtime and define `type Type = Term` for now. However, anything that was `Type` (i.e. `TypeBase<NoRV>`) now needs to be `check_term_type`d against `Term::RuntimeType(TypeBound::Linear)` - this may be a large part of the PR (making up for the loss of the Type/Term distinction in Rust), and/or may just mean combining `check_term_type` with `TypeBase::validate`.
   * (`TypeRV` is more like a list of type; again we have lost Rust automatically checking of what's got RVs vs not, i.e. reverting #1138, but the `ListConcat`/`TupleConcat` is much nicer than what we had before #1138.)
* `Signature` (no toplevel row vars) has `Vec<Term>` (each element of type Term::RuntimeType) for inputs/outputs
* Note difference between SumType (has row vars) and `Vec<Vec<Type>>` (no RowVars); we can convert rowvar-less sumtypes into the latter form by matching Term::List (*not* ListConcat).

**Serialization**
[...]

PR #2784, opened 2026-01-02
feat!: TypeRow: add impl From array of Type, remove From<Type>

acl-cqc:
This somewhat looks forward to #2341 but seems like a good thing right now.

Removal of `impl from Type(RV) for TypeRow(RV)` requires an extra `vec![...]` at each callsite, but if we add `impl <const N: usize> From<[Type; N] for TypeRow` (further parametrized by RV) then only two extra characters `[`...`]` are required, which seems bearable; and this strengthens the distinction between type and list-of-type.

BREAKING CHANGE: implicit conversion from single Type to TypeRow removed, use array of single Type instead.


ss2165 (approved):
I agree pretty strongly that implicitly converting T to Vec<T> makes it much harder for the rust type system to find bugs for us and causes (me) confusion, so I'm very happy with this change. Will defer the utility/performance of TypeRow to future discussions.

Issue #2756, opened 2025-12-12
Unify `Value` with `Term`


acl-cqc:
Prereqs: #2296; ideally #2266, #2353
 
* Add some `Term::SumValue(tag: usize, items: Vec<Term>)` - this `check_term_type`s against `Term::ConstType(Type::Sum(...))` if the `items` each `check_term_type` against `Term::ConstType(...)` of the relevant sum element.
  * We may need to store the target SumType, as `hugr_core::ops::constant::Sum` does, to support an equivalent of `Value::get_type` on Terms that are constants - `Term::get_const_type(&self) -> Option<Type>`, say. (See below for an alternative.)
  * Note hugr-core `Term::ConstType` should serialize to/from a hugr-model `Term::Apply` of `CORE_CONST` but this has not been implemented yet (in either direction)
  * SumValue (Like Value::Sum) serializes as a `CORE_CONST_ADT`, although the docs for CORE_CONST_ADT look dodgy (instead see uses in `fn export_value` and `fn import_value`):
```
/// Constants for runtime algebraic data types.
///
/// - **Parameter:** `?variants : (core.list core.type)`     -- should be (core.list (core.list core.type)) ?
/// - **Parameter:** `?types : (core.list core.static)`     -- written out as Wildcard, ignored when read in
/// - **Parameter:** `?tag : core.nat`
/// - **Parameter:** `?values : (core.tuple ?types)`. -- so perhaps `?types` is intended to be `?variants[?tag]` and inferred
/// - **Result:** `(core.const (core.adt ?variants))`
pub const CORE_CONST_ADT: &str = "core.const.adt";
```
* We are expecting to deprecate `Value::Function` #2353 so hopefully don't need to add that here.
* Probably also add `Term::OpaqueConst` containing a `dyn CustomConst` as a migration aid (serializing to/from  hugr-model `COMPAT_CONST_JSON`)
* Since this follows #2296, we can use the same `Term::Extension` instantiation of a `TermDef`, giving us *inspectable* extension values, rather than opaque ones based on `trait CustomConst`.
* Add new `LoadTerm` op that stores a Term (which must `check_term_type` against some `Term::Const(t)`, and then has a single value outport with type `t`).
   * This requires being able to synthesize the type of the Term (the above `get_const_type`) rather than having LoadTerm store the type as well; another option would be a struct equivalent to `Value` that store both Term and result-type (with methods to build a Value from a TermDef+args or compose Values into sums) - this might be better?
[...]

Issue #2558, opened 2025-09-16
Overhaul or scrap aliases


zrho:
Aliases are currently unused and limited in functionality. Should we keep them around? If so, we should allow aliases to refer to arbitrary terms that are not types. This is already possible in `hugr-model` but can not be represented yet in `hugr-core`.

Issue #2296, opened 2025-06-03
model-style Custom constructors in hugr-core, i.e. TypeDef -> TermDef


zrho:
Follow-up to #2341.

Then, extend `TypeDef` (making it `TermDef`). Presently TypeDef stores a TypeBound, and instances are `Term::RuntimeType(thatTypeBound)`; instead, store a Term, with instances have type of that Term. (How to make this dependent on bounds of the typeargs to the typedef, remains to be determined. Perhaps some enum that's either a Term (perhaps instantiated *using* the args to the typedef) or a TypeDefBound (effectively using the *bounds* of the args to the typedef, which we can't express in the former).

Instantiations of the TypeDef then become arbitrary Terms of that Type, giving us custom TypeArgs / custom constructors. (Thus, the Term::RuntimeExtension from #2341 is now Term::Extension.)

We'll want some validation: a suggested restriction would be that a TypeDef's result type is either (a) Term::RuntimeType(Linear/Copyable), (b) Term::StaticType, (c) an instance of a TermDef *defined in the same extension*

Finally, the Term::RuntimeExtension (introduced in #2341 but always producing a RuntimeType) becomes Term::Extension i.e. producing whatever the instantiated TypeDef says ;)

**Serialization** Extensions are not serialized using hugr-model, so this still relates to json. Basically we are adding a new field to TypeDef, but depending on exactly how the output type is represented in Rust, we might need a custom serializer/deserializer.

PR #2785, opened 2026-01-02
refactor(hugrv2)!: WIP+RFC: Unify Type and Term

### LINKED ISSUE #2341: Merge `Type` and `Term` in `hugr-core`.

acl-cqc:
Follows #2784; closes #2341.

* Add `Term::RuntimeFunction`, `Term::RuntimeSum` and `Term::RuntimeExtension`, removing `Term::Runtime(Type)` as well as `enum Type` (now an alias for `Term`) and `TypeEnum` (RIP).
   * These are named so that what were previously `Type`s are now RuntimeXXX's. However this new convention is not consistent with (does not apply to) `Term::RuntimeType` :-(... so the existing convention, that the name XXXType (e.g. `Term::RuntimeType`, `Term::StringType`) means a type of `Term`s i.e. in the static system, takes priority.
   * I think this nomenclature is confusing and might be better to (a) rename everything XXXType to XXXKind (e.g. `Term::StringKind`, `Term::StaticKind`) and (b) then rename everything RuntimeXXX to XXXType (e.g. `Term::FunctionType`, `Term::ExtensionType`). Thus, `Term::RuntimeType(TypeBound)` -> `Term::TypeKind(TypeBound)`). This will be *even more* breaking as Rust doesn't have good support for renaming enum variants, note....
   * Note also `Term::RuntimeExtension` (which would be `Term::ExtensionType` under renaming proposed above) is planned to become `Term::Extension`in the next PR, i.e. able to produce things other than runtime types
* `SumType` caches its bound; this is the only case where TypeEnum's cache was useful (as Extensions already cache inside `CustomType` and Functions are always Copyable) so then we can drop the Type/TypeEnum wrapping
* Remove the RV hack. This is the big win for now. In some sense it is a step backwards (we lose some Rust static type checking). But RowVars were always a hack, in that they looked like one entry in a list, but were any number of entries. Instead now variables that range over lists are not members of lists, but there is (already) a `Term::ListConcat` all of whose members are lists; this is a much nicer/more-principled solution.
   * The key here is that `TypeRow` remains as a `Vec<Term>` (see below) but **`TypeRowRV` is now a `Term`**: either a `Term::List` (of single types, including variables), a `Term::Variable` (a variable ranging over *lists of types*), or a `Term::ListConcat` (of sub*lists* each being one of these three, and canonicalized to a mix of the previous two only).
   * See 68fba45f781a37f7515281d53eeeedcc3bf1d84c for a way of bringing back some of the ease-of-use/compactness of the RV system using the checking we have implemented in Rust; however, the "improvement" in the user code (/the pain of having to explicitly `Term::new_list_concat`) was not that great so I have reverted...thoughts?
* I removed Aliases; I mean I could add them but they are vestigial and not really working anyway (#2558). Or perhaps we should add them in and support properly now we have linking?
* Significant work has gone into preserving backwards-compatibility for serde/json serialization of types, as IIUC there is no plan for deprecating this (it's required for extension declarations as well as Hugrs). In particular I made much use of `serde_as` with some new markers (`SerType`, `SerTypeRowRV`). This has had some impact, see next two points.
* Abandon the parametrized FuncTypeBase / PolyFuncTypeBase (instead adding macros for code common to Signature/FuncValueType and PolyFuncType / PolyFuncTypeRV), necessary because FuncValueType needs a `#[serde_as(as=SerTypeRowRV)]` declaration. However, see next point.
* **NOTE** the current state of the PR **abuses `TypeRow` significantly**. In general it can store arbitrary Terms (of course), but it *serializes into json* in the legacy format that only supports Types. Meanwhile `TypeRowRV` is just `Term` but places using it tend to use a `#[serde_as(as=SerTypeRowRV)]` which *serializes into json* in the legacy format that supports Types *and RowVariables*.
  * I've given `Signature`, `FuncValueType` and `SumType` checking constructors: `::try_new` checks the contents are actually types/lists of types; `::new` does the same check *and unwraps*; `::new_unchecked` avoids the check (but will break at JSON-serialization time, and shouldn't really happen anyways).
  * A perhaps-better solution is to make `TypeRowRV` be a wrapper-struct around `Term` and instead put these checking constructors on both `TypeRow` (i.e. requiring that the elements are single types) and `TypeRowRV`. However this has problems because of our use of `impl Into<TypeRow>`(/RV) everywhere: either
     * `impl From` would call `new_unchecked` (meaning the checks are bypassed in many cases)
     * `impl From` would call `new` (i.e. `.into()` panics)
     * We change all the `x: impl Into<TypeRow>` into `x: impl TryInto<TypeRow>` and add Results with that error type, but this would be very disruptive.
   * Also `impl TryFrom<Term> for TypeRow` would become `Term::as_list(self) -> Option<Vec<Term>>` i.e. a utility for a single-variant `match` - we have plenty of these so that's fine.
   * We could even make `Type` be a wrapper-struct around `Term` that checks it's a runtime type (a `RuntimeExtension` / `RuntimeFunction` / `RuntimeSum` / `Variable`-of-appropriate-type), again via checking constructors, but I think this would be tooooo painful
**Summary** This PR is substantially painful, although I hope that removing the RV hack can be a significant win in itself right now, if we can do it right. The bigger gains are expected to come in later PRs (#2296 and #2756, eventually allowing "inspectable" custom constants with efficient serialization).

TODOs/unresolved questions:
* Is it time to retire old aliases `TypeArg`, `TypeParam` and the new `Type`? (Perhaps also `TypeRowRV` but the above suggests making it a non-alias struct with guarantees; I guess we could do the same for `TypeRV` too)
* If we don't strengthen `TypeRow` to guarantee its elements are runtime-types, then
   - [ ] rename to `TermRow`....or just remove it
   - [ ] add checking contsructors to all OpTypes, etc., too
- [ ] Update hugr-py
- [x] Update hugr-llvm (perhaps only test snapshots)

BREAKING CHANGE: Term gains `RuntimeFunction`, `RuntimeSum` and `RuntimeExtension` variants; Type is now an alias for Term. Use `check_term_type(tm, &TypeBound::Linear.into())` to check that `tm: &Term` represents a runtime type. No more aliases.

### DIFF
--- a/hugr-core/src/types/type_param.rs
+++ b/hugr-core/src/types/type_param.rs
@@ -262,16 +262,6 @@ impl From<UpperBound> for Term {
     }
 }
 
-/*ALAN delete(?)
-impl<RV: MaybeRV> From<TypeBase<RV>> for Term {
-    fn from(value: TypeBase<RV>) -> Self {
-        match value.try_into_type() {
-            Ok(ty) => Term::Runtime(ty),
-            Err(RowVariable(idx, bound)) => Term::new_var_use(idx, TypeParam::new_list_type(bound)),
-        }
-    }
-}*/
-
 impl From<u64> for Term {
     fn from(n: u64) -> Self {
         Self::BoundedNat(n)


hugr-core/src/ops/constant.rs
```diff
@@ -208,7 +208,7 @@ pub enum Value {
 /// use serde_json::json;
 ///
 /// let expected_json = json!({
-///     "typ": usize_t(),
+///     "typ": {"t": "I"}, // No public way to serialize a Term as a (SerSimple)Type...
```
acl-cqc: This would be `#[serde_as(as="crate::types::serialize::SerType")]` on a rust decl, but we don't have a way to do that here. We could do `"type": SerSimpleType::from(usize_t())` but would have to make `SerSimpleType` public

PR #2791, opened 2026-01-05
docs: add metadata spec document

### LINKED ISSUE #2431: Add metadata specification document with current and planned keys
A number of metadata keys are currently being used ad-hoc. Some are mentioned in the main spec document. We should instead have a structured document that lists the namespaces, keys, value types and meaning.

We also have metadata that can be attached to nodes but also metadata that can be attached to the contents of extension definitions.

Further, the way the model uses metadata is slightly different to how the core uses it. 

### General
- `core.generator` for recording structured HUGR generator names and versions
- `core.used_extensions` for names and versions of used extensions in HUGR

### Model
- `core.meta.description` for descriptions
- `core.order_hint` prefix for recording order edges (is this actually metadata, it does not say `meta`, is this part of stable model spec instead?)
- `compat.meta_json` to string encode existing metadata (like those in the General section above) 

## Proposed
- `core.hint` prefix for adding compiler hints #2328 #2372 
- `core.side_effect` for marking impure OpDefs
- `core.passes.<pass name>` prefix for targeting specific passes
- Debug information, for example guppy source info

ss2165:
Closes #2431

- Remove outdated discussions of metadata implementation options
- Comment out TODOs
- Add reserved keys document
- List model keys first
- Move core keys to new document

### DIFF
--- a/specification/metadata.md
+++ b/specification/metadata.md
@@ -1,6 +1,6 @@
 # HUGR Core Metadata Specification
 
-HUGR nodes can have associated metadata, this is described in the
+HUGR nodes can have associated metadata. This is described in the
 [Extensible Metadata](./hugr.md#extensible-metadata) section of the specification.
 
 This document describes metadata keys reserved by reference HUGR tooling, which are all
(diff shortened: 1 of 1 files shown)


specification/metadata.md
```diff
@@ -0,0 +1,38 @@
...
+The HUGR "model" serialization format has its own notion of node metadata, which
+it uses to encode some data that is considered core to the HUGR specification.
+It uses a special key to store HUGR metadata, so in some sense it is nested.
+
+Here we list the model-level metadata keys reserved by HUGR tooling; the types refer to
+model format types.
+
+| Key                        | Model Type               | Description                                                     |
+|----------------------------|--------------------------|-----------------------------------------------------------------|
+| core.meta.description      | string                   | Human readable descriptions.                                    |
```
cqc-alec:
```suggestion
| core.meta.description      | string                   | Human-readable descriptions.                                    |
```

PR #2792, opened 2026-01-05
fix: more nondeterminism in linking

acl-cqc:
More iteration through HashMaps keyed by source node i.e. adding source functions to the target in a different order...


hugr-core/src/hugr/linking.rs
```diff
@@ -1507,4 +1509,34 @@ mod test {
             );
         }
     }
+
+    #[test]
+    fn determinism() {
+        // Really this just checks that things from the source hugr are added in consistent
+        // order (taking next available indices in the target). Actually linking them with
+        // nodes already (at consistent positions) in the target only reduces variability.
+        let (insert, _, _) = dfg_calling_defn_decl();
```
acl-cqc: This was a handy hugr lying around, with the entrypoint + two defns/decls being inserted as separate "trees" in a forest, there are 3!=6 possible orderings, so the test succeeded around 1 time in 6. We could decrease the odds of a false positive by inserting a bigger Hugr but it would make a longer test.

PR #2817, opened 2026-01-19
feat: Add used_extensions method to Hugr

### LINKED ISSUE #2614: Compute used extensions when compiling packages
Currently we hard-code the tket2 extensions in the packages when compiling a module.
We should detect which extensions are actually being used, and only include those (and add custom ones too).

https://github.com/CQCL/guppylang/blob/473745fcf952fd7bc2b442cb257e6539b30d0e98/guppylang/module.py#L337-L343 

A related feature would be to only include the required op definitions from each extension, to further reduce the package size.

aborgna-q:
Defines `used_extensions(self) -> ExtensionRegistry` methods throughout the definitions.

In contrast to the rust implementation—since python Hugrs con be modified by directly accessing a node's operation—we cannot pre-compute this set and ensure it remains valid.
Instead the method traverses all node operations, and collects extensions from their op, types, and type args.

Closes #2614

### DIFF
--- a/hugr-py/src/hugr/tys.py
+++ b/hugr-py/src/hugr/tys.py
@@ -48,7 +48,7 @@ class TypeParam(Protocol):
         """Get the set of extensions required to define this type parameter.
 
         Raises:
-            UnresolvedExtensionError: if a type parameter contains a
+            UnresolvedExtensionError: if a type parameter contains an
                 :class:`Opaque` type that has not been resolved.
 
         Example:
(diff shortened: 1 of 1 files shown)


hugr-py/src/hugr/tys.py
```diff
@@ -57,6 +79,21 @@ def to_model(self) -> model.Term | model.Splice:
         """Convert the type argument to a model Term."""
         raise NotImplementedError(self)
 
+    def used_extensions(self) -> ExtensionRegistry:
+        """Get the set of extensions required to define this type argument.
+
+        Raises:
+            UnresolvedExtensionError: if a type argument contains a
```
cqc-alec:
```suggestion
            UnresolvedExtensionError: if a type argument contains an
```

cqc-alec (approved):

PR #2836, opened 2026-01-23
feat: update passes to use PassScope where non-breaking

acl-cqc:
Follow-up to #2772 with non-breaking updates to ConstFoldPass, DeadCodeElimPass, NormalizeCFGsPass, RedundantOrderEdgesPass, RemoveDeadFuncsPass, ReplaceTypes, UntuplePass. (Breaking updates to follow in #2871.)

Generally this means keeping pre-existing pass configurating methods but deprecating, and storing such config in an `Either<PassScope, old-config>`, so we can drop the old config in time.

Also deprecated toplevel functions `constant_fold_pass`, `remove_dead_funcs` and `monomorphize`.

### DIFF
--- a/hugr-passes/src/untuple.rs
+++ b/hugr-passes/src/untuple.rs
@@ -107,7 +107,7 @@ impl UntuplePass {
 
     /// Sets the parent node to optimize (overwrites any previous setting)
     ///
-    /// If the pass was previously configured by [Self::with_scope] then
+    /// If the pass was previously configured by [Self::with_scope_internal] then
     /// implicitly `[Self::set_recursive]`'s with [PassScope::recursive]
     #[deprecated(note = "Use with_scope instead", since = "0.25.7")]
     #[expect(deprecated)] // Remove along with UntupleRecursive
@@ -128,7 +128,7 @@ impl UntuplePass {
 
     /// Sets whether the pass should traverse the HUGR recursively.
     ///
-    /// If the pass was last configured via [Self::with_scope], overrides that,
+    /// If the pass was last configured via [Self::with_scope_internal], overrides that,
     /// with `set_parent` of default `None`.
     #[must_use]
     #[deprecated(note = "Use with_scope", since = "0.25.7")]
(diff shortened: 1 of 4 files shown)


hugr-passes/src/dead_code.rs
```diff
@@ -97,11 +103,11 @@ impl<H: HugrView> DeadCodeElimPass<H> {
...
-    /// Mark some nodes as entry points to the Hugr, i.e. so we cannot eliminate any code
-    /// used to evaluate these nodes.
-    /// [`HugrView::entrypoint`] is assumed to be an entry point;
-    /// for Module roots the client will want to mark some of the `FuncDefn` children
-    /// as entry points too.
+    /// Mark some nodes as starting points for analysis, i.e. so we cannot eliminate any code
+    /// used to evaluate these nodes. (E.g. nodes at which we may start executing the Hugr.)
+    ///
+    /// Other starting points are added according to the [PassScope].
+    // TODO should we deprecate this? i.e. require use of PreserveCallback / Hugr edges?
```
acl-cqc: However, deprecating this would make it impossible to support `ConstantFoldPass::with_inputs` which I think is quite a nice feature for debugging. (Perhaps following deprecation both could be made crate-private instead of removing?)

hugr-passes/src/untuple.rs
```diff
@@ -66,74 +91,148 @@ pub struct UntupleResult {
...
+        }
+    }
+
+    /// Create a new untuple pass with the given configuration
+    #[must_use]
+    pub fn new_scoped(scope: PassScope) -> Self {
+        Self {
+            scope: Either::Left(scope),
         }
     }
```
aborgna-q: Do we need this? It's the same as
`UntuplePass::default().with_scope(...)`

hugr-passes/src/untuple.rs
```diff
@@ -66,74 +91,148 @@ pub struct UntupleResult {
...
 
     /// Find tuple pack operations followed by tuple unpack operations
+    /// beneath a specified parent and according to this instance's recursiveness
+    /// ([Self::recursive] or [Self::with_scope] + [PassScope::recursive])
     /// and generate rewrites to remove them.
     ///
     /// The returned rewrites are guaranteed to be independent of each other.
     ///
     /// Returns an iterator over the rewrites.
+    #[deprecated(note = "Use all_rewrites")]
```
aborgna-q:
```suggestion
    #[deprecated(note = "Use all_rewrites", since="0.25.7")]
```

aborgna-q (approved):

Issue #2828, opened 2026-01-21
(hugr-model) Model round-trip panics with `AliasDecl`


cqc-alec:
With the HUGR constructed in [this test](https://github.com/Quantinuum/hugr/blob/15b27cd11b2603f0c4d48826039179be344dc38d/hugr-core/src/builder/module.rs#L289), serialization to model followed by deserialization panics with:
```
thread 'builder::module::test::simple_alias' (1243268) panicked at hugr-core/src/envelope.rs:301:63:
called `Result::unwrap()` on an `Err` value: Payload { source: PayloadError(ModelBinary(Import(ImportError { inner: Context(Context(Context(Context(Context(Invalid("`qubit_state` is not a valid symbol name"), "term 3 as `Type`"), "function inputs"), "function type"), "symbol `_2` defined by node 1"), "`define-func` node with id 1") }))), partial_description: PackageDesc { header: EnvelopeHeader { format: ModelWithExtensions, zstd: true }, modules: [Some(ModuleDesc { num_nodes: Some(5), entrypoint: None, used_extensions_resolved: None, generator: None, used_extensions_generator: None, public_symbols: None })], packaged_extensions: [] } }
```

PR #2839, opened 2026-01-27
fix: Canonicalize more

### LINKED ISSUE #2789: `Hugr::canonicalize_nodes()` doesn't always canonicalize node indices
The function `check_hugr_equality()` (which is used as part of hugr validation to compare a hugr with the result of a serialization/deserialization round-trip) first calls `canonicalize_nodes()` on each hugr, and then compares them node by node. However, `canonicalize_nodes()` does not correct all irrelevant differences in node numbering.

Currently the `validate()` method [works around this](https://github.com/Quantinuum/hugr/blob/191c4736b5fbe869f8d797703bbf1c29f33b3a92/hugr-core/src/hugr/validate.rs#L77-L79) by using the json serialization instead of the model serialization. But when the model serialization is used instead, many of the tests fail because we end up comparing two hugrs containing `Const` and `LoadConstant` nodes that are numbered in a different order, and `canonicalize_nodes()` does not correct this.

cqc-alec:
This allows us to use the binary format instead of the deprecated JSON format in the round-trip equality testing used for HUGR validation.

I've marked one test (`simple_alias`) as `#[ignore]`. Aliases are not properly supported by the model, are not currently used as far as I know, and will probably be removed or at least reworked soon. (See #2828 .)

Some of the benchmark results are a little alarming. It looks like the regressions are mainly in a sorting method being called during JSON serialization.

Closes #2789 .

### DIFF
--- a/hugr-core/src/hugr.rs
+++ b/hugr-core/src/hugr.rs
@@ -313,6 +313,17 @@ impl Hugr {
     }
 }
 
+/// Strategy to use for ordering nodes.
+#[derive(Clone, Copy, Debug)]
+enum CanonicalOrderStrat {
+    /// Do a breadth-first search through the hierarchy, ordering sibling nodes
+    /// according to a partial order that is somewhat arbitrary but fine-grained
+    /// enough to be practically useful for testing.
+    TryHard,
+    /// Quickly order by a breadth-first search through the hierarchy.
+    Quick,
+}
+
 /// Internal API for HUGRs, not intended for use by users.
 impl Hugr {
     /// Add a node to the graph.
@@ -364,42 +375,47 @@ impl Hugr {
     /// descendant nodes of a root, following the graph hierarchy.
     ///
     /// This starts with the root, and then proceeds in BFS order through the
-    /// contained regions. If `try_hard` is `true`, sibling nodes are ordered
-    /// according to a partial order that is somewhat arbitrary but fine-grained
-    /// enough to be practically useful for testing.
+    /// contained regions.
     ///
     /// Used by [`HugrMut::canonicalize_nodes`] and the serialization code.
-    fn canonical_order(&self, root: Node, try_hard: bool) -> impl Iterator<Item = Node> + '_ {
+    fn canonical_order(
+        &self,
+        root: Node,
+        strategy: CanonicalOrderStrat,
+    ) -> impl Iterator<Item = Node> + '_ {
         // Generate a BFS-ordered list of nodes based on the hierarchy
         let mut queue = VecDeque::from([root]);
         iter::from_fn(move || {
             let node = queue.pop_front()?;
             let mut children = self.children(node).collect_vec();
-            if try_hard {
-                let sort_fn = |a: &Node, b: &Node| {
-                    let n_a_inp = self.input_neighbours(*a).count();
-                    let n_b_inp = self.input_neighbours(*b).count();
-                    let n_a_out = self.output_neighbours(*a).count();
-                    let n_b_out = self.output_neighbours(*b).count();
-                    if n_a_inp < n_b_inp {
-                        Ordering::Less
-                    } else if n_a_inp > n_b_inp {
-                        Ordering::Greater
-                    } else if n_a_out < n_b_out {
-                        Ordering::Less
-                    } else if n_a_out > n_b_out {
-                        Ordering::Greater
+            match strategy {
+                CanonicalOrderStrat::TryHard => {
+                    let sort_fn = |a: &Node, b: &Node| {
+                        let n_a_inp = self.input_neighbours(*a).count();
+                        let n_b_inp = self.input_neighbours(*b).count();
+                        let n_a_out = self.output_neighbours(*a).count();
+                        let n_b_out = self.output_neighbours(*b).count();
+                        if n_a_inp < n_b_inp {
+                            Ordering::Less
+                        } else if n_a_inp > n_b_inp {
+                            Ordering::Greater
+                        } else if n_a_out < n_b_out {
+                            Ordering::Less
+                        } else if n_a_out > n_b_out {
+                            Ordering::Greater
+                        } else {
+                            self.get_optype(*a)
+                                .partial_cmp(self.get_optype(*b))
+                                .unwrap_or_else(|| a.cmp(b))
+                        }
+                    };
+                    if self.contains_dsg_or_csg(node) {
+                        children[2..].sort_by(sort_fn);
                     } else {
-                        self.get_optype(*a)
-                            .partial_cmp(self.get_optype(*b))
-                            .unwrap_or_else(|| a.cmp(b))
+                        children.sort_by(sort_fn);
                     }
-                };
-                if self.contains_dsg_or_csg(node) {
-                    children[2..].sort_by(sort_fn);
-                } else {
-                    children.sort_by(sort_fn);
                 }
+                CanonicalOrderStrat::Quick => {}
             }
             queue.extend(children);
             Some(node)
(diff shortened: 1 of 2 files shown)


hugr-core/src/hugr.rs
```diff
@@ -322,25 +324,120 @@ impl Hugr {
...
+                        .partial_cmp(self.get_optype(*b))
+                        .unwrap_or_else(|| a.cmp(b))
+                }
+            };
+            if self.contains_dsg_or_csg(node) {
+                children[2..].sort_by(sort_fn);
+            } else {
+                children.sort_by(sort_fn);
+            }
+            for child in children {
```
acl-cqc: `queue.extend(children)` probably works here

hugr-core/src/hugr.rs
```diff
@@ -322,25 +324,120 @@ impl Hugr {
...
+    fn canonicalize_consts(&mut self) {
+        let const_nodes = self
+            .nodes()
+            .filter(|n| self.get_optype(*n).is_const())
+            .collect_vec();
+        for node in const_nodes {
+            let const_parent = self.get_parent(node).unwrap();
+            let outport = self.node_outputs(node).exactly_one().ok().unwrap();
+            let succ_nodeports = self.all_linked_inputs(node).collect_vec();
+            for (succ_node, succ_inport) in succ_nodeports {
```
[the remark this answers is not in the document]
cqc-alec: I did it this way because this is what the model loader does already (inserts a `Const` node in the same region as the `LoadConst`). LCA could work too, but would be a bit fiddlier.

hugr-core/src/ops.rs
```diff
@@ -71,6 +72,51 @@ pub enum OpType {
...
+impl PartialOrd for OpType {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        let a = optype_id(self);
+        let b = optype_id(other);
+        if a < b {
+            Some(Ordering::Less)
+        } else if a > b {
+            Some(Ordering::Greater)
+        } else {
+            match format!("{:?}", self).cmp(&format!("{:?}", other)) {
```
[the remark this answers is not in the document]
cqc-alec: The sorting should happen once per `finish_hugr()` call, because that does validation which involves the round-trip comparison, which would be expensive if the hugr contains a large number of indistinguishable ops in the same region, so I thought if I did [this](https://github.com/Quantinuum/hugr/pull/2839/changes/ec58b23fa885a6700a18f6772a0fdf4792c8f436) (skipping validation) it would solve the problem, and it did seem to make a difference when I ran `cargo bench` locally, but the results on the PR didn't change much.

hugr-core/src/hugr.rs
```diff
@@ -322,25 +324,118 @@ impl Hugr {
...
     fn canonical_order(&self, root: Node) -> impl Iterator<Item = Node> + '_ {
         // Generate a BFS-ordered list of nodes based on the hierarchy
         let mut queue = VecDeque::from([root]);
         iter::from_fn(move || {
             let node = queue.pop_front()?;
-            for child in self.children(node) {
-                queue.push_back(child);
+            let mut children = self.children(node).collect_vec();
+            let sort_fn = |a: &Node, b: &Node| {
+                let n_a_inp = self.input_neighbours(*a).count();
```
acl-cqc: I am not sure this ordering by number of predecessors/successors adds very much, but no objection.

If we wanted to be more canonical...an algorithm might be to do a topological sort, with our own choice among the legal candidates at each step - our choice being the minimum by a lexicographic ordering of each input's (node-number of source node, output index), followed by optype. (The topsort ensures the source node has been given a number before any consumer of its outputs.)

I think that's more work though so probably not worth it.

hugr-core/src/hugr.rs
```diff
@@ -387,10 +484,12 @@ impl Hugr {
         self.module_root = portgraph::NodeIndex::new(0);
         self.entrypoint = new_entrypoint.unwrap();
 
-        // Finish by compacting the copy nodes.
+        // Compact the copy nodes.
         // The operation nodes will be left in place.
         // This step is not strictly necessary.
         self.graph.compact_nodes(|_, _| {});
+
+        self.order_siblings_by_node_index();
```
acl-cqc: It does seem odd to call this here having already used `canonical_order` earlier in `fn canonicalize_nodes`?

hugr-core/src/ops.rs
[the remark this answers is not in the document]
acl-cqc: if it's doing this sort in the roundtrip that happens during validation (only) when HUGR_TEST_SCHEMA is set, that shouldn't affect benchmarks, should it?

acl-cqc (commented):
Hmmm, still not quite sure we've got to the bottom of this yet, but a few thoughts

hugr-core/src/hugr.rs
cqc-alec: I'm now struggling to understand why this was necessary! May have to come back to it... It certainly _is_ necessary in the sense that tests fail without it...

hugr-core/src/hugr/validate.rs
```diff
@@ -74,9 +74,7 @@ impl<'a, H: HugrView> ValidationContext<'a, H> {
 
             let (mut hugr, node_map) = self.hugr.extract_hugr(self.hugr.module_root());
             hugr.set_entrypoint(node_map.extracted_node(self.hugr.entrypoint()));
-            // TODO: Currently fails when using `hugr-model`
-            //crate::envelope::test::check_hugr_roundtrip(&hugr, EnvelopeConfig::binary());
-            crate::envelope::test::check_hugr_roundtrip(&hugr, EnvelopeConfig::text());
+            crate::envelope::test::check_hugr_roundtrip(&hugr, EnvelopeConfig::binary());
```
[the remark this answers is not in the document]
aborgna-q: Maybe add a TODO to add ModelText once that's stable?

hugr-core/src/hugr.rs
```diff
@@ -322,25 +324,120 @@ impl Hugr {
...
+            let const_parent = self.get_parent(node).unwrap();
+            let outport = self.node_outputs(node).exactly_one().ok().unwrap();
+            let succ_nodeports = self.all_linked_inputs(node).collect_vec();
+            for (succ_node, succ_inport) in succ_nodeports {
+                let succ_parent = self.get_parent(succ_node).unwrap();
+                if succ_parent == const_parent {
+                    continue;
+                }
+                self.disconnect_edge(node, outport, succ_node, succ_inport);
+                let new_node = self.add_node_before(succ_node, self.get_optype(node).clone());
```
aborgna-q: nit: If multiple nodes in the same region use the constant we'll still generate multiple const nodes here.

Maybe we could collect a `BTreeMap<succ_parent: Node, succ_nodes: Vec<(Node, IncomingPort)>>` first?

Not sure if it's worth the extra complexity.

hugr-core/src/hugr.rs
```diff
@@ -322,25 +324,120 @@ impl Hugr {
...
+    fn contains_dsg_or_csg(&self, node: Node) -> bool {
+        matches!(
+            self.get_optype(node),
+            OpType::FuncDefn(_)
+                | OpType::DFG(_)
+                | OpType::DataflowBlock(_)
+                | OpType::TailLoop(_)
+                | OpType::CFG(_)
+                | OpType::Case(_)
+        )
```
aborgna-q: This could be done using tags, if we want to avoid enumerating all cases.
```rs
let tag = self.get_optype(node).tag();
tag <= OpTag::DataflowParent || tag <= OpTag::Cfg
```

cqc-alec: Nice, thanks.

PR #2843, opened 2026-01-29
feat: Auto-embed extensions in stored hugrs

### LINKED ISSUE #2841: `Hugr.to_bytes` should include required extensions
Package definitions include a manually-defined list of extensions required to load the hugr,
that gets included in the envelope when calling `Package.to_bytes`/`to_str`.

When encoding a standalone Hugr, we wrap it in a Package with an empty list of extensions before encoding it.

Now that we can compute the list of extensions used by a hugr dinamically (#2817), we could automatically add those to the created package to ensure that the envelope can be loaded by any target.

I'd propose adding a new parameter to `Hugr.to_bytes` and `to_str`,
```py
class Hugr:
    def to_bytes(self, (...), *, omit_extensions: ExtensionRegistry | None = None) -> bytes: ...
    def to_str(self, (...), *, omit_extensions: ExtensionRegistry | None = None) -> str: ...
```
These run `used_extensions`, and add any one not in `omit_extensions` to the created package.

If one wants better control over the included extensions, it's always possible to create a package manually with the desired bundled extensions.

---

Note that this behaviour differs from the rust definition, where the default `store` explicitly omits storing extra extensions.
https://github.com/Quantinuum/hugr/blob/a21e818c12b3f9ca991c1e46d146e704f7b41c41/hugr-core/src/hugr.rs#L184-L204

The rust interface in general provides finer control over the compiler behaviour, but for the python side we should err on the side of intuitiveness.
If I `to_bytes` a hugr I would expect it can be loaded afterwards wherever I need it. Trimming the extensions included in the envelope should be a later size-optimization.

aborgna-q:
Closes #2841. See that issue for more details.

- Adds `include_extensions` parameters to `Hugr.to_bytes` and `to_str`.
- Adds a `to_package(self, include_extensions: ExtensionRegistry | None = None) -> Package` method to Hugr.

When the registry is `None` for any of those, we compute them using `Hugr.used_extensions`.

### DIFF
--- a/hugr-py/src/hugr/envelope.py
+++ b/hugr-py/src/hugr/envelope.py
@@ -38,6 +38,7 @@ from enum import Enum
 from typing import TYPE_CHECKING, ClassVar
 
 import pyzstd
+from typing_extensions import deprecated
 
 import hugr._hugr.model as rust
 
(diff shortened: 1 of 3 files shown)


hugr-py/src/hugr/envelope.py
```diff
@@ -55,15 +55,10 @@
 _ZSTD_FLAG = 0b0000_0001
 
 
-def make_envelope(package: Package | Hugr, config: EnvelopeConfig) -> bytes:
-    """Encode a HUGR or Package into an envelope, using the given configuration."""
-    from hugr.package import Package
-
+def make_envelope(package: Package, config: EnvelopeConfig) -> bytes:
```
mark-koch: No longer accepting `Hugr` is a breaking change, right?

mark-koch (approved):

PR #2871, opened 2026-02-11
feat!: Update remainder of passes to use PassScope, drop default with_scope

acl-cqc:
Following #2836, the remaining passes, where breakage was inevitable - mostly passes that were an empty struct (without any previous config).

Non-local-edges stuff needed some API changes where I did some deprecation, but we could just break.

RedundantOrderEdgesPass I've removed the old config methods - we should perhaps deprecate first, but some breakage seems inevitable as the empty struct derived Copy.

LocalizeEdges I've deprecated+moved top-level `ensure_no_nonlocal_edges` into a struct method `check_no_nonlocal_edges`, as IIUC it's used, but I've removed top-level `nonlocal_edges` altogether, it doesn't seem to be used.

Also remove the default impl for `ComposablePass::with_scope_internal`, to check we got 'em all.

BREAKING CHANGE: InlineDFGsPass, LocalizeEdgesPass, MonomorphizePass, RedundantOrderEdgesPass must be constructed via ::default(); remove RedundantOrderEdgesPass::recursive and don't derive Copy; ComposablePass::with_scope_internal requires impl; remove toplevel `fn nonlocal_edges`

### DIFF
--- a/hugr-passes/src/redundant_order_edges.rs
+++ b/hugr-passes/src/redundant_order_edges.rs
@@ -68,23 +68,23 @@ impl RedundantOrderEdgesPass {
         // Traverse the region in topological order.
         let (region, node_map) = hugr.region_portgraph(parent);
         let postorder = petgraph::visit::Topo::new(&region);
-        for pg_node in postorder.iter(&region) {
-            let node = node_map.from_portgraph(pg_node);
-            let op = hugr.get_optype(node);
+        for pg_child in postorder.iter(&region) {
+            let child = node_map.from_portgraph(pg_child);
+            let op = hugr.get_optype(child);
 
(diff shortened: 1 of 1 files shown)


hugr-passes/src/non_local.rs
```diff
@@ -57,69 +105,63 @@ pub enum FindNonLocalEdgesError<N> {
     Edges(Vec<(N, IncomingPort)>),
 }
 
-/// Verifies that there are no non local value edges in the Hugr.
+/// Verifies that there are no non local value edges in the Hugr beneath the entrypoint.
+#[deprecated(note = "Use LocalizeEdges::check_no_nonlocal_edges")]
```
acl-cqc: (Seems misnamed)

PR #2895, opened 2026-02-25
refactor(hugrv2)!: combine TypeEnum with Term, no RV parametrization

acl-cqc:
* Merge TypeEnum into Term: remove `Term::Runtime(Type)`, add `Term::RuntimeFunction(FuncValueType)`, `Term::RuntimeSum(SumType)` and `Term::RuntimeExtension(CustomType)` and remove `TypeEnum`. That is, runtime types are now just variants of `Term`.
* `Type` now just wraps a `Term`, but guarantees by construction that the Term inside it represents a single runtime type. (There is no way to bypass this; there is `TryFrom<Term> for Type`; there are utility constructors on `Type` that take `Type`s thus preserving the invariant - the same `new_extension`, `new_tuple`, `new_sum`, `new_function` as before).
* Remove `TypeRV`. It never had a consistent meaning: it was sometimes a type, and sometimes a list. Also remove `trait MaybeRV`, `enum NoRV` and `struct RowVariable`.
* `TypeRowRV` (much like Type) wraps a `Term`, but guarantees it represents a list of types (of perhaps unknown length). That is, it could be a `Term::List` (whose elements are single types), or a `Term::Variable` (a "row variable" i.e. ranging over lists of types, of unknown length), or a `Term::ListConcat` (whose elements are one of these three). Again there are utility constructors preserving the invariant, tho there is also a `TypeRowRV::new_unchecked`.
   * Note prior to 0ac2b5316a0f581082bfb36812da8b5e6d799c70 (see diff LHS) TypeRowRV was merely an alias to Term (with GeneralSum and FuncValueType providing panicking-`new` + `try_new` + `new_unchecked` constructors, removed in that commit). However I think with a few nifty TypeRowRV conversions and methods (`just_row_var` and `concat`) the wrapper-struct is easier to use as well as giving more static checking.
* `TypeRow` remains as a list of `Type`s, i.e. whose length/number-of-types is known
* A new `trait TypeRowLike` allows parametrizing `FuncTypeBase` so it can still do both `Signature` and `FuncValueType`. I have kept this trait crate-hidden, so we keep the `substitute` and `validate` methods hidden (as before), which does mean external code cannot parametrize in this way.

Thus, although the new system offers less Rust-compile-time checking, there is still a reasonable amount, and it's more principled now ;)

Along the way,
* Replace `Type::as_type_enum` with `impl Deref<Target=Term> for Type`, similarly for TypeRowRV
* AliasDecl/Defn and Type::Alias dropped for now, pending #2558
* Since `TermTypeError` is now used more (it's a subclass of SignatureError but captures all the errors that might happen from e.g. turning a Term into a Type), I've added `impl From<TermTypeError> for OpLoadError` (also `for ImportErrorInner`) - because Rust won't let me `impl <T: From<SignatureError>> From<TermTypeError> for T` :-(.
* Generalize `Term::new_list` to take items that are `impl Into<Term>` rather than just `Term`

Proposed follow-ups.... 
* Renaming, in #3019
* Move caching of TypeBound from `Type` into `SumType::General`, in #3022
- [ ] Either remove some redundant SignatureError variants (that also appear in TermTypeError) or perhaps combine the two errors altogether. (Less urgent?)
- [ ] Could add a utility method `TypeRowRV::new_spliced(IntoIterator<Item=Term>)` that checks each Term is *either* a type or a list of types (and panics if not....ok could also have `try_new_spliced`), and then assembles the appropriate lists/concats. commit 68fba45f781a37f7515281d53eeeedcc3bf1d84c shows this done for FuncValueType whereas here/now it would be for TypeRowRV, and I'm not sure that it's much better than the `TypeRowRV::just_row_var`, `concat` and `from([Type])` that this PR has now (in particular, just_row_var/concat have better Rust-compile-time checking than this "splicing" approach). So, probably less urgent / only if TypeRowRV is too awkward in practice.

BREAKING CHANGE: `TypeEnum` and `TypeRV` are no more - use `Term`. `RowVariable`, `MaybeRV` etc. are gone - use appropriate term types and variables. `Type` and `TypeRowRV` merely wrap `Term` with invariants.


cqc-alec (commented):
Thanks @acl-cqc . Have had a general look over this and it looks good to me!

Regarding your questions:

- Does `Kind` have a technical meaning? What would you rename `RuntimeType` to? (I guess `TypeType`?) In any case I think the docstring for `Term` should be expanded, as its current form ("A term in the language of static parameters in HUGR") is rather unclear, at least to me.
- Suggest combining these error types, either as a pre- or post-PR (I think this PR is big enough already without adding more changes).
- Probably worth trying (maybe as a follow-up PR).
- Seems like a good idea (perhaps as a follow-up PR).

hugr-core/src/types/type_param.rs
```diff
@@ -95,9 +93,22 @@ pub enum Term {
     /// The type of static tuples.
     #[display("TupleType[{_0}]")]
     TupleType(Box<Term>),
-    /// A runtime type as a term. Instance of [`Term::RuntimeType`].
+    /// The type of runtime values defined by an extension type.
+    /// Instance of [Self::RuntimeType] for some bound.
+    //
+    // TODO optimise with `Box<CustomType>`?
+    // or some static version of this?
+    RuntimeExtension(CustomType),
```
doug-q: needs a `#[display]` incantation

acl-cqc:
Thanks @cqc-alec !

> Does Kind have a technical meaning?

Normally, kinds describe sets of possible things at the type level just like types describe sets of possible things at the runtime level. So in something like Rust or Haskell, kinds might be "type" (an instance of which would be, say, `int` or `string`) or `type -> type` (an instance of which would be, say, `list` - because list is not a type, but `list<string>` is a type). Of course in practice (e.g. in python) things are slightly more complex because you have "default" element types for list and specifically including the "unknown" type.

> What would you rename RuntimeType to? (I guess TypeType?)

| Current name | Proposed New Name |
| ----- | ----- |
| `StringType` | `StringKind` |
| `FloatType` | `FloatKind` |
| `ListType(Term)` | `ListKind(Term)` |
| `List(Vec<Term>)` | `List(Vec<Term>)` (no change) |
| `RuntimeType(TypeBound)` | `TypeKind(TypeBound)` |
| `Runtime(Type)` (pre-this PR) | `Type(Type)` |
| `RuntimeExtension(CustomType)` (post-this PR) | `ExtensionType(CustomType)` |
| `RuntimeFunction(FuncValueType)` (post-this PR) | `FunctionType(FuncValueType)` |
| `RuntimeSum(SumType)` (post-this PR) | `SumType(SumType)` |

I think the big question here is whether to do this first or after, how breaking it is (do we deprecate old names and release with both sets of names existing in parallel?), etc...or whether it is soooo disruptive not to do it ever, but we did rename `TypeParam::Foo` to `TypeParam::FooType` (defining alias `TypeParam=Term`) before...

> In any case I think the docstring for Term should be expanded, as its current form ("A term in the language of static parameters in HUGR") is rather unclear, at least to me.

Fair :)

> Suggest combining these error types [SignatureError and TermTypeError], either as a pre- or post-PR (I think this PR is big enough already without adding more changes).

Yeah, I think we can do it after. For this PR I've managed to reduce the number of `.map_err(SignatureError::from)` from 20 down to about 2 by adding only `impl From<TermTypeError> for OpLoadError` and `impl From<TermTypeError> for ImportErrorInner`, which seems good enough for now.

hugr-core/src/extension/resolution/types.rs
```diff
@@ -121,23 +120,40 @@ pub(crate) fn collect_op_types_extensions(
...
 
+/// Collect the Extension pointers in the [`CustomType`]s inside a [FuncValueType].
+///
+/// # Attributes
+///
+/// - `func_ty`: The function type to collect the extensions from.
+/// - `used_extensions`: A The registry where to store the used extensions.
+/// - `missing_extensions`: A set of `ExtensionId`s of which the
+///   `Weak<Extension>` pointer has been invalidated.
+pub(crate) fn collect_func_type_exts(
```
acl-cqc: There is some possibility we could combine these by making `trait Substitutable` (crate-private) include gathering extensions but that would move much out of extension/resolution. (Or we could have a local trait just to parametrize over Term vs TypeRow, I guess?)

hugr-core/src/types/type_param.rs
```diff
@@ -308,8 +372,11 @@ impl Term {
 
     /// Creates a new concatenated list.
     #[inline]
-    pub fn new_list_concat(lists: impl IntoIterator<Item = Self>) -> Self {
-        Self::ListConcat(lists.into_iter().collect())
+    pub fn concat_lists(lists: impl IntoIterator<Item = Self>) -> Self {
```
acl-cqc: Hmmmm, should I update the comment? A `ListConcat([x])` normalizes to just `x`

hugr-core/src/types/type_param.rs
```diff
@@ -641,24 +757,17 @@ pub fn check_term_type(term: &Term, type_: &Term) -> Result<(), TermTypeError> {
...
-            })
-        }
-        (Term::ListConcat(lists), Term::ListType(item_type)) => lists
+        (Term::List(elems), Term::ListType(item_type)) => elems
             .iter()
-            .try_for_each(|list| check_term_type(list, item_type)),
+            .try_for_each(|elem| check_term_type(elem, item_type)),
+        (Term::ListConcat(lists), Term::ListType(_)) => lists
+            .iter()
+            .try_for_each(|list| check_term_type(list, type_)), // ALAN this used the element type, which seems very wrong
```
acl-cqc: driveby bug fix; I will delete comment

hugr-llvm/src/extension/collections/list.rs
```diff
@@ -202,7 +202,7 @@ fn emit_list_op<'c, H: HugrView<Node = Node>>(
     op: ListOp,
 ) -> Result<()> {
     let hugr_elem_ty = match args.node().args() {
-        [TypeArg::Runtime(ty)] => ty.clone(),
+        [ty] => ty.clone().try_into().expect("List elements not a type"),
```
aborgna-q: There are some new `expect`s now in the lowering code that used to return an error instead (in `stack_array.rs` too).

Could we `bail!` instead of panicking?

hugr-core/src/types.rs
```diff
@@ -270,43 +267,59 @@ impl SumType {
...
         }
     }
 
-    /// If a sum is an option of a single type, return the type.
-    #[must_use]
-    pub fn as_unary_option(&self) -> Option<&TypeRV> {
-        self.as_option()
-            .and_then(|row| row.iter().exactly_one().ok())
-    }
+    // ALAN removing as_unary_option.
```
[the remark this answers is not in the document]
aborgna-q: Fine with removing.
I only see a single use in `tket`, that can be replaced by `let Some([ty]) = sum.as_option()` 

Remember to remove the comment.

hugr-core/src/types.rs
```diff
@@ -400,58 +364,32 @@ impl<RV: MaybeRV> TypeEnum<RV> {
...
-            (TypeEnum::Alias(a1), TypeEnum::Alias(a2)) => a1 == a2,
-            (TypeEnum::Function(f1), TypeEnum::Function(f2)) => f1 == f2,
-            (TypeEnum::Variable(i1, b1), TypeEnum::Variable(i2, b2)) => i1 == i2 && b1 == b2,
-            (TypeEnum::RowVar(v1), TypeEnum::RowVar(v2)) => v1.as_rv() == v2.as_rv(),
-            (TypeEnum::Sum(s1), TypeEnum::Sum(s2)) => s1 == s2,
-            _ => false,
-        }
-    }
-}
+pub struct Type(Term, TypeBound);
```
aborgna-q, line 363: nit: This may be an opportunity to change the tuple struct to have named fields.

`type.0` doesn't tell much; `type.term` and `type.bound` would be clearer.

hugr-core/src/types.rs
```diff
@@ -505,89 +437,40 @@ impl<RV: MaybeRV> TypeBase<RV> {
...
-            TypeEnum::Alias(_) => Ok(()),
-            TypeEnum::Extension(custy) => custy.validate(var_decls),
-            // Function values may be passed around without knowing their arity
-            // (i.e. with row vars) as long as they are not called:
-            TypeEnum::Function(ft) => ft.validate(var_decls),
-            TypeEnum::Variable(idx, bound) => check_typevar_decl(var_decls, *idx, &(*bound).into()),
-            TypeEnum::RowVar(rv) => rv.validate(var_decls),
-        }
+        self.0.validate(var_decls)?;
+        // ALAN even this should be only a debug-assert really:
```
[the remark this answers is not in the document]
aborgna-q: > Move caching of TypeBound from Type into SumType::General (via a checked struct - prior to https://github.com/Quantinuum/hugr/commit/0ac2b5316a0f581082bfb36812da8b5e6d799c70 there was such, called GeneralSum, so we could reintroduce that and add the bound there). This because of the other Type variants, CustomType stores the bound itself already, and FunctionType is always Copyable. This can come soon after I think.

If we do the move of the bound then we wouldn't need the check here, right?

hugr-core/src/types.rs
```diff
@@ -876,18 +631,15 @@ pub(crate) mod test {
                 // Dummy extension reference.
                 &Weak::default(),
             )),
-            Type::new_alias(AliasDecl::new("my_alias", TypeBound::Copyable)),
+            //Type::new_alias(AliasDecl::new("my_alias", TypeBound::Copyable)),
```
aborgna-q, line 630:
```suggestion
```

hugr-core/src/types.rs
```diff
@@ -905,9 +657,22 @@ pub(crate) mod test {
...
         let opt = option_type([usize_t()]);
 
-        assert_eq!(opt.as_unary_option().unwrap().clone(), usize_t());
         assert_eq!(
-            Type::new_unit_sum(2).as_sum().unwrap().as_unary_option(),
+            opt.as_option().unwrap().clone(),
+            TypeRowRV::from([usize_t()])
+        );
+        // Two empty variants is like an option of empty.
+        // ALAN note there used to be as_unary_option...
```
aborgna-q: This seems OK though. Alternatively, `assert_matches!(..., [])`

hugr-core/src/extension.rs
```diff
@@ -518,9 +517,10 @@ pub enum SignatureError {
     /// A type variable that was used has not been declared
     #[error("Type variable {idx} was not declared ({num_decls} in scope)")]
     FreeTypeVar { idx: usize, num_decls: usize },
-    /// A row variable was found outside of a variable-length row
-    #[error("Expected a single type, but found row variable {var}")]
-    RowVarWhereTypeExpected { var: RowVariable },
+    // ALAN this is now just another TypeArgMismatch
```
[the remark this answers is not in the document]
aborgna-q: Yeah

hugr-core/src/export.rs
```diff
@@ -982,7 +962,16 @@ impl<'a> Context<'a> {
                 let item_types = self.export_term(item_types, None);
                 self.make_term_apply(model::CORE_TUPLE_TYPE, &[item_types])
             }
-            Term::Runtime(ty) => self.export_type(ty),
+            Term::RuntimeExtension(ext) => self.export_custom_type(ext),
+            /*TypeEnum::Alias(alias) => {
+                let symbol = self.resolve_symbol(self.bump.alloc_str(alias.name()));
+                self.make_term(table::Term::Apply(symbol, &[]))
+            }*/
```
aborgna-q, line 969:
```suggestion
```

hugr-core/src/extension/resolution/types.rs
aborgna-q: I like keeping the logic here rather than in an unrelated trait, esp. if we end up making that trait public at some point.

But it's not a strong preference.

hugr-core/src/extension/resolution/types_mut.rs
```diff
@@ -154,34 +167,18 @@ pub(super) fn resolve_type_row_exts<RV: MaybeRV>(
     Ok(())
 }
 
-/// Update all weak Extension pointers in the [`CustomType`]s inside a type.
+/// Update all weak Extension pointers in the [`CustomType`]s inside a type row.
 ///
-/// Adds the extensions used in the type to the `used_extensions` registry.
-pub(super) fn resolve_type_exts<RV: MaybeRV>(
+/// Adds the extensions used in the row to the `used_extensions` registry.
+fn resolve_typerow_rv_exts(
```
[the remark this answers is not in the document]
aborgna-q: I don't think `TypeRowRV::as_mut_term_unchecked` would really be _unchecked_ since there's nothing to check on the call site, right?

But I agree we should try to avoid the footgun.
An option would be to mark the function as `unsafe` to ensure we are careful on the call point.

hugr-core/src/extension/resolution/types_mut.rs
```diff
@@ -211,17 +208,43 @@ pub(super) fn resolve_custom_type_exts(
...
+pub(crate) fn resolve_type_exts(
+    node: Option<Node>,
+    typ: &mut Type,
+    extensions: &WeakExtensionRegistry,
+    used_extensions: &mut WeakExtensionRegistry,
+) -> Result<(), ExtensionResolutionError> {
+    const EMPTY: Type = Type::new_unit_sum(0); // as no Type::default()
+    let mut tm = std::mem::replace(typ, EMPTY).into();
+    let r = resolve_term_exts(node, &mut tm, extensions, used_extensions);
+    *typ = tm.try_into().unwrap();
```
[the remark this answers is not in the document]
aborgna-q: I feel this one is fine as-is. `Type` is 96B, so all is done on the stack.
We don't need to fumble with pointers as with `TypeRowRV`.

hugr-core/src/hugr/serialize/test.rs
```diff
@@ -552,7 +545,7 @@ fn serialize_types_roundtrip() {
 #[case(bool_t())]
 #[case(usize_t())]
 #[case(INT_TYPES[2].clone())]
-#[case(Type::new_alias(crate::ops::AliasDecl::new("t", TypeBound::Linear)))]
+//#[case(Type::new_alias(crate::ops::AliasDecl::new("t", TypeBound::Linear)))]
```
aborgna-q, line 548:
```suggestion
```

hugr-core/src/types.rs
aborgna-q, line 1: Remember to remove the `ALAN` comments

aborgna-q, line 1: I suggested removing all the commented-out alias stuff. Do we plan to re-instate it with the overhaul?

aborgna-q, line 1: The first proposed followup says:

>  I strongly suggest renaming: Term::XXXType -> Term::XXXKind and Term::RuntimeXXX -> Term::XXXType. (Except Term::RuntimeType -> Term::TypeKind.) I think we probably do not want to merge this PR without having that ready to follow immediately after (or at least, before the next Hugr release).

Is that work in progress?

hugr-core/src/ops/handle.rs
```diff
@@ -91,7 +91,8 @@ impl<const DEF: bool, N> AliasID<DEF, N> {
 
     /// Construct new `AliasID`
     pub fn get_alias_type(&self) -> Type {
-        Type::new_alias(AliasDecl::new(self.name.clone(), self.bound))
+        unimplemented!("Type aliases")
+        //Type::new_alias(AliasDecl::new(self.name.clone(), self.bound))
     }
```
aborgna-q: This is a public method. Can we just delete it?
We should not merge the `unimplemented!`.

hugr-core/src/std_extensions/collections/array/array_value.rs
```diff
@@ -94,8 +94,9 @@ impl<AK: ArrayKind> GenericArrayValue<AK> {
 
         // constant can only hold classic type.
         let ty = match typ.args() {
-            [TypeArg::BoundedNat(n), TypeArg::Runtime(ty)] if *n as usize == self.values.len() => {
-                ty
+            // ALAN checking copyable here might be a bugfix but sounds like we should?
```
aborgna-q: It's cheap to check. It doesn't hurt to return an error.

hugr-core/src/types/serialize.rs
```diff
@@ -25,45 +25,69 @@ pub(crate) enum SerSimpleType {
...
+
+impl From<SerSimpleType> for Term {
+    fn from(value: SerSimpleType) -> Self {
+        match value {
+            SerSimpleType::Q => qb_t().into(),
+            SerSimpleType::I => usize_t().into(),
+            SerSimpleType::G(sig) => Type::new_function(*sig).into(),
+            SerSimpleType::Sum(st) => Type::from(st).into(),
+            SerSimpleType::Opaque(o) => Type::new_extension(o).into(),
+            SerSimpleType::Alias(_) => todo!("alias?"),
```
aborgna-q, line 79: I guess this is a breaking change for the serialization, but we don't use aliases anywhere...

hugr-core/src/types/type_param.rs
```diff
@@ -58,7 +56,20 @@ pub type TypeArg = Term;
...
+/// and instantiate functions, ops, and types. (For example, array lengths
+/// are not types but they are static parameters of array types and ops.)
+///
+/// Terms are used for both parameter declarations and arguments fitting those
+/// parameters, e.g. a [Term::FloatType] parameter would be instantiated (statically)
+/// with a [Term::Float] argument. [`check_term_type`] checks that an argument
+/// is valid (of the correct kind) for the parameter.
+// TODO it might be good to have a separate function that tells, for a given Term,
+// whether there is *any* valid argument; we could then rule out using as parameters
+// any Term for which there are no valid arguments.
```
aborgna-q, line 72: Thanks for adding better docs :)

hugr-core/src/types/type_param.rs
```diff
@@ -111,9 +122,14 @@ pub enum Term {
     #[display("{}", _0.into_inner())]
     Float(OrderedFloat<f64>),
     /// A list of static terms. Instance of [`Term::ListType`].
+    /// Note, not a [TypeRow] because `impl Arbitrary for TypeRow` generates only types.
+    ///
+    /// [TypeRow]: super::TypeRow
     #[display("[{}]", {
         use itertools::Itertools as _;
-        _0.iter().map(|t|t.to_string()).join(",")
+        //_0.iter().map(|t|t.to_string()).join(",")
```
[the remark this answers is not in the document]
aborgna-q: I like the spacing personally, so I'd drop the comment.

Display implementations are meant to be readable.

hugr-core/src/types/type_param.rs
```diff
@@ -262,14 +304,37 @@ impl From<Vec<Term>> for Term {
     }
 }
 
+impl From<Vec<Type>> for Term {
```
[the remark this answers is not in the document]
aborgna-q: Is it being used much?

hugr-core/src/types/type_param.rs
aborgna-q: Yes, I would mention that behaviour

hugr-core/src/types/type_param.rs
```diff
@@ -333,122 +400,42 @@ impl Term {
         }
     }
 
-    /// Returns a [`Type`] if the [`Term`] is a runtime type.
-    #[must_use]
-    pub fn as_runtime(&self) -> Option<TypeBase<NoRV>> {
+    pub(crate) fn least_upper_bound(&self) -> Option<TypeBound> {
```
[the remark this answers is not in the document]
aborgna-q: I wouldn't drop the others even if we make this public, they are just helpers with clear names.
Maybe add a comment in their docs referencing `least_upper_bound` (if public).

hugr-core/src/types/type_row.rs
```diff
@@ -127,259 +118,289 @@ impl TypeRow {
...
 
-// Fallibly convert a [TypeArg] to a [TypeRowRV].
-//
-// This will fail if `arg` is of non-sequence kind (e.g. Type).
-impl TryFrom<Term> for TypeRowRV {
-    type Error = SignatureError;
+/// Fallibly convert a [Term] to a [TypeRow].
+///
+/// This will fail if `arg` is not a [Term::List] or any of the elements are not [Type]s
+impl TryFrom<Term> for TypeRow {
```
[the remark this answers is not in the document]
aborgna-q: If the first conversion is also a `TryFrom` then calling the chain may get too boilerplate-y (since it requires explicit types).

hugr-core/src/types/type_row.rs
```diff
@@ -127,259 +118,289 @@ impl TypeRow {
...
+    }
+
+    /// Wraps the given Term, without checking its type.
+    pub fn new_unchecked(t: impl Into<Term>) -> Self {
+        Self(t.into())
+    }
+
+    /// Creates a singleton row with just a row variable
+    /// (a variable ranging over lists of types of any length)
+    pub fn just_row_var(idx: usize, b: TypeBound) -> Self {
```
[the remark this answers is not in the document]
aborgna-q: I think yes. I associate **just** with `Some` rather than "this is a single element row", but that may be because of of haskell...

hugr-core/src/types/type_row.rs
```diff
@@ -127,259 +118,289 @@ impl TypeRow {
...
+}
+
+// This allows an easy syntax for building TypeRowRV's which are all Types
+impl<T: IntoIterator<Item = Type>> From<T> for TypeRowRV {
+    fn from(value: T) -> Self {
+        Self(Term::new_list(value.into_iter().map_into()))
+    }
+}
+
+/*impl FromIterator<Type> for TypeRowRV {
```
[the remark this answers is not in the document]
aborgna-q: Perhaps we should implement this _instead_ of `<T: IntoIterator<Item = Type>> From<T> for TypeRowRV` ?

Having to `collect_vec` seems quite annoying.

hugr-core/src/types/type_param.rs
acl-cqc: It'd be better to change this to call `new_list_from_parts` rather than the other way around (i.e. make `new_list_from_parts` use a closure that's basically this now), so then e.g. `concat_lists([Term::List([a,b]), Term::List([c,d])])` would give `Term::List([a,b,c,d])` - at present it gives `Term::ListConcat([Term::List([a,b]), Term::List([c,d])])`. But that's a separate PR (and would want to more/updated test cases to make them meaningful)

hugr-core/src/types.rs
acl-cqc: Yes, #3019

hugr-core/src/types/type_row.rs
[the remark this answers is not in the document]
acl-cqc: For TypeRow and TypeRowRV, I think we want `From<Vec<Type>>` and similarly for `From<[Type;N]>`, it'll be too painful otherwise. ATM we write lots of parameters as `impl Into<TypeRow>` so that allows e.g. `someTypeRowRV.concat([type1,type2])`.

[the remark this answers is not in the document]
acl-cqc: <strike>However the place for FromIter might be Term...will continue in thread https://github.com/Quantinuum/hugr/pull/2895#discussion_r3038474886</strike> EDIT, no totally not

hugr-core/src/types/type_param.rs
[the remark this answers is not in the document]
acl-cqc: Mostly in tests in `type_param.rs`. A couple of places elsewhere (validate/test.rs, builder/dataflow.rs again tests).

TBH I think it does no harm - `Term::new_list([a,b])` rather than `[a,b].into()` - not much longer, and is more explicit; the latter syntax is now only good for creating `TypeRow` from `Type`s, or `Term` from `Term`s. (I suppose the same argument might suggest removing `impl From<Vec<Term>> for Term` but I've not gone that far.)

That said, this first requires generalizing `Term::new_list` (to take iter of `T: Into<Type>`), which is a great change in itself (see 983470b823720d6aa31778140cff5a44df18fd6d).

b67372af2ae172679e5d67ae3b88fb89fd3d8c69 then removes the Term-from-list-of-Type, which I think is good to reduce confusion.

<strike>Now it might be reasonable to think about FromIterator - replacing Term from list of Term? (Currently there for Vec and array). But that still changes `[a,b].into()` into `[a,b].into_iter().collect()`. Hmmm, and maybe that `collect` is just a bit weird`</strike>. Dunno, any more thoughts?

hugr-core/src/types/type_row.rs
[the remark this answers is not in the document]
acl-cqc: yes, it'd be a double try-from, as the first checks kind and the second checks fixed-arity

hugr-llvm/src/extension/collections/list.rs
acl-cqc: Ah yes, hadn't seen what `bail` did. Nice. Done (*2, think that's all of them?)

hugr-core/src/types/type_param.rs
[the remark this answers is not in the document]
acl-cqc: (have decided FromIterator for Term is too weird, so going back to the other thread)

hugr-core/src/types/type_row.rs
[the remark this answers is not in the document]
acl-cqc: Ok, so I've replaced the `impl <T: IntoIterator<Item=Type>> From<T> for TypeRowRV` with `impl From<Vec<Type>>` and `impl <const N:usize> From<[Type; N]>`, because that's consistent with what we do for `TypeRow` and also (from array/vec of Term) for `Term`.

However, could also make all three use `<T: IntoIterator<Item....>>` ??

(I think it's good to avoid changing a list of *Type*s into a *Term* without `Term::new_list`)

So FromIterator could still be done as a separate thing. Should be done for both TypeRow *and* TypeRowRV if it is for either, so done in 3a41dda87ac36c951dde79f37f51d4a338164a68 - looks like an improvement??

hugr-core/src/types.rs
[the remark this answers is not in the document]
acl-cqc: It'll move into either SumType::validate or, since that doesn't currently exist ATM, into `Term::validate` (which currently doesn't check the elements are types at all...so that would be a better place)

hugr-core/src/types.rs
acl-cqc: I'm hoping to do the SumType change soon after I think, it's fairly straightforward structural stuff so hopefully can be done via some *ahem* magic (cf. Carl Sagan)...then we'll just have `pub struct Type(Term)`. Happy to get that ready before I merge this...

acl-cqc:
@aborgna-q some improvements on the conversions, thanks, including generalization of Term::new_list which was a big win, some questions remaining at https://github.com/Quantinuum/hugr/pull/2895#discussion_r3065010190.

Also unsure about aliases. Searching for commented-out code mentioning alias might be useful when we revamp....is it any good to include `Term::RuntimeAlias(Void, TypeBound)` i.e. for uninstantiable Void? (Maybe not, I don't think we've even figured out the details of that typebound yet, and will we want alias *terms* or only types?)

hugr-core/src/types.rs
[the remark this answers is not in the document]
acl-cqc: See #3022

hugr-core/src/types.rs
acl-cqc: Yes, #3022

acl-cqc:
* Tried FromIterator for both TypeRow and TypeRowRV, see if you think it's worth it? https://github.com/Quantinuum/hugr/pull/2895/commits/3a41dda87ac36c951dde79f37f51d4a338164a68
* Replaced `impl <T: IntoIterator<Item=Type>> From<T> for TypeRow`(and ...`RV`) with from-Vec and from-Array, because this is consistent with Term and so on. Could go for the generic from-T for all three?
* Moving bound cache into (opaque struct inside) SumType::General in #3022 - needs a bit more work but you can see the idea and can have this ready to merge immediately

I think that's all the serious concerns (in particular, all those marked ALAN), plenty of smaller ones which I can have a look over but ATM it looks like they haven't raised any strong opinions ;)

hugr-core/src/types/type_param.rs
```diff
@@ -448,7 +495,7 @@ impl Term {
             Term::ListType(item_type) => Term::new_list_type(item_type.substitute(t)),
             Term::TupleType(item_types) => Term::new_list_type(item_types.substitute(t)),
```
aborgna-q: drive-by: This looks like a bug. Shouldn't it be `Term::new_tuple_type`?

hugr-core/src/types/serialize.rs
aborgna-q: What are we doing with the alias stubs? If this stays here, it should at least give a better error.

hugr-core/src/types/type_row.rs
```diff
@@ -127,259 +150,327 @@ impl TypeRow {
...
+
+    /// Create a new empty row.
+    pub const fn new() -> Self {
+        Self::EMPTY
+    }
+
+    /// Wraps the given Term, without checking its type.
+    pub fn new_unchecked(t: impl Into<Term>) -> Self {
+        Self(t.into())
+    }
```
aborgna-q: Are we OK with this being public? The rest of the code assumes these are valid by construction so this feels like a footgun.

We could have a public `try_new` if needed, an keep the no-checks version internal-only.

aborgna-q (commented):
LGTM! Just some additional comments.

hugr-core/src/types/type_param.rs
acl-cqc: Good spot! Test coverage here is not what we would like

hugr-core/src/types/type_param.rs
acl-cqc: Yes, there are enough drive-by fixes in this PR now not to worry about another one ;-)

hugr-core/src/types/type_row.rs
acl-cqc: I first made `TypeRowRV::new_unchecked` crate-private, then thought - do we really need it? Can't we just use `try_into().expect()`? The latter works fine, and doesn't change anything on CodSpeed (although -10% on capnp roundtrip *either way* is not great, hmmm)

acl-cqc:
* Removed the commented out alias code. Have kept the commit on my own private branch
* Dropped `TypeRowRV::new_unchecked`.
* ....but are we happy with 10% perf drop in capnp::roundtrip or should I investigate before merging?
* Also do you think we should change the `impl From<Vec<...>>` and `impl <const N: usize> From<[...; N]>` to blanket `impl <T: IntoIterator<Item=....> From<T>`s for Type/Term?

hugr-core/src/builder/module.rs
```diff
@@ -352,7 +352,7 @@ mod test {
         };
         assert_matches!(build_result, Ok(_));
         Ok(())
-    }
+    }*/
```
aborgna-q:
```suggestion
```

hugr-core/src/types/type_param.rs
aborgna-q: Ok 👍 
Should we still update the comment with the current behaviour?

aborgna-q (commented):
Another suggestion to remove the vestigial commented-out alias code, but otherwise :shipit:

aborgna-q (approved):

hugr-core/src/types/type_param.rs
acl-cqc: I think I have done, in 90b9fe300a

acl-cqc:
> Another suggestion to remove the vestigial commented-out alias code

I had actually tried to but had missed a couple! Thanks :)

PR #2951, opened 2026-03-16
feat!: do not require an Order edge for each nonlocal `Ext` edge; compute on demand in scheduling_graph()

acl-cqc:
* Removes validation check that order edges are present (for each nonlocal)
* Removes rust+python builder code that adds order edges (for each nonlocal)
* Actually calculates the edges required and adds them to SynEdgeWrapper
* Updates the SiblingSubgraph test from #3015 to show the changes in behaviour
* Updates spec
* Removes many APIs deprecated in #3015

BREAKING CHANGE: Hugrs are no longer guaranteed to have an Order edge representing the ordering constraint from each nonlocal `Ext` edge; require HugrView impls to provide scheduling_graph; remove deprecated region_portgraph(), as_petgraph(), into_region_portgraph(), sibling_subgraph::{TopoConvexChecker, ValidationMode, SiblingSubgraph::validate(), ConvexChecker}, ValidationError::MissingOrderEdge


hugr-core/src/hugr/views.rs
```diff
@@ -852,6 +823,102 @@ impl HugrView for Hugr {
...
+    syn_edges
+}
+
+/// Like [HugrView::scheduling_graph], but returns a [SchedulingGraph] that owns its own view of the region,
+/// taken from the Hugr.
+///
+/// The API is designed for `hugr-persistent` and allows to implement [HugrView] such that
+/// [HugrInternals::RegionPortgraph] is not tied to the original view but rather an owned
+/// temporary [Hugr].
+pub fn owned_scheduling_graph<'a, N: HugrNode, V>(
```
acl-cqc, line 895: This is broadly equivalent to the removed `Hugr::into_region_portgraph` from `internals.rs`; it could be an instance method again (`into_scheduling_graph`) but seems more consistent in this file.

It is definitely the part of the PR that I'm least keen on!! But either we expose more internal details of SchedulingGraph (which I've been trying to avoid in the design, to keep some flexibility to change in the future), and use those details in `hugr-persistent`; or, we put some `hugr-persistent`-specific functionality here. This PR chooses the latter route...
(but it doesn't seem that weird a functionality, even if `hugr-persistent` is the only place we're using it ATM.)

Thoughts welcome!

specification/hugr.md
```diff
@@ -300,9 +300,14 @@ flowchart
 
 In a dataflow graph, the evaluation semantics are simple: all nodes in
 the graph are necessarily evaluated, in some order (perhaps parallel)
-respecting the Dataflow edges. The following operations are used to
+respecting the Dataflow edges.** The following operations are used to
 express control flow, i.e. conditional or repeated evaluation.
 
+**ALAN this might be a place to either say, and respecting the ordering
```
acl-cqc: Thoughts here?

hugr-py/tests/test_hugr_build.py
```diff
@@ -186,6 +186,8 @@ def test_build_nested(snapshot):
 
 
 def test_build_inter_graph(snapshot):
+    # Possibly a bit redundant now, really we're just testing that we *don't* do
+    # anything special anymore, following https://github.com/Quantinuum/hugr/pulls/2951.
```
cqc-alec:
```suggestion
    # anything special anymore, following https://github.com/Quantinuum/hugr/pull/2951.
```

specification/hugr.md
cqc-alec: Yes, could be worth saying that descendent nodes may be interleaved with siblings, subject to the ordering constraints imposed by Dataflow edges.

cqc-alec (approved):

hugr-py/tests/test_hugr_build.py
acl-cqc: I am humbled. I would never have spotted that myself :-)

specification/hugr.md
acl-cqc: I have done - I realize that I was actually planning for that to come later along with tweaks to some Hugr code to enact it, but it's not a correction to the spec (it doesn't change anything that was there before AFAICS), merely a clarification; and whilst there is some Hugr code that sort-of-assumes otherwise, I think updating that can be classed as an improvement or optimization (or just possibly bugfix) i.e. in another PR.

So, would you mind having another quick look over the spec 🙏 ? I also explicitly mentioned respecting Order edges too (but not synthetic/nonlocal edges since there is no need to).

acl-cqc: (@cqc-alec, sorry!)

cqc-alec (approved):
LGTM

Issue #2961, opened 2026-03-18
[Feature]: Make `_scope` and `_composable_pass` modules public


CalMacCQ:
### Problem Statement

These modules have stabilized more or less. We can make them public.

PR #2965, opened 2026-03-18
fix: allow both _composable_pass and composable_pass imports

CalMacCQ:
resolves breakage in #2962 by allowing both `from _composable_pass` and `from composable_pass` imports. This is a problem as in tket-py we access the private module `_composable_pass` for the implementation of `PytketHugrPass` (not the best practice on my part).

I've added a duplicate` _composable_pass.py` file now with a wildcard import (gross)

Will remove this once we have `_composable_pass` usage removed from tket-py. There shouldn't be any `_scope` usage in the tket-py main branch.

Theres probably a nicer way to do this with something like

```python
import sys
sys.modules["_foo"] = sys.modules[__name__]
```
However this wasn't immediately working for me.

Related #2961

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -8,70 +8,70 @@ exclude = ["tests/error", "test_files/"]
 [lint]
 
 select = [
-    "F", # pyflakes
-    "E", # pycodestyle Errors
-    "W", # pycodestyle Warnings
+  "F", # pyflakes
+  "E", # pycodestyle Errors
+  "W", # pycodestyle Warnings
 
-    # "A",    # flake8-builtins
-    # "ANN",  # flake8-annotations
-    # "ARG",  # flake8-unused-arguments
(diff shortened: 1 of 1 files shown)


hugr-py/tests/test_passes.py
```diff
@@ -6,12 +6,13 @@
...
 
 from hugr.hugr.base import Hugr
-from hugr.passes.composable_pass import (
+from hugr.passes._composable_pass import (
     ComposablePass,
     ComposedPass,
     PassResult,
     implement_pass_run,
 )
+from hugr.passes.composable_pass import *
```
CalMacCQ: Added this line to test that it actually works for both import paths

CalMacCQ:
I'm on holiday tomorrow so feel free to edit/merge in my absence if need be.

PR #2971, opened 2026-03-19
feat: Add debug info metadata specification in `hugr-py`

tatiana-s:
Moved from https://github.com/Quantinuum/guppylang/pull/1554

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -317,9 +317,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "capnp"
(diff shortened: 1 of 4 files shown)


hugr-py/src/hugr/debug_info.py
```diff
@@ -0,0 +1,152 @@
...
+    """Debug information for a location, corresponds to call or extension operation
+    node.
+    """
+
+    KIND: ClassVar[str] = "location"
+
+    column: int
+    line_no: int
+
+    def to_json(self) -> dict[str, str]:
```
cqc-alec: Should be `dict[str, JsonType]`?

cqc-alec (approved):

Issue #2946, opened 2026-03-13
[Feature]: Remove requirement to add Order edge for each nonlocal edge


acl-cqc:
### Problem Statement

These order edges give us a lot of problems - they often prevent optimization because algorithms don't deal with them (or bail if order are present), and much complexity in builder and elsewhere to add them and verify that they are there.

(Result of adding an order edge to a DFG and then using InlineDFG is that you get an order edge from the original source, to *every* just-after-Input node from the DFG. Yet we probably need none of those...)

As a parallel(/first) step to removing order edges altogether, we can remove these by generating petgraph "Views" that include synthetic edges (not present in the Hugr) representing exactly these order restrictions.

### Proposed Solution

Add a `region_petgraph` function alongside `region_portgraph`.
   * This returns a new struct, that contains both the current portgraph wrapper of just a region of the Hugr; and its own list of edges, computed in `region_petgraph` by scanning the subtree looking for outgoing nonlocal edges (i.e. which go to within the subtree).
   * We can then implement the relevant petgraph traits to union those two together, using hugr `PetgraphWrapper` as a model
   * for `petgraph::visit::GraphBase` we probably want `type EdgeId = ((T::Node, PortOrSynthetic), (T::Node, PortOrSynthetic));` (i.e. some custom enum that might be a port, for an edge in the hugr, or might be a synthetic....not sure if it's possible for that thing to enforce that either both ends are ports or both are synthetic)
   * Similarly in `petgraph::visit::Data` we probably want ` type EdgeWeight = EdgeKindOrSynthetic` (akin to `Option<EdgeKind>` but again an enum so we can extend in future).
   * `petgraph::visit::IntoNeighbours` and `IntoNeighboursDirected` will need to add new edges
   * *Hopefully* we won't need `GetAdjacencyMatrix`

Then we migrate uses of `region_portgraph` to use `region_petgraph`, and potentially deprecate `region_portgraph`.
* Most uses only use a petgraph `Topo` anyway so should just do that on the new struct
* SiblingSubgraph needs significant work, but note it doesn't support order edges at all ATM, so we may not need to do much. We *must* take account of the synthetic edges in convexity(/cycle) checking, but beyond that, can probably ignore them for purpose of determining the boundary ports of the rewrite - this should make SiblingSubgraph more widely applicable already

### Component

Rust Core

### Effort Estimate

Medium - Moderate complexity

acl-cqc:
Completed with #2951

PR #3015, opened 2026-04-02
refactor: add HugrView::scheduling_graph, deprecate region_portgraph

acl-cqc:
This paves the way for #2946, but does not solve. For that, we will need to add "fake" edges: for each nonlocal edge, from the source to the sibling that is ancestor of the target. (Currently stored in Hugr as order edges but these will be removed). However, it will not be possible to add these fake edges to the `region_portgraph`: being a portgraph (view of hugr), there is no port for the fake edges.

Instead this PR introduces a new method `scheduling_graph` returning a new SchedulingGraph structure: a generic petgraph, which augments a view of the Hugr with an explicit list of extra edges. In this PR, said list is empty, as the Hugr still contains the order edges necessary to ensure topological orderings are sound. #2951 removes the order edges from Hugr and generates a list of those required as part of the SchedulingGraph.

This PR includes moving `sibling_subgraph.rs` tests into a separate `tests.rs`, and also copies the `TopoConvexChecker` from portgraph (reparametrized to work on a petgraph).

Add some tests demonstrating existing behaviour of SiblingSubgraph and `scheduling_graph` with order edges, these are then updated in #2951 to show the (i.e. breaking) changes in behaviour.


hugr-core/src/hugr/views/sibling_subgraph/convex.rs
```diff
@@ -0,0 +1,117 @@
+use petgraph::visit::{
+    GraphBase, IntoNeighborsDirected, IntoNodeIdentifiers, NodeCount, Topo, Visitable, Walker,
+};
+use std::collections::BTreeSet;
+
+use portgraph::{SecondaryMap, UnmanagedDenseMap};
```
acl-cqc, line 6: This file is basically a straight copy of TopoConvexChecker from portgraph then slightly reparametrized

hugr-core/src/hugr/views/sibling_subgraph/tests.rs
```diff
@@ -0,0 +1,778 @@
+use std::collections::BTreeSet;
```
acl-cqc, line 1: Tests moved out of sibling_subgraph.rs

hugr-core/src/hugr/views/syn_edge.rs
```diff
@@ -0,0 +1,231 @@
...
+#[derive(Clone, Debug)]
+pub(super) struct SynEdgeWrapper<T: LinkView> {
+    pub(super) region_view: T,
+    #[allow(clippy::type_complexity)]
+    pub(super) syn_edges: Vec<(NIdx<T::NodeIndexBase>, NIdx<T::NodeIndexBase>)>,
+}
+
+impl<T: LinkView> pv::GraphBase for SynEdgeWrapper<T> {
+    type NodeId = NIdx<T::NodeIndexBase>;
+    type EdgeId = (
```
acl-cqc, line 29: Note this is significantly more informative than portgraph's petgraph-traits, but perhaps it shouldn't be

hugr-core/src/hugr/views/syn_edge.rs
```diff
@@ -0,0 +1,231 @@
...
+
+impl<T: LinkView> pv::NodeCount for SynEdgeWrapper<T> {
+    fn node_count(&self) -> usize {
+        self.region_view.node_count()
+    }
+}
+
+impl<T: LinkView> pv::NodeIndexable for SynEdgeWrapper<T> {
+    fn node_bound(&self) -> usize {
+        // ALAN copied from petgraph; but are `NodeIndex`es always dense?
```
acl-cqc, line 48: I was a bit worried here??

hugr-core/src/hugr/views/syn_edge.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// Turns out the underlying [FlatRegion] has unit node weights, we may want to fix that.
+    ///
+    /// [FlatRegion]: portgraph::view::FlatRegion
+    type NodeWeight = ();
+
+    /// Turns out the underlying [FlatRegion] has unit edge weights; we may want to fix that,
+    /// but at least this distinguishes synthetic edges from original edges.
+    ///
+    /// [FlatRegion]: portgraph::view::FlatRegion
+    type EdgeWeight = MaybeSynEdge<(PortOffset<T::PortOffsetBase>, PortOffset<T::PortOffsetBase>)>;
```
acl-cqc, line 77: Again, portgraph's petgraph-traits does not provide any edge info, not even just the port offsets

hugr-core/src/hugr/views.rs
```diff
@@ -403,6 +403,22 @@ pub trait HugrView: HugrInternals {
         PetgraphWrapper { hugr: self }
     }
 
+    /// A view of a flat region, including ordering constraints from nonlocal edges,
+    /// suitable for use with petgraph algorithms.
+    fn scheduling_graph(&self, parent: Self::Node) -> SchedulingGraph<'_, Self> {
```
acl-cqc, line 408: Name is very much up for grabs, I wasn't too keen on `region_petgraph` (it also has to return a nodemap, and doesn't really say what it's about - edges not in the hugr).

hugr-core/src/hugr/views.rs
```diff
@@ -552,6 +568,42 @@ impl<S: HugrNode> ExtractionResult<S> for HashMap<S, Node> {
     }
 }
 
+/// A graph of a flat region of a Hugr, including ordering constraints from nonlocal edges
+pub struct SchedulingGraph<'a, V: HugrView + ?Sized + 'a> {
```
acl-cqc, line 572: So we could have `fn scheduling_graph` return a pair of the nodemap and some kind of graph, but I wanted to hide the type of `graph` below (!!), and returning a pair of (nodemap, `impl pv::GraphBase+.....`) didn't seem great - for one thing, it'd make it impossible for anyone to store that graph in a struct

hugr-core/src/hugr/views.rs
```diff
@@ -552,6 +568,42 @@ impl<S: HugrNode> ExtractionResult<S> for HashMap<S, Node> {
...
+
+    /// Like [Self::node_map] but extracts the map, discarding the rest of `self`
+    pub fn into_node_map(self) -> V::RegionPortgraphNodes {
+        self.node_map
+    }
+
+    /// Access to the graph, sufficient to allow [pv::Topo]
+    pub fn petgraph(
+        &self,
+    ) -> impl pv::NodeCount
```
acl-cqc, line 610: this type is part of the hiding, and quite deliberate. It could be much more detailed (NodeId/EdgeId types etc.)

hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -789,18 +839,29 @@ impl<N: HugrNode> SiblingSubgraph<N> {
...
-pub enum ValidationMode<'t, 'h, H: HugrView> {
-    /// Check convexity with the given checker.
-    WithChecker(&'t TopoConvexChecker<'h, H>),
-    /// Construct a checker and check convexity.
-    #[default]
-    CheckConvexity,
-    /// Skip convexity check.
-    SkipConvexity,
+mod hidden {
+    #![expect(deprecated)] // Remove enum along with TopoConvexChecker
```
acl-cqc, line 833: This `mod hidden` is so we can do this to capture the deprecation warning from `#[derive(Default)]`

hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -1028,19 +1097,22 @@ pub type TopoConvexChecker<'g, Base> =
 ///
 /// This is a good choice for checking convexity of circuit-like graphs,
 /// particularly when many checks must be performed.
-pub type LineConvexChecker<'g, Base> =
-    ConvexChecker<'g, Base, portgraph::algorithms::LineConvexChecker<CheckerRegion<'g, Base>>>;
+pub type LineConvexChecker<'g, Base> = PortgraphCheckerWithNodes<
```
acl-cqc, line 1090: So I've put some effort into trying to keep this alive, on the grounds that it's potentially much more efficient than TopoConvexChecker. I *think* it works only for graphs with no container nodes, in which case there will be no syn edges, so it'll still be as ok as it ever was.

But, we could also drop it and plan to reimplement line checking somewhere else if we want it, possibly based off https://github.com/Quantinuum/tket2/pull/1289 somehow

hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -1526,797 +1626,97 @@ fn has_unique_linear_ports<H: HugrView>(host: &H, ports: &OutgoingPorts<H::Node>
...
-        let subg = SiblingSubgraph::try_new_dataflow_subgraph::<_, DfgID>(
-            RootChecked::try_new(&h).expect("Root should be DFG."),
+    fn nodes_if_convex(
+        &self,
+        hugr: &impl HugrView<Node = H::Node>,
+        inputs: &IncomingPorts<H::Node>,
+        outputs: &OutgoingPorts<H::Node>,
+        function_calls: &IncomingPorts<H::Node>,
+    ) -> Result<Vec<H::Node>, InvalidSubgraph<H::Node>> {
+        // Compute the nodes inside the boundary ignoring synthetic edges -
```
acl-cqc, line 1636: This is key, can you check that you agree?

hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -1526,797 +1626,97 @@ fn has_unique_linear_ports<H: HugrView>(host: &H, ports: &OutgoingPorts<H::Node>
...
+        // there is no way to specify a synthetic edge as part of the boundary,
+        // and if there are any nonlocal edges not part of the boundary (the condition
+        // that would lead to needing those synthetic edges) then the subgraph is invalid
+        // (nodes would not share the same parent)!
+        let node_indices = make_pg_subgraph::<H>(
+            self.checker.graph().region_view.clone(),
```
acl-cqc, line 1642: So I guess this is the same backdoor as `portgraph_no_syn_edges`, and a slight shame I can't use the same method, but here I think it *is* the right thing to do

hugr-core/src/hugr/views/sibling_subgraph.rs
aborgna-q: You can avoid the module by using `allow` instead of `expect`. (Seems like a bug, but [this works](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1175ef9a822642344b12e955579eea7e))

```rs
#[deprecated]
#[allow(deprecated)]
#[derive(Default)]
pub enum E {
    #[default]
    A
}
```

hugr-core/src/hugr/views/sibling_subgraph.rs
aborgna-q: Nice! If it works with your changes I would keep it.

hugr-core/src/hugr/views/sibling_subgraph.rs
aborgna-q: The logic looks good to me

hugr-core/src/hugr/views.rs
aborgna-q: I like `scheduling_graph`, it conveys well the causal order meaning.

hugr-core/src/hugr/views/syn_edge.rs
aborgna-q: Ouch, that looks wrong.

It should be `self.region_view.node_capacity()` instead, since node indices do not get compacted after a removal.

aborgna-q (approved):
LGTM asides from the node bound bug.

hugr-core/src/hugr/views/sibling_subgraph.rs
acl-cqc: Hah! Yes, that is odd, I had tried expect. Thanks!

PR #3019, opened 2026-04-07
refactor(hugrv2)!: Rename Terms to clarify (runtime) Types from static parameter Kinds

acl-cqc:
[Note: this will textually conflict with #3022, will have to merge whichever is approved second but I think should b straightforward.]

Using `Type` to mean a static parameter (that accepts a static Term argument) is confusing; instead keep Type as meaning a type of *runtime values*

* `Term::RuntimeType` becomes `Term::TypeKind`
* `Term::XXXType` becomes `Term:::XXXKind` (e.g. `StringType` -> `StringKind`, `ListType` -> `ListKind`)
* `Term::RuntimeXXX` becomes `Term::XXXType` (e.g. `Term::RuntimeExtension` -> `Term::ExtensionType)`
* `check_term_type` -> `check_term_kind`
* `TermTypeError` -> `TermKindError`, variant `TypeMismatch` -> `KindMismatch` and field `type_` -> `kind`, reviewed separately in #3053

BREAKING CHANGE: many renames of variants of Term; ...Types are now Kinds, Runtime... are now Types. Also check_term_type is now check_term_kind, Term(Type->Kind)Error::(Type->Kind)Mismatch::{type_->kind}

### DIFF
--- a/hugr-core/src/builder/dataflow.rs
+++ b/hugr-core/src/builder/dataflow.rs
@@ -481,7 +481,7 @@ pub(crate) mod test {
     use crate::metadata::Metadata;
     use crate::ops::{FuncDecl, FuncDefn, OpParent, OpTag, OpTrait, Value, handle::NodeHandle};
     use crate::std_extensions::logic::test::and_op;
-    use crate::types::type_param::{TermTypeError, TypeParam};
+    use crate::types::type_param::{TermKindError, TypeParam};
     use crate::types::{EdgeKind, FuncValueType, Signature, Term, Type, TypeBound, TypeRowRV};
     use crate::utils::test_quantum_extension::h_gate;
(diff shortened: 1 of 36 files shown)


cqc-alec (approved):
Thanks!

> Should we also change TermTypeError::TypeMismatch { term, type_} to TermTypeError::KindMismatch {value, kind}? I guess we should??

Yes I guess we should do that as well.

acl-cqc:
> Thanks!
> 
> > Should we also change TermTypeError::TypeMismatch { term, type_} to TermTypeError::KindMismatch {value, kind}? I guess we should??
> 
> Yes I guess we should do that as well.

I am gonna do that as a followup because I realize there is also TermTypeError (=> TermKindError?)...EDIT done in #3053 and merged in here

PR #3022, opened 2026-04-10
perf!: remove bound-caching from Type, do in SumType::General instead

acl-cqc:
`Type` currently caches the `TypeBound` alongside its Term; there are only three possible Term:: variants, and the cache is pretty much redundant in two (and a half):
* Term::RuntimeFunction --> we know it's TypeBound::Copyable
* Term::RuntimeExtension --> the ExtensionType *also* stores the bound, so we can just read that
* Term::RuntimeSum depends on the SumType
   * SumType::Unary --> we know it's TypeBound::Copyable
   * SumType::General --> the only case where we actually need it

Hence, make SumType::General store a `struct GeneralSum`, with bound and rows (these are kept private to avoid mutating rows without updating the bound). And remove the cache in Type.

BREAKING CHANGE: any match on `SumType::General {rows} ...rows...` becomes `SumType::General(gs) ...gs.rows()... `; construct GeneralSum with ::new


hugr-core/src/types.rs
```diff
@@ -316,7 +335,13 @@ impl Transformable for SumType {
     fn transform<T: TypeTransformer>(&mut self, tr: &T) -> Result<bool, T::Err> {
         match self {
             SumType::Unit { .. } => Ok(false),
-            SumType::General { rows } => rows.transform(tr),
+            SumType::General(general) => {
+                let changed = general.rows.transform(tr)?;
+                if changed {
+                    *general = GeneralSum::new(std::mem::take(&mut general.rows));
```
acl-cqc, line 341: Does this need a comment to say, recompute the bound?

hugr-core/src/types.rs
```diff
@@ -177,8 +177,44 @@ pub enum SumType {
...
+        Self { rows, bound }
+    }
+
+    /// Returns the variant rows of the sum.
+    #[must_use]
+    pub fn rows(&self) -> &[TypeRowRV] {
+        &self.rows
+    }
+
+    pub(crate) fn rows_mut(&mut self) -> &mut [TypeRowRV] {
```
acl-cqc, line 215: Note this allows bypassing the update of the bound, so we keep it crate-private (and use it only for things where we are sure the bound doesn't change - namely, extension resolution)

hugr-core/src/types.rs
```diff
@@ -448,14 +472,6 @@ impl Type {
     /// [TypeDef]: crate::extension::TypeDef
     pub(crate) fn validate(&self, var_decls: &[TypeParam]) -> Result<(), SignatureError> {
         self.0.validate(var_decls)?;
-        // ALAN even this should be only a debug-assert really:
```
acl-cqc, line 448: This temporary debris from #2895 has now moved to a proper home in Term::validate

hugr-core/src/types/type_param.rs
```diff
@@ -428,8 +428,16 @@ impl Term {
     /// - this is left to [check_term_type].
     pub(crate) fn validate(&self, var_decls: &[TypeParam]) -> Result<(), SignatureError> {
         match self {
-            Term::RuntimeSum(SumType::General { rows }) => {
+            Term::RuntimeSum(SumType::General(GeneralSum { rows, bound })) => {
```
doug-q: I think it would be better to not match on GeneralSum here, and instead match `SumType::General(general)` and use methods on general to introspect below.

doug-q (approved):
Looks good Alan.

I would prefer that we don't match on `GeneralSum`, and instead add public methods and use those. This makes it easier to make similar changes in the future, and pushes us to include public functions to help downstream users, who are not able to access the fields of GeneralSum.

hugr-core/src/types/type_param.rs
acl-cqc: I've done a few more instances of this within types.rs/type_param.rs as part of merging with #3019, although I realize that you probably meant this one in particular because it's the (only) one that doesn't have `...` so would have broken if we'd added any more fields. Hope you don't mind a few more uses of the methods!

(And added `pub fn bound()` which seemed to have been missing, oops)

doug-q: No I meant all of them(ref parent comment), and indeed forcing `bound` etc to exist is one reason to do this

PR #3039, opened 2026-04-24
chore(deps-rs): Upgrade inkwell to 0.9.0

### LINKED ISSUE #3038: [hugr-llvm] Update to inkwell 0.9.0
Looks like there are some minor API changes from 0.8.0: see #3037 .

maximilianruesch:
Updates inkwell to 0.9.0 and deprecates functions in the process (thus, once merged, this should be modified to perhaps show up in the changelog?).

Concretely, the following upgrade decisions were made:
- Where bit width is now checked, non-zeroness structs are used and unwrapped since we use literals or constructs for which it is mathematically impossible to be zero. Also, there is an assumption that we never go beyond the maximum supported bit width (and the existing message from inkwell is useful enough if we do).
- The `PassManager` is deprecated in inkwell. Since the replacement is a bit more involved and can be run directly on the module that the using struct exposes via `::module`, the function using it was deprecated (no usage was detected, but it is public). *We could also just remove it imo.*
- The minimum inkwell version is raised to `0.9.0`, since the new API is not backward compatible.

Closes #3038

BEGIN_COMMIT_OVERRIDE
feat: Upgrade inkwell to 0.9.0, and deprecate `hugr-llvm::emit::test::Emission::opt`. Use the new inkwell passes on the module exposed via `::Emission::module` instead.
END_COMMIT_OVERRIDE

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1378,22 +1378,22 @@ dependencies = [
 
 [[package]]
 name = "inkwell"
-version = "0.8.0"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1def4112dfb2ce2993db7027f7acdb43c1f4ee1c70a082a2eef306ed5d0df365"
+checksum = "7decbc9dfa45a4a827a6ff7b822c113b1285678a937e84213417d4ca8a095782"
 dependencies = [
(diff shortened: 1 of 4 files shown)


cqc-alec (approved):
Thanks!

PR #3046, opened 2026-05-01
ci: Build WASM wheels for emscripten v5

aborgna-q:
Updates the target pyodide ABI from [pyodide_2025_0](https://github.com/pyodide/pyodide/blob/main/docs/development/abi/313.md) to [pyemscripten_2026_0](https://github.com/pyodide/pyodide/blob/main/docs/development/abi/314.md).

We're not actually specifying the ABI anywhere, just building compatible wheels.

We no longer require rust nightly to build the ABI-compatible wheels (see the linked ABI docs), but it's still a good idea to do so, so we can pass `-Z build-std` to the compiler and ensure the rust ABI is correct (see https://doc.rust-lang.org/beta/rustc/platform-support/wasm32-unknown-emscripten.html#emscripten-abi-compatibility).

Test run: https://github.com/Quantinuum/hugr/actions/runs/25214250043/job/73931092528

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -273,13 +273,14 @@ jobs:
         # Ideally this should match both the version used by rustc and the pyodide ABI.
         # See <https://doc.rust-lang.org/beta/rustc/platform-support/wasm32-unknown-emscripten.html>
         # and <https://github.com/pyodide/pyodide/blob/main/docs/development/abi.md>
-        # Pyodide `0.28` and `0.29` use 4.0.9
-        # We fix a rustc version compatible with it.
+        #
+        # We instead use the emsdk chosen by the `pyemscripten_2026_0` ABI.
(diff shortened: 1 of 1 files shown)


aborgna-q:
Turns out the current stable pyodide is still on the old v4 ABI, so I'll close this for now.

aidanCQ:
Pyodide v314.0.0 requires emscripten v5 so sheepishly seeking to reopen this 🙏🏻 

```
  File "/lib/python3.14/site-packages/micropip/transaction.py", line 102, in add_requirement_from_url
    check_compatible(wheel.filename)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/lib/python3.14/site-packages/micropip/_utils.py", line 191, in check_compatible
    raise ValueError(
    ...<2 lines>...
    )
ValueError: Wheel was built with Emscripten v4.0.9 but Pyodide was built with Emscripten v5.0.3

```

PR #3047, opened 2026-05-05
chore: Fix cyclic dev dependency in hugr-model

aborgna-q:
#3016 introduced a cyclic dev dependency between `hugr-model` and `hugr-core` to validate the `.edn`s used for testing.

Crates.io does not allow such cycles, so this caused an error when releasing `0.27.1`.

This PR fixes that by moving the validation test to `hugr-core` instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1147,7 +1147,6 @@ dependencies = [
  "bumpalo",
  "capnp",
  "derive_more 2.1.1",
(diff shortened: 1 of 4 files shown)


croyzor (approved):

aborgna-q:
This fix got backported and used to make a fixed `0.21.7` release.

PR #3055, opened 2026-05-12
fix: Replace non-deterministic iterations on hash maps

aborgna-q:
Enables the `iter_over_hash_type` clippy lint, and fixes all cases of iteration over hashmaps and hashsets that could cause non-deterministic outputs.

### DIFF
--- a/hugr-core/src/hugr.rs
+++ b/hugr-core/src/hugr.rs
@@ -11,7 +11,6 @@ pub mod validate;
 pub mod views;
 
 use std::cmp::Ordering;
-use std::collections::HashMap;
 use std::collections::VecDeque;
 use std::io;
 use std::iter;
@@ -444,7 +443,7 @@ impl Hugr {
     /// (since these must be the entry and exit blocks) and in each dataflow
     /// sibling graph (since these must be the input and output nodes).
     fn order_siblings_by_node_index(&mut self) {
-        let mut node_children: HashMap<Node, Vec<Node>> = HashMap::default();
+        let mut node_children = Vec::new();
         for node in self.nodes() {
             let mut children = self.children(node).collect_vec();
             if self.contains_dsg_or_csg(node) {
@@ -452,9 +451,9 @@ impl Hugr {
             } else {
                 children.sort();
             }
-            node_children.insert(node, children);
+            node_children.push((node, children));
         }
-        for (node, children) in &node_children {
+        for (node, children) in node_children {
             self.hierarchy.detach_children(node.into_portgraph());
             for child in children {
                 self.hierarchy
--- a/hugr-core/src/hugr.rs
+++ b/hugr-core/src/hugr.rs
@@ -443,17 +443,16 @@ impl Hugr {
     /// (since these must be the entry and exit blocks) and in each dataflow
     /// sibling graph (since these must be the input and output nodes).
     fn order_siblings_by_node_index(&mut self) {
-        let mut node_children = Vec::new();
-        for node in self.nodes() {
+        // Equivalent to `self.nodes()` that does not borrow the full structure, so we can mutate `self.hierarchy` while iterating.
+        let nodes = self.graph.nodes_iter().map_into();
+        for node in nodes {
             let mut children = self.children(node).collect_vec();
             if self.contains_dsg_or_csg(node) {
                 children[2..].sort();
             } else {
                 children.sort();
             }
-            node_children.push((node, children));
-        }
-        for (node, children) in node_children {
+
             self.hierarchy.detach_children(node.into_portgraph());
             for child in children {
                 self.hierarchy
--- a/hugr-core/src/hugr/linking.rs
+++ b/hugr-core/src/hugr/linking.rs
@@ -91,6 +91,10 @@ pub trait HugrLinking: HugrMut {
             roots.insert(other.entrypoint(), parent);
             other.set_parent(other.entrypoint(), other.module_root());
         };
+        #[expect(
+            clippy::iter_over_hash_type,
+            reason = "roots are stored in a BTreeMap and child pruning is independent"
+        )]
         for (ch, dirv) in children.iter() {
             roots.insert(*ch, self.module_root());
             if matches!(dirv, NodeLinkingDirective::UseExisting(_)) {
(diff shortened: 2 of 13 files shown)


hugr-core/src/hugr/patch/outline_cfg.rs
```diff
@@ -481,7 +481,7 @@ mod test {
         cfg: Node,
         blocks: Vec<Node>,
     ) -> (Node, Node, Node) {
-        let mut other_blocks = h.children(cfg).collect::<HashSet<_>>();
+        let mut other_blocks = h.children(cfg).collect::<BTreeSet<_>>();
```
acl-cqc: Not necessary - we only iterate through this to do an `assert` (line 489, five lines down) - you could either `expect` on the loop or rewrite to an `.all` (if that helps?)

hugr-core/src/hugr/linking.rs
```diff
@@ -759,10 +763,22 @@ fn check_directives<SRC: HugrView, TN: HugrNode>(
     parent: Option<TN>,
     children: &NodeLinkingDirectives<SRC::Node, TN>,
 ) -> Result<Transfers<SRC::Node, TN>, NodeLinkingError<SRC::Node, TN>> {
+    /// Returns the minimum key in `children` for which `pred` returns true, or
+    /// `None` if there is no such key.
+    ///
+    /// Helper function to avoid indeterminism in error reporting from iterating
```
acl-cqc:
```suggestion
    /// Helper function to avoid nondeterminism in error reporting from iterating
```

hugr-core/src/hugr/linking.rs
```diff
@@ -787,10 +803,15 @@ fn check_directives<SRC: HugrView, TN: HugrNode>(
...
-        if other.get_parent(sn) != Some(other.module_root()) {
-            return Err(NodeLinkingError::NotChildOfRoot(sn));
-        }
+    if let Some(sn) = min_directive_key(children, |sn| {
+        other.get_parent(sn) != Some(other.module_root())
+    }) {
+        return Err(NodeLinkingError::NotChildOfRoot(sn));
+    }
+    for sn in other.children(other.module_root()) {
+        let Some(dirv) = children.get(&sn) else {
```
acl-cqc: Eeek! This means we're now going through `other.children` which might be much bigger than `children`.

Nondeterminism only affects the error reported (NodeMultiplyReplaced), but to prevent that, you are probably better off copying `children` into a BTreeSet (avoiding the need for `min_directive_key` and so on)

hugr-core/src/hugr.rs
```diff
@@ -444,17 +443,17 @@ impl Hugr {
     /// (since these must be the entry and exit blocks) and in each dataflow
     /// sibling graph (since these must be the input and output nodes).
     fn order_siblings_by_node_index(&mut self) {
-        let mut node_children: HashMap<Node, Vec<Node>> = HashMap::default();
+        let mut node_children = Vec::new();
```
acl-cqc: I'm not convinced there's any actual nondeterminism here; processing each element of `node_children` seems independent.

However, `node_children` seems like an enormous structure. Why not just merge the loops (move the `self.hierarchy` stuff in the second loop, into the first)? You'd need to do `for node in self.nodes().collect::<Vec<_>>()` but that vec would be *much smaller* than `node_children` after all!

hugr-persistent/src/persistent_hugr.rs
```diff
@@ -187,13 +188,10 @@ impl PersistentHugr {
...
         // Check that `replacement` does not conflict with siblings at any of its
         // parents
-        let new_invalid_nodes = replacement
-            .subgraph()
-            .nodes()
-            .iter()
-            .map(|&PatchNode(id, node)| (id, node))
-            .into_grouping_map()
-            .collect::<BTreeSet<_>>();
+        let mut new_invalid_nodes = IndexMap::<_, BTreeSet<_>>::new();
```
acl-cqc, line 191: This feels independent/driveby? Map on LHS is BTreeMap not HashMap?

acl-cqc (commented):
Love it, thanks @aborgna-q - a few suggestions but happy to approve unless you'd rather @mark-koch had a final look over?

hugr-core/src/hugr.rs
aborgna-q: Done! We're only mutating the hierarchy, so the borrow checker is happy if we iterate over `self.graph`'s nodes instead.

hugr-persistent/src/persistent_hugr.rs
aborgna-q: The groups themselves are stored using hashes.
`.into_grouping_map().collect<T>()` produces a `HashMap<T>`.

hugr-core/src/hugr/linking.rs
aborgna-q: Rolled back, and added some `expect`s.

hugr-core/src/hugr/patch/replace.rs
```diff
@@ -400,6 +400,10 @@ impl<HostNode: HugrNode> PatchHugrMut for Replacement<HostNode> {
         h.remove_node(inserted_entrypoint);
 
         // 6. Transfer to keys of `transfers` children of the corresponding values.
+        #[expect(
+            clippy::iter_over_hash_type,
```
acl-cqc, line 408: Of course, if the same *value* (old_parent) appeared twice in the map, we'd get completely different results - they'd get moved to the first new_parent (and then there'd to nothing to move to whichever new_parent came second)....

hugr-core/src/hugr/patch/replace.rs
```diff
@@ -409,6 +413,10 @@ impl<HostNode: HugrNode> PatchHugrMut for Replacement<HostNode> {
         }
 
         // 7. Remove remaining nodes
+        #[expect(
```
acl-cqc: Agreed this is harmless. But we could remove the `expect` by changing private `fn get_removed_nodes` to return a BTreeSet....

hugr-core/src/hugr/linking.rs
```diff
@@ -91,6 +91,10 @@ pub trait HugrLinking: HugrMut {
             roots.insert(other.entrypoint(), parent);
             other.set_parent(other.entrypoint(), other.module_root());
         };
+        #[expect(
```
acl-cqc, line 94: You could move the `roots.insert` out of the loop using `extend`, which avoids the lint (whether because it really knows that the keys must all be unique, or it just doesn't understand extend, I'm not sure..). But yeah, removing the lint requires `collect`ing all the children-whose-grandchildren-to-remove into a list, which again, makes me doubt that the linter is actually all that clever...

hugr-core/src/hugr/patch/replace.rs
aborgna-q: Ouch, good catch.

Repeated values should cause a hard error.
I added the check (previously it was only a `debug_assert`), and documented it in `Replacement::adoptions`.

hugr-core/src/hugr/linking.rs
acl-cqc: Have you? You're still going through `other.children`, I mean I guess it's not terrible (you'll copy them all when linking happens)...you decided not to just put children into a BTreeThing

hugr-core/src/hugr/patch/replace.rs
aborgna-q: done

PR #3056, opened 2026-05-12
feat: SiblingSubgraphs with polymorphic signatures

maximilianruesch:
In some situations (e.g. when removing tuple/untuple where at least one element contains a type argument), a SiblingSubgraph may need to represent itself using a polymorphic signature. This PR enables this, keeping the old `::signature` with a check that the type parameters are empty.

Related to: https://github.com/Quantinuum/tket2/issues/1486

### DIFF
--- a/hugr-core/src/hugr/patch/simple_replace.rs
+++ b/hugr-core/src/hugr/patch/simple_replace.rs
@@ -45,6 +45,11 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
 
     /// Create a new [`SimpleReplacement`] specification.
     ///
+    /// The given replacement should have an entrypoint that is a dataflow container (such as a DFG
+    /// or a function definition). If the entrypoint is a function, its (potentially polymorphic)
+    /// signature will be checked against the subgraph signature, otherwise the inner function type
+    /// of the entrypoint will be checked against the subgraph signature.
+    ///
     /// Return a [`InvalidReplacement::InvalidSignature`] error if `subgraph`
     /// and `replacement` have different signatures.
     pub fn try_new(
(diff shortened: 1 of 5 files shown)


hugr-core/src/hugr/patch/simple_replace.rs
```diff
@@ -45,25 +45,36 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
...
+            .or(Some(
+                replacement
+                    .inner_function_type()
+                    .unwrap()
+                    .into_owned()
+                    .into(),
+            ))
+            .ok_or(InvalidReplacement::InvalidDataflowGraph {
+                node: replacement.entrypoint(),
+                op: Box::new(replacement.get_optype(replacement.entrypoint()).to_owned()),
```
aborgna-q:
```suggestion
                op: Box::new(replacement.entrypoint_optype().to_owned()),
```
https://docs.rs/hugr/latest/hugr/hugr/trait.HugrView.html#method.entrypoint_optype

hugr-core/src/hugr/patch/simple_replace.rs
```diff
@@ -45,25 +45,36 @@ impl<HostNode: HugrNode> SimpleReplacement<HostNode> {
...
-                .inner_function_type()
-                .ok_or(InvalidReplacement::InvalidDataflowGraph {
-                    node: replacement.entrypoint(),
-                    op: Box::new(replacement.get_optype(replacement.entrypoint()).to_owned()),
-                })?;
-        if &subgraph_sig != repl_sig.as_ref() {
+        let subgraph_sig = subgraph.poly_func_type(host);
+        let repl_sig = replacement
+            .poly_func_type()
+            .or(Some(
```
aborgna-q: This should use `or_else`, otherwise we're constructing and unwrapping the `inner_function_type` even if we don't use it.

hugr-core/src/types/type_param.rs
```diff
@@ -453,6 +453,30 @@ impl Term {
...
     }
 
+    /// Checks whether the term is parametric, i.e. it (recursively) contains variables.
+    pub(crate) fn is_parametrized(&self) -> bool {
+        match self {
+            Term::RuntimeSum(SumType::General { rows }) => {
+                rows.iter().any(|row| row.is_parametrized())
+            }
+            Term::RuntimeSum(SumType::Unit { .. }) => false, // No leaves there
+            Term::RuntimeExtension(custy) => !custy.args().is_empty(),
```
aborgna-q: Should we recurse and look for variables instead?

Concrete extension types like `array_type(4, qb_t())` do not depend on external variables.

aborgna-q (approved):

PR #3057, opened 2026-05-12
feat: Metadata key aliases for non-breaking migrations

aborgna-q:
Adds a new `ALIASES` array to both the rust and python `Metadata` trait/protocol definitions.

When reading the metadata value for a node, we use the new aliases as alternative keys if the main one is not found.
This allows us to do backwards-compatible renames of the metadata keys.

### DIFF
--- a/hugr-core/src/hugr/views.rs
+++ b/hugr-core/src/hugr/views.rs
@@ -105,10 +105,14 @@ pub trait HugrView: HugrInternals {
 
     /// Returns the metadata associated with a node.
     ///
+    /// Looks up metadata under `M::KEY`, then under `M::ALIASES` in order.
+    ///
     /// For a non type-safe accessor use [`HugrView::get_metadata_any`] instead.
     #[inline]
     fn get_metadata<M: Metadata>(&self, node: Self::Node) -> Option<<M as Metadata>::Type<'_>> {
-        self.get_metadata_any(node, <M as Metadata>::KEY)
+        std::iter::once(<M as Metadata>::KEY)
+            .chain(<M as Metadata>::ALIASES.iter().copied())
+            .find_map(|key| self.get_metadata_any(node, key))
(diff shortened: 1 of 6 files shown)


hugr-core/src/hugr/views.rs
```diff
@@ -105,10 +105,14 @@ pub trait HugrView: HugrInternals {
 
     /// Returns the metadata associated with a node.
     ///
+    /// Looks up metadata under `M::KEY`, then under `M::ALIASES` in order.
```
maximilianruesch:
```suggestion
    /// Looks up metadata under `M::KEY`, then under entries of `M::ALIASES` in order.
```

hugr-core/src/metadata.rs
```diff
@@ -37,6 +37,12 @@ pub type RawMetadataValue = serde_json::Value;
 pub trait Metadata {
     /// Key associated with the metadata entry.
     const KEY: &'static str;
+    /// Other aliases of the metadata key.
+    ///
+    /// Typed metadata reads use these, in order, as fallbacks when [`Metadata::KEY`]
+    /// is not present. This is intended for backward compatibility when renaming
+    /// metadata keys.
```
maximilianruesch, line 44: We should perhaps explicitly mention that typed metadata writes do not use these. Same for Python

hugr-py/src/hugr/metadata.py
maximilianruesch, line 107: Add a small docstring explaining the fallback? Highly optional, but there is one on the Rust side.

maximilianruesch (approved):
Love it

hugr-py/src/hugr/metadata.py
aborgna-q: Added docstrings here and on the other class methods too

cqc-alec (approved):

PR #3058, opened 2026-05-13
feat!: Insert fewer order edges when inlining and update spec for order edges

acl-cqc:
**Currently** the algorithm for inserting Order edges is that any node linked via an order edge to a DFG being inlined, becomes linked to *every* input-successor/output-predecessor (by any edge type, not just order); and any node inside the DFG linked via an order edge to the input/output, becomes linked to *every* predecessor/successor (not just order) of the DFG. This embodies a couple of ideas in the spec:
1. that an Order edge enforces that *the predecessor finishes executing before the successor begins*...oh, except where the successor is a DFG containing nodes not linked from the Input (e.g. LoadConstant), which can happen earlier
2. that nodes "are necessarily evaluated, in some order respecting the Dataflow and Order edges" (i.e. a partial order). Note this essentially implies strict evaluation (node cannot execute until all inputs are ready), i.e. atomically with respect to their predecessors and successors. Except for LoadConstants again, of course

What we actually care about is Order edges enforcing ordering of *nodes with side effects*, so the algorithm above takes a very conservative interpretation of what might have side effects.

**Instead**, link DFG Order-predecessors only to Input Order-successors, and Output Order-predecessors to DFG Order-successors. This is exactly the way we treat other (Value) edges, modulo the multi in-port; and embodies the idea that nodes do not have to execute atomically even wrt. their inputs/outputs.

- [ ] TODO: if we inline a DFG with internal order edges (from Input/to Output) when there are no order edges to/from the DFG itself. Previously the internal Order-targets would be ordered after all value inputs to the DFG. Now they'll have no Order dependencies. What should we do? Connect these to the containing Input? Was the input DFG meaningful??

Is this breaking? Well...
* We will need to update RemoveRedundantOrderEdges pass, which will need to be updated not to touch order edges whose source/target is Input, Output, *or any container node or a Call*. 
* That done, Guppy programs should be ok, as they always have neat chains of order edges
* Hugrs constructed by other means may not and might be broken by this PR; I've had a look at InQuanto's, and they take a "flat circuit" function from pytket, (Qubit^N,Bool^M) -> (Qubit^N',Bool^M') and wire up N `QAlloc`s and M `Const(false)`s to the inputs, and N' QFree's and M' `result`s to the outputs. No order edges which looks like they'd be broken by DeadCodeElimination anyway...

Possible mitigations?
* Rename DFG and Call to WeaklyOrderedDFG and WeaklyOrderedCall at the same time ;-). Then everyone will know to look very carefully at what they're doing ;).
* Add metadata: `weakly_ordered=true` enables the new policy, false/absent keeps the old policy. InlineCall needs to merge metadata from Call and callee. Switch guppy to add the metadata. Nice that this gives a fallback, but the long-term strategy to move the metadata to default-true (and then eventually remove it) is unclear.
* Just hope that everyone is using guppy

BREAKING CHANGE: the implicit contract that DFGs, Calls, etc. execute atomically, is hereby broken; if you need that, add more Order edges, or barriers.


hugr-core/src/hugr/patch/inline_dfg.rs
```diff
@@ -307,32 +277,36 @@ mod test {
...
+        #[values(false, true)] o3: bool,
+        #[values(false, true)] o4: bool,
+    ) -> Result<(), Box<dyn std::error::Error>> {
         /*      -----|-----|-----
          *           |     |
          *          H_a   H_b
-         *           |.    /         NB. Order edge H_a to nested DFG
-         *           | .  |
+         *           |.   |?         NB. Order edge H_a to nested DFG
+         *           | .  |?         NB. Optional(o1) Order edge from H_b to nested DFG
```
acl-cqc, line 299: I am starting to think that perhaps unparametrizing the test - with a fixture returning the Hugr, and all 5 relevant nodes/ops, allowing individual tests to add extra order edges, apply the patch, and test what Order edges are present after, might be clearer. But, a lot longer, I think...

cqc-alec:
Interesting. It isn't clear to me that the existing behaviour is wrong. Doesn't an order edge mean that nothing in the target node should execute until everything in the source node has executed -- even if one or both of those nodes is a DFG?

Or am I missing the point?

acl-cqc:
> Interesting. It isn't clear to me that the existing behaviour is wrong. Doesn't an order edge mean that nothing in the target node should execute until everything in the source node has executed -- even if one or both of those nodes is a DFG?
> 
> Or am I missing the point?

Yes and no. Prior to this PR, an order edge indeed means "nothing in the target (of the Order edge) should execute until everything in the source". That is, every DFG includes an implicit order-barrier across the inputs and outputs (not quite a normal barrier: dependencies solely by value edges are not barriered). Note there is no way not to have that barrier!

Post this PR, there is no such barrier: Input/Output/DFG are transparent and just forward edges (the edge into the DFG feeds directly to the corresponding output edge(s) from the Input, only). If you want barriers, add them explicitly.

Which is why this is "breaking", but I'm not sure (need to make certain about) to what extent the spec actually says the former (and needs updating), or whether the spec is unclear (needs changing, i.e. this is a clarification), or whether this is actually a pure "fix"

cqc-alec:
> > Interesting. It isn't clear to me that the existing behaviour is wrong. Doesn't an order edge mean that nothing in the target node should execute until everything in the source node has executed -- even if one or both of those nodes is a DFG?
> > Or am I missing the point?
> 
> Yes and no. Prior to this PR, an order edge indeed means "nothing in the target (of the Order edge) should execute until everything in the source". That is, every DFG includes an implicit order-barrier across the inputs and outputs (not quite a normal barrier: dependencies solely by value edges are not barriered). Note there is no way not to have that barrier!
> 
> Post this PR, there is no such barrier: Input/Output/DFG are transparent and just forward edges (the edge into the DFG feeds directly to the corresponding output edge(s) from the Input, only). If you want barriers, add them explicitly.
> 
> Which is why this is "breaking", but I'm not sure (need to make certain about) to what extent the spec actually says the former (and needs updating), or whether the spec is unclear (needs changing, i.e. this is a clarification), or whether this is actually a pure "fix"

Either way, the spec does need updating to make it clear what an order edge to or from a container node means. I must confess I don't really know what it means if the container node is not atomic with respect to order. (Even if it wasn't made explicit I had always assumed this.)

specification/hugr.md
```diff
@@ -138,10 +138,20 @@ edges; see [operations](#node-operations).
 
 ### `Order` edges
 
-`Order` edges represent explicit constraints on ordering between nodes
-(e.g. useful for stateful operations). These can be seen as
-local value edges of unit type `()`, i.e. that pass no data, and where
-the source and target nodes must have the same parent. There can be at
+`Order` edges provide a way to express constraints on ordering between nodes
```
acl-cqc, line 141: I wonder if "express constraints on ordering" should be "to express hidden/stateful dependences" - they no longer mean the first node has to finish before the second node starts, unless the nodes are atomic....thoughts?

specification/hugr.md
```diff
@@ -717,6 +720,30 @@ flowchart
     linkStyle 12,13,14,15,16,17 stroke:#ff3,stroke-width:4px;
 ```
 
+(evaluation-semantics)=
+## Evaluation Semantics
+
+In a dataflow graph, the evaluation semantics are simple: all nodes in the
```
acl-cqc: This moved from its previous location which seemed a bit out-of-place, and significantly expanded, but I kept the original definition of all nodes being evaluated; I am tempted to say "at least those nodes required to produce all inputs to the Output node must be evaluated; more may be at the discretion of the implementation", although this is really targetted at (a) Tierkreis (b) removing nonterminating computations whose results are unused; thoughts?

Moreover, our current implementation of dead-code elimination transforms the hugr to another that *does not conform* to these evaluation semantics (removed nodes are not evaluated!); should we say something like "evaluating a dataflow graph behaves as if all nodes were evaluated"?

hugr-core/src/hugr/patch/inline_dfg.rs
```diff
@@ -73,82 +73,46 @@ impl<N: HugrNode> PatchHugrMut for InlineDFG<N> {
...
+
+        // DFG Inputs.
         for inp in h.node_inputs(n).collect::<Vec<_>>() {
-            if inp == oth_in {
-                continue;
-            }
-            // Hugr is invalid if there is no output linked to the DFG input.
-            let (src_n, src_p) = h.single_linked_output(n, inp).unwrap();
+            let dfg_preds = h.linked_outputs(n, inp).collect::<Vec<_>>();
+            assert!(inp == oth_in || dfg_preds.len() == 1); // Any number of order preds
```
doug-q:
```suggestion
            debug_assert!(inp == oth_in || dfg_preds.len() == 1); // Any number of order preds
```

hugr-core/src/hugr/patch/inline_dfg.rs
```diff
@@ -73,82 +73,46 @@ impl<N: HugrNode> PatchHugrMut for InlineDFG<N> {
...
             h.disconnect(n, inp); // These disconnects allow permutations to work trivially.
             let outp = OutgoingPort::from(inp.index());
-            let targets = h.linked_inputs(input, outp).collect::<Vec<_>>();
+            let mut targets = h.linked_inputs(input, outp).collect::<Vec<_>>();
             h.disconnect(input, outp);
+            if inp == oth_in && !internal_order_path {
```
doug-q: This seems dodgy. It would be very nice to do the same thing here independently of whether there is an "internal_order_path". Said property seems pretty ad-hoc.

I don't think we should be treating "has internal order path from input->output" differently to "has any path from input->output". We are planning to replace order edges with others and we don't want this branch to change right?

Given that we should not treat order edges differently, consider a DFG with 1 input directly forwarded to output, and some other ops, which are not connected to input.

Then this branch executes on said DFG, but deleting the forwarded input/output causes this branch to not execute

acl-cqc: Interesting, thanks.

> It would be very nice to do the same thing here independently of whether there is an "internal_order_path".

I.e. connect all Order-predecessors of the DFG to the Order-successors of the DFG? Yes, we can do that. If there is an order path, those edges will be redundant (essentially you'll get A->B->C and also A->C), in fact that was what we were doing before, so no strong objection. However I'll note....

> Said property seems pretty ad-hoc.

Well, what we need to guarantee is that if we have A->(DFG)->B along Order edges before inlining, then we necessarily have *some* path from A to B after inlining. So I think this is the minimal edges we need to add, but we could add more.

The other point is that when we have *linear* token types, those will guarantee there is a path; this is only to handle some odd cases e.g. DFGs that ignore their inputs and return a LoadConstant. (Essentially the rule in this PR says that if there is no order path previously, pretend there is an order edge from Input to Output; the old rule says, *always* pretend there was an order edge from Input to Output.)

> Given that we should not treat order edges differently, consider a DFG with 1 input directly forwarded to output, and some other ops, which are not connected to input.
>
> Then this branch executes on said DFG, but deleting the forwarded input/output causes this branch to not execute

Not quite. This only looks for *order* paths, not general paths. Does that change anything?

acl-cqc:
@cqc-alec wrote:
> Either way, the spec does need updating to make it clear what an order edge to or from a container node means.
> I must confess I don't really know what it means if the container node is not atomic with respect to order.
> (Even if it wasn't made explicit I had always assumed this.)

I've now updated the wording in the spec quite a bit, there was and is nothing specific to DFGs or containers, so I could add something - can you have another read and see if you still think this is necessary? (If so - in the new evaluation semantics section, or spread among DFG/Conditional/CFG/etc. nodes?)

We'd be saying something that amounts to: containers (conditonal/CFG/TailLoop) treat their Order ports linearly (the value fed into the TailLoop's order inport / its Output node's order inport, is forwarded to *either* the TailLoop's Input node's order outport *or* the TailLoop's order outport, and only when the predicate is known, sort of thing). This amounts to saying, nodes with incoming Order edges cannot be "speculatively executed" and others can, which I think is reasonable.

hugr-core/src/hugr/patch/inline_dfg.rs
doug-q: What you say makes sense but I'm still struggling to understand (it's complicated!).

I think this inlining is subtle and fragile, so I would prefer to be conservative in changing it. I would prefer that we accept adding more order edges than necessary here in exchange for simpler semantics. The separate "remove excessive order edges" should be relied upon to cleanup.

> Well, what we need to guarantee is that if we have A->(DFG)->B along Order edges before inlining, then we necessarily have some path from A to B after inlining. So I think this is the minimal edges we need to add, but we could add more.
> Well, what we need to guarantee is that if we have A->(DFG)->B along Order edges before inlining, then we necessarily have some path from A to B after inlining. So I think this is the minimal edges we need to add, but we could add more.

This is very clear. Another way to accomplish this is to unconditionally add a `Noop` with order connections to A and B.

> The other point is that when we have linear token types, those will guarantee there is a path; this is only to handle some odd cases e.g. DFGs that ignore their inputs and return a LoadConstant. (Essentially the rule in this PR says that if there is no order path previously, pretend there is an order edge from Input to Output; the old rule says, always pretend there was an order edge from Input to Output.)

Yes these "odd" cases are the interesting ones here. That it would be nice(simpler) to be able to treat them uniformly is the point. "Always pretend there is an order edge from Input to Output" seems like the simplest thing to do?

> Not quite. This only looks for order paths, not general paths. Does that change anything?

I don't understand why we need treat order paths differently.

specification/hugr.md
```diff
@@ -717,6 +720,30 @@ flowchart
...
+graph are necessarily evaluated. (For hierarchical nodes, this may entail
+evaluating some children, according to the node.) Evaluation may occur in
+some order, and for nodes that execute strictly and atomically (according to
+implementation) this will respect the Dataflow and Order edges. However *in
```
cqc-alec:
```suggestion
evaluating some children, according to the node.) For nodes that execute strictly and atomically (according to
implementation) the order of execution will respect the Dataflow and Order edges. However *in
```
Is this clearer, or am I missing a subtlety?

cqc-alec:
> @cqc-alec wrote:
> 
> > Either way, the spec does need updating to make it clear what an order edge to or from a container node means.
> > I must confess I don't really know what it means if the container node is not atomic with respect to order.
> > (Even if it wasn't made explicit I had always assumed this.)
> 
> I've now updated the wording in the spec quite a bit, there was and is nothing specific to DFGs or containers, so I could add something - can you have another read and see if you still think this is necessary? (If so - in the new evaluation semantics section, or spread among DFG/Conditional/CFG/etc. nodes?)
> 
> We'd be saying something that amounts to: containers (conditonal/CFG/TailLoop) treat their Order ports linearly (the value fed into the TailLoop's order inport / its Output node's order inport, is forwarded to _either_ the TailLoop's Input node's order outport _or_ the TailLoop's order outport, and only when the predicate is known, sort of thing). This amounts to saying, nodes with incoming Order edges cannot be "speculatively executed" and others can, which I think is reasonable.

I think this is fine with the spec changes, thanks.

hugr-core/src/hugr/patch/inline_dfg.rs
acl-cqc: > I think this inlining is subtle and fragile, so I would prefer to be conservative in changing it.
> I would prefer that we accept adding more order edges than necessary here in exchange for simpler semantics.
> The separate "remove excessive order edges" should be relied upon to cleanup.

Yes, well, the adding extra edges direct from the DFG's Order-predecessors to its Order-successors, was added in #3072 as an essential bugfix. The check that we only add said if there is no Order path from Input to Output (i.e. if the extra edges are not redundant) stems from that PR, is not new here. So I've broken out https://github.com/Quantinuum/hugr/pull/3096 to do just that (plus extend the tests), that can go in first and it'll make the "diff" much clearer here.

> I don't understand why we need to treat Order paths differently

Ok, so: consider
A -> DFG -> B by Order edges
(some value inputs/outputs to/from DFG, maybe from/to A/B, maybe not)
(DFG Input), (Constant Zero) -> multiply -> (DFG Output) by Value edges, no Order edges

If we inline but don't add extra edges, we lose the Order path from A->B. The two may be connected by value edges (depending on the DFG's value inputs/outputs)...until we rewrite multiply-by-zero to constant zero, whereupon A and B are completely disconnected!

(With linear token types of course this wouldn't apply, the token would have to be passed from DFG Input to DFG Output. And your suggestion of *always* adding A->B works too. But in terms of how often you add the extra Order edges, "always" > "only if no order path" != "only if no value path" > "only if no path", a diamond lattice, and we need either "always" or "only if no order path")

hugr-core/src/hugr/patch/inline_dfg.rs
acl-cqc: Test was updated in #3096 and I think this shows the reduction in edges vs. that PR quite well now.

hugr-core/src/hugr/patch/inline_dfg.rs
acl-cqc: Done the change to always add the edges bypassing the DFG, even if there was an internal order path, in #3096 - that makes this PR a bit simpler and the changes clearer

acl-cqc:
Just tried this on Chloe's example, we handcooked/fudged a bit (running NormalizeGuppy without inline-dfgs), to get the attached hugr with 15 order edges (no one having >1 incoming).

Without this PR, tket2 InlineDFGsPass (i.e. the InlineDFG patch here, on every dfg) ballooned this to 2084 order edges, with some node(s) having over 300 incoming edges.

With this PR, after InlineDFGsPass we get total of 12 order edges; some node(s) has/have 2 incoming order edges, but none any more than that. (This probably results from the adding of edges "around" the DFG as well as through the contents.)

Hoping we can test Chloe's resynth pass on the output but there are a couple of other updates (no relation to this) needed there first.

specification/hugr.md
```diff
@@ -717,6 +724,33 @@ flowchart
...
+  this means that side-effects will occur in a partial/total order respecting
+  the `Order` edges, but the `Order` edges do not restrict pure functional
+  computation (that does not depend on any global state or have any side effects).
+
+This applies both to extension ops and core constructs such as DFG, CFG, or a
+Call to a function, but may be refined (introducing more precise guarantees
+such as atomicity or strictness) for specific ops and/or by specific
+implementations.
+
+For ops such as `Conditional`, `TailLoop` and `CFG`/`BasicBlock`, the order input to the container node, or to any `Output` node within, must be treated linearly: (only) after all Order-predecessors have (finished mutating state/produced their Order outputs), may the Order output (of exactly one `Input` node or the container itself) be ready - chosen according to the appropriate predicate. **Thus**, nodes reachable along Order edges from an `Input` may not be speculatively evaluated (perform any stateful update); but nodes which are not `Order`-reachable may be.
```
cqc-alec:
```suggestion
For ops such as `Conditional`, `TailLoop` and `CFG`/`BasicBlock`, the order input to the container node, or to any `Output` node within, must be treated linearly: only after all Order-predecessors have finished mutating state and produced their Order outputs may the Order output (of exactly one `Input` node or the container itself) be ready - chosen according to the appropriate predicate. Thus, nodes reachable along Order edges from an `Input` may not be speculatively evaluated (perform any stateful update); but nodes which are not Order-reachable may be.
```

specification/hugr.md
cqc-alec: Yes, if we say something like "evaluating a dataflow graph behaves as if all nodes were evaluated", this covers all bases I think?

specification/hugr.md
```diff
@@ -717,6 +724,33 @@ flowchart
...
+
+In a dataflow graph, the evaluation semantics are simple: all nodes in the
+graph are necessarily evaluated. (For hierarchical nodes, this may entail
+evaluating some children, according to the node.) For nodes that execute
+strictly (according to implementation) this will happen in an order that
+respects the Dataflow and Order edges; this will be a total order if the
+nodes execute atomically. However *in general* Hugr allows that
+* Nodes may execute non-atomically, perhaps in parallel/overlapping other nodes
+* Nodes may execute partially or totally before all inputs are ready (if the
+  op is able to "do" anything without all its inputs). **However** note that
```
cqc-alec:
```suggestion
  op is able to "do" anything without all its inputs). But note that
```
Just a style thing (avoiding nested "howevers").

cqc-alec (approved):

specification/hugr.md
```diff
@@ -717,6 +724,33 @@ flowchart
...
 ```
 
+(evaluation-semantics)=
+## Evaluation Semantics
+
+In a dataflow graph, the evaluation semantics are simple: all nodes in the
+graph are necessarily evaluated. (For hierarchical nodes, this may entail
+evaluating some children, according to the node.) For nodes that execute
+strictly (according to implementation) this will happen in an order that
+respects the Dataflow and Order edges; this will be a total order if the
```
cqc-alec: Can you clarify what this means? In particular, "For nodes that execute strictly (according to implemention)".

cqc-alec: (See also https://github.com/Quantinuum/tket2/discussions/1649 .)

cqc-alec (approved):
I feel we may have not quite nailed down the execution semantics, but don't want to hold up merging the code change.

specification/hugr.md
acl-cqc: Yes, much better, thanks

specification/hugr.md
acl-cqc: I did a bit more bracket removal and cleanup around here and ended up with this being the only However. Kept the emphasis though as it's not a consequence but its own rule.

specification/hugr.md
acl-cqc: There were two things I was trying to get at:
* For nodes that execute strictly, they execute in an order respecting the edges; it's a total order if they execute atomically (i.e. no overlapping).
   * ....I guess if non-atomic then it might be a partial order, but then, if nonstrict you could say a partial order too, and both the non-total cases are stretching the meaning of a "partial order" - the partial ordering where nothing is before or after anything else is still partial but pretty useless - so I didn't mention that
* Both strictness and also atomicity depend on the specific toolchain or implementation

In the end I expanded, rephrased, deleted some repetition, and added a footnote clarifying "strict", so this did change a bit (again), but only in presentation not meaning, promise ;-)

acl-cqc: Ok removed the business about total order, it's not really helpful, and may just confuse. The key point is that without strictness, evaluation order doesn't even need to respect the edges.

PR #3060, opened 2026-05-13
fix: Export dataflow blocks with CFG node link merged into input port when possible

### LINKED ISSUE #3059: [Bug]: PanicException: assertion failed: links.len() <= optype.port_count(direction)
### Description

This exception is raised when we try to normalise the following program:
```python
@guppy(unitary=True)
def bar(q: qubit) -> None:
    x(q)


@guppy
def main() -> None:
    controllers: array[qubit, 3] = array(qubit(), qubit(), qubit())
    t = qubit()

    h(controllers[0])
    h(controllers[1])
    h(controllers[2])

    with control(controllers):
        bar(t)

    state_result("r", controllers[0], controllers[1], controllers[2], t)

    discard_array(controllers)
    discard(t)
```

This failure occurs only when we normalise programs with an array, so it is probably due to that.

maximilianruesch:
Fixes a bug in the Python HUGR model export, where a CFG entry block is part of a loop and thus already has an input port. Adding a "virtual link" from the CFG parent to the entry block adds a second port. Rust expects the entry block to have such an input port regardless of whether it is used, but also validates during import that the number of incoming links is not greater than the port count (which is 1).

The Python export is now fixed to always use the first port (mirroring the Rust side) for the CFG-entry link. When there is no first port (because in Python (different to Rust), no port is inserted if the entry block has no predecessors), a new port is created along with a link from the CFG node.

As a driveby, fixes a missing check in the CI required checks, and improves the debug assertion message to also report the failed node.

Closes #3059


aborgna-q (approved):
:shipit:

PR #3063, opened 2026-05-14
feat!: Add extension version to custom ops and types

aborgna-q:
First step towards #2996. Closes #2999.

Adds `version_extension` fields to (rust and python) custom ops and opaque types, as well as rust typedefs and opdefs. Python defs do not require the version field since they always have access to their extension.

For the operation definitions, the fields are `Option<Version>`s so we're able to deserialize them from JSON. The field then gets filled in immediately after loading the extension. We should be able to remove the `Option` once json is dropped.

For ops and types, we use `Option<Version>`s to be backwards compatible with older serialized hugrs.
The public API requires a valid version, only deserialization may leave the field empty.

This PR only adds the field. We don't store the versions yet in the hugr model.

BREAKING CHANGE: `OpaqueOp::new` and `CustomType::new` now require an extension version.

Not a breaking change for python, so we override the breaking flag in the title for release-please:
BEGIN_COMMIT_OVERRIDE
feat: Add extension version to custom ops and types (#3063)
END_COMMIT_OVERRIDE


aborgna-q:
Benchmark changes only affect json definitions, due to the extra optional field deserialization.
We can safely ignore the regression since JSON encoding is only kept for backwards compat.

hugr-llvm/src/extension/collections/snapshots/hugr_llvm__extension__collections__static_array__test__emit_static_array_of_static_array@llvm21.snap
```diff
@@ -5,24 +5,24 @@ expression: mod_str
 ; ModuleID = 'test_context'
 source_filename = "test_context"
 
-@sa.inner.ac73413c.0 = constant { i64, [0 x i64] } zeroinitializer
```
mark-koch, line 8: These changes are quite surprising to me, any idea why the LLVM snapshots should be affected?

hugr-py/src/hugr/_serialization/ops.py
```diff
@@ -6,6 +6,7 @@
 from typing import Any, Literal
 
 from pydantic import ConfigDict, Field, RootModel
+from pydantic_extra_types.semantic_version import SemanticVersion  # noqa: TCH002
```
mark-koch: I assume the non-typing import is to make pydantic validation work?

hugr-py/src/hugr/_serialization/tys.py
```diff
@@ -51,6 +54,16 @@ def _json_custom_error_validator(
 ExtensionId = str
 ExtensionSet = list[ExtensionId]
 
+
+def to_semantic_version(version: Version | None) -> SemanticVersion | None:
+    """Convert an application semver value to the pydantic serialization adapter."""
+    if version is None:
+        return None
+    if isinstance(version, SemanticVersion):
```
mark-koch, line 64: So should the type annotation be `Version | SemanticVersion | None`?

hugr-core/src/std_extensions/arithmetic/float_types.rs
```diff
@@ -26,6 +26,7 @@ pub fn float64_custom_type(extension_ref: &Weak<Extension>) -> CustomType {
         FLOAT_TYPE_ID,
         vec![],
         EXTENSION_ID,
+        VERSION,
```
mark-koch, line 29: Should this version also be bumped?

hugr-core/src/std_extensions/arithmetic/int_types.rs
```diff
@@ -33,6 +33,7 @@ pub fn int_custom_type(
         INT_TYPE_ID,
         [width_arg.into()],
         EXTENSION_ID,
+        VERSION,
```
mark-koch, line 36: Same here?

mark-koch (approved):

hugr-core/src/std_extensions/arithmetic/float_types.rs
aborgna-q: The float type and int type haven't changed, because they do not use other types in their definition so their json is exactly the same as before.

hugr-py/src/hugr/_serialization/ops.py
aborgna-q: Yep, added a comment

hugr-py/src/hugr/_serialization/tys.py
aborgna-q: Yeah. We don't really need this branch, but It doesn't hurt 🤷

hugr-llvm/src/extension/collections/snapshots/hugr_llvm__extension__collections__static_array__test__emit_static_array_of_static_array@llvm21.snap
aborgna-q: Those ids are derived from type hashes, which now have the extra version field.

PR #3064, opened 2026-05-15
feat!: Keep multiple incompatible extension versions in ExtensionRegistry

aborgna-q:
Part of #2996. Closes #3000. Followup to #3063.

#3005 added support for multiple versions for the same extension in `ExtensionRegistry`. This PR implements the same for the rust side.

drive-by: Added missing extensions in `hugr-cli gen-extensions` test, and ensured the test won't get outdated again.

BREAKING CHANGE: `ExtensionRegistry::register` now returns an insertion result instead of failing on existing entries. Removed `ExtensionRegistry::register_updated` 
BREAKING CHANGE: `hugr-cli gen-extensions` now adds the extension version as suffix to the json files. Use `--unversioned` to get the old behaviour.

Added some python helper methods to rust the new ones in rust, but this is not a breaking change there.
BEGIN_COMMIT_OVERRIDE
feat: Keep multiple incompatible extension versions in ExtensionRegistry (#3064)
END_COMMIT_OVERRIDE

### DIFF
--- a/hugr-cli/src/extensions.rs
+++ b/hugr-cli/src/extensions.rs
@@ -1,6 +1,7 @@
 //! Dump standard extensions in serialized form.
 use anyhow::Result;
 use clap::Parser;
+use hugr::Extension;
 use hugr::extension::ExtensionRegistry;
 use std::{io::Write, path::PathBuf};
 
@@ -20,32 +21,66 @@ pub struct ExtArgs {
         help = "Output directory."
     )]
     pub outdir: PathBuf,
+    /// Add the extension version to the output filename, e.g. "foo-1.2.3.json".
+    #[arg(
+        long,
+        help = "Add the extension version to the output filename.",
+        default_value_t = false
+    )]
+    pub versioned: bool,
 }
 
 impl ExtArgs {
     /// Write out the standard extensions in serialized form.
+    ///
     /// Qualified names of extensions used to generate directories under the specified output directory.
-    /// E.g. extension "foo.bar.baz" will be written to "OUTPUT/foo/bar/baz.json".
+    /// The extension version is added to the json filename.
+    ///
+    /// E.g. extension "foo.bar.baz" will be written to "OUTPUT/foo/bar/baz-1.2.3.json".
     pub fn run_dump(&self, registry: &ExtensionRegistry) -> Result<()> {
-        let base_dir = &self.outdir;
-
-        for ext in registry {
-            let mut path = base_dir.clone();
-            for part in ext.name().split('.') {
-                path.push(part);
-            }
-            path.set_extension("json");
-
-            std::fs::create_dir_all(path.clone().parent().unwrap())?;
-            // file buffer
-            let mut file = std::fs::File::create(&path)?;
-
-            serde_json::to_writer_pretty(&mut file, &ext)?;
-
-            // write newline, for pre-commit end of file check that edits the file to
-            // add newlines if missing.
-            file.write_all(b"\n")?;
+        if self.versioned {
+            registry
+                .iter_all()
+                .try_for_each(|ext| self.dump_extension(ext))?;
+        } else {
+            // The STD registry should only contain one version of each
+            // extension, so we can just take the latest version for each
+            // extension name and avoid overriding files.
+            registry
+                .iter()
+                .map(|ext| ext.latest())
+                .try_for_each(|ext| self.dump_extension(ext))?;
         }
+
+        Ok(())
+    }
+
+    fn dump_extension(&self, ext: &Extension) -> Result<()> {
+        let mut path = self.outdir.clone();
+        let mut parts = ext.name().split('.').peekable();
+        while let Some(part) = parts.next() {
+            if parts.peek().is_some() {
+                path.push(part);
+                continue;
+            }
+
+            // Optionally add the extension version to the output filename, e.g. "foo-1.2.3.json".
+            if self.versioned {
+                path.push(format!("{part}-{}.json", ext.version()));
+            } else {
+                path.push(format!("{part}.json"));
+            }
+        }
+
+        std::fs::create_dir_all(path.clone().parent().unwrap())?;
+        // file buffer
+        let mut file = std::fs::File::create(&path)?;
+
+        serde_json::to_writer_pretty(&mut file, &ext)?;
+
+        // write newline, for pre-commit end of file check that edits the file to
+        // add newlines if missing.
+        file.write_all(b"\n")?;
 
         Ok(())
     }
--- a/hugr-cli/src/extensions.rs
+++ b/hugr-cli/src/extensions.rs
@@ -21,13 +21,17 @@ pub struct ExtArgs {
         help = "Output directory."
     )]
     pub outdir: PathBuf,
-    /// Add the extension version to the output filename, e.g. "foo-1.2.3.json".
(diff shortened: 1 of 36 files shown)


hugr-cli/tests/gen_extensions.rs
```diff
@@ -33,7 +33,42 @@ fn test_extension_dump(mut cmd: Command) {
         "arithmetic/float.json",
         "arithmetic/conversions.json",
         "collections/array.json",
+        "collections/borrow_arr.json",
         "collections/list.json",
+        "collections/static_array.json",
+    ];
```
doug-q: Is it worth asserting that these are the only files?

hugr-cli/src/extensions.rs
```diff
@@ -20,32 +21,66 @@ pub struct ExtArgs {
         help = "Output directory."
     )]
     pub outdir: PathBuf,
+    /// Add the extension version to the output filename, e.g. "foo-1.2.3.json".
+    #[arg(
+        long,
+        help = "Add the extension version to the output filename.",
+        default_value_t = false
```
doug-q, line 32: Why default false? To be nonbreaking? I would expect we would just always want this

hugr-cli/src/extensions.rs
```diff
@@ -20,32 +21,66 @@ pub struct ExtArgs {
...
+    ///
+    /// E.g. extension "foo.bar.baz" will be written to "OUTPUT/foo/bar/baz-1.2.3.json".
     pub fn run_dump(&self, registry: &ExtensionRegistry) -> Result<()> {
-        let base_dir = &self.outdir;
+        if self.versioned {
+            registry
+                .iter_all()
+                .try_for_each(|ext| self.dump_extension(ext))?;
+        } else {
+            // The STD registry should only contain one version of each
```
doug-q: This is OK, but it seems fragile to this policy changing in the future

hugr-cli/tests/gen_extensions.rs
```diff
@@ -33,7 +33,42 @@ fn test_extension_dump(mut cmd: Command) {
...
+    // check all paths exist
```
doug-q: should we assert that no other paths exist? It looks like we forgot to add two extensions, and you are fixing that.

hugr-cli/tests/gen_extensions.rs
```diff
@@ -33,7 +33,42 @@ fn test_extension_dump(mut cmd: Command) {
...
+fn test_extension_versioned_dump(mut cmd: Command) {
+    let temp_dir = assert_fs::TempDir::new()
+        .expect("temp dir creation failure.")
+        .into_persistent_if(std::env::var_os("HUGR_CLI_TEST_PERSIST_FILES").is_some());
+    cmd.arg("-o");
+    cmd.arg(temp_dir.path());
+    cmd.arg("--versioned");
+    cmd.assert().success();
+
+    let expected_paths = [
```
doug-q: This will break whenever we bump versions or add extensions. Is there a better way? We could generate these from the `STD` global.

hugr-core/src/extension/resolution/weak_registry.rs
```diff
@@ -17,7 +17,94 @@ use crate::extension::{ExtensionId, ExtensionRegistry};
...
+pub struct WeakExtensionVersions {
+    id: ExtensionId,
+    versions: BTreeMap<Version, Weak<Extension>>,
+}
+
+impl WeakExtensionVersions {
+    fn new(id: ExtensionId, ext: Weak<Extension>) -> Self {
+        let version = ext
+            .upgrade()
+            .expect("weak extension must be upgradeable when registered")
```
doug-q: Nit: I was confused by this, I thought you were talking about upgrading extensions not upgrading Arcs

doug-q: If you are going to panic if the Weak fails to upgrade, you should make the caller upgrade by taking an Arc

hugr-core/src/extension.rs
```diff
@@ -177,12 +178,187 @@ impl Clone for ExtensionRegistry {
...
+    ///
+    /// If `requested` is `Some(version)`, returns the highest retained
+    /// extension compatible with `version`. If `requested` is `None`, returns
+    /// the latest retained extension.
+    ///
+    /// This is used when deserializing older HUGRs that may not have a version
+    /// specified for their extensions.
+    #[must_use]
+    pub(crate) fn get_req(&self, requested: Option<&Version>) -> Option<&Arc<Extension>> {
+        requested
```
doug-q: I think this is complicated enough that it's worth adding an `assert` or `debug_assert` that the returned extension has a compatible version

hugr-core/src/extension.rs
```diff
@@ -177,12 +178,187 @@ impl Clone for ExtensionRegistry {
...
+                version: compatible_version,
+            }
+        } else {
+            ExtensionRegisterResult::ReplacedExact
+        }
+    }
+
+    // Find a registered extension version that is in the same semver
+    // compatibility group as `version`, if any.
+    fn compatible_registered_version(&self, version: &Version) -> Option<Version> {
```
doug-q, line 291: I would expect this to not return `1.0.0` as compatible with `1.0.1`. We can substitute a newer minor version, but not an older one.

I don't claim anything is broken (need to check use sites), but this seems like a bit of a landmine

doug-q: No I'm wrong, this does what I said it should do. But the comment is a bit wrong. It's not semver compatibility group, it's " a registered extension version that is semver-compatibile with `version`

doug-q (approved):
Looks good, just some nits and questions

hugr-cli/tests/gen_extensions.rs
aborgna-q: Yeah, changed the test to compare set of existing vs expected files instead 👍

hugr-cli/tests/gen_extensions.rs
aborgna-q: Yes. See my other comment

hugr-cli/src/extensions.rs
aborgna-q: Yes, it was to keep the current behaviour.

I guess we can make it a breaking change there too.
`tket-exts` uses this to export a dump with all the extensions.

But now with multi-version support we may even want to export more than one instance for the same tket extension 🤔 

I'll change the flag to `unversioned`, so it gets versioned by default.

hugr-cli/src/extensions.rs
aborgna-q: Updated the comment. Since we're doing versioned-by-default outputs, writing only the latest versions here when `--unversioned` is passed makes more sense.

hugr-cli/tests/gen_extensions.rs
aborgna-q: Merged into the other test, and replaced the explicit lists with one generated from the std registry

hugr-core/src/extension.rs
aborgna-q: Yes, I expanded the method docs to clarify this.

hugr-core/src/extension.rs
aborgna-q: I cleared up the implementation using a `match` instead. Hopefully now it's straightforward

PR #3066, opened 2026-05-19
fix: Publish parametrisation check for types

maximilianruesch:
It seems that in #3056, I forgot to make this function public, and I need it to be public for work in https://github.com/Quantinuum/tket2/pull/1609.


aborgna-q (approved):
-.-'

PR #3070, opened 2026-05-20
feat(py): Linking packages serialised as bytes

maximilianruesch:
Offers a `hugr-py` utility to link packages serialised as bytes, in particular without deserialising them. This offers massive performance improvements, especially when throwing the result directly into Selene, as serialising and deserialising in Python is very slow atm.

Example trace of a client program before the change:
image

After the change:
image


hugr-py/tests/test_linking.py
cqc-alec, line 1: Can we add a test for `link_packages()` at the python level?

hugr-py/rust/linking.rs
```diff
@@ -1,18 +1,53 @@
...
+#[derive(derive_more::Display, Debug)]
+enum LinkError {
+    #[display("Could not link modules: {_0}")]
+    LinkError(String),
+    #[display("Cannot link two executable modules together.")]
+    BothExecutable,
+}
+
+fn link(hugr_into: &mut Hugr, hugr_from: Hugr) -> Result<(), LinkError> {
+    let into_executable = hugr_into.entrypoint() != hugr_into.module_root();
```
cqc-alec: Aside: Maybe we could do with a `Hugr.is_executable()` method.

cqc-alec (commented):
Nice.

hugr-py/tests/test_linking.py
maximilianruesch: This would be a very similar set of tests as `Package.link`, since that is implemented using `link_packages` (and serialisation). Do you want to add more? Would be fair enough to do so, just gathering some thoughts.

hugr-py/rust/linking.rs
maximilianruesch: Not sure whether we want to spread this "executable" wording far, since any hugr is executable if you know the function name. Thoughts? I ma also change this to "into_non_root" or similar.

hugr-py/tests/test_linking.py
cqc-alec: Fair enough, maybe just a simple test that the API function works.

hugr-py/rust/linking.rs
cqc-alec: Good point, let's not add this now but maybe use something like `non_root_entrypoint` here for clarity.

hugr-py/tests/test_linking.py
maximilianruesch: Added basic tests (my inner monk made me add three instead of one). I also changed the signature of `link_packages` to be variadic, if you think that is worse lmk.

hugr-py/tests/test_linking.py
```diff
@@ -81,6 +82,48 @@ def test_link_modules_linking_error():
...
+    pkg1 = Package(
+        modules=[build_module(entrypoint=False), build_module(entrypoint=True)]
+    )
+    pkg2 = Package(
+        modules=[build_module(entrypoint=True), build_module(entrypoint=False)]
+    )
+
+    with pytest.raises(
+        HugrLinkingError,
+        match="Could not link module 2: Cannot link two modules with non-root entrypoints together",  # noqa: E501
```
cqc-alec, line 122: Hmm, shouldn't this be module 1 (assuming 0-based counting)? Bit confused why we see 2 here.

maximilianruesch: Because a. we skip the first module and take it as the one to merge all others into and b. the module at index 1 is successfully linked in (because the module at index 0 does not already have an entrypoint), but it is only when the module at index 2 comes in (which also has a non-root entrypoint) that things go wrong.

cqc-alec: Ah, I was mixing up modules and packages, and thought it was complaining about package 2. OK, makes sense.

cqc-alec (approved):

PR #3072, opened 2026-05-20
fix: Preserve ordering in InlineDFG

acl-cqc:
A corner case when inlining a DFG that's disconnected (no path from Input to Output, e.g. that just loads a constant) could lead to the chain/web of order edges going through the DFG being cut where the DFG was.

Fix: if there is no order path from Input to Output, then explicitly connect order-predecessors of the DFG to order-successors (we might use other criteria for the "if", but this seems a reasonable one).

### DIFF
--- a/hugr-core/src/hugr/patch/inline_dfg.rs
+++ b/hugr-core/src/hugr/patch/inline_dfg.rs
@@ -158,6 +158,9 @@ impl<N: HugrNode> PatchHugrMut for InlineDFG<N> {
     }
 }
 
+/// Determines whether there is a path along [Order] edges only from `src` to `tgt`.
+///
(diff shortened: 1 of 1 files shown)


hugr-core/src/hugr/patch/inline_dfg.rs
```diff
@@ -144,6 +158,21 @@ impl<N: HugrNode> PatchHugrMut for InlineDFG<N> {
     }
 }
 
+fn is_order_reachable<H: HugrView>(h: &H, src: H::Node, tgt: H::Node) -> bool {
```
aborgna-q, line 164: Pls add a docstring.

aborgna-q (approved):

PR #3080, opened 2026-05-29
feat!: Add extension versions to hugr-model elements

### LINKED ISSUE #3077: Encode ext versions for ops and types in hugr-model
Suppose a user has a HUGR that contains

- Operation `opA` from `extA@0.2.3`

- Operation `opB` from `extB@0.1.0` that depends on a custom type from `extA@0.2.3`.

The extensions are managed by different projects, and the user wants to migrate to newer versions of both extensions:

- Project A releases `extA@0.3.0` that is not semver-compatible with `0.2.3`

- Project B releases `extB@0.2.0` that depends on `extA@0.3.0`.

**What's the workflow for the user to migrate their HUGR to use the newer versions of the extensions?**

We propose a few requirements for this workflow:

Each project defines a migration procedure that upgrades the specific extension operations.

- A Hugr may contain multiple versions of the same extension operation while the migration is being run.

- A Hugr may contain multiple versions of the same extension type between migration steps (e.g. between migrating extA and extB above).

- Migration can be either Rust or Python code. We must assume there may be serialization steps between partially migrated HUGRs.

The runtime HUGR representations added support for co-existing op and type versions in #3063 and #3064, so the remaining question is how to support multiple versions on the wire format.

This document proposes a design for supporting multiple extension versions in the serialization formats.

## Current Encoding

Currently encoded HUGRs can define custom operations and types that are linked to some extension ID, but do not carry information about the actual extension version they were defined against.

We have defined an optional `core.used_extensions` metadata that stores entry that declares a superset of the extensions required to define a HUGR, but this is more a debugging annotation. It not being part of the model means that we cannot reliably depend on it when loading and processing HUGRs.

### Text format

The text (EDN) format uses symbol names to refer to extension operations and types. Symbols are brought into scope with import statements, and referenced as term applications at use sites. Operation and type declarations may also appear in modules when the extension's signature is known at export time.

None of these carry version information today:

```lisp
; Import an extension operation by name (no version).
(import arithmetic.int.iadd)

; A type is a term application of a symbol to arguments.
; Here `collections.static_array` is a type constructor applied to a type arg.
(collections.static_array prelude.usize)

; An operation call is also a term application: symbol + static args.
(arithmetic.int.iadd 6)

; When an OpDef has a known signature, the exporter emits an operation
; declaration. This introduces the symbol for resolution, but carries
; no extension version.
(declare-operation
  arithmetic.int.iadd
  (param ?n core.nat)
  (core.fn
    [(arithmetic.int.types.int ?n) (arithmetic.int.types.int ?n)]
    [(arithmetic.int.types.int ?n)]))

; Type constructors are declared similarly.
(declare-ctr collections.static_array
  (param ?t core.type)
  (param ?n core.nat)
  core.type)
```

### Cap'n Proto binary format

The binary format mirrors the text format structurally. Operations and types are encoded as term applications pointing to a resolved symbol node. Imports carry only the symbol name as a text field, and declarations carry a Symbol struct with name, parameters, and signature — but no version.

```capnp
struct Operation {
  union {
    # A custom operation node references a term (which is a Term::Apply
    # pointing to the resolved import/declaration symbol node).
    custom @0 :TermId;
    # An import brings an external symbol into scope by name.
    # No version information.
    import @11 :Text;
    # Type constructor declarations.
    constructorDecl @12 :Symbol;
    # Operation declarations (emitted when OpDef has a known signature).
    operationDecl @13 :Symbol;
    # ... other variants (dfg, cfg, funcDefn, funcDecl, etc.)
  }
}

struct Symbol {
  name @0 :Text;
  params @1 :List(Param);
  constraints @2 :List(TermId);
  signature @3 :TermId;
  visibility @4 :Visibility;
  # No version field today.
}

struct Term {
  union {
    # Types and operations are term applications referencing a symbol node.
    apply :group {
      symbol @0 :NodeId;
      args @1 :List(TermId);
    }
    # ... other variants (variable, list, literal, etc.)
  }
}
```

There is no place for extension versions anywhere in the current binary schema.

The in-memory Rust representations (both the capnp-like resolved table and the text-like unresolved AST) mirror this structure and are equally unversioned.

### Design Goals

- Preserve read compatibility with current encoded HUGRs that do not specify extension versions.
- Encode an optional extension version for both custom operations and extension types.
- Allow one model package to mention multiple incompatible versions of the same extension id.
- Keep the text format readable.
- Keep import errors informative when a requested version cannot be resolved.

### Text format changes

Extension imports and declarations gain an optional `@major.minor.patch` version suffix. The `@version` is part of the model syntax, not part of the symbol name itself.

```lisp
; Versioned imports — declares which extension versions this module uses.
(import arithmetic.int.iadd@0.2.3)
(import arithmetic.int.iadd@0.3.0)

; At use sites, @version disambiguates when multiple versions are in scope.
(arithmetic.int.iadd@0.2.3 6)  ; explicitly uses 0.2.3
(arithmetic.int.iadd 6)        ; unversioned: resolved later by registry-aware import

; Declarations also carry a version to record provenance.
(declare-operation
  arithmetic.int.iadd@0.3.0
  (param ?n core.nat)
  (core.fn
    [(arithmetic.int.types.int ?n) (arithmetic.int.types.int ?n)]
    [(arithmetic.int.types.int ?n)]))

; Type constructors likewise.
(declare-ctr collections.static_array@0.3.0
  (param ?t core.type)
  (param ?n core.nat)
  core.type)
```

Parsing `arithmetic.int.iadd@0.2.3` produces:

- symbol name: `arithmetic.int.iadd`
- extension version: `0.2.3`
- arguments: unchanged

#### Text encoding export rules

We don't always need to include the `@version` field, if it can be deduced from the context.
This makes the format easier to read by avoiding spamming every symbol with `@v` tags.

- Always include `@version` on import/declaration lines.
- At use sites
  - Omit `@version` if we're using the highest version between all imports/declarations of the symbol.
  - Include the `@version` otherwise.

It is always valid to include @version explicitly, the omissions are optional.

#### Text encoding import rules

- When reading, we resolve unversioned symbols to the highest declared versions.
- If no version has been declared, `hugr-model` leaves the field empty. This should only happen when loading older pre-versioning HUGRs.
- It's the responsibility of the import code in `hugr-core` and `hugr-py` to resolve these extensions, using the context from their extension registries.

### Cap'n Proto binary format changes

We make three backwards-compatible schema changes:

- `import` becomes a group retaining its existing `@11` text slot, adding an
optional version field. This is a backwards compatible addition.
- `Symbol` gains an optional version field so that operation/constructor
declarations can record their extension version.
- `Version` gains a patch field (needed for full semver; the existing
major/minor-only struct is currently only used for the model format versioning).

```capnp
struct Operation {
  union {
    import :group {
      # Retains the existing import payload slot for wire compatibility.
      name @11 :Text;
      # Optional extension version for this imported symbol.
      version @15 :Version;
    }
    # Operation declarations now carry a version via their Symbol.
    operationDecl @13 :Symbol;
    constructorDecl @12 :Symbol;
    # ... other variants unchanged
  }
}

struct Symbol {
  name @0 :Text;
  params @1 :List(Param);
  constraints @2 :List(TermId);
  signature @3 :TermId;
  visibility @4 :Visibility;
  # NEW: optional extension version.
  version @5 :Version;
}
```

Term.apply does not need a version field because it references a concrete
symbol node by NodeId; the version lives on that symbol's import or
declaration.

### Backwards compatibility:

New readers can read old binaries: missing version fields are treated as None.

Old readers ignore new fields per normal Cap'n Proto evolution rules. This means the new hugrs can be loaded, but we may still see errors due to mismatched extension versions in the reader.

Upgrading `import @11 :Text` to an `import :group` preserving name `@11` is a compatible Cap'n Proto schema evolution (moving an existing field into a newly introduced group is explicitly allowed).

## Bridging the gap: Progressive migrations

This will probably be a followup issue.

Going back to the first example, where we have

- Operation `opA` from `extA@0.2.3`
- Operation `opB` from `extB@0.1.0` that depends on a custom type from `extA@0.2.3`

Let’s imagine we have a HUGR that uses opA and opB.

```lisp
; Import the operation symbols
(import extA.opA@0.2.3)
(import extB.opB@0.1.0)

; `opA` outputs a `typeA` value that gets consumed by `opB`
(extA.opA@0.2.3 [] [%1]) ; %1 has type `typeA@0.2.3`
(extB.opB@0.1.0 [%1] [])
```

We want to upgrade to `extA@0.3.0` and `extB@0.2.0`, but the operation migration code is done by different tools.

To keep the encoding valid, we’ll need a temporary adapter op in the middle. First we run the extA migration:

```lisp
(import extA.opA@0.3.0)
(import extB.opB@0.1.0)
; We also need the type identifiers now so we can use them in `castType`
(import extA.typeA@0.2.3)
(import extA.typeA@0.3.0)

(extA.opA@0.3.0 [] [%1]) ; %1 has type `typeA@0.3.0`
(std.compat.castType extA.typeA@0.3.0 extA.typeA@0.2.3 [%1] [%2]) ; %2 has type `typeA@0.2.3`
(extB.opB@0.1.0 [%2] [])
```

This HUGR gets serialized and sent to the migration tool for extB, that generates the final form:

```lisp
(import extA.opA@0.3.0)
(import extB.opB@0.2.0)

(extA.opA@0.3.0 [] [%1])
(extB.opB@0.2.0 [%1] [])
```

The castType operation can be defined as an extension op in HUGRs ext, it doesn’t need core support.
We may however want to validate in the normal compilation workflow that HUGRs are fully migrated to the current extension versions, and do not contain compatibility artifacts.
This should be easily done by checking that we’re using a single version per extension, and that they match the versions used by the compiler.

aborgna-q:
This PR adds extension version information to the encoded custom op and types in `hugr-model`.
We'll use these to implement extension migration in a followup PR.

Part of #2996. Closes #3077.

See #3077 for a description of what we changed in this PR.

BREAKING CHANGE: The capnp import/export code has been updated to the format version `v1.1.0`.


hugr-model/src/v0/ast/hugr.pest
```diff
@@ -4,6 +4,8 @@ identifier         = @{ identifier_initial ~ identifier_rest* }
 identifier_initial = @{ ASCII_ALPHA | "_" | "-" }
 identifier_rest    = @{ ASCII_ALPHANUMERIC | "_" | "-" }
 symbol_name        = @{ !(reserved ~ !(identifier_rest | ".")) ~ (identifier ~ ("." ~ identifier)*) }
+version            = @{ ASCII_DIGIT+ ~ "." ~ ASCII_DIGIT+ ~ "." ~ ASCII_DIGIT+ }
+symbol_ident         = { symbol_name ~ ("@" ~ version)? }
```
cqc-alec:
```suggestion
symbol_ident       = { symbol_name ~ ("@" ~ version)? }
```
Indentation?

hugr-py/src/hugr/package.py
```diff
@@ -126,10 +126,24 @@ def from_model(
...
         pkg = Package([Hugr.from_model(hugr) for hugr in package.modules], extensions)
 
-        # Resolve extensions in the modules and inside the bundled extension definitions
-        pkg.resolve_extensions(resolve_from or ExtensionRegistry())
+        # Resolve extensions in the modules and inside the bundled extension
+        # definitions.
+        #
+        # TODO: Remove the None compatibility path once encoded HUGRs without
+        # extension version information are no longer supported.
+        # <http://github.com/Quantinuum/hugr/issues/???>
```
cqc-alec: Create issue for this?

hugr-core/src/export.rs
```diff
@@ -84,8 +87,15 @@ struct Context<'a> {
     symbols: model::scope::SymbolTable<'a>,
 
     /// Mapping from implicit imports to their node ids.
+    ///
+    /// These symbol exclude extension op and type imports. See `extension_exports` for those.
```
cqc-alec:
```suggestion
    /// These symbols exclude extension op and type imports. See `extension_exports` for those.
```

hugr-core/src/export.rs
```diff
@@ -1143,20 +1214,51 @@ impl<'a> Context<'a> {
...
+                    },
                     ..table::Node::default()
                 })
             }),
         }
     }
 
+    /// Resolve an extension symbol identifier in the model, creating an implicit import if needed.
+    ///
+    /// The symbols are linked to an specific extension version.
```
cqc-alec:
```suggestion
    /// The symbols are linked to a specific extension version.
```

hugr-core/src/import.rs
```diff
@@ -140,6 +142,27 @@ enum ExtensionError {
...
     },
+
+    /// An unversioned extension symbol cannot be resolved.
+    ///
+    /// These symbols originate from older encoded HUGRs that did not include
+    /// version information for extension symbols.
+    ///
+    /// TODO: Remove the unversioned compatibility path once encoded HUGRs
+    /// without extension version information are no longer supported.
+    /// <http://github.com/Quantinuum/hugr/issues/???>
```
cqc-alec: Add issue?

hugr-core/src/import.rs
```diff
@@ -520,6 +544,33 @@ impl<'a> Context<'a> {
         Ok(name)
     }
 
+    /// Get the identifier and version of the symbol introduced by a model node.
+    ///
+    /// Only extension op and type symbols may include version information.
+    ///
+    /// TODO: Remove the `None` path once encoded HUGRs without extension
+    /// version information are no longer supported.
+    /// <http://github.com/Quantinuum/hugr/issues/???>
```
cqc-alec: Add issue?

hugr-core/src/import.rs
```diff
@@ -1330,25 +1381,74 @@ impl<'a> Context<'a> {
...
+    ///   in the [`Context::extensions`] registry that contains the symbol and
+    ///   return it if it exists.
+    /// - If the model symbol is unversioned and we don't have a matching
+    ///   extensions in [`Context::extensions`], return `None`.
+    ///
+    /// TODO: Remove the unversioned fallback once encoded HUGRs without
+    /// extension version information are no longer supported. Return an
+    /// [`ExtensionError::UnversionedSymbol`] error instead of `Ok(None)` in
+    /// that case.
+    /// <http://github.com/Quantinuum/hugr/issues/???>
```
cqc-alec: Add issue?

hugr-core/src/import.rs
```diff
@@ -1330,25 +1381,74 @@ impl<'a> Context<'a> {
...
+        // Return the latest extension version that defines the symbol, if any.
+        for ext in versions.iter().rev() {
+            if ext.get_op(name).is_some() || ext.get_type(name).is_some() {
+                return Ok(Some(ext.version().clone()));
+            }
+        }
+
+        // TODO: Remove this fallback once encoded HUGRs without extension
+        // version information are no longer supported.
+        // <http://github.com/Quantinuum/hugr/issues/???>
```
cqc-alec: Make issue?

hugr-model/src/v0/ast/resolve.rs
```diff
@@ -18,11 +19,20 @@ pub struct Context<'a> {
...
 impl<'a> Context<'a> {
+    /// Create an empty resolver context backed by `bump`.
+    ///
+    /// The context records explicit symbol versions from imports and
+    /// declarations. It does not fill missing versions from any external
+    /// registry; unresolved legacy versions remain `None` in the table model.
+    ///
+    /// TODO: Remove the `None` compatibility path once encoded HUGRs without
+    /// extension version information are no longer supported.
+    /// <http://github.com/Quantinuum/hugr/issues/???>
```
cqc-alec: Issue?

hugr-model/src/v0/ast/resolve.rs
```diff
@@ -336,21 +353,35 @@ impl<'a> Context<'a> {
...
+    ///
+    /// Versioned references resolve exactly. Unversioned references first use
+    /// any unversioned binding, then the latest visible versioned binding.
+    /// If no binding exists, the implicitly created import preserves the
+    /// reference version as written, including `None` for legacy unversioned
+    /// references.
+    ///
+    /// TODO: Remove the legacy unversioned fallback once encoded HUGRs without
+    /// extension version information are no longer supported.
+    /// <http://github.com/Quantinuum/hugr/issues/???>
```
cqc-alec: Issue?

hugr-model/src/v0/scope/symbol.rs
```diff
@@ -160,6 +262,32 @@ impl<'a> SymbolTable<'a> {
...
+/// Symbol lookup key for a name and optional extension version.
+///
+/// The version is optional so the table can represent both legacy unversioned
+/// symbols and modern versioned extension symbols. `None` is a real binding key,
+/// not a wildcard; wildcard/default behavior is handled explicitly by
+/// [`SymbolTable::resolve`].
+///
+/// TODO: Remove the legacy unversioned path once encoded HUGRs without
+/// extension version information are no longer supported.
+/// <http://github.com/Quantinuum/hugr/issues/???>
```
cqc-alec: Add issue?

cqc-alec (approved):

hugr-py/src/hugr/package.py
aborgna-q: Ouch. Forgot to update those -.-

I created #3086 and replaced all the missing links.

PR #3081, opened 2026-05-29
fix: Properly validate extension definitions

acl-cqc:
Previously we were not validating signatures of ops in extensions, for example that type variables referred to in the signature body were declared as type parameters, or any of the checks made in `PolyFuncTypeRV::validate`.

First commit adds test, second fixes.


hugr-core/src/extension/op_def.rs
```diff
@@ -409,7 +409,10 @@ impl OpDef {
     ) -> Result<(), SignatureError> {
         let temp: PolyFuncTypeRV; // to keep alive
         let (pf, args) = match &self.signature_func {
-            SignatureFunc::CustomValidator(ts) => (&ts.poly_func, args),
+            SignatureFunc::CustomValidator(custom) => {
+                custom.validate.validate(args, self)?;
```
acl-cqc: Seems like it might be better to do the other validation (against PF declared params) first, but this matches `compute_signature`

mark-koch (approved):

PR #3089, opened 2026-06-03
feat!: Remove deprecated pure json HUGR APIs

### LINKED ISSUE #3075: [Feature]: Remove deprecated json read/write interfaces for Hugrs
The APIs for reading and writing HUGRs as json files have been deprecated in both rust and python for a while now. We should remove them.

Note that this **does not** mean we can remove the serde implementations yet, as they are used in two other paths;
- Hugr JSON can still be loaded from `EnvelopeFormat::PackageJson` envelopes. This format was deprecated on 2026-03-30 (#2991) and released in `hugr 0.27.0 (2026-03-31)` and `hugr-py 0.16.0 (2026-04-01)`.
- We still use `serde` for encoding HUGR extension definitions. This requires not only `opDef` and `typeDef` serialization, but also `Type` serde-derives to specify in the operation signatures.

We should be able to remove the deprecated envelope formats soon, and partially do the serde removals (#3007).
Transitioning out of JSON extensions will take longer.

aborgna-q:
Removes the deprecated public APIs for writing and loading raw json hugrs.

Closes #3075.

This does not yet remove the `Json` envelope format, only the pure jsons with no envelope header.
While that component is still included, old json hugrs may be migrated to envelopes by prepending and appending the following strings:
```rs
PREPEND = r#"HUGRiHJv?@{"modules": ["#;
APPEND = r#"],"extensions": []}"#;
```

The removal of envelope json support is tracked in #3007.

BREAKING CHANGE: Removed pure HUGR JSON store/decode APIs

### DIFF
--- a/hugr-cli/src/hugr_io.rs
+++ b/hugr-cli/src/hugr_io.rs
@@ -1,12 +1,12 @@
 //! Input/output arguments for the HUGR CLI.
 
 use clio::Input;
+use hugr::Extension;
 use hugr::envelope::description::PackageDesc;
 use hugr::envelope::read_envelope;
 use hugr::extension::ExtensionRegistry;
 use hugr::extension::resolution::WeakExtensionRegistry;
 use hugr::package::Package;
-use hugr::{Extension, Hugr};
 use std::io::{BufReader, Read};
 use std::path::PathBuf;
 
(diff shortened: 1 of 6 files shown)


cqc-alec (approved):
We should probably put the advice about the `PREPEND` and `APPEND` strings in the changelog.

PR #3090, opened 2026-06-03
fix: Re-link opdefs when doing extension resolution

aborgna-q:
When doing extension resolution on `ExtensionOp`s, ensure we update the `OpDef` reference to the one in the resolution extension registry.

Previously, when an `ExtensionOp` was defined with a different extension and we then ran extension resolution with a different Extension for the same ID, we didn't update the `Weak<Extension>` reference.

After doing resolution, replacing the extension registry inside the hugr may cause the op's Weak extension pointer to be dropped, causing errors down the line.

### DIFF
--- a/hugr-core/src/extension/resolution/ops.rs
+++ b/hugr-core/src/extension/resolution/ops.rs
@@ -7,11 +7,13 @@
 
 use std::sync::Arc;
 
+use semver::Version;
+
 use super::{Extension, ExtensionCollectionError, ExtensionResolutionError};
 use crate::Node;
-use crate::extension::{ExtensionId, ExtensionRegistry};
+use crate::extension::{ExtensionId, ExtensionRegistry, OpDef};
 use crate::ops::custom::OpaqueOpError;
-use crate::ops::{DataflowOpTrait, ExtensionOp, NamedOp, OpType};
+use crate::ops::{DataflowOpTrait, ExtensionOp, NamedOp, OpNameRef, OpType};
 
 /// Returns the extension in the registry required by the operation.
 ///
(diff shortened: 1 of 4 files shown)


hugr-core/src/extension/resolution/ops.rs
```diff
@@ -64,78 +64,85 @@ pub(crate) fn resolve_op_extensions<'e>(
...
-        return Ok(extension);
-    };
-
-    // Fail if the Extension is not in the registry, or if the Extension was
-    // found but did not have the expected operation.
-    let extension = extension.expect("OpaqueOp should have an extension");
-    let Some(def) = extension.get_op(opaque.unqualified_id()) else {
-        return Err(OpaqueOpError::OpNotFoundInExtension {
+    let ext_not_found_error =
+        |ext_id: &ExtensionId, op_id| ExtensionResolutionError::MissingOpExtension {
```
acl-cqc: You might want to make this a closure that takes an ExtensionId, version, qualified id and unqualified id, and returns a `Result<&OpDef, AppropriateError>` as the two cases below that call both closures seem very similar (doing very similar stuff besides calling the closures if there are errors)

hugr-core/src/extension/resolution/ops.rs
```diff
@@ -64,78 +64,85 @@ pub(crate) fn resolve_op_extensions<'e>(
...
+        };
 
-    let ext_op = ExtensionOp::new_with_cached(def.clone(), opaque.args().to_vec(), opaque)
-        .map_err(|e| OpaqueOpError::SignatureError {
-            node,
-            name: opaque.name().clone(),
-            cause: e,
-        })?;
+    match op {
+        OpType::ExtensionOp(ext_op) => {
```
acl-cqc, line 103: Alternatively you might call ext_op.make_opaque() and proceed as per second case, although I agree that does clone more stuff so is probably less efficient. (Would save a lot of code tho as you wouldn't need `relink_def` either)

hugr-core/src/extension/resolution/test.rs
```diff
@@ -191,6 +191,81 @@ fn make_extension_self_referencing(name: &str, op_name: &str, type_name: &str) -
     })
 }
 
+/// Create two extensions where one operation's signature uses the other's type.
+fn make_dependent_extensions() -> (Arc<Extension>, Arc<Extension>) {
+    let dep_ext =
+        Extension::new_test_arc(ExtensionId::new_unchecked("dummy.dep"), |ext, ext_ref| {
+            ext.add_type(
+                "dep_type".into(),
```
acl-cqc, line 199: Oh no!! Dependent types!! 😆

acl-cqc (approved):
Yes, definitely agree this is something we should be doing, thanks! :)

Can you fix the description?

> Previously, when an ExtensionOp was defined with a different extension instantiation

Probably you mean "definition" or "version" rather than "instantiation" - or maybe just "an extension"....

> and we then ran extension resolution

...and then "with a different Extension for the same ID" here

Secondly:

> we didn't update the Weak<Extension> reference. After doing the resolution, replacing the extension registry inside the hugr may cause

finish the sentence ;)

hugr-core/src/hugr/validate.rs
```diff
@@ -201,6 +202,16 @@ impl<'a, H: HugrView> ValidationContext<'a, H> {
         if let OpType::Const(c) = op_type {
             c.validate()?;
         }
+        if let OpType::ExtensionOp(ext_op) = op_type {
```
acl-cqc, line 205: One might say that adding this check makes this a breaking PR...

hugr-core/src/extension/resolution/ops.rs
aborgna-q: Uhm. I'd be clean, but I would try to avoid the clonings. This is an expensive traversal already (with a lot of pointer dereferences).

PR #3096, opened 2026-06-05
refactor: simplify InlineDFG, extend tests

acl-cqc:
#3072 added extra order edges from the DFG's Order-predecessors to its Order-successors only if there was no path from the DFG's Input to the DFG's Output.

However, it's fairly harmless to add them anyway - the extra ones would be totally redundant in such case and should be removed by tket2's RemoveRedundantOrderEdgesPass - moreover the extra edges are from (nodes that already have other outgoing Order edges) and to (nodes that already have incoming Order edges), and do not place any extra restrictions on ordering, so should not inhibit optimization at all.

While about it, also extend the test (and remove the extra order edge added "in the middle" in #3072)


doug-q (approved):
Thanks Alan, I think this is a definite improvement.

PR #3101, opened 2026-06-10
fix: Quote non-parsable Symbol names in S-expressions

aborgna-q:
Guppy has tests that generated functions with non-parsable symbol identifiers;

```erlang
(define-func
  public
  tests.integration.test_basic.test_implicit_return.<locals>.ret
  ...
)
```

`hugr-model` happily exported those strings, but the `pest` grammar was not able to parse them back when loading them.

This PR adds an optional symbol name escaping similar to rust r-strings, so those symbols can still be supported.

```erlang
(define-func
  public
  r#"tests.integration.test_basic.test_implicit_return.<locals>.ret"#
  ...
)
```
More `#`s can be added as needed, `r###"some symbol with two "## but not three"###`.

We do not use bare strings to avoid confusing symbols with string values.


hugr-model/src/v0/ast/hugr.pest
```diff
@@ -3,7 +3,10 @@ COMMENT            = _{ ";" ~ (!("\n") ~ ANY)* ~ "\n" }
 identifier         = @{ identifier_initial ~ identifier_rest* }
 identifier_initial = @{ ASCII_ALPHA | "_" | "-" }
 identifier_rest    = @{ ASCII_ALPHANUMERIC | "_" | "-" }
-symbol_name        = @{ !(reserved ~ !(identifier_rest | ".")) ~ (identifier ~ ("." ~ identifier)*) }
+bare_symbol_name    = @{ !(reserved ~ !(identifier_rest | ".")) ~ (identifier ~ ("." ~ identifier)*) }
+raw_symbol_name     = ${ "r" ~ PUSH("#"+) ~ "\"" ~ raw_symbol_name_raw ~ "\"" ~ POP }
```
croyzor, line 7: neat

hugr-model/src/v0/ast/parse.rs
```diff
@@ -42,9 +42,26 @@ mod pest_parser {
...
+}
+
 fn parse_symbol_name(pair: Pair<Rule>) -> ParseResult<SymbolName> {
     debug_assert_eq!(Rule::symbol_name, pair.as_rule());
-    Ok(SymbolName(pair.as_str().into()))
+    let pair = pair.into_inner().next().unwrap();
+
+    Ok(match pair.as_rule() {
+        Rule::bare_symbol_name => SymbolName(pair.as_str().into()),
+        Rule::raw_symbol_name => SymbolName(parse_raw_symbol_name(pair)),
```
croyzor, line 62: Why not the same as the above?

croyzor (approved):
yeee

hugr-model/src/v0/ast/parse.rs
aborgna-q: We need to strip the `#` pre/postfix that are not part of the symbol.

PR #3110, opened 2026-06-11
fix: Allow TypeRow to be constructed from a ListConcat term

doug-q:


hugr-core/src/types/type_row.rs
```diff
@@ -468,4 +468,12 @@ mod test {
 
         assert_eq!(term.try_into(), Ok(type_row));
     }
+
+    #[test]
+    fn test_typerow_from_concatenated_type_row_rvs() {
+        let tr1 = TypeRowRV::from(vec![Type::UNIT]);
+        let tr2 = TypeRowRV::from(vec![bool_t()]);
+        let concatenated: TypeRow = tr1.clone().concat(tr2).try_into().unwrap();
```
acl-cqc: super-nit: can probably drop the clone here

acl-cqc (approved):
Nice one Doug, thanks :). Suggest you add a test of a failing case i.e. a `TypeRowRV::new_var_use` (or concatenation including one), but otherwise looks good to me

PR #3121, opened 2026-06-19
feat: Allow instantiated ops with missing compute funcs when validating

aborgna-q:
Follow up to #3118.

Extension ops may define a signature computation function.
This binary function cannot cross serialization boundaries, so it's represented with a `SignatureFunc::MissingComputeFunc`.

If an extension gets defined in python, there is no way for the rust-based validator to run the function.
Previously this caused an error, but it should instead ignore the missing function.
The already-instantiated op will have a cached signature when deserialized.

### DIFF
--- a/hugr-core/src/extension/op_def.rs
+++ b/hugr-core/src/extension/op_def.rs
@@ -422,7 +422,7 @@ impl OpDef {
                 temp = custom.compute_signature(static_args, self)?;
                 (Some(&temp), other_args)
             }
-            // Validation functions cannot cross the serialization boundary.
+            // Signature computation and validation functions cannot cross the serialization boundary.
             // If they are missing, we ignore them and use the signature as-is.
             //
             // TODO raise warning: https://github.com/CQCL/hugr/issues/1432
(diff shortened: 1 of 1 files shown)


maximilianruesch (approved):

hugr-core/src/extension/op_def.rs
```diff
@@ -408,27 +408,31 @@ impl OpDef {
...
                 let (static_args, other_args) =
                     args.split_at(min(custom.static_params().len(), args.len()));
                 static_args.iter().try_for_each(|ta| ta.validate(&[]))?;
                 check_term_kinds(static_args, custom.static_params())?;
                 temp = custom.compute_signature(static_args, self)?;
-                (&temp, other_args)
+                (Some(&temp), other_args)
             }
-            SignatureFunc::MissingComputeFunc => return Err(SignatureError::MissingComputeFunc),
             // Validation functions cannot cross the serialization boundary.
```
cqc-alec:
```suggestion
            // Signature computation and validation functions cannot cross the serialization boundary.
```

cqc-alec (approved):

PR #3128, opened 2026-06-29
feat!: Remove metadata object from python Nodes

### LINKED ISSUE #1757: [Bug]: Rust-backed python pass drops module-root metadata
### Description

Looks like somewhere in the py->rs->py roundtrip we're dropping the module root metadata from the hugr.

### Steps to Reproduce

```py
from guppylang import OptimizationLevel, guppy
from hugr.metadata import HugrGenerator, HugrUsedExtensions
from tket import passes


@guppy
def main() -> None:
    pass


# Compile without Guppy's default optimization path so we can isolate tket.
original = main.compile().modules[0]

opt_pass = passes.NormalizeGuppy(
    resolve_modifiers=False,
    simplify_cfgs=False,
    remove_tuple_untuple=False,
    constant_folding=False,
    remove_dead_funcs=False,
    inline_dfgs=False,
    remove_redundant_order_edges=False,
    squash_borrows=False,
)

optimized = opt_pass.run(original, inplace=False).hugr

before = original.module_root.metadata
after = optimized.module_root.metadata

print("before metadata:", before)
print("after metadata: ", after)

assert HugrGenerator.KEY in before
assert HugrUsedExtensions.KEY in before

# These fail: NormalizeGuppy dropped root metadata.
assert HugrGenerator.KEY in after
assert HugrUsedExtensions.KEY in after
```

### Expected Behaviour

Module root metadata is preserved

### Actual Behaviour

Module root metadata gets dropped

### Component

_No response_

### Environment

```markdown
- guppylang == 1.0.0-a7
- hugr == 0.17.1
- tket == 0.14.2
```

### Additional Context

_No response_

aborgna-q:
Fixes https://github.com/Quantinuum/tket2/issues/1757

Python `Node`s kept a `NodeMetadata` object internally to record info about the node.

When a node got added / linked to a `Hugr`, the metadata object was replaced with a reference to the one inside the corresponding `NodeData` in `Hugr._nodes`.

The problem with this is that if someone does
```
hugr[some_node].metadata = new_metadata_object
```
then the object references get out of sync.

This was the case with `hugr.entrypoint` and `hugr.module_root` that didn't get updated after decoding a hugr, but could have shown in user code too.

The simplest solution here, to avoid users accidentally breaking the sync, is to remove the metadata object from `Node`, and fetch it from the appropriate Hugr instead.

BREAKING CHANGE: Removed `ToNode.metadata` method. Use `Hugr[node].metadata` instead.

### DIFF
--- a/hugr-py/src/hugr/hugr/base.py
+++ b/hugr-py/src/hugr/hugr/base.py
@@ -211,11 +211,7 @@ class Hugr(Mapping[Node, NodeData], Generic[OpVarCov]):
         return n
 
     def __iter__(self) -> Iterator[Node]:
-        return (
-            self._node_handle(idx, data.metadata)
-            for idx, data in enumerate(self._nodes)
-            if data is not None
-        )
+        return (Node(idx) for idx, data in enumerate(self._nodes) if data is not None)
 
     def __len__(self) -> int:
         return self.num_nodes()
@@ -356,24 +352,14 @@ class Hugr(Mapping[Node, NodeData], Generic[OpVarCov]):
         if self._free_nodes:
             node = self._free_nodes.pop()
             self._nodes[node.idx] = node_data
-            node._bind_hugr(self)
         else:
-            node = self._node_handle(len(self._nodes), node_data.metadata)
+            node = Node(len(self._nodes))
             self._nodes.append(node_data)
         node._num_out_ports = num_outs
-        node._metadata = node_data.metadata
         if parent:
             self[parent].children.append(node)
 
         self._update_node_outs(node, num_outs)
-        return node
-
-    def _node_handle(
-        self, idx: NodeIdx, metadata: NodeMetadata, num_outs: int | None = None
-    ) -> Node:
-        """Create a node handle bound to this HUGR's canonical node table."""
-        node = Node(idx, metadata, num_outs)
-        node._bind_hugr(self)
         return node
 
     def _update_node_outs(self, node: Node, num_outs: int | None) -> Node:
@@ -485,9 +471,6 @@ class Hugr(Mapping[Node, NodeData], Generic[OpVarCov]):
             self._links.delete_left(_SubPort(out))
 
         weight, self._nodes[node.idx] = self._nodes[node.idx], None
-
-        # Free up the metadata dictionary
-        node._metadata = NodeMetadata()
 
         self._free_nodes.append(node)
         return weight
@@ -977,7 +960,7 @@ class Hugr(Mapping[Node, NodeData], Generic[OpVarCov]):
             serial_idx = len(nodes)
 
             # non contiguous indices will be erased
-            nodes.append(data._to_serial(Node(serial_idx, NodeMetadata())))
+            nodes.append(data._to_serial(Node(serial_idx)))
             metadata.append(data.metadata.as_dict() if data.metadata else None)
             if self.entrypoint == node:
                 entrypoint = serial_idx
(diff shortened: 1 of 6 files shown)


hugr-py/src/hugr/hugr/node_port.py
```diff
@@ -155,11 +157,41 @@ class Node(ToNode):
...
+    # If the node is not linked to a Hugr, the metadata information.
+    #
+    # This gets overriden by the Hugr's node info if `self._hugr` is alive.
     _metadata: NodeMetadata = field(
         repr=False, compare=False, default_factory=NodeMetadata
     )
+    # Number of output ports for this node, or None if the number is not fixed.
     _num_out_ports: int | None = field(default=None, compare=False, repr=False)
+    # The Hugr that owns this node, if any.
+    _hugr: weakref.ReferenceType[Any] | None = field(
```
acl-cqc: does this have to be `[Any]`? Can we not have `weakref.ReferenceType[Hugr]` ? (Would need to update _bind_hugr too)

hugr-py/src/hugr/hugr/base.py
```diff
@@ -1040,7 +1049,9 @@ def get_meta(idx: int) -> NodeMetadata:
                 continue
 
             node_meta = get_meta(idx)
-            parent: Node | None = Node(serial_node.root.parent)
+            parent: Node | None = hugr._node_handle(
```
acl-cqc: Why the `| None`? `def _node_handle` never returns None, and I don't see `parent` being reassigned?

hugr-py/src/hugr/hugr/base.py
```diff
@@ -356,8 +356,9 @@ def _add_node(
         if self._free_nodes:
             node = self._free_nodes.pop()
             self._nodes[node.idx] = node_data
+            node._bind_hugr(self)
```
acl-cqc: AFAICS this is the only place we `_bind_hugr` on a node that already exists, rather than being freshly created. And it's just come from `_free_nodes`....

Enforcing that all _free_nodes have a valid weakref to *this Hugr* would require
* Checking delete_node is only given nodes in the current hugr (via weakref). This seems like a sensible check anyway (albeit breaking)...?
   * Indeed adding a check in `Hugr.__getitem__` that the supplied node belongs to the same Hugr might be good too...??
* Processing the list in `_overwrite_hugr` to update the refs in the free list (which is also transferred)

acl-cqc: Ok so maybe we want `Node`s to be transferable between Hugrs, hmmm

hugr-py/src/hugr/hugr/node_port.py
acl-cqc: Consider `kw_only=True` but you *can* assign this in constructor without using `_bind_hugr`

hugr-py/src/hugr/hugr/base.py
```diff
@@ -356,8 +356,9 @@ def _add_node(
...
         else:
-            node = Node(len(self._nodes), NodeMetadata())
+            node = self._node_handle(len(self._nodes), node_data.metadata)
             self._nodes.append(node_data)
         node._num_out_ports = num_outs
         node._metadata = node_data.metadata
```
acl-cqc: This can go in the `if self._free_nodes` branch, the `else` branch does it already

acl-cqc (commented):
So I think this fixes the case where:
```
n = next(iter(hugr))
hugr[n].metadata = ...new...
print(n.metadata)
```

but not the case where:
```
n = next(iter(hugr))
hugr[n].metadata = ...new...
hugr = None # Or otherwise cause hugr to be GC'd
print(n.metadata)
```
?

I think it'd be good to see both of these in tests.

Is an alternative approach to put `_data: NodeData` into `class Node`, give `@property def metadata` that reads that? There is no `hugr[n] = newNodeData` so that is the point in the tree of data at which the reference cannot become stale

aborgna-q:
After some irl discussion, I removed both the old metadata object and the new hugr reference from `Node` / `ToNode`, as it was really prone to desync bugs when using nodes to id into different structures.

Instead, metadata must now be set via the `Hugr`. I added some helper methods on the builder definitions to ease up that.

acl-cqc (approved):
The builder tweaks are genius as then no test changes!! Super :)

Please update PR description and title before merging but otherwise LGTM

PR #3138, opened 2026-07-07
feat: Ignore invalid non-local Order edges in InlineDFG reachability check

aborgna-q:
`is_order_reachable` in `InlineDFGs` only needs to check connectivity in the local region.

This is a simple change making the `is_order_reachable` ignore non-local edges into other regions, to avoid doing unnecessary work.

### DIFF
--- a/hugr-core/src/hugr/patch/inline_dfg.rs
+++ b/hugr-core/src/hugr/patch/inline_dfg.rs
@@ -127,6 +127,7 @@ impl<N: HugrNode> PatchHugrMut for InlineDFG<N> {
 ///
 /// [Order]: crate::types::EdgeKind::StateOrder
 fn is_order_reachable<H: HugrView>(h: &H, src: H::Node, tgt: H::Node) -> bool {
+    let parent = h.get_parent(src);
     let mut visited = HashSet::new();
     let mut to_visit = vec![src];
     while let Some(n) = to_visit.pop() {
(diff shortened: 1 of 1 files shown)


hugr-core/src/hugr/patch/inline_dfg.rs
```diff
@@ -135,7 +136,12 @@ fn is_order_reachable<H: HugrView>(h: &H, src: H::Node, tgt: H::Node) -> bool {
                 return true;
             }
             let order_outport = h.get_optype(n).other_output_port().unwrap();
-            to_visit.extend(h.linked_inputs(n, order_outport).map(|(n, _)| n));
+            to_visit.extend(
+                h.linked_inputs(n, order_outport)
+                    .map(|(n, _)| n)
+                    // Ignore non-local edges
```
acl-cqc, line 142:
```suggestion
                    // Ignore (invalid) non-local order edges
```

acl-cqc (approved):
Slightly ugh but yeah ok. Let's not confuse anyone by suggesting non-local edges can include order edges tho, these are invalid edges....

EDIT: hmmm, I suppose there is a broader issue here, is it a good idea for NormalizeCFGs to run InlineDFG on an invalid hugr at all? I mean, what kind of behaviour should it expect? This'll mean we assume the order path doesn't exist, but if the path were via a sibling of the DFG being inlined, that will be a valid order path *after* the InlineDFG completes....

aborgna-q:
Nvm. The check should never be `false` in valid hugrs.

PR #3148, opened 2026-07-10
perf: Reduce extension resolution work in envelope decoding

aborgna-q:
Avoid registering the same extension multiple times while decoding envelopes, as well as looking up the same terms repeatedly.

When running benchmarks locally with callgrind, this showed a `9.48%` cycle reduction when decoding a large real-world hugr.


cqc-alec:
@codspeedbot explain why this is faster

aborgna-q:
@codspeedbot explain why this is faster

cqc-alec (approved):

PR #3149, opened 2026-07-10
perf: Avoid cloning signatures for simple port checks

aborgna-q:
Avoids creating a whole signature object when calling `OpType::value_port_count` and `OpType::port_kind`.

Adds an internal `ValuePortOp` trait to dispatch the call to each optype variant.

Part of #1551


aborgna-q:
@codspeedbot explain why this is faster

hugr-core/src/ops/dataflow.rs
```diff
@@ -233,6 +266,12 @@ impl DataflowOpTrait for Call {
         }
     }
 }
+
+impl ValuePortOp for Call {
+    fn value_port_signature(&self) -> Option<&Signature> {
```
acl-cqc: Does this mean that if you pass in the static-input port you'll get back None as the edge kind?

acl-cqc: Oh, sorry, that might be deliberate....

hugr-core/src/ops.rs
```diff
@@ -539,6 +537,51 @@ pub(crate) trait NamedOp {
...
+/// allocation of the complete signature.
+#[enum_dispatch]
+pub(crate) trait ValuePortOp {
+    /// Returns a cached signature when the operation stores one directly.
+    fn value_port_signature(&self) -> Option<&Signature> {
+        None
+    }
+
+    /// Returns the kind of one value port, constructing only its requested type.
+    fn value_port_kind(&self, port: Port) -> Option<EdgeKind> {
```
acl-cqc: if it's only value ports we're interested in, then just return `Option<Type>` not `Option<EdgeKind>`. Or, rename to include static ports too...

hugr-core/src/ops/dataflow.rs
aborgna-q: Yes, this is trait replaces querying the signature for info on the value ports. Other ports are already tracked directly in the public `OpTrait` implementation.

Ideally the new `ValuePortOp` trait methods would be part of `OpTrait`, but I wanted to avoid breaking changes.

I can try unifying things more, or leave a TODO to seal `OpTrait` and add the methods there.

aborgna-q: I guess adding default-implemented methods to `OpTrait` wouldn't be a breaking change 🤔

acl-cqc (commented):
* Main concern here is just about duplicating logic in two places (the new method impls, and the old ones that constructs a Signature)
* Shame that this is internal to hugr-core; I don't see why external clients (e.g. some tket2 passes!) wouldn't benefit from the same
* Makes me think that we should make something like this be the *main* way to query a node - I think that means some mix of
   * Deprecating `signature()`
   * Changing the type `fn signature(&self) -> impl ValuePortThing` (...maybe instead of the methods on the op)
   * Making `Signature` implement this trait too, and `fn signature` just builds it as a cache of what the op methods are returning
* Note that there is `HugrView::`{`in_`,`out_`,}`value_types` as well, that calls the "expensive" (sometimes) Signature...maybe they just need updating

aborgna-q:
Refactored the change, adding the new methods to `OpTrait` instead.

hugr-core/src/ops.rs
```diff
@@ -577,6 +575,23 @@ pub trait OpTrait: Sized + Clone {
...
+    }
+
+    /// Returns the number of value ports in one direction.
+    ///
+    /// Implementations may override this to avoid constructing a complete
+    /// [`Signature`] when only its size is needed.
+    fn value_port_count(&self, dir: Direction) -> usize {
+        self.dataflow_signature()
+            .map_or(0, |signature| signature.port_count(dir))
+    }
```
aborgna-q, line 628: The new OpTrait methods are here.
They default to querying the signature so it's not a breaking change.

hugr-core/src/hugr/views/sibling_subgraph.rs
```diff
@@ -648,16 +648,16 @@ impl<N: HugrNode> SiblingSubgraph<N> {
             .iter()
             .map(|part| {
                 let &(n, p) = part.iter().next().expect("is non-empty");
-                let sig = hugr.signature(n).expect("must have dataflow signature");
-                sig.port_type(p).cloned().expect("must be dataflow edge")
+                let op = hugr.get_optype(n);
+                op.value_input_type(p).expect("must be dataflow edge")
```
aborgna-q, line 652: I'm wondering if we should have a shorthand for this on `HugrView`.

aborgna-q:
> Main concern here is just about duplicating logic in two places (the new method impls, and the old ones that constructs a Signature)

Done, we're using the existing traits now

> Shame that this is internal to hugr-core; I don't see why external clients (e.g. some tket2 passes!) wouldn't benefit from the same

It's now in the public trait

> Makes me think that we should make something like this be the main way to query a node - I think that means some mix of
> - Deprecating signature()
> - Changing the type fn signature(&self) -> impl ValuePortThing (...maybe instead of the methods on the op)
> - Making Signature implement this trait too, and fn signature just builds it as a cache of what the op methods are returning

Could be interesting, I'll track it in a separate issue since it would be a breaking change.

> Note that there is HugrView::{in_,out_,}value_types as well, that calls the "expensive" (sometimes) Signature...maybe they just need updating

Updated!

hugr-core/src/hugr/views/sibling_subgraph.rs
acl-cqc: For `get_optype(n).value_(in/out)put_type`, for `value_(in/out)put_type.expect`, or all three? Not opposed to any of those

acl-cqc: Or `(Node) -> impl Iterator<Item=Type>`? (+direction, or *2 for in/out)

hugr-core/src/hugr/views.rs
```diff
@@ -486,9 +486,9 @@ pub trait HugrView: HugrInternals {
     /// Iterator over all outgoing ports that have Value type, along
     /// with corresponding types.
     fn value_types(&self, node: Self::Node, dir: Direction) -> impl Iterator<Item = (Port, Type)> {
-        let sig = self.signature(node).unwrap_or_default();
-        self.node_ports(node, dir)
-            .filter_map(move |port| sig.port_type(port).map(|typ| (port, typ.clone())))
+        let op = self.get_optype(node);
+        op.value_ports(dir)
```
acl-cqc: yeah, definitely feels like giving back types as well as ports would be good here

hugr-core/src/ops/controlflow.rs
```diff
@@ -37,6 +37,28 @@ impl DataflowOpTrait for TailLoop {
         Cow::Owned(Signature::new(inputs, outputs))
     }
 
+    fn value_port_type(&self, port: Port) -> Option<Type> {
+        let (head, tail) = match port.direction() {
+            Direction::Incoming => (&self.just_inputs, &self.rest),
+            Direction::Outgoing => (&self.just_outputs, &self.rest),
+        };
+        head.get(port.index())
```
acl-cqc: consider `chain`ing two and then `.get` on that?

hugr-core/src/ops/sum.rs
```diff
@@ -49,6 +50,21 @@ impl DataflowOpTrait for Tag {
         ))
     }
 
+    fn value_port_type(&self, port: Port) -> Option<Type> {
+        match port.direction() {
+            Direction::Incoming => self.variants.get(self.tag)?.get(port.index()).cloned(),
```
acl-cqc: Just `None` if `self.tag` is out-of-bounds? In `fn signature` we panic (via `expect`)

hugr-core/src/ops.rs
```diff
@@ -694,3 +723,99 @@ macro_rules! impl_validate_op {
...
+    #[rstest]
+    #[case::input(input())]
+    #[case::output(output())]
+    #[case::call_indirect(call_indirect())]
+    #[case::load_constant(load_constant())]
+    #[case::load_function(load_function())]
+    #[case::tag(tag())]
+    #[case::tail_loop(tail_loop())]
+    #[case::conditional(conditional())]
+    fn value_ports_match_signature(#[case] op: OpType) {
```
acl-cqc, line 828: Good test. You might want to consider `call` and `dfg`

acl-cqc (commented):
Like this, good work @aborgna-q :)

Happy to approve like this but what do you think about having an iterator over (in/out-Port, Type) or even just Type for a node/direction? (`value_ports` only returns ports)

Could be on OpType instead of the method taking a in/out-Port, or as well, or I suppose on the Hugr(View)...

hugr-core/src/ops/controlflow.rs
aborgna-q: Chaining iterators produces a linear-time `.get`.

I rewrote this with simpler `if`s.

hugr-core/src/ops/sum.rs
aborgna-q: Updated, and gave it a better error message for all cases.

hugr-core/src/hugr/views/sibling_subgraph.rs
aborgna-q: Looking more into it, I think it's fine to leave the the port-specific type getters in `OpType`.

This PR already adds `HugrView::value_types(node, dir) -> Iterator<(Port, Type)>` and `in_`/`out_` variants. That should be enough to simplify relevant calls.

acl-cqc: Hang on....HugrView already defines https://github.com/Quantinuum/hugr/blob/e609c66507b9b7d0dde12ba2b7b681c800ec697d/hugr-core/src/hugr/views.rs#L496 ....that's not new in this PR ? [EDIT: ah yes, but I see you have reimplemented those on top of your new methods 👍 ]

hugr-core/src/ops/dataflow.rs
```diff
@@ -320,6 +374,23 @@ impl DataflowOpTrait for CallIndirect {
         Cow::Owned(s)
     }
 
+    fn value_port_type(&self, port: Port) -> Option<Type> {
+        match port.direction() {
```
acl-cqc, line 378: You might consider
```
if port.direction() == Direction::Incoming {
  if port.index() == 0 {return Some(Type::function(.....))}
  port = Port::new(Direction::Incoming, port.index() - 1); // harder than it should be?
}
self.signature.port_type(port)
```

hugr-core/src/ops/sum.rs
```diff
@@ -25,6 +26,18 @@ impl Tag {
     pub fn new(tag: usize, variants: Vec<TypeRow>) -> Self {
         Self { tag, variants }
     }
+
+    /// Return the TypeRow of the selected variant.
+    ///
+    /// Panics if the tag is out of bounds.
+    fn variant(&self) -> &TypeRow {
```
acl-cqc: Maybe call this `variant_row` or similar to avoid any ambiguity

acl-cqc (approved):
👍 Yah ok, agreed this is the best design that works 😆

PR #3150, opened 2026-07-10
perf: Export op signatures directly to hugr-model

aborgna-q:
Avoid cloning signatures when exporting ops.

Part of #1551.


aborgna-q:
Not a significant improvement vs the amount of code bloat

PR #3151, opened 2026-07-10
chore: Organize serialization benchmarks

aborgna-q:
Cleans up the serialization benchmarks

- Moves the serialization-specific benchmarks from `::benchmarks::hugr` to a new `::benchmarks::serialization` module.
- Drops benchmarks for deprecated json format
- Adds SExpression envelope benchmarks
- Disables zstd in the model encoding benchmarks
- Adds benchmarks for the envelope formats without the extra extension encoding.
- Adds a real-world `t_factory` example hugr, in addition to the syntetic ones.

- drive-by: Moves the common example hugrs from `benchmarks::hugr::examples` to `::examples`.


test_files/guppy_examples/t_factory.py
```diff
@@ -0,0 +1,78 @@
+# /// script
+# requires-python = ">=3.13"
+# dependencies = [
+#     "guppylang==1.0.0a5",
+#     "guppylang-internals==1.0.0a5",
```
maximilianruesch: Upgrade to the latest rc?

hugr/benches/examples/mod.rs
maximilianruesch, line 1: Do I go right in seeing that this change (and the one in subgraph.rs) is largely disconnected from the serialisation changes? If they are completely disconnected, I would prefer them as separate PRs...

hugr/benches/benchmarks/serialization.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// Benchmarks every representative HUGR used by the serialization suite.
+fn bench_serialization(c: &mut Criterion) {
+    bench_hugr(c, "t_factory", &t_factory());
+    bench_hugr(c, "simple_cfg", &simple_cfg_hugr());
+
+    for byte_size in [1024, 1024 * 1024] {
+        bench_hugr(c, &format!("big_hugr/{byte_size}"), &big_hugr(byte_size));
+    }
+}
```
maximilianruesch, line 97: What happened to the circuit based benchmarks?

hugr/benches/benchmarks/serialization.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Returns every serialization format covered by these benchmarks.
+///
+/// Compression is explicitly disabled so the timings only cover serialization.
+///
+/// Each format is measured with and without embedding extension definitions.
+fn serializers() -> [Serializer; 4] {
+    [
+        Serializer {
+            name: "capnp/without_extensions",
+            config: EnvelopeConfig::new(EnvelopeFormat::Model).disable_compression(),
```
maximilianruesch: Can you instead call `.disable_compression` before serialising, so there is no chance to miss this when adding new serialisers?

maximilianruesch (dismissed):
Gotta love some new benchmarks!

hugr/benches/examples/mod.rs
aborgna-q: I added the t factory example here. The rest is just renaming the file to a more shared space.

I don't think it's worth splitting a PR just for that (listed it as `drive-by` in the PR).

hugr/benches/benchmarks/serialization.rs
aborgna-q: I removed it, to balance a bit the number of new benchmars (this PR bumps the count from 33 to 45. The `circuit` ones would be 20 more.

The original idea for that test case was to see the asymptotic time grow for similar circuits of different sizes. Unfortunately codspeed doesn't have a way to visualize that nicely, so I don't think it was a good tradeoff against the extra CI time.

cqc-alec (approved):

PR #3152, opened 2026-07-10
perf: Register used extensions through a special resolver and lazy coalesce versions

maximilianruesch:
Initial status: Extension resolution takes about a second on a (private) medium size hugr:
image

The refactor to using a specialised extension resolver avoids new `ExtensionResult` objects and allows some function calls to be cached in the Python JIT cache, since they suddenly return `None`. This already brings the performance up by 37%:
image

Furthermore, the refactor allows to easily lazify version coalesce, which was very expensive due to how often individual versions had to be compared. Thus, performance jumps by another 8x:
image

### DIFF
--- a/hugr-py/src/hugr/ops.py
+++ b/hugr-py/src/hugr/ops.py
@@ -29,7 +29,11 @@ if TYPE_CHECKING:
 
     from hugr import ext
     from hugr._serialization.ops import BaseOp
-    from hugr.ext import ExtensionRegistry, ExtensionResolutionResult, UsedExtensionResolver
+    from hugr.ext import (
+        ExtensionRegistry,
+        ExtensionResolutionResult,
+        UsedExtensionResolver,
+    )
     from hugr.tys import Visibility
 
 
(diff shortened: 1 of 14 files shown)


hugr-py/src/hugr/ops.py
```diff
@@ -84,24 +88,23 @@ def name(self) -> str:
...
         extension definitions, to avoid infinite recursion.
 
         Args:
+            resolver: The resolver to report used extensions to.
             registry: A registry to resolve unresolved extensions from.
 
         Returns:
             - The resolved operation
             - A result type with both resolved and unresolved extensions
               referenced by the operation.
```
aborgna-q:
```suggestion
```

hugr-py/src/hugr/val.py
```diff
@@ -43,23 +43,21 @@ def type_(self) -> tys.Type:
...
 
         The value is modified in-place.
 
         Args:
+            resolver: The resolved to report used extensions to.
             registry: A registry to resolve unresolved extensions from.
                 If None, opaque operations and types will not be resolved.
 
         Returns:
             The resolution result containing used and unresolved extensions.
```
aborgna-q:
```suggestion
```

aborgna-q (approved):
Nice! It's a shame we don't have python benchmarks here too see the improvements

PR #3154, opened 2026-07-13
fix: Fix partial const folding of logic.eq

### LINKED ISSUE #1831: [Bug]: measurement != False drops control flow
### Description

This problem was originally reported by @jemguhit-qntm.

When comparing a measurement outcome with `!= False`, control flow drops and the branch is always taken regardless of the value of the LHS. This was originally detected with `!= comptime(False)` but I appear to get the same behaviour regardless if using runtime False or comptime(False).

Example guppy (v1):
```python
@guppy
def main() -> None:
    q = qubit()
    m = measure(q)
    a = m.read()
    if a != comptime(False):
        exit("Bug", 1)
    result("Okay", 1)
```

Some digging:
- Using a classical False instead of the measurement result on the LHS produces expected behaviour
- A conditional inversion on an unrealistic True value produces expected behaviour:
```python
    if a != False:
        a = False
    if a != comptime(False):
        exit("Bug", 1)
 ```
- Checking with `== True` produces expected behaviour
- Flipping the qubit and checking with `== False` produces expected behaviour
- Flipping the qubit and checking with `!= True` produces expected behaviour

### Steps to Reproduce

1. Use the following guppy:
```python
@guppy
def main() -> None:
    q = qubit()
    m = measure(q) # .read() for v1
    if m != False:
        exit("Failure", 1)
```
2. Compile to QIS

### Expected Behaviour

1. The output LLVM should have a conditional jump based on the measurement outcome
2. In noise-free simulation the exit path should never be chosen.

### Actual Behaviour

1. The output LLVM does not have a conditional jump based on the measurement outcome
2. Regardless of simulation config (noisy or not, statevector or coinflip) the exit path is _always_ chosen.

### Component

Rust library

### Environment

```markdown
- tket version: Problem seen on 0.19.0 and above, may be older
```

### Additional Context

_No response_

aborgna-q:
Partial folding `unknown == False` returned `False` instead of `unknown`.

drive-by: Cover all the symmetric cases in the const folding tests for logic ops.

Fixes https://github.com/Quantinuum/tket2/issues/1831

### DIFF
--- a/hugr-core/src/std_extensions/logic.rs
+++ b/hugr-core/src/std_extensions/logic.rs
@@ -43,10 +43,10 @@ impl ConstFold for LogicOp {
             }
             Self::Eq => {
                 let inps = read_inputs(consts)?;
-                let res = inps.iter().copied().reduce(|a, b| a == b)?;
-                // If we have only some inputs, we can still fold to false, but not to true
-                (!res || inps.len() as u64 == 2)
-                    .then_some(vec![(0.into(), ops::Value::from_bool(res))])
+                let [lhs, rhs] = inps.as_slice() else {
+                    return None;
+                };
+                Some(vec![(0.into(), ops::Value::from_bool(lhs == rhs))])
             }
             Self::Not => {
                 let inps = read_inputs(consts)?;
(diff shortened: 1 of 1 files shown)


cqc-alec (approved):

acl-cqc:
The code looks here as if it was expecting to have to cope with >2 inputs...?!?!

PR #3181, opened 2026-07-27
fix: Narrow the type definition for hugr.qsystem.result.DataValue

isobelhooper:
The existing type definitions:
```
DataPrimitive = int | float | bool
DataValue = DataPrimitive | list[DataPrimitive]
```

allow DataValue to include heterogeneous lists of ints, floats and bools. However, as far as I know, the actual results are limited to being homogeneous lists - either list[int], list[float] or list[bool]. The existing definition is a little too lax.


cqc-alec (approved):
Good spot.

PR #3192, opened 2026-08-04
fix: avoid poisoned values when using ishl/ishr whith large shift amounts

### LINKED ISSUE #3155: [Bug]: Fix LLVM lowering of ishl/ishr whith large shift amounts
### Description

As described in #3135, our `IntOpDef::ishl` and `ishr` operations do not limit the shift count.

We lower these to llvm's `shl`/`lshr` define those cases as poison. However, we are lowering those operations directly.

LLVM docs: https://llvm.org/docs/LangRef.html#shl-instruction / https://llvm.org/docs/LangRef.html#lshr-instruction

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

nicolaassolini-qntm:
closes #3155

To ensure consistent behaviour with respect to the hugr declared semantics, we lower the shift operation into the following LLVM sequence:
```llvm
%shift_is_valid = icmp ult i64 %rhs, 64
%shifted = shl i64 %lhs, %rhs
%shift_result = select i1 %shift_is_valid, i64 %shifted, i64 0
```
where we first check the validity of the shift, and then we pick the shifted value only if the shift was valid (0 otherwise).

### DIFF
--- a/hugr-llvm/src/extension/collections/borrow_array.rs
+++ b/hugr-llvm/src/extension/collections/borrow_array.rs
@@ -2874,10 +2874,12 @@ mod test {
     #[rstest]
     fn exec_shift_by_64(mut exec_ctx: TestContext, #[values(false, true)] right: bool) {
         const SHIFTED_VAL: u64 = 35;
-        // This test generates LLVM code to shift a 64-bit value right/left by 64 places.
-        // This is a no-op in LLVM, rather than producing 0 as one might expect.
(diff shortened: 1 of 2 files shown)


aborgna-q (approved):

aborgna-q:
@codspeedbot Why is this slower?

nicolaassolini-qntm:
I guess that the benchmark are fine?

PR #3194, opened 2026-08-05
fix: Fix pointer to hugr cli function

croyzor:
In #2764, the `_hugr` module in `hugr-py` was updated, but this script was missed!


aborgna-q (approved):
-.-

PR #3199, opened 2026-08-10
feat(hugr-py): Render extension version with render dot

nicolaassolini-qntm:
First step to solve https://github.com/Quantinuum/hugr/issues/3001

We also extend the option of rendering qualified op names to types

### DIFF
--- a/hugr-py/src/hugr/hugr/render.py
+++ b/hugr-py/src/hugr/hugr/render.py
@@ -18,7 +18,7 @@ from hugr.tys import (
     OrderKind,
     ValueKind,
 )
-from hugr.utils import name_w_args_str
+from hugr.utils import name_w_args_render
 
 from .node_port import InPort, Node, OutPort
 
@@ -301,14 +301,26 @@ class DotRenderer:
             else ""
         )
 
-        # NICOLA: TODO: change here, we want to render the version also of the types args
         match hugr[node].op:
-            case AsExtOp() as op if not self.config.qualify_op_name:
-                op_name = name_w_args_str(op.op_def().name, op.type_args())
+            case AsExtOp() as op:
+                name = (
+                    op.op_def().qualified_name()
+                    if self.config.qualify_op_name
+                    else op.op_def().name
+                )
+                op_name = name_w_args_render(
+                    name,
+                    op.type_args(),
+                    self.config.display_node_extension_version,
+                )
             case Case() as op if sibling_order is not None:
                 op_name = f"{op.name()}[{sibling_order}]"
             case Custom() as op:
-                op_name = name_w_args_str(op.name(), op.args)
+                op_name = name_w_args_render(
+                    op.name(),
+                    op.args,
+                    self.config.display_node_extension_version,
+                )
             case op:
                 op_name = op.name()
 
--- a/hugr-py/src/hugr/hugr/render.py
+++ b/hugr-py/src/hugr/hugr/render.py
@@ -398,6 +398,8 @@ class DotRenderer:
         match kind:
             case ValueKind(ty):
                 if self.config.display_link_label:
+                    # TODO: Figure out why some nested types result in all type fields
+                    # being printed instead of just the `__str__` representation.
                     label = html.escape(
                         ty.render(self.config.display_edge_extension_version)
                     )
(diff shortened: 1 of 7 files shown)


hugr-py/src/hugr/utils.py
```diff
@@ -255,8 +255,21 @@ def comma_sep_repr_paren(items: Iterable[T]) -> str:
...
+    name: str,
+    args: Iterable["TypeArg"],
+    include_extension_version: bool,
+) -> str:
+    """Format a type name with recursively rendered type arguments."""
+    args = list(args)
+    if not args:
+        return name
+    rendered_args = (arg.render(include_extension_version) for arg in args)
+    return f"{name}<{comma_sep_str(rendered_args)}>"
```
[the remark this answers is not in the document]
nicolaassolini-qntm: Pretty sure we dont need this

hugr-py/src/hugr/hugr/render.py
```diff
@@ -370,7 +400,9 @@ def _viz_link(
                 if self.config.display_link_label:
                     # TODO: Figure out why some nested types result in all type fields
                     # being printed instead of just the `__str__` representation.
```
aborgna-q: I would imagine this has gone away now that we have a new recursive rendering code.

```suggestion
```

hugr-py/src/hugr/ops.py
```diff
@@ -629,6 +629,11 @@ def _resolve_used_extensions(
 
         return ExtOp(self._op_def, new_signature, new_args)
 
+    def get_extension_version(self) -> stys.Version | None:
```
aborgna-q, line 632: The other methods do not use `get` (`op_def`, `outer_signature`, etc.), so I would drop it here.

hugr-py/src/hugr/ops.py
```diff
@@ -629,6 +629,11 @@ def _resolve_used_extensions(
...
+        """Get the extension version."""
```
aborgna-q: Document when this may be None

hugr-py/src/hugr/tys.py
```diff
@@ -123,6 +128,10 @@ def _to_serial(self) -> stys.BaseType:
     def _to_serial_root(self) -> stys.Type:
         return stys.Type(root=self._to_serial())  # type: ignore[arg-type]
 
+    def render(self, include_extension_version: bool) -> str:
```
aborgna-q: Avoid having unlabeled bool parameters, `fn_call(true)` doesn't give much context on what's the parameter doing.

nit: We could also shorten the param name.

```suggestion
    def render(self, *, extension_version: bool = False) -> str:
```

Some for `TypeArg`.

hugr-py/src/hugr/utils.py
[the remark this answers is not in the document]
aborgna-q: Well, it *is* a breaking change.

No hurt in going back to `name_w_args`.

hugr-py/src/hugr/utils.py
```diff
@@ -255,8 +255,21 @@ def comma_sep_repr_paren(items: Iterable[T]) -> str:
...
     """Helper to format a type name with type arguments."""
     if len(args) == 0:
         return name
     return f"{name}<{comma_sep_str(args)}>"
+
+
+def name_w_args_render(
+    name: str,
+    args: Iterable["TypeArg"],
+    include_extension_version: bool,
```
aborgna-q:
```suggestion
    *,
    include_extension_version: bool = False,
```

hugr-py/src/hugr/tys.py
```diff
@@ -123,6 +128,10 @@ def _to_serial(self) -> stys.BaseType:
...
+        """Render the type, optionally including extension versions."""
```
aborgna-q: Add that we're rendering the qualified name for extension types. (Same for ops)

Should we have a parameter to disable that to?

hugr-py/src/hugr/ops.py
nicolaassolini-qntm: I added the method `self._op_def._extension` because it may be None

hugr-py/src/hugr/tys.py
nicolaassolini-qntm: Maybe... types are always very long

aborgna-q (approved):
:shipit:

Issue #3189, opened 2026-07-30
[Bug]: Error serializing hugr with auto-wrapped DFG taking generic type


cqc-alec:
### Description

Serializing a hugr constructed with `Dfg.hugr` fails for some valid DFGs.

### Steps to Reproduce

```python
from hugr.build.dfg import Dfg
from hugr.std.int import int_t, INT_OPS_EXTENSION
from hugr.tys import BoundedNatArg

dfg = Dfg(int_t(6))
neg_op = INT_OPS_EXTENSION.operations["ineg"].instantiate([BoundedNatArg(6)])
node = dfg.add_op(neg_op, *dfg.inputs())
dfg.set_outputs(node)
h = dfg.hugr
h.to_bytes()
```

### Expected Behaviour

The hugr should be serialized to bytes.

### Actual Behaviour

An error is raised from `to_bytes()`:
```
[...]
    return rust.package_to_bytes(self)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^
ValueError: unknown var: ?_0
```

### Component

hugr-py

### Additional Context

The error occurs only when the type input to the DFG is an instantiated generic type.

croyzor:
Looking at the `to_str()` representation of this example, we can see this `?_0` is used as the width of the ints in the `ineg` signature
```clojure
HUGRiHJv)@[](hugr 0)

(mod)

(define-func
  private
  _1
  (core.fn
    [(arithmetic.int.types.int@0.1.0 6)]
    [(arithmetic.int.types.int@0.1.0 ?_0)])
  (meta (core.title "main"))
  (dfg [%_1] [%_2]
    (signature
      (core.fn
        [(arithmetic.int.types.int@0.1.0 6)]
        [(arithmetic.int.types.int@0.1.0 ?_0)]))
    (dfg [%_1] [%_2]
      (signature
        (core.fn
          [(arithmetic.int.types.int@0.1.0 6)]
          [(arithmetic.int.types.int@0.1.0 ?_0)]))
      (meta core.entrypoint)
      (dfg [%_3] [%_4]
        (signature
          (core.fn
            [(arithmetic.int.types.int@0.1.0 6)]
            [(arithmetic.int.types.int@0.1.0 ?_0)]))
        ((arithmetic.int.ineg@0.1.1 6) [%_3] [%_4]
          (signature
            (core.fn
              [(arithmetic.int.types.int@0.1.0 ?_0)]
[...]

PR #3200, opened 2026-08-11
feat!: Reject early bad calls to `OpDef.instantiate`

croyzor:
`OpDef.instantiate` requires a concrete signature to be provided in order to add a polymorphic operation to the graph. Otherwise, types in the operation's signature will be left out, meaning the resulting hugr wont validate, as in #3189.

Here, I try to raise this error earlier, by checking if the concrete signature is necessary, as well as raising errors for an incorrect number of type args.

---
### Aside
The signature of instantiate: 
```python
instantiate(args: Sequence[TypeArg] | None, sig: FunctionType | None)
```
should really be
```python
instantiate(concrete: tuple[Sequence[TypeArg], FunctionType] | None)
```
since the `FunctionType` needs to be provided whenever any type args are given. But this isn't usually done in python, so I thought I'd get pushback for the ugly API

---

BREAKING CHANGE: Raise errors in bad calls to `OpDef.instantiate`

### DIFF
--- a/hugr-py/src/hugr/ext.py
+++ b/hugr-py/src/hugr/ext.py
@@ -263,33 +263,40 @@ class OpDef(ExtensionObject):
             args: Type arguments corresponding to the type parameters of the definition.
             concrete_signature: Concrete function type of the operation, only required
                 if the operation is polymorphic.
+
+        Raises:
+            ValueError: if a `concrete_signature` arg is not provided to a `OpDef`
+                which requires type parameters, or if the wrong number of args are
+                provided for the `OpDef`.
+
         """
         num_args = len(args) if args else 0
         if self.signature.poly_func is not None:
             params = self.signature.poly_func.params
             if num_args > len(params):
                 msg = (
-                    f"Too many args: Op {self.name} takes {len(params)} args, "
-                    f"but was given {num_args}"
+                    f"Too many args: Op {self.qualified_name()} takes {len(params)}"
+                    f" args, but was given {num_args}"
                 )
-                raise TypeError(msg)
+                raise ValueError(msg)
             elif len(params) > num_args:
                 msg = (
-                    f"Not enough args: Op {self.name} takes {len(params)} args, "
-                    f"but was given {num_args}"
+                    f"Not enough args: Op {self.qualified_name()} takes {len(params)}"
+                    f" args, but was given {num_args}"
                 )
-                raise TypeError(msg)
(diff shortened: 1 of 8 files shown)


hugr-py/src/hugr/ext.py
```diff
@@ -264,6 +264,33 @@ def instantiate(
             concrete_signature: Concrete function type of the operation, only required
                 if the operation is polymorphic.
         """
```
aborgna-q, line 272: Add a `Raises` section to the docstring

hugr-py/src/hugr/ext.py
```diff
@@ -264,6 +264,33 @@ def instantiate(
...
+        num_args = len(args) if args else 0
+        if self.signature.poly_func is not None:
+            params = self.signature.poly_func.params
+            if num_args > len(params):
+                msg = (
+                    f"Too many args: Op {self.name} takes {len(params)} args, "
+                    f"but was given {num_args}"
+                )
+                raise TypeError(msg)
```
aborgna-q: These should rather be `ValueError`s, the types are correct.

hugr-py/src/hugr/ext.py
```diff
@@ -264,6 +264,33 @@ def instantiate(
             concrete_signature: Concrete function type of the operation, only required
                 if the operation is polymorphic.
         """
+        num_args = len(args) if args else 0
+        if self.signature.poly_func is not None:
+            params = self.signature.poly_func.params
+            if num_args > len(params):
+                msg = (
+                    f"Too many args: Op {self.name} takes {len(params)} args, "
```
aborgna-q: Use `self.qualified_name()`, so the op is clearly disambiguated.

aborgna-q (approved):

hugr-py/src/hugr/ext.py
aborgna-q, line 1: Do we need to make this a breaking change?

Cases that would error out now produced invalid hugrs before, so we are only breaking invalid uses 🤔

croyzor: Before this PR, we are successfully making the hugrs, but with unsolved type variables. I remember that at some point, when parsing the model we wanted to be able to silently solve these variables when possible, so I'm not sure if we can say the hugrs are invalid?

My thinking about the breakage is that people could have programs generating hugrs that they haven't validated, which would now surprisingly break

PR #3207, opened 2026-08-13
feat: Add configuration settings to mermaid render

### LINKED ISSUE #3001: [Feature] Render extension version for ops
In both mermaid and dot rendering (py).

This should be an opt-in flag.

nicolaassolini-qntm:
closes #3001 
by updating the mermaid rendering.

This PR allows you to decide when to print the qualified name, extension version and type arguments of nodes and edges during mermaid rendering.

This PR also deprecate `fn dot_string` in `HugrView`


hugr-core/src/hugr/views.rs
```diff
@@ -430,6 +430,14 @@ pub trait HugrView: HugrInternals {
...
         self.mermaid_string_with_formatter(self.mermaid_format())
     }
+    /// Return the mermaid representation of the underlying hierarchical graph
+    /// using the provided rendering configuration.
+    ///
+    /// This method allows customizing the appearance of node labels and other
+    /// elements in the mermaid diagram.
+    fn mermaid_string_with_config(&self, config: RenderStringConfig) -> String {
+        self.mermaid_string_with_formatter(self.mermaid_format().with_render_string_config(config))
+    }
```
nicolaassolini-qntm, line 441: Not sure if a method like this may be useful

hugr-core/src/ops/dataflow.rs
```diff
@@ -2,7 +2,7 @@
```
nicolaassolini-qntm, line 2: Most of the Op here contains types, not sure for which one should I change the rendering, since the information about the type is in the edge and the name is self explicable

hugr-core/src/ops/sum.rs
```diff
@@ -37,6 +37,10 @@ impl DataflowOpTrait for Tag {
         "Tag Sum operation"
     }
 
+    fn render_str(&self, _config: RenderStringConfig) -> String {
+        self.name().to_string()
+    }
```
nicolaassolini-qntm, line 42: Is it fine? should I render the types of the Type row?

hugr-core/src/types/custom.rs
```diff
@@ -181,9 +181,9 @@ impl Display for CustomType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{}", self.id)?;
         if !self.args.is_empty() {
-            write!(f, "(")?;
+            write!(f, "<")?;
             crate::utils::display_list(&self.args, f)?;
-            write!(f, ")")?;
+            write!(f, ">")?;
         }
```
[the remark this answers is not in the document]
nicolaassolini-qntm: I changed it for consistency, but we can change it back

hugr-core/src/hugr/views.rs
```diff
@@ -430,6 +430,14 @@ pub trait HugrView: HugrInternals {
...
```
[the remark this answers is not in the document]
nicolaassolini-qntm: Is `dot_string()` used for real in rust?

hugr-core/src/ops.rs
```diff
@@ -548,12 +548,26 @@ pub trait StaticTag {
...
+/// Configuration for rendering an operation as a string.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct RenderStringConfig {
+    /// Include the version of the extension defining the operation.
+    pub extension_version: bool,
+    /// Include the operation's type arguments.
+    pub print_type_args: bool,
+    /// Qualify operation name with their extension identifier.
+    pub qualify_name: bool,
+}
```
aborgna-q, line 560: This belongs in `render.rs`, or somewhere similar

hugr-core/src/ops.rs
```diff
@@ -548,12 +548,26 @@ pub trait StaticTag {
...
+
 #[enum_dispatch]
 /// Trait implemented by all `OpType` variants.
 pub trait OpTrait: Sized + Clone {
     /// A human-readable description of the operation.
     fn description(&self) -> &str;
 
+    /// Returns an exhaustive string representation of the operation.
+    fn render_str(&self, config: RenderStringConfig) -> String;
```
aborgna-q, line 569: Adding a new non-default-implemented method to a public trait is a breaking change (since it would break user code implementing it).

For now I'd leave either a `panic!()` or a default string here.
We should leave a TODO for sealing this trait in a breaking release too.

hugr-core/src/types/type_param.rs
```diff
@@ -177,6 +178,87 @@ impl Term {
     /// An empty list of Terms.
     pub const EMPTY_LIST: Self = Self::List(vec![]);
 
+    /// Returns a string representation of this term.
+    ///
+    /// Composite terms recursively render each of their nested terms.
+    pub fn render_str(&self, config: RenderStringConfig) -> String {
```
aborgna-q, line 184: Some of this should be dispatched to the internal structs instead.

We have `Display` impls for `SumType`, `CustomType`, etc. that do almost-but-not-quite the same logic.
Those could be either merged, or at least put the impls beside each other.

hugr-core/src/types/poly_func.rs
```diff
@@ -102,6 +103,28 @@ impl<T> PolyFuncTypeBase<T> {
 }
 
 impl<T: TypeRowLike> PolyFuncTypeBase<T> {
+    /// Render the polymorphic function type using the supplied configuration.
+    pub fn render_str(&self, config: RenderStringConfig) -> String {
+        let params = if self.params.is_empty() {
+            Cow::Borrowed("")
```
aborgna-q, line 109: Why the cow? Seems like a `&str` would be enough here.

hugr-core/src/ops/sum.rs
aborgna-q: It should print the `tag` index at least.
The type is shown in the edges, so I'd say it's not necessary.

hugr-core/src/ops/dataflow.rs
```diff
@@ -22,6 +22,9 @@ pub trait DataflowOpTrait: Sized {
     /// A human-readable description of the operation.
     fn description(&self) -> &str;
 
+    /// Returns a string representation of the operation.
+    fn render_str(&self, config: RenderStringConfig) -> String;
```
aborgna-q, line 26: Same here re. breaking changes.
This should have a default implementation.

hugr-core/src/ops/custom.rs
```diff
@@ -228,6 +228,20 @@ impl DataflowOpTrait for ExtensionOp {
...
 
+    fn render_str(&self, config: RenderStringConfig) -> String {
+        let name = render_name_with_args(
+            &self.qualified_id(),
+            self.unqualified_id(),
+            self.args(),
+            config,
+        );
+        if config.extension_version {
+            format!("{}@{}", name, self.extension_version())
```
aborgna-q, line 239: nit: `format!` is a bit slow for straight concatenation.
```suggestion
            name + "@" + &self.extension_version().to_string()
```

There's a bunch of cases of this. We don't want to make the code harder to read, but it may be worth it to change some of those.

hugr-core/src/ops/constant.rs
```diff
@@ -83,6 +83,10 @@ impl OpTrait for Const {
         "Constant value"
     }
 
+    fn render_str(&self, _config: RenderStringConfig) -> String {
+        self.name().to_string()
```
aborgna-q, line 87: We should print some info on the value.

See `Sum.__repr__` in `val.py`. For `CustomConst` we'll have to rely on `Debug`.

hugr-core/src/hugr/views.rs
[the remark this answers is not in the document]
aborgna-q: Not really, we should probably deprecate it.

hugr-core/src/hugr/views.rs
aborgna-q: We have the formatter call already, which is barely used. So I'd omit this one.

hugr-core/src/hugr/views/render.rs
```diff
@@ -26,6 +26,8 @@ pub struct MermaidFormatter<'h, H: HugrInternals + ?Sized = Hugr> {
     type_labels_in_edges: bool,
     /// A node to highlight as the graph entrypoint.
     entrypoint: Option<H::Node>,
+    /// How operation names are rendered in node labels.
+    render_string_config: RenderStringConfig,
```
aborgna-q, line 30: It makes sense to have the Config struct to pass around internally, but we don't really need the extra nesting on the public API.

Can we inline the options (with better names) here, and add the `with_*` methods?

hugr-core/src/ops.rs
```diff
@@ -548,12 +548,26 @@ pub trait StaticTag {
...
 
+/// Configuration for rendering an operation as a string.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct RenderStringConfig {
+    /// Include the version of the extension defining the operation.
+    pub extension_version: bool,
+    /// Include the operation's type arguments.
+    pub print_type_args: bool,
+    /// Qualify operation name with their extension identifier.
+    pub qualify_name: bool,
```
aborgna-q, line 559: Should we keep this on by default, to keep the existing behaviour?

hugr-core/src/ops/custom.rs
```diff
@@ -383,6 +420,21 @@ impl DataflowOpTrait for OpaqueOp {
         "Opaque operation"
     }
 
+    fn render_str(&self, config: RenderStringConfig) -> String {
```
aborgna-q, line 423: Before we prepended `OpaqueOp:` to the name.
OpaqueOps shouldn't appear on user hugrs, so it's nice to have the explicit marker to differentiate them from `ExtensionOp`.

aborgna-q (changes requested):
Can you add some ` ```mermaid ` examples to the main comment?

hugr-core/src/ops/custom.rs
nicolaassolini-qntm: Is ok without space?

hugr-core/src/ops.rs
nicolaassolini-qntm: put as true

hugr-core/src/types/poly_func.rs
nicolaassolini-qntm: `display_params` use `Cow` too, should I change?

hugr-core/src/types/type_param.rs
nicolaassolini-qntm: changes done:
- Custom type logic moved in `CustomType.render_str` and unified with the display logic
- Sum type logic moved in `SumType.render_str` and unified with the display logic
- FuncTypeBase I didnt move out the logic since it is simple (potentially we can)

hugr-core/src/hugr/views/render.rs
```diff
@@ -9,10 +9,71 @@ use portgraph::{LinkView, MultiPortGraph, NodeIndex, PortIndex, PortView};
...
 use crate::core::HugrNode;
 use crate::hugr::internal::HugrInternals;
-use crate::ops::{NamedOp, OpType};
+use crate::ops::OpTrait;
 use crate::types::EdgeKind;
 use crate::{Hugr, HugrView, Node};
 
+/// Configuration for rendering an operation as a string.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct RenderStringConfig {
```
nicolaassolini-qntm, line 18: I moved the class here, also refactor a bit the definition, now the fields are private

aborgna-q, line 18: Do we need this to be public?
```suggestion
pub(crate) struct RenderStringConfig {
```

hugr-core/src/hugr/views/render.rs
```diff
@@ -26,6 +87,12 @@ pub struct MermaidFormatter<'h, H: HugrInternals + ?Sized = Hugr> {
     type_labels_in_edges: bool,
     /// A node to highlight as the graph entrypoint.
     entrypoint: Option<H::Node>,
+    /// Include the version of the extension defining the operation.
+    pub extension_version: bool,
+    /// Include the operation's type arguments.
+    pub print_type_args: bool,
+    /// Qualify operation name with their extension identifier.
+    pub qualify_name: bool,
```
aborgna-q: The other flags use setters instead
```suggestion
    /// Include the version of the extension defining the operation.
    extension_version: bool,
    /// Include the operation's type arguments.
    print_type_args: bool,
    /// Qualify operation name with their extension identifier.
    qualify_name: bool,
```

hugr-core/src/ops/sum.rs
```diff
@@ -50,6 +51,10 @@ impl DataflowOpTrait for Tag {
         "Tag Sum operation"
     }
 
+    fn render_str(&self, _config: RenderStringConfig) -> String {
+        "Tag(".to_string() + &self.tag.to_string() + ")"
```
aborgna-q: Suggestion: We can pretty-print the kind of tagging operation here, as we do in python already

```suggestion
        let content = match &self.variants[..] {
            [val] => {
                if val.is_empty() {
                    "Unit"
                } else {
                    "Tuple"
                }
            }
            [left, right] => {
                if left.is_empty() && right.is_empty() {
                    if self.tag == 0 { "False" } else { "True" }
                } else if left.is_empty() {
                    if self.tag == 0 { "None" } else { "Some" }
                } else {
                    if self.tag == 0 { "Left" } else { "Right" }
                }
            }
            _ => &self.tag.to_string(),
        };
        "Tag(".to_string() + content + ")"
```

aborgna-q (approved):
Can you share the rendered mermaid for some complex hugr here? (Maybe `repeat_until_success.hugr` in `test_files`?)

hugr-core/src/hugr/views/render.rs
nicolaassolini-qntm: if I do the change:
```
warning: type `RenderStringConfig` is more private than the item `TypeRowLike::render_str`
  --> hugr-core/src/types/type_row.rs:94:9
   |
94 |         fn render_str(&self, config: RenderStringConfig) -> String;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method `TypeRowLike::render_str` is reachable at visibility `pub`
   |
note: but type `RenderStringConfig` is only usable at visibility `pub(crate)`
  --> hugr-core/src/hugr/views/render.rs:18:1
   |
18 | pub(crate) struct RenderStringConfig {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
so I guess the answer is yes?

aborgna-q: Indeed, we use it on public traits so no way around it.

PR #3208, opened 2026-08-13
perf: Avoid computing unused sub offsets during link establishment

maximilianruesch:
We previously also had scenarios where sub offsets contained holes (e.g. when links were broken as a last act), so it should make no real difference in functionality where there are more or less holes in sub offsets. In Python, it should also not have much of a performance impact. This change seems to pass all tests in `guppylang` as well, so I presume it is fine.

### DIFF
--- a/hugr-py/src/hugr/hugr/base.py
+++ b/hugr-py/src/hugr/hugr/base.py
@@ -43,7 +43,6 @@ from hugr.ops import (
     is_dataflow_op,
 )
 from hugr.tys import Kind, OrderKind, Type, ValueKind
-from hugr.utils import BiMap
 
 from .node_port import (
     Direction,
@@ -54,6 +53,7 @@ from .node_port import (
     PortOffset,
     ToNode,
     _SubPort,
+    _NodeLinks,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


hugr-py/src/hugr/hugr/base.py
aborgna-q: I think the rest of the code is not really prepared to handle sparse subport indices. (It will work, but a lot of methods assume the indices are mostly dense).

We need to do a full review that there's no other perf issues.

hugr-py/src/hugr/hugr/node_port.py
```diff
@@ -234,3 +236,42 @@ class _SubPort(Generic[P]):
...
+
+
+_SO = _SubPort[OutPort]
+_SI = _SubPort[InPort]
+
+
+@dataclass(frozen=True)
+class _NodeLinks:
+    _inner: BiMap[_SO, _SI] = field(init=False, default_factory=BiMap)
+    _max_subs: dict[OutPort | InPort, int] = field(init=False, default_factory=dict)
```
aborgna-q: _max_subs should be excluded from `__eq__`, `__hash__`, etc.

doug-q:
Here is a failing test. Do you agree this should pass?
```
diff --git a/hugr-py/tests/test_hugr_build.py b/hugr-py/tests/test_hugr_build.py
--- a/hugr-py/tests/test_hugr_build.py
+++ b/hugr-py/tests/test_hugr_build.py
@@ -112,6 +112,19 @@ def test_multiport(snapshot):
     validate(h.hugr, snap=snapshot)
 
 
+def test_reuses_deleted_link_sub_offset():
+    dfg = Dfg(tys.Bool)
+    source = dfg.input_node.out(0)
+    first_target = dfg.output_node.inp(0)
+    second_target = dfg.output_node.inp(1)
+
+    dfg.hugr.add_link(source, first_target)
+    dfg.hugr.delete_link(source, first_target)
+    dfg.hugr.add_link(source, second_target)
+
+    assert list(dfg.hugr.linked_ports(source)) == [second_target]
+
+
 def test_add_op(snapshot):
     h = Dfg(tys.Bool)
     (a,) = h.inputs()
```

aborgna-q (approved):

Issue #2266, opened 2025-05-28
Terms in `core`, esp ref-counting


zrho:
This issue is to propose moving towards a system of terms in `hugr-core` similar to that of `hugr-model`. While this is a major change, it simplifies the compiler, increases its expressivity and enables substantial optimisations.

- **Simplicity**: Using terms in `core` enables us to dramatically simplify the compiler data structures. Terms subsume `Type`s, `TypeArg`s, `TypeParam`s, `TypeRow`s, metadata and constants and unify their representation. 
- **Expressivity**: Despite being simpler, implementing the `model` term language in `core` would increase the expressivity of `HUGR`. In particular, it would enable two features that have already been requested: constants depending on static parameters and operations/functions parameterised by statically known functions. 
- **Performance**: We have the chance to design a term representation that allows for deduplication and cheap comparisons. This will enable a significant performance improvement over the current `core` representation, especially when importing hugrs.

## Immutable Terms

Terms are immutable and have no identity beyond structural equality. Two structurally equal terms should be considered logically indistinguishable. Equal terms can share their representation and this sharing can be observed to enable optimisations. Practical programs, and especially quantum circuits, refer to the same term many times over; sharing can lead to vast performance improvements here.

## Terms have Unique Types within a HUGR

A term has a unique type within the context of a hugr. In particular, local variables refer to their binding site. The same term may have different types when interpreted in the context of different hugrs. This enables caching types and constraints in a side table instead of requiring them to be stored inline with the term.

## View Types

Terms contain no custom Rust data structures or dynamic dispatch. This allows us to write library code that works on all terms and to optimise the representation of terms more aggressively.

As a convenient way to manipulate terms for concretely known types, views can be created that shallowly "parse" a term. Conversely, these view types can be used to construct terms. For instance, the `core.fn` constructor would have a view type like this:

```rust
struct CoreFn {
    pub inputs: Term,
    pub outputs: Term
}
```

At first, these view types and their conversions from and to terms would be written by hand or via `macro_rules`. Ultimately, view types for simple constructors can be created by codegen from declarative extensions.

[...]

PR #3215, opened 2026-08-28
perf!: Reuse shared types during HUGR import

aborgna-q:
Use an `Arc` under `Type` to avoid multiple decodings of the same type.

BREAKING CHANGE: Some `Type` methods are no longer `const`

### DIFF
--- a/hugr-core/src/extension/resolution/types_mut.rs
+++ b/hugr-core/src/extension/resolution/types_mut.rs
@@ -225,13 +225,89 @@ pub(super) fn resolve_type_exts(
     extensions: &WeakExtensionRegistry,
     used_extensions: &mut WeakExtensionRegistry,
 ) -> Result<(), ExtensionResolutionError> {
-    const EMPTY: Type = Type::new_unit_sum(0); // as no Type::default()
-    let mut tm = std::mem::replace(typ, EMPTY).into();
+    if collect_current_type_exts(typ, extensions, used_extensions) {
+        return Ok(());
+    }
+
+    let mut tm = std::mem::replace(typ, Type::new_unit_sum(0)).into();
     let r = resolve_term_exts(node, &mut tm, extensions, used_extensions);
     *typ = tm
         .try_into()
         .expect("Resolving extensions cannot change kind from RuntimeType");
     r
+}
+
+/// Collects extension references when they already match the target registry.
+///
+/// Imported types already contain the correct weak references. Detecting that
+/// case before requesting mutable access preserves their shared backing.
+fn collect_current_type_exts(
+    typ: &Type,
+    extensions: &WeakExtensionRegistry,
+    used_extensions: &mut WeakExtensionRegistry,
+) -> bool {
+    if !term_extension_refs_match(typ, extensions) {
+        return false;
+    }
+
+    let mut current_extensions = WeakExtensionRegistry::default();
+    let mut missing_extensions = ExtensionSet::new();
+    collect_term_exts(typ, &mut current_extensions, &mut missing_extensions);
+
+    if !missing_extensions.is_empty() {
+        return false;
+    }
+
+    for (id, version, extension) in current_extensions {
+        used_extensions.register(id, version, extension);
+    }
+    true
+}
+
+/// Checks every custom type reference against the target registry.
+fn term_extension_refs_match(term: &Term, extensions: &WeakExtensionRegistry) -> bool {
+    match term {
+        Term::ExtensionType(custom) => {
+            let current = custom.extension_ref();
+            current.upgrade().is_some()
+                && extensions
+                    .get_req(custom.extension(), custom.extension_version())
+                    .is_some_and(|(_, expected)| current.ptr_eq(expected))
+                && custom
+                    .args()
+                    .iter()
+                    .all(|arg| term_extension_refs_match(arg, extensions))
+        }
+        Term::FunctionType(func) => {
+            term_extension_refs_match(&func.input, extensions)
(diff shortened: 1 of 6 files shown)


hugr-core/src/extension/resolution/types_mut.rs
```diff
@@ -225,15 +225,91 @@ pub(super) fn resolve_type_exts(
...
     *typ = tm
         .try_into()
         .expect("Resolving extensions cannot change kind from RuntimeType");
     r
 }
 
+/// Collects extension references when they already match the target registry.
+///
+/// Imported types already contain the correct weak references. Detecting that
+/// case before requesting mutable access preserves their shared backing.
```
aborgna-q, line 243: This adds an extra traversal of the type tree to decide whether to do the resolution (and create a new clone).

This is a temporary change to keep this PR small. I'll follow up with a better refactor of the resolution logic.

hugr-core/src/import.rs
```diff
@@ -267,18 +267,7 @@ pub(crate) fn import_described_hugr(
 ) -> (ModuleDesc, Result<Hugr, ImportError>) {
     // TODO: Module should know about the number of edges, so that we can use a vector here.
     // For now we use a hashmap, which will be slower.
```
cqc-alec: This comment seems out of place now?

hugr-core/src/extension/resolution/types_mut.rs
aborgna-q: #3217

cqc-alec (approved):

acl-cqc:
Note that we have a longstanding issue #2266 but that would be a significantly bigger change (i.e. *every* recursive reference in `enum Term` - I'd probably just go for Arc for simplicity, though, rather than Lukas' preferred servo-arc etc.). However, that might fit better with the underlying structure of Terms in the  "table" (at least that was the idea). Is it worth prototyping? (a fairly mechanical change to the enum, so try getting an AI to follow that through?)

I do like this for the change being so localized, though! :)

Another approach to consider is https://github.com/Quantinuum/hugr/issues/1740

hugr-core/src/types.rs
```diff
@@ -454,18 +515,18 @@ impl Type {
...
     pub fn new_var_use(idx: usize, bound: TypeBound) -> Self {
-        Self(Term::new_var_use(idx, bound))
+        Self::from_term(Term::new_var_use(idx, bound))
     }
 
     /// Report the least upper [`TypeBound`]
     #[inline(always)]
-    pub const fn least_upper_bound(&self) -> TypeBound {
-        self.0.least_upper_bound().unwrap()
+    pub fn least_upper_bound(&self) -> TypeBound {
```
acl-cqc, line 527: Doesn't this make this a breaking PR? Is it not possible to keep it `const` via a `match`?

hugr-core/src/types.rs
```diff
@@ -509,15 +570,41 @@ impl Type {
 
 impl Transformable for Type {
     fn transform<T: TypeTransformer>(&mut self, tr: &T) -> Result<bool, T::Err> {
-        self.0.transform(tr)
+        self.0.make_mut().transform(tr)
```
acl-cqc, line 583: Given e.g. `ReplaceTypes` acts as a TypeTransformer, note this will unshare everything when we use that. I think that's probably not for this PR but probably worth acting on at some point.

hugr-core/src/types.rs
```diff
@@ -401,17 +399,80 @@ impl From<SumType> for Type {
...
+        }
+
+        let Self::Shared(term) = self else {
+            unreachable!("static type storage was converted to shared storage")
+        };
+        Arc::make_mut(term)
+    }
+
+    /// Consumes the storage, avoiding a clone when shared storage is unique.
+    fn into_term(self) -> Term {
```
acl-cqc: Could just be an `impl Into`

hugr-core/src/types.rs
```diff
@@ -401,17 +399,80 @@ impl From<SumType> for Type {
...
+}
 
 impl Type {
     /// An empty `TypeRow` or `TypeRowRV`. Provided here for convenience
     pub const EMPTY_TYPEROW: TypeRow = TypeRow::new();
     /// Unit type (empty tuple).
-    pub const UNIT: Self = Self(Term::SumType(SumType::Unit { size: 1 }));
+    pub const UNIT: Self = Self(TypeStorage::Static(&UNIT_TERM));
+
+    /// Wraps a validated runtime type term in shared storage.
```
acl-cqc, line 458: Wraps any term; we expect the client to validate first!
```suggestion
    /// Wraps a term in shared storage. Caller should check term is a valid type first.
```

And/or consider adding a `debug_assert!(matches!(term, Term::FunctionType(_) | Term::SumType(_) | Term::ExtensionType(_)))`

hugr-core/src/types.rs
aborgna-q: Ugh, right. We can't const deref an Arc, so `match` wouldn't help here.

hugr-core/src/types.rs
aborgna-q: Yeah, worth keeping in mind.
I added a comment / TODO

hugr-core/src/types.rs
aborgna-q: After looking more into it, I don't think there's a clean way to keep this non-breaking (specially due to `Type::new_extension` not being const anymore).

I'll mark this as breaking.

PR #3217, opened 2026-08-28
refactor: Add a shared TypeResolverContext to avoid repeated resolution

aborgna-q:
Follow up to #3215.

Improves extension resolution in a hugr by keeping an internal context struct with cached resolution of type definitions.
Since the linked PR, types now share definitions via internal Arcs. That means that once we've process an instantiation of a Type during resolution we can quickly update any other type using the same Arc, avoiding unnecessary traversals.

### DIFF
--- a/hugr-core/src/extension/resolution.rs
+++ b/hugr-core/src/extension/resolution.rs
@@ -26,10 +26,7 @@ pub use weak_registry::WeakExtensionRegistry;
 
 pub(crate) use ops::{collect_op_extension, resolve_op_extensions};
 pub(crate) use types::{collect_op_types_extensions, collect_signature_exts, collect_term_exts};
-pub(crate) use types_mut::resolve_op_types_extensions;
-use types_mut::{
-    resolve_custom_type_exts, resolve_term_exts, resolve_type_exts, resolve_value_exts,
-};
+pub(crate) use types_mut::TypeExtensionResolver;
 
 use derive_more::{Display, Error, From};
 
@@ -46,8 +43,7 @@ pub fn resolve_type_extensions(
     typ: &mut Type,
     extensions: &WeakExtensionRegistry,
 ) -> Result<(), ExtensionResolutionError> {
-    let mut used_extensions = WeakExtensionRegistry::default();
-    resolve_type_exts(None, typ, extensions, &mut used_extensions)
+    TypeExtensionResolver::new(extensions).resolve_type(None, typ)
 }
 
 /// Update all weak Extension pointers in a custom type.
@@ -55,8 +51,7 @@ pub fn resolve_custom_type_extensions(
     typ: &mut CustomType,
     extensions: &WeakExtensionRegistry,
 ) -> Result<(), ExtensionResolutionError> {
-    let mut used_extensions = WeakExtensionRegistry::default();
-    resolve_custom_type_exts(None, typ, extensions, &mut used_extensions)
+    TypeExtensionResolver::new(extensions).resolve_custom_type(None, typ)
 }
 
 /// Update all weak Extension pointers inside a type argument.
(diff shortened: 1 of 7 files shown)


hugr-core/src/extension/resolution/types_mut.rs
```diff
@@ -3,348 +3,437 @@
...
 ///
-/// Returns an iterator over the used extensions.
+/// It wraps a [`Type`], but compares and hashes based on the underlying storage
+/// pointer rather than the type's structural content.
+#[derive(Clone)]
+struct TypeStorageKey(Type);
+
+impl TypeStorageKey {
+    /// Creates a key retaining the type's current backing allocation.
+    fn new(typ: &Type) -> Self {
```
doug-q: I think it's better to take by value here. If callers have an owned value to spare they can avoid a clone.

hugr-core/src/extension/resolution/types_mut.rs
```diff
@@ -3,348 +3,437 @@
...
+    }
+}
+
+impl PartialEq for TypeStorageKey {
+    fn eq(&self, other: &Self) -> bool {
+        std::ptr::eq(self.0.storage_ptr(), other.0.storage_ptr())
+    }
+}
+
+impl Eq for TypeStorageKey {}
```
doug-q: this can be derived

hugr-core/src/types.rs
```diff
@@ -462,16 +462,29 @@ impl Type {
...
+    pub(crate) fn storage_ptr(&self) -> *const Term {
+        match &self.0 {
+            TypeStorage::Static(term) => std::ptr::from_ref(*term),
+            TypeStorage::Shared(term) => Arc::as_ptr(term),
+        }
+    }
+
+    /// Returns mutable access to the backing term using copy-on-write.
+    ///
+    /// Callers must preserve the invariant that the term is a runtime type.
```
maximilianruesch, line 477: I suppose since you are handing off complete control there is no way to insert a debug assertion for this?

hugr-core/src/extension/resolution.rs
```diff
@@ -46,35 +43,31 @@ pub fn resolve_type_extensions(
...
 }
 
 /// Update all weak Extension pointers in a custom type.
 pub fn resolve_custom_type_extensions(
     typ: &mut CustomType,
     extensions: &WeakExtensionRegistry,
 ) -> Result<(), ExtensionResolutionError> {
-    let mut used_extensions = WeakExtensionRegistry::default();
-    resolve_custom_type_exts(None, typ, extensions, &mut used_extensions)
+    TypeExtensionResolver::new(extensions).resolve_custom_type(None, typ)
```
doug-q, line 54: These functions seem to kind of defeat the point of this work. Surely we need to keep them to avoid breakage. Should we deprecate them? Or mention in their docs that they defeat the sharing

doug-q (approved):

doug-q:
Also, OpaqueOp is not covered by tests

hugr-core/src/extension/resolution/types_mut.rs
aborgna-q: This is a private fn called only in places we have a `&mut Type`, so the clone is unavoidable.

I'll make the change, it may help if the code changes in the future.

hugr-core/src/extension/resolution.rs
aborgna-q: Unfortunately these are required by extension constant definitions, since they are defined on their own without access to the hugr they'll go in :/

It's one of the parts we'll need to work on improving to make full use of the type sharing.
(along with type transformations, and some cloning on serialization)

hugr-core/src/types.rs
aborgna-q: Yeah, with `&mut` we lose all control. The method is private, but the writers are distributed along the `resolve_term` call tree.

At least we can check for correctness locally in the file.

PR #3220, opened 2026-08-28
perf: Avoid various intermediate clones during op resolution

aborgna-q:


hugr-core/src/ops/custom.rs
```diff
@@ -53,28 +53,36 @@ impl ExtensionOp {
...
             Err(SignatureError::MissingComputeFunc) => {
                 // TODO raise warning: https://github.com/CQCL/hugr/issues/1432
-                opaque.signature().into_owned()
+                Ok(opaque.signature().into_owned())
             }
-            Err(e) => return Err(e),
-        };
-        Ok(Self {
+            Err(e) => Err(e),
+        }
```
maximilianruesch:
```suggestion
        match def.compute_signature(opaque.args()) {
            Err(SignatureError::MissingComputeFunc) => {
                // TODO raise warning: https://github.com/CQCL/hugr/issues/1432
                Ok(opaque.signature().into_owned())
            }
            other => other,
        }
```
or even using a `.map_err` to make this more explicit

hugr-core/src/ops/custom.rs
```diff
@@ -374,6 +382,11 @@ impl OpaqueOp {
     pub(crate) fn args_mut(&mut self) -> &mut [TypeArg] {
         self.args.as_mut_slice()
     }
+
+    /// Move out the argument allocation after resolution has succeeded.
+    pub(crate) fn take_args(&mut self) -> Vec<TypeArg> {
+        std::mem::take(&mut self.args)
+    }
```
maximilianruesch, line 387:
```suggestion
    /// Move out the argument allocation after resolution has succeeded.
    pub(crate) fn into_args(self) -> Vec<TypeArg> {
        self.args
    }
```
Can we just consume the entire object? At its usage site we do not need it anymore.

maximilianruesch (approved):

hugr-core/src/ops/custom.rs
aborgna-q: The usage site has a &mut reference to the wrapping OpType, but I agree it's clearer to move the `mem::take` there and have `into_args` here.

aborgna-q: Ah, nvm. OpaqueOp is not `Default`, so we cannot `mem::take` it :/

PR #3221, opened 2026-08-28
perf: Reuse shared types during HUGR import

aborgna-q:
LLM'd backport of #3215.

Use an `Arc` under `Type` to avoid multiple decodings of the same type.

This branch is pre-term unification, so some changes needed to be made to work with it.
We also needed to store a `*const TypeEnum<RV>` in the Type wrapper to avoid un`const`ing functions and breaking the API.

### DIFF
--- a/hugr-core/src/extension/resolution/test.rs
+++ b/hugr-core/src/extension/resolution/test.rs
@@ -13,7 +13,10 @@ use crate::builder::{
 use crate::envelope::EnvelopeConfig;
 use crate::extension::prelude::{ConstUsize, bool_t, usize_custom_t, usize_t};
 use crate::extension::resolution::WeakExtensionRegistry;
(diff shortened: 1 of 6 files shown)


aborgna-q:
cargo-semver-checks is wrong here, those functions have not been modified in this branch.
It's using an incorrect baseline commit... https://github.com/Quantinuum/hugr/actions/runs/33185685901/job/98899304997?pr=3221#step:3:124

cqc-alec (approved):
LGTM, though it's a pity about the unsafe things and it would be nice to have a good performance comparison -- there are warnings about differing baselines, but the numbers seem good even if taken with a pinch of salt.

aborgna-q:
Did a last review of the code.
Adding `unsafe`s is annoying, but things seem sound here.

I'll go ahead and merge it.

PR #3230, opened 2026-09-01
feat(hugr-model): Allow parsing literals as constants

croyzor:


aborgna-q (approved):
Is this following some bug report?

croyzor:
> Is this following some bug report?

No, I was trying to make a `panic` in the model and found I had no way to construct a string for the error message! I just fixed it without opening an issue

PR #3231, opened 2026-09-02
feat!: Read ext version as optional in ExtensionDesc metadata

aborgna-q:
Pre-breaking release cleanup.

This was left as a breaking TODO when first implemented.

BREAKING CHANGE: `ExtensionDesc::version` is now an optional version.

Not a python breaking release
BEGIN_COMMIT_OVERRIDE
feat: Read ext version as optional in ExtensionDesc metadata (#3231)
END_COMMIT_OVERRIDE

### DIFF
--- a/hugr-core/src/envelope.rs
+++ b/hugr-core/src/envelope.rs
@@ -198,10 +198,14 @@ pub enum ExtensionBreakingError {
     Deserialization(#[from] serde_json::Error),
 }
 
-/// If HUGR metadata contains a list of used extensions, under the key [`USED_EXTENSIONS_KEY`],
-/// and extension is registered in the given registry, check that the
-/// version of the extension in the metadata matches the registered version.
-/// Version compatibility is defined by [`compatible_versions`].
+/// If HUGR metadata contains a list of used extensions, under the key
+/// [`USED_EXTENSIONS_KEY`], and extension is registered in the given registry,
+/// check that the version of the extension in the metadata matches the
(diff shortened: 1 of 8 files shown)


hugr-py/src/hugr/envelope.py
```diff
@@ -446,15 +446,10 @@ def to_json(self) -> dict[str, str]:
         """Encodes the extension as a dictionary of native types that can be
         serialized by `json.dump`.
         """
-        return {
-            "name": self.name,
-            # TODO: We set a valid version here for backwards compatibility with
-            # `hugr-rs <=0.26.1`.
-            #
```
maximilianruesch, line 453: I quite liked the previous json style expression with an inline ternary, but I do not feel that strongly about it

hugr-core/src/envelope.rs
```diff
@@ -210,14 +210,17 @@ fn check_breaking_extensions<'e>(
         let Some(registered) = registry.get(ext.name.as_str()) else {
             continue; // Extension not registered, ignore
         };
-        if !compatible_versions(registered.version(), &ext.version) {
+        let Some(used) = &ext.version else {
+            continue; // Extension version not specified, compatibility cannot be checked
+        };
```
maximilianruesch, line 219: Perhaps in the future we may want another function that allows to error instead of skip in such cases? For right now, this behaviour should be specified in the docstring (even though this function is private).

maximilianruesch (approved):

PR #3235, opened 2026-09-02
perf: Improve pest AST parsing

aborgna-q:
https://github.com/Quantinuum/hugr/pull/3212 tried to bump the `pest` version to `2.9.0` but found out that a soundness fix in it produced a 15-25% slowdown on sexpr decoding.

This PR pins the pest version to `=2.8.8`, so `2.9.0` doesn't get used for now.

While looking into that problem, I found out that a slight semantic-preserving change in the grammar may speed up things, so it's included here.


maximilianruesch (approved):
Approved assuming our own CI performance warning goes away.

Is there some issue from pest you can link here? (maybe via invisible link so that it does not show up in the pest repo?)

aborgna-q:
The relevant pest fix is at [pest#1184](https://www.github.com/pest-parser/pest/pull/1184).
That change caused the big performance drop in our parser.

Now that the benchmarks ran with the pinned `2.8.8` version, the slowdowns are gone.

PR #3241, opened 2026-09-03
perf: Cache extension type helpers

aborgna-q:
Caches the `qb_t` / `bool_t` / etc. type helpers so they don't create a fresh `Arc` each time they get called.

### DIFF
--- a/hugr-core/src/extension/prelude.rs
+++ b/hugr-core/src/extension/prelude.rs
@@ -170,20 +170,24 @@ pub(crate) fn qb_custom_t(extension_ref: &Weak<Extension>) -> CustomType {
     )
 }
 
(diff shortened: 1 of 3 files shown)


maximilianruesch:
I will be waiting for the benchmark to see how much of an impact this has

aborgna-q:
> Merging this PR will degrade performance by 8.82%

I blame allocator noise :P (see flamegraphs).
I'm tempted to enable [Allocator Exclusion](https://codspeed.io/docs/instruments/cpu/allocation-exclusion) to avoid these benchmark variances.

Asides from that, the benches are positive / neutral.

maximilianruesch (approved):

PR #3246, opened 2026-09-04
ci: Update cargo-dist version to fix CI erors

aborgna-q:
Backport of #3100

`hugr-cli` release builds are failing due to an old `cargo-dist` trying to use rust `1.89`, lower than our `1.91` MSRV.

https://github.com/Quantinuum/hugr/actions/runs/27023075469/job/79756088308#step:9:16

This should fix it. I just installed a newer `cargo-dist` locally and ran `dist init` to regenerate the files.

I'm not sure how to check it actually works though, the workflow run for PRs only generates a manifest that doesn't state the rust versions used. But it should normally work correctly...

### DIFF
--- a/.github/workflows/hugr-cli-v-release.yml
+++ b/.github/workflows/hugr-cli-v-release.yml
@@ -56,7 +56,7 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           persist-credentials: false
           submodules: recursive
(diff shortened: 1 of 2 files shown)


.github/workflows/hugr-cli-v-release.yml
```diff
@@ -56,17 +56,17 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
```
maximilianruesch, line 59: This change is a lot noisier than just a cargo-dist upgrade, is that intentional?

aborgna-q: `hugr-cli-v-release.yml` is auto-generated by cargo-dist

maximilianruesch (approved):

PR #3247, opened 2026-09-04
feat: Improve extension resolution errors

aborgna-q:
Improves the error message when an op/type cannot be resolved from the set of extensions.

Possible messages include:

- `Could not resolve operation foo.bar: Extension foo@0.1.2 is required, but the available version 0.1.1 is too old`
- `Could not resolve type foo.baz in Node(4): Extension foo@0.1.2 is required, but the available versions [0.0.1, 1.2.3] are incompatible`
- `Could not resolve operation foo.bar: Extension foo@0.1.2 is required, but it was not found. The available extensions are: X@0.0.1, Y@0.1.2`

### DIFF
--- a/hugr-core/src/envelope/reader.rs
+++ b/hugr-core/src/envelope/reader.rs
@@ -105,8 +105,9 @@ impl<R: BufRead> EnvelopeReader<R> {
 
     /// Handle extension resolution errors by recording missing extensions in the description.
     ///
-    /// This function inspects the error and adds any missing extensions to the module description
-    /// without a version.
+    /// This function inspects the error and adds any missing extensions to the module description,
+    /// preserving the required version when one was supplied.
+    #[expect(deprecated)]
     fn handle_resolution_error(desc: &mut ModuleDesc, err: &ExtensionResolutionError) {
         match err {
             ExtensionResolutionError::MissingOpExtension {
@@ -117,6 +118,12 @@ impl<R: BufRead> EnvelopeReader<R> {
             } => desc.extend_used_extensions_resolved([ExtensionDesc::new_unversioned(
                 missing_extension,
             )]),
+            ExtensionResolutionError::UnresolvedOpExtension { description, .. }
+            | ExtensionResolutionError::UnresolvedTypeExtension { description, .. } => desc
+                .extend_used_extensions_resolved([ExtensionDesc::new(
+                    &description.required_extension,
+                    description.required_version.clone(),
+                )]),
             ExtensionResolutionError::InvalidConstTypes {
                 missing_extensions, ..
             } => desc.extend_used_extensions_resolved(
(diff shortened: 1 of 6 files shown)


hugr-core/src/envelope/reader.rs
```diff
@@ -105,8 +105,9 @@ impl<R: BufRead> EnvelopeReader<R> {
 
     /// Handle extension resolution errors by recording missing extensions in the description.
     ///
-    /// This function inspects the error and adds any missing extensions to the module description
-    /// without a version.
+    /// This function inspects the error and adds any missing extensions to the module description,
+    /// preserving the required version when one was supplied.
+    #[expect(deprecated)]
```
maximilianruesch: Can the docstring detail what exactly is deprecated here, and what should be used instead?

hugr-core/src/extension/resolution/test.rs
```diff
@@ -152,6 +154,162 @@ fn resolve_custom_type_uses_highest_compatible_extension() {
...
+            ],
+        }
+    );
+    assert!(
+        error
+            .to_string()
+            .contains(&format!("{ext_id}@{required_version}"))
+    );
+    assert!(error.to_string().contains(expected_reason));
+    assert!(!error.to_string().contains("other_ext@9.0.0"));
```
maximilianruesch:
```suggestion
    assert!(!error.to_string().contains("{other_id}@9.0.0"));
```
for a bit more robustness?

hugr-core/src/extension/resolution.rs
```diff
@@ -180,6 +215,150 @@ impl<N: HugrNode> ExtensionResolutionError<N> {
...
+            .iter()
+            .copied()
+            .filter(|version| {
+                *version < &self.required_version
+                    && semver_compatible(&self.required_version, version)
+            })
+            .max()
+        {
+            return format!(
+                "extension {} is required, but the available version {version} is too old",
```
maximilianruesch: Perhaps this should list the other available versions as well, for transparency?

hugr-core/src/extension/resolution.rs
```diff
@@ -180,6 +215,150 @@ impl<N: HugrNode> ExtensionResolutionError<N> {
...
+        }
+    }
+
+    /// Format the required extension as `id@version`.
+    fn required(&self) -> String {
+        format!("{}@{}", self.required_extension, self.required_version)
+    }
+
+    /// Format every available extension as `id@version`.
+    fn available(&self) -> String {
```
maximilianruesch: Perhaps `all_available` since it is independent of the extension id?

maximilianruesch (approved):

hugr-core/src/extension/resolution/test.rs
aborgna-q: `.contains(&format!("{other_id}@9.0.0")))` is less readable imho, but sure

PR #3257, opened 2026-09-10
chore: Use a supported dependabot strategy for uv ecosystem

maximilianruesch:
`widen` is not supported for the `uv` ecosystem yet:
```
The property '#/updates/1/versioning-strategy' value "widen" did not match one of the following values: lockfile-only, auto, increase, increase-if-necessary
```
Thus we use `increase-if-necessary` for now, since that is really similar to `widen`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -46,7 +46,19 @@ updates:
     labels:
       - "X-dependabot"
       - "A-python"
-    versioning-strategy: "increase-if-necessary"
+    # Do not bump the minimal version of dependencies.
+    #
+    # Bumping the minimum can be a breaking change for downstream python users,
+    # as they may break their dependency resolution and prevent them from updating.
+    #
+    # Ideally we'd widen the upper bound.
+    # The dependabot documentation indicates this can be done by setting versioning-strategy to "widen",
+    # but this returned errors saying that the value is invalid.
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -46,7 +46,7 @@ updates:
     labels:
       - "X-dependabot"
       - "A-python"
-    versioning-strategy: "widen"
+    versioning-strategy: "increase-if-necessary"
```
aborgna-q: I would leave this as `lockfile-only` for automated changes. Bumping across unsupported dependencies should be done manually in a breaking PR.

Also, copying the comment blob [from tket](https://github.com/Quantinuum/tket2/blob/e70a548932b0abc1cb4ce88e074fc0dc71d121cd/.github/dependabot.yml#L68).

```suggestion
    # Do not bump the minimal version of dependencies.
    #
    # Bumping the minimum can be a breaking change for downstream python users,
    # as they may break their dependency resolution and prevent them from updating.
    #
    # Ideally we'd widen the upper bound.
    # The dependabot documentation indicates this can be done by setting versioning-strategy to "widen",
    # but this returned errors saying that the value is invalid.
    # (https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#versioning-strategy--)
    #
    # See the uv versioning-strategy issue:
    # <https://github.com/dependabot/dependabot-core/issues/15290>
    versioning-strategy: "lockfile-only"
```

aborgna-q (approved):

PR #3262, opened 2026-09-16
ci: Fix semver-checks check permission

aborgna-q:
We added a check in the `rs-semver-checks` action to reject runs from `pull_request_target`s, since that's a security issue (we'd be running untrusted fork PRs with access to the repo's secrets).

That means we need to update the workflow here.
The check will still work on fork PRs, but no comment will be posted (since we cannot access a token with pr write access from there).


aborgna-q:
(The check failing in this PR is still running with `pull_request_target`, since that takes precedence).

cqc-alec (approved):

